PR #1739, opened 2015-11-19
Change 'Select All' button to 'Deselect All' when all items are selected

AMeng:
If all items in a list are selected, we're left with an unhelpful "Select All" button. This changes that button to a "Deselect All" if every item is already selected. If only some of the items are selected, it remains "Select All".

image
image
image

### DIFF
--- a/app/scripts/modules/core/forms/checklist/checklist.directive.html
+++ b/app/scripts/modules/core/forms/checklist/checklist.directive.html
@@ -20,7 +20,7 @@
     </label>
   </li>
   <li ng-if="includeSelectAllButton && items.length > 1">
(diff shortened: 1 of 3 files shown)


ajordens:
:thumbsup:

Thanks for the PR!

@zanthrash This look reasonable to you?

tomaslin:
would be nice if the tests were updated to include changes - https://github.com/AMeng/deck/blob/deselect-all-button/app/scripts/modules/core/forms/checklist/checklist.directive.spec.js

AMeng:
@tomaslin, tests added.

tomaslin:
LGTM @anotherchrisberry PTAL

PR #1748, opened 2015-11-23
Support an explicit rollback action

ajordens:
- `Rollback` becomes an server group action on enabled server groups

### DIFF
--- a/app/scripts/modules/amazon/common/footer.directive.js
+++ b/app/scripts/modules/amazon/common/footer.directive.js
@@ -3,15 +3,15 @@
 let angular = require('angular');
 
 module.exports = angular
-  .module('spinnaker.amazon.serverGroup.details.resize.footer.directive', [
+  .module('spinnaker.amazon.footer.directive', [
   ])
-  .directive('awsResizeFooter', function () {
+  .directive('awsFooter', function () {
     return {
       restrict: 'E',
-      templateUrl: require('./resizeFooter.directive.html'),
+      templateUrl: require('./footer.directive.html'),
       scope: {},
       bindToController: {
-        resize: '&',
+        action: '&',
         isValid: '&',
         cancel: '&',
       },
--- a/app/scripts/modules/amazon/serverGroup/details/rollback/rollbackServerGroup.html
+++ b/app/scripts/modules/amazon/serverGroup/details/rollback/rollbackServerGroup.html
@@ -1,0 +1,37 @@
+<div modal-page class="confirmation-modal">
+  <task-monitor monitor="taskMonitor"></task-monitor>
+  <form role="form">
+    <modal-close></modal-close>
+    <div class="modal-header">
+      <h3>Rollback {{serverGroup.name}}</h3>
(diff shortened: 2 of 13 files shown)


ajordens:
PTAL @anotherchrisberry.

As far as those directives go, I pulled them up a bit but left them in `aws`. I'm sure there's a more appropriate long-term place.

ajordens:
image

image

app/scripts/modules/amazon/common/footer.directive.js
```diff
@@ -3,15 +3,15 @@
 let angular = require('angular');
 
 module.exports = angular
-  .module('spinnaker.amazon.serverGroup.details.resize.footer.directive', [
+  .module('spinnaker.amazon.footer.directive', [
   ])
-  .directive('awsResizeFooter', function () {
+  .directive('awsFooter', function () {
```
anotherchrisberry: This should probably be "modalFooter", if we are going to abstract it away from the resize modal, but I will do that as part of a separate PR this afternoon.

app/scripts/modules/amazon/serverGroup/details/rollback/rollbackServerGroup.html
```diff
@@ -0,0 +1,37 @@
...
+          </ui-select>
+        </div>
+      </div>
+
+      <div class="row" ng-if="command.platformHealthOnlyShowOverride">
+        <div class="col-sm-10 col-sm-offset-1">
+          <platform-health-override command="command"
+                                    platform-health-type="'Amazon'"
+                                    show-help-details="true"
+                                    field-columns="12">
```
anotherchrisberry: `field-columns` doesn't do anything any more (but I need to update the directive definition, so I will clean this up at that point).

app/scripts/modules/core/serverGroup/serverGroup.write.service.js
```diff
@@ -60,6 +60,23 @@ module.exports = angular
       });
     }
 
+    function rollbackServerGroup(serverGroup, application, params={}) {
+      params.asgName = serverGroup.name;
+      params.serverGroupName = serverGroup.name;
```
anotherchrisberry: Do we need both here? You just wrote the Orca functionality - does it expect the "asgName" field to be populated from the client?

app/scripts/modules/core/serverGroup/serverGroup.write.service.js
```diff
@@ -60,6 +60,23 @@ module.exports = angular
...
+      params.type = 'rollbackServerGroup';
+      params.regions = [serverGroup.region];
+      params.region = serverGroup.region;
```
anotherchrisberry: Same as above - do we need `regions` and `region`?

ajordens: `orca` is just passing almost directly through to cloud driver hence following the same convention here.

Just looking at CD to see if we can't tidy up some of this.

PR #1787, opened 2015-12-09
Add a user-menu

riltsken:
Currently only displays the logged in users name. Could do more later (eg logout). Does not
show if the authEnabled setting is false.

### DIFF
--- a/app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
+++ b/app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
@@ -1,3 +1,14 @@
-<li ng-show="authEnabled" class="user-menu">
-  <a class="user-name">{{ user.name }}</a>
+<li ng-if="authEnabled" class="user-menu">
+  <!-- Smaller screens use this -->
+  <a href class="hidden-lg user-name-icon dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="true">
+    <span class="glyphicon glyphicon-user"></span>
+    <ul class="dropdown-menu">
+      <li>
+        <a class="user-name-small">{{ user.name }}</a>
+      </li>
+    </ul>
+  </a>
+
+  <!-- Larger screens use this -->
+  <a class="hidden-xs hidden-sm hidden-md user-name-large">{{ user.name }}</a>
 </li>
--- a/app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
+++ b/app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
@@ -1,0 +1,3 @@
+<li ng-show="authEnabled" class="user-menu">
+  <a class="user-name">{{ user.name }}</a>
+</li>
--- a/app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
+++ b/app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
@@ -1,6 +1,6 @@
 <li ng-if="authEnabled" class="user-menu">
   <!-- Smaller screens use this -->
-  <a href class="hidden-lg user-name-icon dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="true">
+  <a href class="hidden-lg remove-border-top dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="true">
     <span class="glyphicon glyphicon-user"></span>
     <ul class="dropdown-menu">
       <li>
--- a/app/scripts/modules/core/authentication/userMenu/userMenu.directive.js
+++ b/app/scripts/modules/core/authentication/userMenu/userMenu.directive.js
@@ -11,7 +11,7 @@ module.exports = angular.module('spinnaker.core.authentication.userMenu.directiv
       restrict: 'E',
       replace: true,
       templateUrl: require('./userMenu.directive.html'),
-      link: function(scope, element) {
+      link: function(scope) {
         scope.authEnabled = settings.authEnabled;
         scope.user = authenticationService.getAuthenticatedUser();
       }
--- a/app/scripts/modules/core/authentication/userMenu/userMenu.directive.js
+++ b/app/scripts/modules/core/authentication/userMenu/userMenu.directive.js
@@ -1,0 +1,19 @@
+'use strict';
+
+let angular = require('angular');
+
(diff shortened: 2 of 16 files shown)


riltsken:
image

zanthrash:
deck_ _clusters
Thanks for the PR.  This is what I'm seeing in Firefox 42.0 and Chrome 47.0 and Safari 9.0.1 on OS X.

If you add the "replace: true" to the userMenu directive and remove the isolate scope block it should render correctly.

zanthrash:
Sorry didn't mean to close

riltsken:
@zanthrash Thanks. I updated the commit to reflect your suggestions.

riltsken:
(should actually display correctly now like you said)

anotherchrisberry:
When we first implemented authentication, we had a similar navbar item, but ended up removing it because it felt funny wasn't providing much functionality. That said, it was a menu dropdown, not just the text, which is part of the reason it felt funny. This is cleaner, and makes more sense.

When we get to the point (someday) where we offer other functionality (e.g. logging out, user preferences, clearing caches), we can switch it to a dropdown menu.

The only problem with this implementation is it breaks down a bit when the `netflixMode` feature flag is turned on and the screen width is < 1200px:
screen shot 2015-12-09 at 6 32 32 pm

We can probably get around this by adding a `hidden-md` class to the `whatsNew` and `feedback` directives, and maybe changing the HTML of this new directive to just show the `glyphicon-user` icon with a `uib-tooltip` to provide the user name, using a similar approach to the `whatsNew` and `feedback` directives.

Sorry to ask you to update Netflix-specific components when you won't actually use them, but I think we want to get those changes in place before merging.

app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
```diff
@@ -0,0 +1,3 @@
+<li ng-show="authEnabled" class="user-menu">
```
anotherchrisberry: We prefer `ng-if` to `ng-show`, unless we really want the internal DOM to be rendered and hidden via CSS (for example, we want some of the wizard pages to be rendered in order to initialize their controllers, but that is a whole other huge mess). That's not to say we _haven't_ littered the codebase with `ng-show`s that should be `ng-if`s, just that we haven't cleaned them up and are trying to avoid them in new code.

Anyway, short story is: this should probably be an `ng-if` unless it needs to be an `ng-show`, which I don't think is the case.

app/scripts/modules/core/authentication/userMenu/userMenu.directive.spec.js
```diff
@@ -0,0 +1,46 @@
...
+      spyOn(authenticationService, 'getAuthenticatedUser').and.returnValue({'name': 'sam mulligan'});
+
+      domNode = $compile('<user-menu></user-menu>')($scope);
+      $scope.$digest();
+
+      expect(domNode.hasClass('ng-hide')).toBe(false);
+      expect(domNode.find('.user-name').text()).toBe('sam mulligan');
+    });
+  });
+});
```
anotherchrisberry: thank you for including tests!

app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
riltsken: Totally agree with the `ng-if` actually. The problem I had was when `replace: true` was set on the directive and I ran through the tests the `$compile` portion wouldn't render the dom element and instead returned a html comment `<!-- ng-if authEnabled -->`. I'll look into this more but if you know offhand a solution to this it would be greatly appreciated :)

riltsken:
Yep I actually started out with a dropdown nav but wanted to keep it simple for now in case this wasn't wanted. Other things that could be useful are your ACL and account access being exposed, which is easier than going to `/auth/info/` if you are trying to debug something with auth.

I don't mind doing the changes with netflix mode enabled. I figured something might come up after opening the PR :)

app/scripts/modules/core/authentication/userMenu/userMenu.directive.html
anotherchrisberry: You could wrap the `<user-menu></user-menu>` tag in another tag, e.g. `<div><user-menu></user-menu></div>`, then test if the parent node contains an `<li>`. Give that a shot and see if it works okay.

riltsken:
Almost done with this just finishing up the tests, but here is what it will look like and wanted to run that through here asap in case there are changes wanted.

__Nextflix mode enabled__

Large screen


Medium screen


Small screen



__Netflix mode enabled with whats new updates unread__

Large screen


Med / Sm screens remove the timestamp

__Core mode__

Large screen


Medium screen


Small screen

riltsken:
It says anonymous only because I removed auth so I could test locally. Getting auth enabled with saml proved nontrivial locally :)

riltsken:
Should be ready for review fyi.

anotherchrisberry:
Just getting back to reviewing this - thanks for your patience, @riltsken

app/scripts/modules/core/authentication/userMenu/userMenu.directive.js
```diff
@@ -0,0 +1,19 @@
...
+module.exports = angular.module('spinnaker.core.authentication.userMenu.directive', [
+  require('../../config/settings.js'),
+  require('../authentication.service.js'),
+])
+  .directive('userMenu', function(settings, authenticationService) {
+    return {
+      restrict: 'E',
+      replace: true,
+      templateUrl: require('./userMenu.directive.html'),
+      link: function(scope, element) {
```
anotherchrisberry: `element` can be omitted here.

anotherchrisberry:
Looks like we lost some formatting on the `whatsNew` directive by switching it to `replace: true` - take a look at:
* https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/presentation/main.less#L948-L959
* https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/presentation/main.less#L1004-L1022

They're in separate rule declarations because our CSS is a mess, not because they should be.

Can you update the styles so that we don't lose that formatting?

Otherwise, this looks good - nice work!

riltsken:
Should be ready for another look @anotherchrisberry

anotherchrisberry:
LGTM - just needs a rebase

riltsken:
Rebased

anotherchrisberry:
Oh no, I am sorry, looks like it needs another merge :(

PR #1832, opened 2015-12-30
Surface Google Autoscaler information in server group details.

duftler:


app/scripts/modules/google/serverGroup/details/scalingPolicy/scalingPolicy.directive.html
```diff
@@ -0,0 +1,22 @@
+<dl ng-class="InsightFilterStateModel.filtersExpanded ? '' : 'dl-horizontal dl-wide'" style="margin-bottom: 20px">
+  <dt>
+    Target {{policy.basis}}
+    <help-field ng-if="policy.basisHelpKey" key="{{policy.basisHelpKey}}"></help-field>
```
anotherchrisberry: You shouldn't need the `ng-if` here - if the `key` is not found, the help bubble will not render.

anotherchrisberry:
LGTM once that `ng-if` is removed.

app/scripts/modules/google/serverGroup/details/scalingPolicy/scalingPolicy.directive.html
duftler: Done.

duftler:
image

PR #1869, opened 2016-01-14
The bake stage is now restartable

ajordens:


ajordens:
FYI @duftler - This is the `deck` side of what we were talking about earlier.

PTAL @anotherchrisberry.

PR #1870, opened 2016-01-14
SPIN-150: For Cluster Filters: Filtering down the region specific AZ's when a region filter is selected.

zanthrash:
@anotherchrisberry Please take a look when you have a moment.


anotherchrisberry:
Looks good - I think the only thing missing is clearing the tags at the top of the groups (in the main section of the content) when an option is no longer available.

zanthrash:
That should be working too.  It's all tied off the selected filter in the model.

zanthrash:
Closing to cleanp

PR #1888, opened 2016-01-21
Add userdata field to the UI

AMeng:
The AWS deploy accepts an optional field "base64UserData" in the pipeline's JSON. This just adds that field to the UI so that users don't need to edit the JSON manually. The "Advanced Settings" tab seemed appropriate for this. The help text informs the users that the string must be base64 encoded.

image

### DIFF
--- a/app/scripts/modules/amazon/serverGroup/configure/wizard/advancedSettings.html
+++ b/app/scripts/modules/amazon/serverGroup/configure/wizard/advancedSettings.html
@@ -64,6 +64,15 @@
                                    ng-model="command.iamRole"/></div>
     </div>
     <div class="form-group">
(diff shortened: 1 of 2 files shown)


anotherchrisberry:
Does this need to be restricted to pipeline deployments, or will it work on a create/clone operation?

AMeng:
Oh, yes, those interfaces use this same template, so the field will appear for those as well. For example:

image

anotherchrisberry:
Right, sorry, I just want to verify the functionality will work on create/clone operations as well as pipeline deployments.

LGTM - just needs a rebase.

AMeng:
Gotcha. I just both a create and a clone, and my new user data from the UI was pushed through as expected.

PR #1940, opened 2016-02-05
UI for defining git triggers

tomaslin:


zanthrash:
@tomaslin :+1:

app/scripts/modules/core/pipeline/config/triggers/git/gitTrigger.html
```diff
@@ -0,0 +1,46 @@
...
+      Project
+      <help-field key="pipeline.config.git.project"></help-field>
+    </label>
+
+    <div class="col-md-9">
+      <input type="text" class="form-control input-sm"
+             ng-model="vm.trigger.project"
+             name="project"
+             required
+             placeholder="Enter your repository project name, i.e. spinnaker for github.com/spinnaker/echo"
```
ajordens: Inline examples are helpful but are we sure we've got the terminology right?

`spinnaker` is the owner ... not really a project at least in the case of github which generally labels repositories themselves as `projects` AFAIK. (On the contrary, Stash does treat groups of repositories as a project).

Similarly below it might be more common to ask for the repository slug name vs "source repository name".

Not a huge deal either way.

tomaslin: yes, was trying to keep consistent with the project attributes screen. Makes sense to use the terminology for each scm type in the UI even if we call it something else in the backend

app/scripts/modules/core/pipeline/config/triggers/git/gitTrigger.html
```diff
@@ -0,0 +1,46 @@
+<ng-form name="cronForm">
+  <div class="form-group">
+    <label class="col-md-3 sm-label-right">
+      Repository Type
+      <help-field key="pipeline.config.git.repotype"></help-field>
+    </label>
+    <div class="col-md-6">
+      <select
+        class="form-control input-sm"
+        ng-options="repoType for repoType in ['stash', 'github']"
```
anotherchrisberry: This should be configurable - while Stash makes sense for our needs, I doubt many users outside Netflix are going to want it as an option, much less the default option.

tomaslin:
updated the github screen so now it says
 
screen shot 2016-02-08 at 10 40 22 am

sources now configurable via settings.gitSources

tomaslin:
@anotherchrisberry @ajordens  ^^

PR #1941, opened 2016-02-06
Fix for "The deck debian package does not install apache2 in the right order" Issue#723

vjames-bmc:
Previously apache2 was added as "Depends" in the control file (through build.gradle). Because of this, the actual deployment of apache2 was happening only after the deck files are laid out. This resulted in the index.html file being overwritten by the one from apache2.
Fix was to add apache2 as Pre-Depends instead of Depends in the debian control file thereby making sure that apache2 is installed and configured fully before deck is installed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,7 +65,8 @@ ospackage {
   packageName = "spinnaker-deck"
   version = toVers(project.version.toString())
   release '3'
-  into "/var/www/html"
+  into "/opt/deck/html"
   requires "apache2"
   from "build/webpack"
 }
+


anotherchrisberry:
@cfieber can you take a look at this? Seems reasonable to me.

cfieber:
It seems like pre-depends is frowned upon ( from https://www.debian.org/doc/debian-policy/ch-relationships.html )

If the issue is a conflicting index.html maybe there is a more standard way of plugging in the site content. For example changing it so we build into /var/www/html/deck then adding `/etc/apache2/vhosts.d/deck.vhost` with:

```
<VirtualHost _default_:*>
DocumentRoot /var/www/html/deck/
</VirtualHost>
```

Would appreciate anyone else's thoughts on the 'right' way to do this though ( @skorten @dpeterka @ewiseblatt )?

ewiseblatt:
SeanK might have a suggestion.

skorten:
I agree with @cfieber, Pre-Depends should be avoided.  The installation behavior described is strange and does not match the documentation: "Depends: This declares an absolute dependency. A package will not be configured unless all of the packages listed in its Depends field have been correctly configured" so perhaps something else is going on there.  @vjames-bmc were you able to replicate this on a clean/fresh install?

Regardless, I would still suggest the vhost strategy similar to the one outlined by @cfieber as installing directly to /var/www/html is not great as it does conflict with the apache package.  This actually causes an installation failure when working with RPMs.

The approach I recommend would be to change the installation location of deck from /var/www/html to /opt/deck/html and update the vhost configuration accordingly.  This would also necessitate changes to the spinnaker meta package.

vjames-bmc:
@cfieber  I agree, changing the root for deck to  folder other than /var/www/html would be the ideal solution. I will try that...
@skorten I was able to reproduce the issue on a clean machine by installing just the deck debian package. It doesn't happen when you use the install_spinnaker script because, it installs apache2 upfront as a dependency.

Regarding the "depends" flag, the documentation warns that the dependent packages might be in unconfigured stage. See below. 
"A Depends field takes effect only when a package is to be configured. It does not prevent a package being on the system in an unconfigured state while its dependencies are unsatisfied, and it is possible to replace a package whose dependencies are satisfied and which is properly installed with a different version whose dependencies are not and cannot be satisfied; when this is done the depending package will be left unconfigured (since attempts to configure it will give errors) and will not function properly. If it is necessary, a Pre-Depends field can be used, which has a partial effect even when a package is being unpacked, as explained in detail below. "

vjames-bmc:
As @cfieber and @skorten mentioned, I am changing the deck install folder to a different one. As @skorten  suggested, I am changing it to /opt/deck/html. Does the new location look ok? I like it because, now we will have all the spinnaker components under /opt. Let me now if you prefer a different folder instead.

cfieber:
I like /opt/deck/html

vjames-bmc:
Thanks @cfieber I have made corresponding changes in the spinnaker project as well as @skorten pointed out. I have sent a pull request for that change as https://github.com/spinnaker/spinnaker/pull/740

claymccoy:
It looks like this just needs to be rebased and then both it and #740 can be merged at the same time.

vjames-bmc:
@claymccoy I have rebased and squashed the commits as you requested. Thanks

anotherchrisberry:
@vjames-bmc sorry, can you rebase again? ping me when you do and I'll try to get this merged before anything else goes in on top of it.

vjames-bmc:
Thanks @anotherchrisberry I have rebased it.

anotherchrisberry:
Awesome, thanks for working through this!

vjames-bmc:
Thanks @anotherchrisberry . There is one more PR related to this which is https://github.com/spinnaker/spinnaker/pull/740 in the spinnaker project. These two PRs were done for fixing Issue#723. Do I have to do something for that issue entry so that it can be closed?

anotherchrisberry:
Thanks for pointing that out - just merged https://github.com/spinnaker/spinnaker/pull/740

PR #1947, opened 2016-02-08
ForceCacheRefresh tasks should not be hidden

ajordens:


ajordens:
Mr. @anotherchrisberry PTAL

PR #1969, opened 2016-02-12
Show a warning if a deprecated region is selected for deploy

ajordens:


app/scripts/modules/amazon/serverGroup/configure/serverGroupConfiguration.service.js
```diff
@@ -54,6 +54,15 @@ module.exports = angular.module('spinnaker.aws.serverGroup.configure.service', [
         return command.suspendedProcesses.indexOf(process) !== -1;
       };
 
+      command.regionIsDeprecated = function() {
+        if (command.backingData === undefined) {
```
ajordens: Probably a better way to do this ...

ajordens:
image

Here it is @anotherchrisberry

app/scripts/modules/amazon/serverGroup/configure/serverGroupConfiguration.service.js
```diff
@@ -54,6 +54,15 @@ module.exports = angular.module('spinnaker.aws.serverGroup.configure.service', [
...
+          return false;
+        }
+
+        var matchedRegion = _.find(command.backingData.filtered.regions, {name: command.region});
+        return matchedRegion ? matchedRegion.deprecated === true : false;
+      };
```
anotherchrisberry: Give this a shot, eh?
```javascript
command.regionIsDeprecated = () => command.backingData && 
  command.backingData.filtered.regions.filter((region) => region.name === command.region && 
    region.deprecated).length;
```

ajordens:
Updated PR with your suggestion @anotherchrisberry.

Peter also asked us to clarify that it was the region in the account that was deprecated so I've adjusted the warning message text slightly.

app/scripts/modules/core/region/regionSelectField.directive.html
```diff
@@ -7,7 +7,7 @@
             ng-model="component[field]"
             ng-change="onChange()"
             required>
-      <option ng-repeat="region in regions" value="{{region.name}}" ng-selected="component[field] === region.name">{{region.name}}</option>
+      <option ng-repeat="region in regions" value="{{region.name}}" ng-selected="component[field] === region.name">{{region.name}} {{region.deprecated ? '(deprecated)' : ''}}</option>
```
anotherchrisberry: Do we have the account name available here? Might be nice to include it, i.e. `(deprecated in test)`

No big deal if not.

PR #1982, opened 2016-02-17
Move ng-if logic intro controller methods.

duftler:
fyi @tonyrzhang


duftler:

tonyrzhang:
Thanks for making the change lgtm

PR #1987, opened 2016-02-18
Gce zone selector fix

duftler:
Cherry-pick's @zanthrash's commit from his pending PR and adds a couple of additional tweaks.
Would have pushed to the same branch/pr, but I don't have permissions for that fork.

### DIFF
--- a/app/scripts/modules/core/account/account.service.js
+++ b/app/scripts/modules/core/account/account.service.js
@@ -94,6 +94,42 @@ module.exports = angular.module('spinnaker.core.account.service', [
       return deferred.promise;
     });
 
+    let getUniqueRegionsForAllAccounts = _.memoize((provider) => {
+      return getRegionsKeyedByAccount(provider)
+        .then(function(regionsByAccount) {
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/account/account.service.js
```diff
@@ -94,6 +94,42 @@ module.exports = angular.module('spinnaker.core.account.service', [
...
+            .flatten()
+            .map(reg => reg.name)
+            .uniq()
+            .value();
+
+          return regions;
+        });
+    });
+
+    let getUniqueGceZonesForAllAccounts = _.memoize((provider) => {
```
duftler: @anotherchrisberry Saw your comment on the pending pr and agree. We should reconcile the differences between the different representations of regions/zones on an account. Took a quick look, and it looks doable. Going to be working through some of the zonal/regional bits on GCE starting next week or so. If you're ok leaving this method as-is until then, would like to tackle this at the same time.

zanthrash:
This looks good to me.  I'm going to let @anotherchrisberry look it over and do the merge.

PR #2039, opened 2016-02-29
Add 'Prefer AMI Block Device Mappings' control to Advanced Settings t…

duftler:
…ab of AWS wizard.

If set, the AMI block device mappings are used for the newly-provisioned server group.

Relates to https://github.com/spinnaker/clouddriver/pull/364.


duftler:
image

PR #2077, opened 2016-03-11
Fix CAP error that caused namespace collision between Kubernetes and …

duftler:
…Google resize server group controllers.


duftler:
@lwander please review.

PR #2128, opened 2016-03-24
Making pipeline stage restartable.

jcusa:


jcusa:
orca change is here: https://github.com/spinnaker/orca/pull/776

jcusa:
accidentally included extra file. will remove

PR #2131, opened 2016-03-25
Making pipeline stage restartable.

jcusa:


jcusa:
orca change in https://github.com/spinnaker/orca/pull/777

PR #2169, opened 2016-04-06
Use 'region' instead of 'zone' on Cloud Foundry server group operations.

duftler:


duftler:
@gregturn please review.

app/scripts/modules/core/pipeline/config/stages/destroyAsg/cf/cfDestroyAsgStage.js
```diff
@@ -21,7 +21,7 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.cf.destroyAsgStag
         },
         { type: 'requiredField', fieldName: 'cluster' },
         { type: 'requiredField', fieldName: 'target', },
-        { type: 'requiredField', fieldName: 'zones', },
+        { type: 'requiredField', fieldName: 'regions', },
```
gregturn: Shouldn't this just be "region" in the singular?

app/scripts/modules/core/pipeline/config/stages/destroyAsg/cf/cfDestroyAsgStage.js
```diff
@@ -32,7 +32,7 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.cf.destroyAsgStag
 
     $scope.state = {
       accounts: false,
-      zonesLoaded: false
+      regionsLoaded: false
```
gregturn: "regionLoaded"?

app/scripts/modules/core/pipeline/config/stages/destroyAsg/cf/cfDestroyAsgStage.js
```diff
@@ -42,13 +42,13 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.cf.destroyAsgStag
 
     ctrl.accountUpdated = function() {
       accountService.getAccountDetails(stage.credentials).then(function(details) {
-        stage.zones = [details.org];
+        stage.regions = [details.org];
```
gregturn: "region"?

gregturn:
In general, is region in the deck plural or singular? Other than that, LGTM.

duftler:
For those stages, the typical setup is that the user configures a set of regions, and then orca synthesized per-region tasks for each region: 
image

duftler:
So for each request, it is singular, but for pipelines stages that support the selection of multiple regions, it can be plural.

gregturn:
Okay.

PR #2173, opened 2016-04-06
Add Webhook trigger to UI: PR for code review

netwolfuk:
This PR is not ready for merge yet, it's a vehicle for discussion and to check I am on the right track.

The main point of discussion is probably that I added a  "category" to the Trigger.
This is because the EventListener in echo usually checks the "type" of the trigger.
The initial proposal by @tomaslin was for the webhook URL to contain the type and source, but makes it difficult to determine that the trigger should only fire for webhook triggers, not any trigger.

Currently the status is:
- Can choose Webhook from trigger list
- Can enter category and source attributes for trigger
- Need to add validation, example payload and dynamic URL generation
- Need to handle when the same category and source are created on same pipeline. Should probably prevent that.


ajordens:
Can we get this branch squashed and rebased.

Does this jive @tomaslin with your original thoughts?

netwolfuk:
Sure. I'll try to do that tonight.

netwolfuk:
I'd like to automatically add the "pipeline" name and the "application" into the default constraints created when the webhook trigger is added in Deck. What do I need to pass into the controller so that I can access the application and pipeline name? https://github.com/netwolfuk/deck/blob/9de75f067c45b35537781a97d2fcad46ee6c088b/app/scripts/modules/core/pipeline/config/triggers/webhook/webhookTrigger.module.js#L19

netwolfuk:
There are a few commits of mine, and a bunch of commits from others.
Do I squash everything into one picked commit, or leave other people's commits as picks and just squash my ones?

netwolfuk:
I have rebased and created a new branch and PR #2214 
I'll close this one.

PR #2204, opened 2016-04-21
Execution details view for the 'applySourceServerGroupCapacity' stage

ajordens:
- This stage is not executable by itself, rather it's synthetically injected after a deploy stage


ajordens:
image

This is related to spinnaker/orca#788 which has added a new synthetic `applySourceServerGroupCapacity` stage after a regular deploy stage.

The orca PR introduces some behavior where a server group will be deployed with `min = desired` capacity. Only after the deploy successfully completes will the `min` capacity be restored to match the source server group.

PR #2268, opened 2016-05-14
provider/google: First pass at support for regional server groups. Fe…

duftler:
…ature-flagged off for now.


duftler:
Only visible effect for now should be that the Zone selector is down past Capacity, instead of up by Region: 
image

duftler:
When the feature is enabled, there is a new checkbox: 
image

duftler:
And when the option is checked, it looks like so: 
image

duftler:
@lwander or @anotherchrisberry please review.

lwander:
Nice, LGTM

PR #2291, opened 2016-05-25
Support for <optional> pipeline stages

ajordens:
- an expression can be used to govern whether a stage should execute or be skipped

### DIFF
--- a/app/scripts/modules/core/delivery/executions/executions.less
+++ b/app/scripts/modules/core/delivery/executions/executions.less
@@ -85,4 +85,8 @@ executions {
     fill: @stage-running;
     background-color: @stage-running;
   }
+  &.execution-marker-skipped {
+    fill: @stage-succeeded;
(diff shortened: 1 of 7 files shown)


ajordens:
image

PTAL @zanthrash

app/scripts/modules/core/delivery/executions/executions.less
```diff
@@ -85,4 +85,8 @@ executions {
     fill: @stage-running;
     background-color: @stage-running;
   }
+  &.execution-marker-skipped {
+    fill: @stage-succeeded;
```
ajordens: Wasn't convinced we needed a different color to indicate skipped stages, thoughts?

zanthrash: Might be worth have a conversation with JT, but I'm okay with this for now.

PR #2292, opened 2016-05-26
Show 'SKIPPED' rather than 'UNKNOWN' as the Canary Status

ajordens:
- Fixed a lint error


ajordens:
image

Might be nicer to make skipped stages more distinct ... but this is better than showing UNKNOWN for canaries.

PR #2293, opened 2016-05-26
Azure default settings: LB port, storage, health

milismsft:
set default port settings for LB to 80 (HTTP)
change default storage for server group disk images to premium
bypass "platformHealthOnly" check when targeting azure


milismsft:
@spinnaker/netflix-reviewers @anotherchrisberry PTAL and merge if LGTY

PR #2298, opened 2016-05-31
provider/google: Prune zone list for regional server group from the r…

duftler:
…ight.


duftler:
@lwander or @ttomsu please review.

lwander:
LGTM

PR #2302, opened 2016-06-01
Azure related bug fixes

milismsft:
Fixed a bug in deploy SG stage: "basic settings" was improperly marked as dirty causing the final validation to fail unless it was re-selected.
Fixed couple minor issues in the create server group modal: LB and security group controllers were too long.
Fixed couple minor bug in load balancer details: typo in the DNS handling, delete LB should be disabled if server groups are still attached to the load balancer.


milismsft:
@spinnaker/netflix-reviewers @anotherchrisberry PTAL

PR #2306, opened 2016-06-02
docker trigger - added intermediate organization field to the trigger…

tomaslin:
…, sorted all the fields and changed trigger.registry to trigger.account

### DIFF
--- a/app/scripts/modules/core/pipeline/config/triggers/docker/dockerTrigger.html
+++ b/app/scripts/modules/core/pipeline/config/triggers/docker/dockerTrigger.html
@@ -1,13 +1,13 @@
 <ng-form name="dockerForm">
   <div class="form-group">
     <div class="col-md-3 sm-label-right">
-      Registry
+      Registry Name
     </div>
     <div class="col-md-6">
-      <ui-select ng-model="trigger.registry" class="form-control input-sm">
-        <ui-select-match>{{trigger.registry}}</ui-select-match>
-        <ui-select-choices repeat="registry in registries | filter: $select.search | orderBy">
-          <span ng-bind-html="registry | highlight: $select.search"></span>
+      <ui-select ng-model="trigger.account" class="form-control input-sm">
+        <ui-select-match>{{trigger.account}}</ui-select-match>
+        <ui-select-choices repeat="account in accounts | filter: $select.search | orderBy">
+          <span ng-bind-html="account | highlight: $select.search"></span>
         </ui-select-choices>
       </ui-select>
     </div>
@@ -19,6 +19,19 @@
               <span ng-class="{'glyphicon-spinning':viewState.imagesRefreshing}"
                     class="glyphicon glyphicon-refresh"></span>
       </a>
+    </div>
+  </div>
+  <div class="form-group">
+    <div class="col-md-3 sm-label-right">
+      Organization
+    </div>
+    <div class="col-md-6">
+      <ui-select ng-model="trigger.organization" class="form-control input-sm">
+        <ui-select-match>{{trigger.organization}}</ui-select-match>
+        <ui-select-choices repeat="organization in organizations | filter: $select.search | orderBy">
+          <span ng-bind-html="organization | highlight: $select.search"></span>
+        </ui-select-choices>
+      </ui-select>
     </div>
   </div>
   <div class="form-group">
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/pipeline/config/triggers/docker/dockerTrigger.html
```diff
@@ -3,11 +3,11 @@
     <div class="col-md-3 sm-label-right">
       Registry
     </div>
-    <div class="col-md-5">
-      <ui-select ng-model="trigger.registry" class="form-control input-sm">
```
tomaslin: replaced registry with account, will make corresponding changes to igor and echo downstream

tomaslin: rationale here is that the registry URL might change and it doesn't make sense to break all the pipelines because the URL changes. ( Change could be due to versioning, http vs https, domain name differences etc )

app/scripts/modules/core/pipeline/config/triggers/docker/dockerTrigger.module.js
```diff
@@ -60,28 +73,41 @@ module.exports = angular.module('spinnaker.core.pipeline.trigger.docker', [
...
+          let parts = image.repository.split('/');
+          let repository = parts.pop();
+          let org = parts.join('/');
+          if (all.indexOf(org) < 0) {
+            map[key] = all.concat(org);
+          }
+          return map;
+        }, {});
+        $scope.accounts = Object.keys($scope.accountMap);
+        $scope.organizationMap = images.reduce((map, image) => {
```
tomaslin: added intermediate organization field to sort by

app/scripts/modules/core/pipeline/config/triggers/docker/dockerTrigger.module.js
```diff
@@ -60,28 +73,41 @@ module.exports = angular.module('spinnaker.core.pipeline.trigger.docker', [
...
+          let repository = parts.pop();
+          let key = `${image.account}/${parts.join('/')}`;
+          let all = map[key] || [];
           if (all.indexOf(image.repository) < 0) {
             map[key] = all.concat(image.repository);
           }
           return map;
         }, {});
-        $scope.registries = Object.keys($scope.registryMap);
-        updateRepositoryList();
```
tomaslin: didn't see reason to call update repository and then update tags if update repository calls update tags, moved it all to updateOrganizationsList() in line 110 of new commit

tomaslin:
screen shot 2016-06-02 at 1 44 24 pm

app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
```diff
@@ -168,7 +168,8 @@ module.exports = angular.module('spinnaker.kubernetes.clusterCommandBuilder.serv
         return {
           fromTrigger: true,
           repository: trigger.repository,
-          registry: trigger.registry,
+          account: trigger.account,
```
tomaslin: wasn't too clear where this is being used. would imagine the image selection in kubernetes would also need to be updated to have organizational filtering

anotherchrisberry:
LGTM

tomaslin:
also changed registry to 'registry name' and repository to 'image'

tomaslin:
DO NOT MERGE - still need igor and echo changes

tomaslin:
@lwander PTAL

app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
lwander: Could we keep this for now? It's just used to render where the image is coming from and will break the UI. I'll switch it to account, and when I do I'll remove this `registry:` line. You can drop a TODO with my name in it here.

app/scripts/modules/core/pipeline/config/triggers/docker/dockerTrigger.module.js
lwander: Good call

lwander:
LGTM afeter comment on `registry` in image description.

app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
```diff
@@ -168,6 +168,8 @@ module.exports = angular.module('spinnaker.kubernetes.clusterCommandBuilder.serv
         return {
           fromTrigger: true,
           repository: trigger.repository,
+          account: trigger.account,
+          organization: trigger.organization,
           registry: trigger.registry,
```
tomaslin: restored registry here

PR #2315, opened 2016-06-06
Load only first N Docker images in Deck for Kubernetes provider

danielpeach:
n-docker-images

Load only first N Docker images in Deck for Kubernetes provider. @lwander PTAL


app/scripts/modules/kubernetes/image/image.reader.js
```diff
@@ -4,7 +4,7 @@ let angular = require('angular');
 
 module.exports = angular.module('spinnaker.kubernetes.image.reader', [])
   .factory('kubernetesImageReader', function ($q, Restangular) {
-    function findImages(params) {
+    function findImages(params) {
```
lwander: I think the linter may yell about this extra space.

lwander:
Very nice, LGTM after comment

danielpeach:
@lwander fixed & squashed

PR #2336, opened 2016-06-14
added instance health status to load balancer ui

danielpeach:
- added instance health status to load balancer ui
kubernetes-loadbalancer-health

@lwander PTAL


lwander:
Nice! LGTM

PR #2349, opened 2016-06-21
run job stage fix: provider choice

danielpeach:
Run job stage fix: no longer asks for provider choice unnecessarily when adding jobs.

@lwander PTAL (this was more straightforward than I thought).


lwander:
Huh, nice! LGTM

PR #2355, opened 2016-06-22
docker trigger - remove tag dropdown from definition and make it into a input that takes in a regex

tomaslin:


tomaslin:
@lwander PTAL

tomaslin:
tested the config without tag and triggering still works with latest echo

lwander:
Does it still work with regex?

tomaslin:
@lwander tag is now in as an input box that will default to null if the string is empty. Seems to fix the triggering.

app/scripts/modules/core/pipeline/config/triggers/docker/dockerTrigger.module.js
```diff
@@ -53,32 +52,15 @@ module.exports = angular.module('spinnaker.core.pipeline.trigger.docker', [
...
         trigger.repository = null;
       }
-      updateTagList();
+      updateTag();
     }
 
-    function updateTagList() {
-      if (!$scope.repositoryMap) {
-        return;
+    function updateTag(){
```
lwander: Missing space before `{`

app/scripts/modules/core/pipeline/config/triggers/docker/dockerTrigger.module.js
```diff
@@ -53,32 +52,15 @@ module.exports = angular.module('spinnaker.core.pipeline.trigger.docker', [
...
+      if(trigger.tag !== null && trigger.tag.trim() === ''){
```
lwander: See above comment

lwander:
LGTM, verified locally too.

PR #2357, opened 2016-06-22
(gce) fixes bug that prevented clone wizard instance type population

danielpeach:
(gce) fixes bug that prevented clone wizard instance type population

@duftler PTAL

This line was supposed to clear the instance type when the custom fields are malformed (e.g. when you pick a zone that doesn't support the vCPU count), but the rest of the function handles this logic properly (and doesn't cause the bug).


duftler:
LGTM.

PR #2362, opened 2016-06-23
Revert "Revert "provider/kubernetes: Fix docker trigger""

lwander:
Reverts spinnaker/deck#2360

@duftler (don't make fun of me)


duftler:
LGTM.

PR #2363, opened 2016-06-23
(gce) prevent invalid selection of instance types for location

danielpeach:
(gce) prevent invalid selection of instance types for location

"getAvailableTypesForRegions" is kept as an alias of "getAvailableTypesForLocations" because it is one of the core functions expected of a provider.

@duftler PTAL


duftler:
LGTM.

PR #2368, opened 2016-06-23
(gce) fixed baseIpAddress in instance details controller

danielpeach:
@duftler PTAL. I tested this on top of chris's PR (#2364) and it fixed the problem.


duftler:
LGTM.

PR #2373, opened 2016-06-24
(kubernetes) instance refresh fix

danielpeach:
(kubernetes) instance refresh fix

This was preventing instances from being updated. 

It was also the source of the "$scope.$on is not a function" error that Steven reported.

@lwander PTAL


lwander:
Awesome, LGTM

PR #2382, opened 2016-06-27
(k8s) fixes job termination and console error

danielpeach:
(k8s) fixes job termination and console error

@lwander PTAL


lwander:
LGTM

PR #2425, opened 2016-07-08
(core) increased notifier size

danielpeach:
@anotherchrisberry 
A user felt that the logged out notification was way too discreet. 

notifier


anotherchrisberry:
What if we made the font a little bigger, too? Maybe 120%?

danielpeach:
Sure -- sounds good to me.
screen shot 2016-07-10 at 3 32 13 pm

anotherchrisberry:
Thanks!

PR #2429, opened 2016-07-10
(k8s) http header support in wizard

danielpeach:
(k8s) http header support in wizard. Headers are copied properly for cloning and pipelines.

@lwander PTAL

screen shot 2016-07-10 at 6 50 32 pm


lwander:
LGTM. Have you verified that a pod with a probe configured to use these
headers loads them correctly during a clone?

danielpeach:
Yes -- both pod and rc YAMLs have the headers after a clone.

PR #2440, opened 2016-07-13
provider/google: Now that we do a disable before destroy, we need to …

duftler:
…ask for health provider in confirmation dialog.


duftler:
@danielpeach please review.

danielpeach:
LGTM

PR #2441, opened 2016-07-13
Fixed up SVG Health and LB Status detail reporting.

jtk54:
Fixed up SVG Health and LB Status detail reporting. 'Status' was not reported properly in certain situations, and accessing the instance health states looks different for SVGs and LBs, so I split them up. Additionally, only 'out of service' instance statuses were reported for LBs. @danielpeach @anotherchrisberry PTAL.


jtk54:
I'll close this and sync with Dan (who actually knows what he's doing) next week when he's back. This is apparently only an issue with representing Google L7 LBs in the status details, L3s are fine.

PR #2454, opened 2016-07-18
(k8s) copy template fix

danielpeach:
(k8s) copy template fix
@lwander PTAL


app/scripts/modules/kubernetes/serverGroup/configure/configuration.service.js
```diff
@@ -42,6 +56,10 @@ module.exports = angular.module('spinnaker.serverGroup.configure.kubernetes.conf
       return `*${query}*`;
     }
 
+    function grabImageAndTag(imageId) {
+      return imageId.split('/')[2];
```
lwander: Valid image ID's include
`gcr.io/image:tag` <- that will fail here.
`gcr.io/org/image:tag`

Instead I recommend splitting by `'/'`, and taking the last item from the list.

lwander:
Good catch on this. This should close https://github.com/spinnaker/spinnaker/issues/978

lwander:
LGTM

PR #2464, opened 2016-07-21
(gce) ad hoc add, edit, delete autoscaler

danielpeach:
(gce) ad hoc add, edit, delete autoscaler operations

@duftler PTAL


duftler:
Shouldn't the Update button be disabled when the "At least one metric is required." error msg is showing?
image

duftler:
I tested Adding/Editing/Deleting Autoscalers for both zonal and regional server groups. All worked as expected.
This came out really nice.
LGTM after addressing the one comment.

duftler:
Wondering if it would be more user-friendly to expect the various target metrics in whole numbers from 1-99 as opposed to decimals from 0.1-0.99. We do render them in the server group details as percentages.

danielpeach:
Addressed your comments.

PR #2468, opened 2016-07-22
core: added a section to the application config tab to serialize the application infrastructure

nwwebb:
serialize_example

Added a section to the config tab to serialize the application. Feature flag for serialize must be set to true in settings.js in order to view the section. PTAL @lwander @duftler @danielpeach

### DIFF
--- a/app/scripts/modules/core/application/config/serializeApplicationSection.component.js
+++ b/app/scripts/modules/core/application/config/serializeApplicationSection.component.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const angular = require('angular');
+
+module.exports = angular
+  .module('spinnaker.core.application.config.serialize.component', [
+    require('angular-ui-router'),
+    require('./../../serialize/serialize.write.service.js'),
+    require('../../confirmationModal/confirmationModal.service.js'),
+    require('../../overrideRegistry/override.registry.js'),
+  ])
+  .component('serializeApplicationSection', {
+    templateUrl: require('./serializeApplicationSection.component.html'),
+    bindings: {
+      application: '=',
+    },
+    controller: function ($state, serializeWriter, confirmationModalService) {
+      if (this.application.notFound) {
+        return;
+      }
+
+      this.serializeApplication = () => {
+
+        var submitMethod = () => {
+          return serializeWriter.serializeApplication(this.application.attributes);
+        };
+
+        var taskMonitor = {
+          application: this.application,
+          title: 'Serializing ' + this.application.name,
+          hasKatoTask: true,
+          onTaskComplete: () => {
+            $state.go('home.infrastructure');
+          }
+        };
+
+        confirmationModalService.confirm({
+          header: 'Are you sure you want to serialize: ' + this.application.name + '?',
+          buttonText: 'Serialize',
+          provider: 'gce',
+          taskMonitorConfig: taskMonitor,
+          submitMethod: submitMethod
+        });
+      };
+    }
+  });
--- a/app/scripts/modules/core/serialize/serialize.write.service.js
+++ b/app/scripts/modules/core/serialize/serialize.write.service.js
@@ -1,0 +1,51 @@
+'use strict';
+
+let angular = require('angular');
+
+module.exports = angular
+  .module('spinnaker.serialize.write.service', [
+    require('../task/taskExecutor.js'),
+    require('../utils/lodash.js'),
+    require('../account/account.service.js')
+  ])
+  .factory('serializeWriter', function($q, taskExecutor, accountService, _) {
+
+    let serializationProviders = ['gce'];
+
+    function buildJobs(app, accountDetails) {
+      let jobs = [];
+      accountDetails.forEach((accountDetail) => {
+        if (serializationProviders.includes(accountDetail.cloudProvider)) {
+          jobs.push({
+            type: 'serializeApplication',
+            credentials: accountDetail.name,
+            applicationName: app.name,
+            cloudProvider: accountDetail.cloudProvider,
+          });
+        }
+      });
+      return jobs;
+    }
+
+    function loadAccountDetails(app) {
+      let accounts = _.isString(app.accounts) ? app.accounts.split(',') : [];
+      let accountDetailPromises = accounts.map(account => accountService.getAccountDetails(account));
+      return $q.all(accountDetailPromises);
+    }
+
+    function serializeApplication(app) {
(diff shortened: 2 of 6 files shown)


app/scripts/modules/core/application/config/serializeApplicationSection.directive.html
```diff
@@ -0,0 +1,18 @@
+<div class="row">
+  <div class="col-md-7">
+    <h4>Serialize Application</h4>
+    <div class="panel panel-default">
+      <div class="panel-body">
+        <p>
+          Serializing the application will save the current state of the infrastructure using Terraform in a bucket. <br>
```
lwander: Maybe word to "... of the infrastructure using [Terraform's config language](https://www.terraform.io/docs/configuration/index.html). in a bucket." (making sure to include that link).

duftler:
I like this approach.

app/scripts/modules/core/application/config/serializeApplicationSection.directive.html
duftler: For some reason, "...Terraform in a bucket." reads funny to me. Maybe we could reword the sentence a bit?

app/scripts/modules/core/application/config/serializeApplicationSection.directive.js
```diff
@@ -0,0 +1,52 @@
...
+const angular = require('angular');
+
+module.exports = angular
+  .module('spinnaker.core.application.config.serialize.directive', [
+    require('angular-ui-router'),
+    require('./../../serialize/serialize.write.service.js'),
+    require('../../confirmationModal/confirmationModal.service.js'),
+    require('../../overrideRegistry/override.registry.js'),
+  ])
+  .directive('serializeApplicationSection', function (overrideRegistry) {
```
duftler: I think we're using components instead of directives these days. Please check with @danielpeach to get some direction on the right way to go here.

settings.js
```diff
@@ -97,5 +97,6 @@ window.spinnakerSettings = {
     // whether stages affecting infrastructure (like "Create Load Balancer") should be enabled or not
     infrastructureStages: process.env.INFRA_STAGES === 'enabled',
     jobs: false,
+    serialize: false,
```
duftler: Feels like `serialization` might read better here. Or not. Now I've repeated both words so many times I don't think I have an opinion anymore.

app/scripts/modules/core/application/config/serializeApplicationSection.directive.html
```diff
@@ -0,0 +1,18 @@
...
+          Currently only GCE and GCS are supported.
+        </p>
+
+        <button class="btn btn-link" ng-click="vm.serializeApplication()"><span
+          class="glyphicon glyphicon-cloud-upload"></span> Serialize Application
```
duftler: If we use the 'cloud-upload' icon for serialization, what will we use for restore/converge?

app/scripts/modules/core/application/config/serializeApplicationSection.directive.html
duftler: Or:

Serializing the application will save the current state of the
infrastructure to a storage bucket using Terraform's config language. (with
the link)

app/scripts/modules/core/application/config/serializeApplicationSection.directive.html
lwander: cloud-download?

duftler: heh, what I was getting at was that it feels flipped...

But if you both think it's better this way, I'm good with it.

lwander: Ah I see what you mean... it's kind of funny that we're moving the definition from one cloud to another...

lwander: Let's leave it up to @nwwebb

app/scripts/modules/core/application/config/serializeApplicationSection.directive.js
danielpeach: Components are new, so the code base is inconsistent, but use components instead of directives unless you need them (e.g. for raw DOM manipulation).

app/scripts/modules/core/application/config/serializeApplicationSection.directive.js
```diff
@@ -0,0 +1,52 @@
...
+    return {
+      restrict: 'E',
+      templateUrl: overrideRegistry.getTemplate('serializeApplicationSectionDirective', require('./serializeApplicationSection.directive.html')),
```
danielpeach: This will probably confuse people, since it implies that there's an override somewhere. You can just `require` your template.

app/scripts/modules/core/serialize/serialize.write.service.js
```diff
@@ -0,0 +1,54 @@
...
+          });
+        }
+      });
+      return jobs;
+    }
+
+    function loadAccountDetails(app) {
+      let accountDetailPromises = [];
+      let accounts = _.isString(app.accounts) ? app.accounts.split(',') : [];
+      accounts.forEach((account) => {
```
danielpeach: `.map` would be cleaner here:
`let accountDetailPromises = accounts.map(account => accountService.getAccountDetails(account));`

app/scripts/modules/core/application/config/serializeApplicationSection.component.js
```diff
@@ -0,0 +1,49 @@
...
+
+module.exports = angular
+  .module('spinnaker.core.application.config.serialize.component', [
+    require('angular-ui-router'),
+    require('./../../serialize/serialize.write.service.js'),
+    require('../../confirmationModal/confirmationModal.service.js'),
+    require('../../overrideRegistry/override.registry.js'),
+  ])
+  .component('serializeApplicationSection', {
+    restrict: 'E',
```
danielpeach: For the component config, you can drop `restrict`, `scope`, and `controllerAs` (by default, it will use `$ctrl` in the template).

danielpeach:
LGTM after comment

nwwebb:
is this ready to merge?

duftler:
LGTM. @anotherchrisberry or @zanthrash any objections to merging?

anotherchrisberry:
LGTM - just needs to be rebased

nwwebb:
rebased

PR #2470, opened 2016-07-25
(k8s) moved pipeline stages to k8s module

danielpeach:
@lwander please review.

I tested the refactored stages (for both configuration and execution) and they worked as expected.


lwander:
Nice, LGTM

PR #2495, opened 2016-07-29
(gce) fix host and path rules modal

danielpeach:
Fixes host and path rules modal.
Allows L7 load balancer details panel to open from the clusters tab. (The approach here isn't great, but the core pieces assume that serverGroup.region === loadBalancer.region. I'll wait until the rest of the L7 pieces are in place before coming up with a better approach.)

@duftler @jtk54 please review.


duftler:
LGTM.

jtk54:
LG2M2

PR #2498, opened 2016-07-29
(gce) load balancing policy

danielpeach:
@jtk54 this is a first pass. Please test and give me feedback.


jtk54:
Can we make the loadbalancing policy required if an L7 is selected? Otherwise LGTM.

PR #2531, opened 2016-08-08
(k8s) fix job clone wizard

danielpeach:
Fixes the job clone wizard (bug surfaces when trying to edit a job in a pipeline). I made a similar fix in the server group wizard controller but forgot to update jobs.

Also updated the markComplete calls, since the job wizard does not have a replicas page.

@lwander PTAL


lwander:
LGTM, thanks!

PR #2534, opened 2016-08-08
Find Image from Tags

ajordens:
- Similar to `Find Image from Cluster` but supports scenarios where an image is not currently deployed (but is tagged)
- Relates to the `Tag Image` stage

### DIFF
--- a/app/scripts/modules/core/pipeline/config/stages/findImageFromTags/findImageFromTagsStage.html
+++ b/app/scripts/modules/core/pipeline/config/stages/findImageFromTags/findImageFromTagsStage.html
@@ -1,0 +1,12 @@
+<div class="form-horizontal">
+  <stage-config-field label="Package">
+    <input type="text" class="form-control input-sm"
+           ng-model="stage.packageName"/>
+  </stage-config-field>
+  <stage-config-field label="Regions">
+    <checklist items="regions" model="stage.regions" inline="true" include-select-all-button="true"></checklist>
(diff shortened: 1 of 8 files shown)


ajordens:
image

---

image

ajordens:
Need to add some validation constraints (ie. regions/packageName/tags are required).

app/scripts/modules/core/pipeline/config/stages/findImageFromTags/findImageFromTagsStage.html
```diff
@@ -0,0 +1,12 @@
+<div class="form-horizontal">
+  <stage-config-field label="Package" help-key="pipeline.config.findImageFromTags.packageName">
```
anotherchrisberry: No need for the `help-key` if there's no help contents backing it.

ajordens: Ahh yes but there should be (and will be momentarily).

anotherchrisberry:
LGTM

app/scripts/modules/core/pipeline/config/stages/findImageFromTags/findImageFromTagsStage.html
ajordens: Actually opt'd for no help-key ... will revise if it proves confusing to people.

PR #2538, opened 2016-08-09
(gce) more l7 fixes

danielpeach:
@jtk54 please review

- makes backend service, port name mapping, and capacity metrics required in the server group wizard.

- disables deregistering from load balancer if server group is attached to an l7 load balancer.


app/scripts/modules/google/instance/details/instance.details.controller.js
```diff
@@ -215,7 +217,17 @@ module.exports = angular.module('spinnaker.instance.detail.gce.controller', [
 
     this.canDeregisterFromLoadBalancer = function() {
       var instance = $scope.instance;
-      if (!instance.loadBalancers || !instance.loadBalancers.length) {
+      // TODO(dpeach): remove when it's possible to degister from l7 load balancer.
```
jtk54: Instance deregister won't be available until we can add individual instances to MIGs. So not in the near future.

jtk54:
LGTM after rebase.

PR #2540, opened 2016-08-09
(gce) refactored autoscaler components

danielpeach:
@duftler 
Refactored autoscaling work into smaller components in preparation for including autoscaler configuration in server group wizard. The behavior in Deck should be the same (and I've tested that it is).


duftler:
LGTM (needs a rebase).

PR #2551, opened 2016-08-10
(gce) update server group wizard hidden metadata

danielpeach:
(gce) update server group wizard hidden metadata

@jtk54 please review


jtk54:
LGTM

PR #2552, opened 2016-08-10
(gce) forces user to pick zone before instance type

danielpeach:
(gce) forces user to pick zone before instance type
also fixes a load balancer refresh bug.

@duftler PTAL


duftler:
LGTM.

PR #2553, opened 2016-08-10
Support tagging of multiple images

ajordens:
image

---

image


ajordens:
This coincides with the support I'm about to PR in `orca` around being able to tag multiple images in one shot.

PTAL @anotherchrisberry

PR #2578, opened 2016-08-15
(gce) update clone server group wizard for server groups with autoscaling policies

danielpeach:
(gce) update clone server group wizard for server groups with autoscaling policies

@duftler please review


duftler:
Nice. LGTM.

PR #2581, opened 2016-08-16
(gce) health check for each backend in load balancer details

danielpeach:
*(gce) health check for each backend in load balancer details
*(gce) server group details does not include non-existent loadBalancingPolicy

@jtk54 please review
backend_service_healthcheck


jtk54:
LGTM

PR #2582, opened 2016-08-16
(core) added retry service, updated docker image reader with retries

danielpeach:
@lwander please review.

### DIFF
--- a/app/scripts/modules/core/retry/retry.service.js
+++ b/app/scripts/modules/core/retry/retry.service.js
@@ -1,0 +1,27 @@
+'use strict';
+
+let angular = require('angular');
+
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/retry/retry.service.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+let angular = require('angular');
+
+module.exports = angular.module('spinnaker.deck.core.retry.service', [])
+  .factory('retryService', function ($q, $timeout) {
+    function buildRetrySequence (fn, retryConditionFn, retryCount, interval) {
```
lwander: What is the time unit of the interval? This should be documented here

lwander:
LGTM, thanks @danielpeach

PR #2584, opened 2016-08-17
guard against undefined while loading docker image

jcusa:
noticed this issue when I was using docker registry as trigger. The docker 'organizations' were showing fine, but the 'image' drop down showed nothing. It was because the returned list contains {} elements. This fixes the issue.


danielpeach:
LGTM (needs a rebase).

danielpeach:
Thanks @jcusa

PR #2588, opened 2016-08-17
SpEL autocomplete

zanthrash:
Allows for context specific autocomplete for the SpEL in textareas and text inputs

At this time it shows a list of the functions, helper params, and stage names.
Also uses the last execution and walks the json and builds out the paths to the leafs for autocomplete.  
If no execution is found it will use the pipeline config to build the paths to the leafs.

There is context aware autocomplete around the #stage, #judgement helper functions as well as the 'parameters' and 'triggers' helper params.

NOTE:  I have plans to do some refactoring to remove some of the boilerplate.

@anotherchrisberry PTAL

### DIFF
--- a/app/scripts/modules/core/delivery/service/execution.service.js
+++ b/app/scripts/modules/core/delivery/service/execution.service.js
@@ -260,6 +260,18 @@ module.exports = angular.module('spinnaker.core.delivery.executions.service', [
       }
     }
 
+    function getLastExecutionForApplicationByConfigId(appName, configId) {
+      return getFilteredExecutions(appName, {}, 1 )
+        .then((executions) => {
+          return executions.filter((execution) => {
+            return execution.pipelineConfigId === configId;
+          });
+        })
+        .then((executionsByConfigId) => {
+          return executionsByConfigId[0];
+        });
+    }
+
     return {
       getExecutions: getExecutions,
       getExecution: getExecution,
@@ -276,5 +288,6 @@ module.exports = angular.module('spinnaker.core.delivery.executions.service', [
       getProjectExecutions: getProjectExecutions,
       addExecutionsToApplication: addExecutionsToApplication,
       updateExecution: updateExecution,
+      getLastExecutionForApplicationByConfigId: getLastExecutionForApplicationByConfigId,
     };
   });
--- a/app/scripts/modules/core/widgets/spelText/jsonListBuilder.spec.js
+++ b/app/scripts/modules/core/widgets/spelText/jsonListBuilder.spec.js
@@ -1,0 +1,113 @@
+'use strict';
+
+describe('jsonListBuilder', function () {
+
+  let builder;
+
+  beforeEach(
+    window.module(
+      require('./jsonListBuilder')
+    )
+  );
+
+  beforeEach(
+    window.inject(function(jsonListBuilder) {
+      builder = jsonListBuilder;
+    })
+  );
+
+  describe('create a list with only the leaf keys', function () {
+    it('a simple json with one attribute', function () {
+      let json = { name: 'foo'};
+      let result = builder.convertJsonKeysToBracketedList(json);
+      expect(result).toEqual([`['name']`]);
+    });
+
+
+    it('a simple json with multiple attributes', function () {
+      let json = { name: 'foo', bar: 'baz'};
+      let result = builder.convertJsonKeysToBracketedList(json);
+      expect(result).toEqual([`['name']`, `['bar']`]);
+    });
+
+    it('nested objects', function () {
+      let json = { name: {foo: 'bar'}};
+      let result = builder.convertJsonKeysToBracketedList(json);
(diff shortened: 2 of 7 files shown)


settings.js
```diff
@@ -94,7 +94,7 @@ window.spinnakerSettings = {
     vpcMigrator: true,
     clusterDiff: true,
     roscoMode: false,
-    netflixMode: false,
+    netflixMode: true,
```
tomaslin: you probably want to revert ths lne

app/scripts/modules/core/delivery/service/execution.service.js
```diff
@@ -260,6 +260,18 @@ module.exports = angular.module('spinnaker.core.delivery.executions.service', [
       }
     }
 
+    function getLastExecution(appName, configId) {
+      return getExecutions(appName)
```
anotherchrisberry: If you change this to `getFilteredExecutions(appName, {}, 1)`, it should be quicker. Also probably just want to return the first item or null - the name implies that's what would happen.

Otherwise, might want to rename the method to `getExecutionsForConfigId` or something, since it's returning all executions (or whatever the `ExecutionFilterModel.sortFilter.count` is set to when the call happens).

app/scripts/modules/core/delivery/service/execution.service.js
```diff
@@ -260,6 +260,18 @@ module.exports = angular.module('spinnaker.core.delivery.executions.service', [
...
+        .then((executions) => {
+          return executions.filter((execution) => {
+            return execution.pipelineConfigId === configId;
+          });
+        })
+        .then((executionsByConfigId) => {
+          return executionsByConfigId;
+        });
```
anotherchrisberry: Doesn't look like this `.then` is doing anything, but could be tweaked to return just the first item (or null).

app/scripts/modules/core/widgets/spelText/jsonListBuilder.spec.js
```diff
@@ -0,0 +1,106 @@
...
+      let json = { name: [{foo: 2}] };
+      let result = builder.convertJsonKeysToBracketedList(json);
+      expect(result).toEqual([`['name'][0]['foo']`]);
+    });
+
+    it('json with an array of two object', function () {
+      let json = { name: [{foo: 2}, {foo:3}] };
+      let result = builder.convertJsonKeysToBracketedList(json);
+      expect(result).toEqual([`['name'][1]['foo']`, `['name'][0]['foo']`]);
+    });
```
anotherchrisberry: Might be good to include a test verifying that we're handling keys with dots, e.g. `notification.type` or `deploy.server.groups`

app/scripts/modules/core/widgets/spelText/spelAutocomplete.service.js
```diff
@@ -0,0 +1,327 @@
...
+
+module.exports = angular
+  .module('spinnaker.core.widget.spelAutocomplete', [
+    require('./jsonListBuilder'),
+    require('../../delivery/service/execution.service'),
+  ])
+  .factory('spelAutocomplete', function($q, jsonListBuilder, executionService) {
+    let brackets = [{open: '(', close: ')'}, {open: '[', close: ']'}, ];
+    let quotes = [{open: '\'', close: '\'' }, {open: '\"', close: '\"' }];
+    let helperFunctions = ['alphanumerical', 'fromJson', 'fromUrl', 'jsonFromUrl', 'judgement', 'stage', 'toBoolean', 'toFloat', 'toInt', 'toJson'];
```
anotherchrisberry: I think `judgement` is `judgment` in Orca.

anotherchrisberry:
Really nice.

ajordens:
fromJson -> readJson (just found this out today)

ajordens:
Super excited about this!

app/scripts/modules/core/widgets/spelText/spelAutocomplete.service.js
zanthrash: Bokay

app/scripts/modules/core/delivery/service/execution.service.js
zanthrash: Nice didn't see that one.  Thanks

app/scripts/modules/core/widgets/spelText/spelAutocomplete.service.js
```diff
@@ -0,0 +1,327 @@
...
+    let helperParams = ['execution', 'parameters', 'trigger', 'scmInfo', 'scmInfo.sha1', 'scmInfo.branch', 'deployedServerGroups'];
```
tomaslin: can we restrict scmInfo to only Jenkins triggers? doesn't work with anything else, really

duftler:
Very cool @zanthrash .

gregturn:
👍 !!!

PR #2593, opened 2016-08-18
(kubernetes) Only strip underscores from names

lwander:
@danielpeach @leelasharma

Turns out the only invalid character in a docker repo name is the underscore: https://github.com/docker/docker/blob/f63cdf0260cf6287d28a589a79d3f947def6a569/runtime.go#L33 - so we can keep the dash.


danielpeach:
LGTM

PR #2598, opened 2016-08-19
(core) update versions and file cleanup

icfantv:
* clean up trailing commas in webpack config file
* update angular to latest
* update babel to latest


icfantv:
@spinnaker/reviewers PTAL.

the angular upgrade is innocuous as 1.5.8 is the version being pulled down via `npm install`.  the babel upgrade is fairly significant but doesn't appear to have any negative side effects.  we can also use the `babel-plugin-transform-strict-mode` plugin to allow us to not have to use `use strict` at the top of every JS file...but....baby steps.

zanthrash:
@icfantv LGTM

PR #2608, opened 2016-08-23
provider/google: Cache http health checks.

duftler:
provider/google: Add support for configuring AutoHealering during create/clone.

### DIFF
--- a/app/scripts/modules/google/serverGroup/configure/serverGroupConfiguration.service.js
+++ b/app/scripts/modules/google/serverGroup/configure/serverGroupConfiguration.service.js
@@ -13,11 +13,13 @@ module.exports = angular.module('spinnaker.serverGroup.configure.gce.configurati
   require('../../instance/gceInstanceType.service.js'),
   require('./../../instance/custom/customInstanceBuilder.gce.service.js'),
   require('../../loadBalancer/elSevenUtils.service.js'),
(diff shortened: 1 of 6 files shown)


duftler:
@danielpeach please review.

app/scripts/modules/google/serverGroup/configure/serverGroupConfiguration.service.js
```diff
@@ -263,6 +269,29 @@ module.exports = angular.module('spinnaker.serverGroup.configure.gce.configurati
...
+      if (command.credentials === null) {
+        return result;
+      }
+
+      filteredData.httpHealthChecks = _(command.backingData.httpHealthChecks[0].results)
+        .filter({provider: 'gce', account: command.credentials})
+        .pluck('name')
+        .valueOf();
+
+      if (_.has(command, 'autoHealingPolicy.healthCheck') && !_(filteredData.httpHealthChecks).contains(command.autoHealingPolicy.healthCheck)) {
```
danielpeach: I think `contains` checks for equality only by reference.

danielpeach:
LGTM after comment

app/scripts/modules/google/serverGroup/configure/serverGroupConfiguration.service.js
duftler: `healthCheck` is a string, so this should be fine.

PR #2615, opened 2016-08-24
Avoid setting a default `interestingHealthProviderNames` (even if pla…

ajordens:
…tformHealthOnly is true)

Rather, `interestingHealthProviderNames` (unless overridden) should be determined when the
task is actually executed in orca.

Protects against a situation where someone:
1) Sets `platformHealthOnly: true`
2) Creates a pipeline
3) Sets `platformHealthOnly: false` at some point in the future (think days/weeks)
4) Runs pipeline

Currently, the pipeline will snapshot `interestingHealthProviderNames` at the point in time that
a stage is created.

### DIFF
--- a/app/scripts/modules/amazon/instance/details/instance.details.controller.js
+++ b/app/scripts/modules/amazon/instance/details/instance.details.controller.js
@@ -255,9 +255,6 @@ module.exports = angular.module('spinnaker.instance.detail.aws.controller', [
       };
 
       var submitMethod = (params = {}) => {
-        if (app.attributes && app.attributes.platformHealthOnly) {
-          params.interestingHealthProviderNames = ['Amazon'];
-        }
         return instanceWriter.rebootInstance(instance, app, params);
       };
 
--- a/app/scripts/modules/amazon/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
+++ b/app/scripts/modules/amazon/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
@@ -41,10 +41,6 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.aws.cloneServerGr
     stage.cloudProvider = 'aws';
     stage.cloudProviderType = 'aws';
 
-    if (stage.isNew && $scope.application.attributes.platformHealthOnly) {
-      stage.interestingHealthProviderNames = ['Amazon'];
-    }
-
     if (!stage.credentials && $scope.application.defaultCredentials.aws) {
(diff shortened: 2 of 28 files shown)


ajordens:
Putting this out there for feedback ... PTAL @anotherchrisberry, @lwander, @tomaslin, @duftler.

I have a related change for orca that injects a new `DetermineHealthProvidersTask` into the relevant stages (create, clone, resize, enable, disable, shrink, rollback, reboot, etc.).

The current behavior has some rather unfortunate implications as the pipeline will proceed without waiting for healthiness even tho the application was set to `platformHealthOnly: false`.

lwander:
This change would break the Kubernetes stages you changed without the `DetermineHealthProvidersTask`. Also, it might not make sense for Kubernetes since the only source of health is "platform" health, but that might be worth revisiting in the future.

ajordens:
@lwander As far as k8s goes, I only removed the `interestingHealthProviderNames` from places where it was being set based on `platformHealthOnly`. This would be replaced by the `DetermineHealthProvidersTask` in the orca stages. Are you thinking something else would break?

As for it only supporting platform-health right now, you probably want to always be setting `interestingHealthProviderNames: ["KubernetesPod"]`?

It also has the side-effect of making third-party API calls behave similar to an in-app UI operation (_right now we typically have to remind people to include the `interestingHealthProviderNames` attribute in their payloads_).

lwander:
Whoops, that's my bad for setting it based on `platformHealthOnly` for that stage alone. Some stages also set combination of `"KubernetesContainer"` and `"KubernetesService"` (without consulting `platformHealthOnly`) which are separate "health" metrics for individual container status and network status respectively. Splitting the health reports into these three categories has made it easiest to make stage intuitively succeed or fail based on the intended behavior, and I'm worried that exposing this to the user adds needless complexity for this provider. So, if all of this can be moved to an Orca task that's for the best.

duftler:
This change sounds ok to me.
Just to confirm, both 'Instance Health' options would remain on the Create/Edit Application dialog and the 'Consider only $platform?' checkbox would still be available on each relevant stage/ad-hoc dialog, right?

The net change is that the `interestingHealthProviderNames` attribute will be injected into the context by orca instead of being included by the caller.

ajordens:
@lwander Is what you're saying that depending on the particular operation, you might want KubernetesService/KubernetesContainer/KubernetesPod (or some combination thereof)?

If that's the case, I might revert the kubernetes changes in this PR as it's a little beyond where I wanted to go.

lwander:
Yup that's right.

PR #2638, opened 2016-08-29
(core) update pipeline trigger timestamp

icfantv:
* alter displayed timestamp to be human readable
* add timestamp tooltip to display real (original) timestamp.


icfantv:
rebased.  @spinnaker/reviewers PTAL

nit from NF internal re displaying timestamps.

icfantv:
thanks @danielpeach

PR #2640, opened 2016-08-30
(k8s) run container as run job stage

danielpeach:
@lwander please review
runjob_configuration
executiondetails
container_search


lwander:
LGTM, worked when I ran it locally. 

FYI @leelasharma

PR #2646, opened 2016-09-01
Support rudimentary templating in default IAM roles

ajordens:
- Currently only handles {[application}}

### DIFF
--- a/app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.service.js
+++ b/app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.service.js
@@ -34,6 +34,9 @@ module.exports = angular.module('spinnaker.aws.serverGroupCommandBuilder.service
           var keyPair = credentials ? credentials.defaultKeyPair : null;
           var applicationAwsSettings = _.get(application, 'attributes.providerSettings.aws', {});
 
+          var defaultIamRole = settings.providers.aws.defaults.iamRole || 'BaseIAMRole';
+          defaultIamRole = defaultIamRole.replace('{{application}}', application.name);
+
           var command = {
             application: application.name,
             credentials: defaultCredentials,
@@ -51,7 +54,7 @@ module.exports = angular.module('spinnaker.aws.serverGroupCommandBuilder.service
             instanceMonitoring: false,
             ebsOptimized: false,
             selectedProvider: 'aws',
-            iamRole: 'BaseIAMRole', // TODO: should not be hard coded here
+            iamRole: defaultIamRole,
             terminationPolicies: ['Default'],
             vpcId: null,
             availabilityZones: availabilityZones,
--- a/app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.spec.js
+++ b/app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.spec.js
@@ -9,11 +9,12 @@ describe('awsServerGroupCommandBuilder', function() {
     )
   );
 
-  beforeEach(window.inject(function(awsServerGroupCommandBuilder, accountService, $q, $rootScope, subnetReader, instanceTypeService) {
+  beforeEach(window.inject(function(awsServerGroupCommandBuilder, accountService, $q, $rootScope, subnetReader, instanceTypeService, _settings_) {
     this.awsServerGroupCommandBuilder = awsServerGroupCommandBuilder;
     this.$scope = $rootScope;
     this.instanceTypeService = instanceTypeService;
     this.$q = $q;
+    this.settings = _settings_;
     spyOn(accountService, 'getPreferredZonesByAccount').and.returnValue($q.when(AccountServiceFixture.preferredZonesByAccount));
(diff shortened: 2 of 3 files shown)


ajordens:
PTAL one of @zanthrash @anotherchrisberry @icfantv :)

Essentially I'd like to see the default IAM role be application-based rather than BaseIAMRole.

They'll be a related PR for `clouddriver`.

app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.spec.js
```diff
@@ -9,11 +9,12 @@ describe('awsServerGroupCommandBuilder', function() {
     )
   );
 
-  beforeEach(window.inject(function(awsServerGroupCommandBuilder, accountService, $q, $rootScope, subnetReader, instanceTypeService) {
+  beforeEach(window.inject(function(awsServerGroupCommandBuilder, accountService, $q, $rootScope, subnetReader, instanceTypeService, _settings_) {
```
icfantv: i'm not seeing this injected function argument used anywhere.  did i miss something?

app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.spec.js
```diff
@@ -9,11 +9,12 @@ describe('awsServerGroupCommandBuilder', function() {
...
     this.awsServerGroupCommandBuilder = awsServerGroupCommandBuilder;
     this.$scope = $rootScope;
     this.instanceTypeService = instanceTypeService;
     this.$q = $q;
+    this.settings = _settings_;
```
icfantv: ditto.

app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.service.js
```diff
@@ -34,6 +34,9 @@ module.exports = angular.module('spinnaker.aws.serverGroupCommandBuilder.service
           var keyPair = credentials ? credentials.defaultKeyPair : null;
           var applicationAwsSettings = _.get(application, 'attributes.providerSettings.aws', {});
 
+          var defaultIamRole = settings.providers.aws.defaults.iamRole || '{{application}}InstanceProfile';
+          defaultIamRole = defaultIamRole.replace('{{application}}', application.name);
```
icfantv: you can simplify these two lines to just `const defaultIamRole = settings.providers.aws.defaults.iamRole || `${application.name}InstanceProfile`;

app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.spec.js
```diff
@@ -26,6 +27,7 @@ describe('awsServerGroupCommandBuilder', function() {
 
     it('initializes to default values, setting usePreferredZone flag to true', function () {
       var command = null;
+      this.settings.providers.aws.defaults.iamRole = '{{application}}IAMRole';
```
icfantv: I need to think about this some more tomorrow.  There should be a cleaner way of doing parameterized strings.

app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.service.js
ajordens: Oops ... this comment caused me to realize the default should still be `BaseIAMRole`. I accidentally modified it when doing some local testing and neglected to put it back.

app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.spec.js
ajordens: Sure ... I had just borrowed this pattern from the couple other places in `deck` that we're doing parameterized strings.

The particular format `{{application}}` is probably worth preserving as it's also used in `clouddriver` for other similar replacements.

zanthrash: My initial reaction is not to use the double curly brace '{{..}}' for a string substitution token since it already has a defined meaning within a Angular 1.x application and can lead to confusion.  With that said, if we have this as an already established pattern in the app then I'm okay with it.

ajordens: Ya ... we've already got it in a couple places in `deck` and `clouddriver`. Eventually I could see the defaults being sourced from `clouddriver` rather than duplicated in here (_but would put off doing that for now_).

ajordens:
Going to get this merged down ... if we come up with a better way to do templating, I'd have no qualms around revisiting.

app/scripts/modules/amazon/serverGroup/configure/serverGroupCommandBuilder.spec.js
icfantv: @zanthrash @ajordens i'm inclined to agree w/ you guys for now.  what do you think about switching to a Logback like syntax:  `Logger.log("{} i am in a {} mood", "today", "happy"); // today i am in a happy mood`?

PR #2648, opened 2016-09-01
Clarified rollback behavior

ajordens:
image

image


ajordens:
FYI @duftler / @gregturn.

We got a bit of feedback indicating that users were uncomfortable with the orchestrated rollbacks because they didn't know what it was going to do.

This attempts to address that.

I noticed that Google/Kubernetes/CloudFoundry have near verbatim copies of `rollbackServerGroup.html` ... it might be worth considering incorporating a similar change into your respective cloud providers.

PR #2649, opened 2016-09-02
provider/core: changed name from serialize application to save snapshot

nwwebb:
Changed the name from serializeApplication to saveSnapshot. Not too much here because Dan changed most of the names when he added the diff view. PTAL @lwander


lwander:
LGTM

PR #2651, opened 2016-09-02
Tidied up the default hosts and urls in settings.js

ajordens:


ajordens:

FYI @ewiseblatt -- Don't believe this would have any impact on the OSS packaging, we just wanted to remove the internal netflix urls.

tomaslin:
LGTM. I like the split between bakeryHost and details.

PR #2654, opened 2016-09-02
de-dupe docker tags

robfletcher:


icfantv:
Just a note that this was a band-aid until the server portion can be sorted out by Lars on Tuesday.

PR #2684, opened 2016-09-12
fix width of region field in titus run job

tomaslin:
screen shot 2016-09-12 at 2 40 26 pm


tomaslin:
@duftler @aglover FYI

PR #2688, opened 2016-09-14
changed size of ssl field so full text for terminated_https displays

kmarquardsen:
Also removed ssl reference as it's not an option for externalProtocol

Before:
screen shot 2016-09-14 at 8 42 20 am

After:
screen shot 2016-09-14 at 8 36 15 am


varikin:
Should the internal port default to 443 for terminated HTTPS? I would figure it would be 80.

PR #2689, opened 2016-09-14
changed size of ssl field so full text for terminated_https displays

kmarquardsen:
changed default listener protocol to http
removed ssl reference since it's not an option for externalProtocol


jcwest:
LGTM

PR #2699, opened 2016-09-15
Azure bug fix: failed to create server group using selected vnet/subnet

milismsft:
Bug fix: we are failing to submit the expected configuration when creating the server group using the user selected vnet and subnet.


milismsft:
@spinnaker/netflix-reviewers @anotherchrisberry PTAL and merge if it looks good.

PR #2701, opened 2016-09-16
(core) add day selector to execution window

icfantv:
* add days-of-week selector widget to pipeline execution restriction window

### DIFF
--- a/app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.spec.js
+++ b/app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.spec.js
@@ -1,0 +1,90 @@
+import { DAYS_OF_WEEK} from './daysOfWeek';
+
+describe('Component:  Execution Window Day Picker', () => {
+
+  let $componentController;
+  beforeEach(window.module(require('./executionWindowDayPicker.component.js')));
+  beforeEach(window.inject((_$componentController_) => $componentController = _$componentController_));
+
+  function constructController(bindings) {
+    return $componentController('executionWindowDayPicker', null, bindings);
+  }
+
+  it('should not have any days selected when the days property is not defined', () => {
+    const ctrl = constructController({days: undefined});
+    expect(ctrl.days).toBeUndefined();
+  });
+
+  it('should not have any days selected when the days property is null', () => {
+    const ctrl = constructController({days: null});
+    expect(ctrl.days).toBe(null);
+  });
+
+  it('should not have any days selected when the days property is an empty array', () => {
+    const ctrl = constructController({days: []});
+    expect(ctrl.days.length).toBe(0);
+  });
+
+  DAYS_OF_WEEK.forEach((day) => {
+    it(`should have ${day.key} selected when the days property contains ${day.ordinal}`, () => {
+      const ctrl = constructController({days: [day.ordinal]});
+      expect(ctrl.days.length).toBe(1);
+      expect(ctrl.days[0]).toBe(day.ordinal);
+    });
+  });
+
+  it('should select all the days when the all button is clicked', () => {
+    const ctrl = constructController({});
+    ctrl.all();
+    expect(ctrl.days.length).toBe(7);
+  });
+
+  it('should select none of the days when the none button is clicked', () => {
+    const ctrl = constructController({});
+    ctrl.all();
+    expect(ctrl.days.length).toBe(7);
+    ctrl.none();
+    expect(ctrl.days.length).toBe(0);
+  });
+
+  it('should select just the weekdays when the weekday button is clicked', () => {
+    const ctrl = constructController({});
+    ctrl.weekdays();
+    expect(ctrl.days.length).toBe(5);
+    expect(ctrl.days).toEqual([2,3,4,5,6]);
+  });
+
+  it('should select just the weekend when the weekend button is clicked', () => {
+    const ctrl = constructController({});
+    ctrl.weekend();
+    expect(ctrl.days.length).toBe(2);
+    expect(ctrl.days[0]).toBe(1); // sunday
+    expect(ctrl.days[1]).toBe(7); // saturday
+  });
+
+  it('should specify whether or not a day is selected', () => {
+    const ctrl = constructController({
+      days: [1]
+    });
+    expect(ctrl.daySelected(1)).toBe(true);
+    expect(ctrl.daySelected(2)).toBe(false);
+
+    ctrl.days = [1, 2];
+    expect(ctrl.daySelected(2)).toBe(true);
+    expect(ctrl.daySelected(3)).toBe(false);
+  });
+
+  it('should add a day to the model when selected', () => {
+    const ctrl = constructController({});
+    ctrl.updateModel(DAYS_OF_WEEK[0]);
+    expect(ctrl.daySelected(1)).toBe(true);
+  });
+
+  it('should remove a day from the model when unselected', () => {
+    const ctrl = constructController({
+      days: [1] // pre-select sunday
+    });
+    ctrl.updateModel(DAYS_OF_WEEK[0]);
+    expect(ctrl.daySelected(1)).toBe(false);
+  });
+});
--- a/app/scripts/modules/core/pipeline/config/stages/executionWindows/executionWindows.html
+++ b/app/scripts/modules/core/pipeline/config/stages/executionWindows/executionWindows.html
@@ -11,6 +11,24 @@
 <div ng-if="stage.restrictExecutionDuringTimeWindow">
   <div class="row">
     <div class="col-md-10 col-md-offset-1">
+      <h5>
+        <strong>Days of the Week</strong>
+        <small>(No days selected implies the stage will execute on any day if triggered)</small>
+      </h5>
+      <p class="execution-window-days-of-week">
+        <execution-window-day-picker days="stage.restrictedExecutionWindow.days"></execution-window-day-picker>
+      </p>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-md-10 col-md-offset-1">
+      <p>
+        <strong>Time of Day</strong>
+      </p>
+    </div>
+  </div>
+  <div class="row">
+    <div class="col-md-10 col-md-offset-1">
       <div class="execution-window-graph">
         <div class="execution-day">
           <div ng-repeat="window in timelineWindows"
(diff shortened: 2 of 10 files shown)


icfantv:
@spinnaker/netflix-reviewers @spinnaker/google-reviewers, PTAL, this is a first attempt at switching to ES6 module loading and class syntax for items which will allow us to move away from using `require` everywhere.

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.js
```diff
@@ -0,0 +1,89 @@
+import angular from 'angular';
+
+class ExecutionWindowDayPickerController {
+
+  constructor() {
+    this.DAYS = [{
```
anotherchrisberry: It's a little confusing the have both `DAYS` and `days` on the controller. Maybe change this to `dayOptions`?

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.js
```diff
@@ -0,0 +1,89 @@
...
+    this.days = updated;
+  }
+}
+
+const EXECUTION_WINDOW_DAY_PICKER_COMPONENT = {
+  bindings: {
+    days: '='
+  },
+  controller: ExecutionWindowDayPickerController,
+  controllerAs: 'dayPicker',
```
anotherchrisberry: We are not using `controllerAs` syntax in components - better to stick with the default (`$ctrl`).

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.js
```diff
@@ -0,0 +1,89 @@
...
+  templateUrl: require('./ExecutionWindowDayPickerComponent.html')
+};
+
+const MODULE_NAME = 'spinnaker.core.pipeline.stage.executionWindows.dayPicker';
+angular.module(MODULE_NAME, [])
+  .component('executionWindowDayPicker', EXECUTION_WINDOW_DAY_PICKER_COMPONENT);
+export default MODULE_NAME;
```
anotherchrisberry: To be consistent with the rest of the code base, it's probably better to rewrite this block (L76-88) as:
```javascript
module.exports = angular.module('spinnaker.core.pipeline.stage.executionWindows.dayPicker', [])
  .component('executionWindowDayPicker', {
    bindings: {
      days: '='
    },
    controller: ExecutionWindowDayController,
    templateUrl: require('./executionWindowDayPicker.component.html')
});
```

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.html
```diff
@@ -0,0 +1,13 @@
+<div class="btn-group">
```
anotherchrisberry: File should be named `executionWindowDayPicker.component.html` to be consistent with other components.

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.js
```diff
@@ -0,0 +1,89 @@
+import angular from 'angular';
```
anotherchrisberry: Should be renamed `executionWindowDayPicker.component.js` to be consistent with other components.

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.spec.js
```diff
@@ -0,0 +1,87 @@
+import dayPickerComponent from './ExecutionWindowDayPickerComponent';
+
+fdescribe('Component:  Execution Window Day Picker', () => {
```
anotherchrisberry: `fdescribe` should just be `describe`

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.js
```diff
@@ -0,0 +1,89 @@
...
+
+class ExecutionWindowDayPickerController {
+
+  constructor() {
+    this.DAYS = [{
+      dow: 'sunday',
+      abbr: 'Sun'
```
anotherchrisberry: To be consistent with the rest of the codebase, these should probably be `key` and `label`

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.spec.js
```diff
@@ -0,0 +1,87 @@
...
+
+  it('should select just the weekdays when the weekday button is clicked', () => {
+    const ctrl = $componentController('executionWindowDayPicker', null, {});
+    ctrl.weekdays();
+    expect(ctrl.days.length).toBe(5);
+
+    const days = new Set(ctrl.days);
+    for (let i = 2; i < 7; i++) {
+      expect(days.has(i)).toBe(true);
+    }
```
anotherchrisberry: These four lines read better as:
```javascript
expect(ctrl.days).toEqual([2,3,4,5,6]);
```

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.js
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  updateModel() {
+    const updated = [];
+    this.DAYS.forEach((day, index) => {
+      if (this.dayModel[day.dow]) {
+        updated.push(index + 1);
+      }
+    });
+    this.days = updated;
```
anotherchrisberry: Does this get applied to the underlying stage? Seems like, by reassigning the value on the controller here, it won't end up on the stage, and the stage's `days` field will remain unchanged.

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.js
```diff
@@ -0,0 +1,89 @@
...
+      abbr: 'Fri'
+    }, {
+      dow: 'saturday',
+      abbr: 'Sat'
+    }];
+    this.dayModel = {};
+  }
+
+  $onInit() {
+    console.log('executing onInit');
```
icfantv: ack!

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.spec.js
icfantv: ack!

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.spec.js
icfantv: Aside from being cleaner, this would imply that order matters for the `days` property, which, on the client-side, it does not.  Can you confirm that days is order-dependent on the server side?  In fact, I'm wondering if this should be a `Set`... thoughts?

app/scripts/modules/core/pipeline/config/stages/executionWindows/ExecutionWindowDayPickerComponent.js
icfantv: @anotherchrisberry yes, this works and i've confirmed across page reloads that saved values are re-retrieved and set correctly.

icfantv:
@anotherchrisberry @zanthrash updated per comments/discussions.  PTAL.  Thanks.

app/scripts/modules/core/pipeline/config/stages/executionWindows/executionWindows.html
```diff
@@ -75,13 +93,11 @@
...
+  <div class="row">
+    <div class="col-md-6 col-md-offset-1">
+      <button class="btn btn-block btn-add-trigger add-new" ng-click="executionWindowsCtrl.addExecutionWindow()">
+        <span class="glyphicon glyphicon-plus-sign"></span> Add an execution window
+      </button>
     </div>
   </div>
-
-</div>
+git st</div>
```
anotherchrisberry: remove `git st`

anotherchrisberry:
Looking better - can you add something to the execution details view around https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/pipeline/config/stages/executionWindows/executionWindowsDetails.html#L7 to show which days (if any) the window is restricted to?

app/scripts/modules/core/pipeline/config/stages/executionWindows/executionWindows.html
icfantv: yea, i caught this afterwards...not sure what happend.

icfantv:
@anotherchrisberry this isn't quite ready yet - i found an issue with the revert button that required completely changing how we initially render the buttons, set the states, and manage the internal model due to `$onChanges` only being fired when a _reference_ to a collection changes, rather than the contents due to the performance ramifications of doing a deep watch on the collections [as reported here](https://github.com/angular/angular.js/issues/14378).

app/scripts/modules/core/pipeline/config/stages/executionWindows/executionWindows.html
icfantv: ok, well....i KNOW what happened....

icfantv:
The huge advantage that this exposed (the `$onChanges` not being fired, was that we now effectively bind the days model directly to the UI and we no longer are using the UIBS checkbox buttons.

icfantv:
screen shot 2016-09-19 at 11 51 46 am
screen shot 2016-09-19 at 11 52 09 am

icfantv:
@anotherchrisberry @zanthrash i think we're done here, PTAL.

app/scripts/modules/core/pipeline/config/stages/executionWindows/executionWindowsDetails.controller.js
```diff
@@ -13,9 +15,28 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.executionWindows.
...
+      return DAYS_OF_WEEK.filter(day => daySet.has(day.ordinal)).map(day => day.label);
+    }
+
+    // ditto
+    function getDayText() {
+      let dayText = 'Everyday';
+      const days = $scope.stage.context.restrictedExecutionWindow.days;
+      if (days && (days.length > 0)) {
+        const daysAsText = replaceDays(days);
+        dayText = daysAsText.toString().replace(/,/g, ', ');
```
anotherchrisberry: If you weren't using a `Set` above, you could just do `days.join(', ')` here. But this works, too, and I know how much you really like your `Set`s.

icfantv: I'll fix.  I'd forgotten about `join`.

icfantv: fixed.

PR #2706, opened 2016-09-16
(gce) use locationToInstanceTypeMap on credentials to determine avail…

danielpeach:
…able instance types

- removed vCpuMaxByLocation list
- fixed console error when selecting custom instance type

@duftler please review


app/scripts/modules/google/serverGroup/configure/wizard/instanceType/instanceType.html
```diff
@@ -1,7 +1,7 @@
 <div ng-show="!!((command.viewState.disableImageSelection || command.image) && (command.zone || command.regional))">
   <v2-instance-archetype-selector command="command"></v2-instance-archetype-selector>
   <div style="padding: 0 15px">
-    <v2-instance-type-selector ng-show="command.viewState.instanceProfile && !ctrl.specialInstanceProfiles.has(command.viewState.instanceProfile)" command="command"></v2-instance-type-selector>
```
danielpeach: `ng-show` here allowed the directive to still render, causing a console error.

duftler:
LGTM.
I tested by manually manipulating the instance types returned by the backend to include some 'larger' types, and then verified that I could create custom type up to the largest size represented.

PR #2707, opened 2016-09-16
Bug fix - failed to select vnet in arbitrary resource group

milismsft:
When creating an azure load balancer and server group, the selected virtual network can be from a different resource group then the one corresponding to the application.


milismsft:
@spinnaker/netflix-reviewers @anotherchrisberry PTAL and merge if looks good.

PR #2717, opened 2016-09-20
(openstack) adding in server group tags

drmaas:


drmaas:
@spinnaker/reviewers ^^^

PR #2726, opened 2016-09-21
(google) Also render 'Desired' attribute when autoscaler is configured.

duftler:


duftler:
@danielpeach please review.

duftler:
image

PR #2742, opened 2016-09-23
(k8s & docker) fixes for rxjs update

danielpeach:
@lwander 

fixes k8s image lookup in wizard.
fixes docker trigger modal.


lwander:
woohoo, thanks, LGTM

PR #2743, opened 2016-09-23
(k8s) add anchor tags for load balancer ips

danielpeach:
@lwander @skim1420 fyi

(k8s) add anchor tags for load balancer ips

### DIFF
--- a/app/scripts/modules/kubernetes/loadBalancer/details/details.html
+++ b/app/scripts/modules/kubernetes/loadBalancer/details/details.html
@@ -89,7 +89,9 @@
       <dl ng-if="loadBalancer.service.spec.clusterIP">
         <dt>Cluster IP</dt>
         <dd>
-        {{loadBalancer.service.spec.clusterIP}}
+          <a target="_blank" href="//{{loadBalancer.service.spec.clusterIP}}">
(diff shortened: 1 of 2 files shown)


app/scripts/modules/kubernetes/loadBalancer/details/details.html
```diff
@@ -89,7 +89,9 @@ <h3 select-on-dbl-click>
       <dl ng-if="loadBalancer.service.spec.clusterIP">
         <dt>Cluster IP</dt>
         <dd>
-        {{loadBalancer.service.spec.clusterIP}}
+          <a target="_blank" href="http://{{loadBalancer.service.spec.clusterIP}}">
```
lwander: Why prefix everything with `http`? Is that necessary? It's not necessarily the case that these endpoints are serving traffic over HTTP

danielpeach: Good point. These should probably be changed on the GCE side too.

danielpeach: Updated the hrefs for kubernetes and GCE

PR #2744, opened 2016-09-23
(gce) add certificate selector for http load balancer wizard

danielpeach:
@duftler @jtk54 please review
certificate


jtk54:
Depending on how users like this, maybe we can add a checkbox in the future to explicitly choose between Http/s LBs. Currently, if you select a cert you'll get an Https LB, otherwise Http. This is explained in the tooltip by 'Certificate'.

jtk54:
LGTM

PR #2756, opened 2016-09-27
(core) webpack config s/rx/rxjs

danielpeach:
@icfantv


icfantv:
LGTM.

This was overlooked in upgrading Rx and moving to ES6 module loading.  Since the package name changed, the `vendor` array also needed to be updated to ensure the Rx library goes in the `vendor.js` file.  Good catch by @gregturn and @danielpeach.

PR #2762, opened 2016-09-28
(Openstack) adding rollback functionality for openstack

SteveKMin:
Closes issue 1114

Rollback functionality details:
Rollback must be done on an enabled server group. After clicking rollback under server group actions, you must select a disabled server group within the same cluster in the drop down. Running rollback will enable the selected disabled server group, resize it to match the size of the active selected server group, and then disable the selected server group.


jcwest:
LGTM, but needs a rebase.

PR #2763, opened 2016-09-28
(k8s) fixes error causing command backing data to be included in pipe…

danielpeach:
…line JSON

GCE also copies the command before entering the edit modal:
https://github.com/spinnaker/deck/blob/master/app/scripts/modules/google/serverGroup/configure/serverGroupCommandBuilder.service.js#L361

@lwander


lwander:
Awesome, LGTM - thanks Dan

PR #2768, opened 2016-09-29
(gce) help text for http lb listener named port

danielpeach:
@jtk54 

Help text contents:
Incoming traffic is directed through a named port (e.g., http). The mapping from named port to port number is specified per server group and can be configured within the server group creation dialogue under Port Name Mapping.


jtk54:
We don't currently support named ports other than 'http', so we should make that clear here I think.

danielpeach:
Updated:

Incoming traffic is directed through a named port (for Spinnaker, the named port is http). The mapping from named port to port number is specified per server group and can be configured within the server group creation dialogue under Port Name Mapping.

jtk54:
LGTM

PR #2779, opened 2016-09-30
Set 'Consider only...' checkbox by default if both app-level settings are enabled.

duftler:


duftler:
This is what the new warnings look like (they must be acknowledged before the create/edit application dialogs can be submitted):
image
image

duftler:
@ajordens and @danielpeach please review.

ajordens:
Can we only show the warning when the user changes the instance health settings (vs when the application is created?)

app/scripts/modules/titus/serverGroup/configure/ServerGroupCommandBuilder.js
```diff
@@ -40,6 +40,10 @@ module.exports = angular.module('spinnaker.titus.serverGroupCommandBuilder.servi
         }
       };
 
+      if (application.attributes && application.attributes.platformHealthOnlyShowOverride && application.attributes.platformHealthOnly) {
+        command.interestingHealthProviderNames = ['Titus'];
```
ajordens: I'm not sure we actually need this for `titus`.

duftler: Backed out these changes for titus.

duftler:
Reworded the message when enabling showOverrideWarning as discussed offline.

duftler:
Relates to https://github.com/spinnaker/deck/pull/2615 and https://github.com/spinnaker/orca/pull/957.

PR #2781, opened 2016-09-30
(gce) rewrite http load balancer wizard

danielpeach:
@jtk54 @duftler 

- Create http load balancer with multiple listeners
- Create and edit backend services and health checks
- Session affinity for backend services
- Delete http load balancer
- View http load balancer details
- Attach server group to load balancer by url map name (with https://github.com/spinnaker/clouddriver/pull/1092).


danielpeach:
Updated: 
- Aligned fields, minor UX details
- Listener names are now validated against existing listener names.

jtk54:
Nice work, LGTM.

PR #2792, opened 2016-10-03
(gce) backend services in load balancer details

danielpeach:
@jtk54 
backend_service_details
 
What do you think about just listing all of the health checks (after deduping, without the "Backend Service Name" header), since the relationship between health checks and backend services should be clear now?


danielpeach:
Also changed the placement of the listener help text (it only shows up once now, instead of on every listener).

jtk54:
LGTM, needs a rebase.

PR #2800, opened 2016-10-04
(openstack) adding instance details view and operations

drmaas:
Will work correctly when https://github.com/spinnaker/clouddriver/pull/1102 is merged.

@spinnaker/reviewers


drmaas:
screen shot 2016-10-04 at 11 31 09 am
screen shot 2016-10-04 at 11 31 29 am

PR #2806, opened 2016-10-04
(core): correct tsc errors

icfantv:
* correct errors spit out by tsc


icfantv:

PR #2814, opened 2016-10-05
(gce) fixed http lb wizard refresh bugs

danielpeach:
@jtk54 
Turns out there were some bugs. They showed up when I tried to refresh backend services or health checks when a bes/hc wasn't fully configured.


jtk54:
LGTM, needs a rebase.

PR #2819, opened 2016-10-06
(gce) fix network lb association

danielpeach:
@duftler @jtk54 please review.

This was a mistake. Here are my assumptions now:
- The json payload needs to have names of the network load balancers under the key `loadBalancers`.
- Including the name of a url map will cause the operation to fail.
- The listener names only need to be included within the metadata.


jtk54:
Those assumptions make sense to me -- LGTM. Needs a rebase.

PR #2821, opened 2016-10-06
(core) fixes check precondition for openstack

SteveKMin:
$scope.precondition.cloudProvider was never set during this pipeline stage, so I have added code to set the variable to the proper cloud provider during the "accountUpdated" step.

### DIFF
--- a/app/scripts/modules/core/pipeline/config/preconditions/selector/preconditionSelector.directive.js
+++ b/app/scripts/modules/core/pipeline/config/preconditions/selector/preconditionSelector.directive.js
@@ -60,6 +60,16 @@ module.exports = angular.module('spinnaker.core.pipeline.config.preconditions.se
 
       let accountFilter = (cluster) => cluster.account === $scope.precondition.context.credentials;
       $scope.regions = appListExtractorService.getRegions([$scope.application], accountFilter);
+
+      //Setting cloudProvider when account is updated
+      let providerFilter = (account) => account.name === $scope.precondition.context.credentials;
+      let accountsArray = $scope.accounts;
+      if (accountsArray !== null && accountsArray !== undefined) {
+        let account = accountsArray.find(providerFilter);
+        if (account !== null && account !== undefined) {
+          $scope.precondition.cloudProvider = account.type;
+        }
+      }
     };
 
     this.reset = () => {


SteveKMin:
@spinnaker/reviewers PTAL

app/scripts/modules/core/pipeline/config/preconditions/selector/preconditionSelector.directive.js
```diff
@@ -60,6 +60,16 @@ module.exports = angular.module('spinnaker.core.pipeline.config.preconditions.se
 
       let accountFilter = (cluster) => cluster.account === $scope.precondition.context.credentials;
       $scope.regions = appListExtractorService.getRegions([$scope.application], accountFilter);
+
+      //Setting cloudProvider when account is updated
+      let providerFilter = (accounts) => accounts.name === $scope.precondition.context.credentials;
+      var accountsArray = $scope.accounts;
+      if(accountsArray !== null) {
+        var account = accountsArray.find(providerFilter);
```
danielpeach: 1). With `var` (as opposed to `let`), `account` will be defined outside the `if` block. Overall, probably better to use `let`, but it's not breaking anything here.
2). If `.find` doesn't find a match, it will return `undefined`, not null, which will cause an error on line 70.

app/scripts/modules/core/pipeline/config/preconditions/selector/preconditionSelector.directive.js
```diff
@@ -60,6 +60,16 @@ module.exports = angular.module('spinnaker.core.pipeline.config.preconditions.se
 
       let accountFilter = (cluster) => cluster.account === $scope.precondition.context.credentials;
       $scope.regions = appListExtractorService.getRegions([$scope.application], accountFilter);
+
+      //Setting cloudProvider when account is updated
+      let providerFilter = (accounts) => accounts.name === $scope.precondition.context.credentials;
+      var accountsArray = $scope.accounts;
+      if(accountsArray !== null) {
```
danielpeach: Would be preferable as `if (` instead of `if(` here and below.

app/scripts/modules/core/pipeline/config/preconditions/selector/preconditionSelector.directive.js
```diff
@@ -60,6 +60,16 @@ module.exports = angular.module('spinnaker.core.pipeline.config.preconditions.se
 
       let accountFilter = (cluster) => cluster.account === $scope.precondition.context.credentials;
       $scope.regions = appListExtractorService.getRegions([$scope.application], accountFilter);
+
+      //Setting cloudProvider when account is updated
+      let providerFilter = (accounts) => accounts.name === $scope.precondition.context.credentials;
```
danielpeach: Very minor, but you're iterating over an array of `accounts`, and this filter will be called with an `account` (rather than all `accounts`).

PR #2844, opened 2016-10-11
Allow a user to deploy w/ only an ami id (backing image may not be in…

ajordens:
…dexed yet)

Probably an "advanced" feature but common enough that we should support it.

- All instance types are available (cannot filter w/o virtualization type)


ajordens:
image

Thoughts @anotherchrisberry?

You need to be in `Search All Images` mode.

anotherchrisberry:
Makes sense to me. I'm not crazy about showing the AMI ID as the image name, but I can't think of anything to put in there that would be less confusing, and, as you said, the operation will fail fast if it's not a valid AMI.

👍

PR #2872, opened 2016-10-14
(openstack) adding in support for assigning floating ips to asg members

drmaas:
@spinnaker/reviewers

### DIFF
--- a/app/scripts/modules/openstack/serverGroup/configure/wizard/access/AccessSettings.controller.js
+++ b/app/scripts/modules/openstack/serverGroup/configure/wizard/access/AccessSettings.controller.js
@@ -6,7 +6,7 @@ module.exports = angular.module('spinnaker.serverGroup.configure.openstack.acces
   require('angular-ui-router'),
   require('angular-ui-bootstrap'),
   require('../../../../common/appCacheBackedMultiSelectField.directive.js'),
-]).controller('openstackServerGroupAccessSettingsCtrl', function($scope, loadBalancerReader, securityGroupReader, v2modalWizardService /* TODO: unused, $rootScope, infrastructureCaches, cacheInitializer */) {
+]).controller('openstackServerGroupAccessSettingsCtrl', function($scope, loadBalancerReader, securityGroupReader, networkReader, v2modalWizardService) {
 
   function updateFilter() {
     $scope.filter = {
--- a/app/scripts/modules/openstack/serverGroup/configure/wizard/access/accessSettings.html
+++ b/app/scripts/modules/openstack/serverGroup/configure/wizard/access/accessSettings.html
@@ -28,6 +28,36 @@
                                             required="true"
                                             model="command.securityGroups"></os-app-cache-backed-multi-select-field>
 
+    <div class="form-group">
+      <div class="col-md-5 sm-label-left">
+        <b>Associate Floating IP Address</b>
+        <help-field key="openstack.network.floatingip"></help-field>
+      </div>
+      <div class="col-md-2 radio">
+        <label>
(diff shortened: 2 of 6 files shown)


app/scripts/modules/openstack/serverGroup/configure/wizard/access/AccessSettings.controller.js
```diff
@@ -6,7 +6,7 @@ module.exports = angular.module('spinnaker.serverGroup.configure.openstack.acces
   require('angular-ui-router'),
   require('angular-ui-bootstrap'),
   require('../../../../common/appCacheBackedMultiSelectField.directive.js'),
-]).controller('openstackServerGroupAccessSettingsCtrl', function($scope, loadBalancerReader, securityGroupReader, v2modalWizardService /* TODO: unused, $rootScope, infrastructureCaches, cacheInitializer */) {
+]).controller('openstackServerGroupAccessSettingsCtrl', function($scope, loadBalancerReader, securityGroupReader, networkReader, v2modalWizardService /* TODO: unused, $rootScope, infrastructureCaches, cacheInitializer */) {
```
danielpeach: I think the TODO could be removed.

app/scripts/modules/openstack/serverGroup/configure/wizard/access/accessSettings.html
```diff
@@ -28,6 +28,36 @@
...
                                             model="command.securityGroups"></os-app-cache-backed-multi-select-field>
 
+    <div class="form-group">
+      <div class="col-md-5 sm-label-left">
+        <b>Associate Floating IP Address</b>
+        <help-field key="openstack.network.floatingip"></help-field>
+      </div>
+      <div class="col-md-2 radio">
+        <label>
+          <input type="radio" ng-model="command.associatePublicIpAddress" ng-value="true" id="associatePublicIpAddressTrue"/>
```
danielpeach: Are these `id`s doing anything? They are a little generically named, and it would be unfortunate if anyone came up with a matching CSS rule down the road.

app/scripts/modules/openstack/serverGroup/configure/wizard/access/accessSettings.html
```diff
@@ -28,6 +28,36 @@
...
+      </div>
+      <div class="col-md-2 radio">
+        <label>
+          <input type="radio" ng-model="command.associatePublicIpAddress" ng-value="null" id="associatePublicIpAddressDefault"/>
+          Default
+        </label>
+      </div>
+    </div>
+
+    <div ng-if="command.associatePublicIpAddress == true" class="form-group">
```
danielpeach: Please use `===` here -- there won't be any problems as it is written, but it looks dangerous, since the statement will be true for any `command.associatePublicIpAddress` that is JS true.

app/scripts/modules/openstack/serverGroup/details/serverGroupDetails.openstack.controller.js
```diff
@@ -346,5 +348,13 @@ module.exports = angular.module('spinnaker.serverGroup.details.openstack.control
       });
     };
 
+    this.applyFloatingIpDetails = () => {
+      return networkReader.listNetworksByProvider('openstack').then((networks) => {
+        ctrl.floatingNetworkName = (_.find(networks, (net) => {
```
danielpeach: This block would read a little better if it used either all `ctrl`s or all `this`s

app/scripts/modules/openstack/serverGroup/configure/wizard/access/accessSettings.html
drmaas: Copy-paste from the Azure code, let's blame Microsoft :)

PR #2879, opened 2016-10-17
(core) make room for long image name in server group label

danielpeach:
@anotherchrisberry let me know what you think about this. If the image has a sha, then the label will always wrap, since there's only room for ~50 characters.

Before:
before

After:
after

### DIFF
--- a/app/scripts/modules/core/cluster/rollups.less
+++ b/app/scripts/modules/core/cluster/rollups.less
@@ -366,6 +366,9 @@ load-balancers-tag {
 
   .section-title {
     min-height: 24px;
+    span {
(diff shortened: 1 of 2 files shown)


anotherchrisberry:
Unfortunately, there are some scenarios where `col-md-3` is not quite big enough for load balancers + health counts...
screen shot 2016-10-17 at 12 29 10 pm

If we set the column counts based on the length (or even presence) of `images`, that would avoid breaking us and satisfy long sha rendering, e.g. `class="col-md-{{viewModel.images ? 3 : 4}}`

danielpeach:
Done

PR #2887, opened 2016-10-20
(google): Fixed L7 type check.

jtk54:
L7 check was 'if global or HTTP', since SSL LBs are global, this needs modified. @danielpeach please review.


danielpeach:
LGTM

PR #2903, opened 2016-10-26
(gce) create/update ilb wizard

danielpeach:
Please do not merge yet. @jtk54


danielpeach:
I've killed off the bugs that Jake found on his first pass. @jtk54 or @duftler please review.

This only includes the ILB create/update wizard. The details panel is still missing a few pieces.

jtk54:
LGTM also.

danielpeach:
Network is hard-coded as default for now in both Deck and Clouddriver. It's not possible to switch the network mode for the default network to "custom" (this took some poking around), which is the mode that allows you to add subnets, so subnets are also hardcoded for now.

PR #2905, opened 2016-10-26
(gce) render backend service and health check in ilb details

danielpeach:
@jtk54
- a few changes to make the ILB details render properly.
- also a quick fix to make L7 upsert work properly.


jtk54:
LGTM.

PR #2912, opened 2016-10-27
(titus): convert to docker image selector

icfantv:
- convert docker selector widget to typescript
- convert titus ASG basic settings docker image from text field to
  widget.
- update kubernetes to reflect changes in API.
- update docker trigger in pipelines to reflect changes in API


tomaslin:
your commit stream is not squashed

icfantv:
@tomaslin yea, i screwed something up....trying to unbork it.

icfantv:
ok, i think i have it fixed...without losing any commits.

@tomaslin @anotherchrisberry @zanthrash PTAL.  I opted for disabling the selects rather than `ng-if`-ing them for better UX.  One more tweak we could (should?) probably make is to disabled the subsequent selects when the parent hasn't yet been set, e.g., disable the tag select box until the image has been selected.

@danielpeach I'm pretty sure I borked Kubernetes, can you please double check this.  I had to change the binding on `registry` to one way so that `$onChanges` is called.  Angular components will only fire the `$onChanges` event for `<` and `@` bindings.  I think this means you won't get the selected registry back.  The recommended way of dealing with this is to pass in yet another binding in the form of a callback.  I don't love it, but can't think of a better way short of using an object instead of a string.  PLMK

app/scripts/modules/docker/image/dockerImageAndTagSelector.component.ts
```diff
@@ -0,0 +1,235 @@
...
+  }
+
+  private initializeImages() {
+    const imageConfig: IImageConfig = {
+      provider: 'dockerRegistry',
+      account: this.showRegistry ? this.account : this.registry
+    };
+
+    if (!this.viewState.imagesLoading) {
+      console.log('performing image refresh');
```
anotherchrisberry: remove

app/scripts/modules/docker/image/docker.image.reader.service.ts
```diff
@@ -0,0 +1,25 @@
...
+              private retryService: any) {}
+
+  public findImages(params: any): ng.IPromise<IDockerImage[]> {
+    return this.retryService.buildRetrySequence(() => this.API.all('images/find').getList(params), (results: any) => (results.length > 0), 10, 1000)
+      .then((results: any) => results)
+      .catch((): any => []);
+  }
+}
+
+export const DOCKER_IMAGE_READER_SERVICE = 'spinnaker.docker.image.readerr';
```
anotherchrisberry: `readerr` should be `reader`

app/scripts/modules/docker/image/dockerImageAndTagSelector.component.ts
icfantv: yep, done.

app/scripts/modules/docker/image/dockerImageAndTagSelector.component.ts
```diff
@@ -0,0 +1,233 @@
...
+
+        this.updateOrganizationMap(images);
+        this.organizations.push(...Object.keys(this.organizationMap));
+        this.updateOrganizationsList();
+
+        this.updateRepositoryMap(images);
+
+        this.viewState.imagesLoading = false;
+        this.viewState.imagesLoaded = true;
+        this.viewState.imagesRefreshing = false;
```
icfantv: ditto

anotherchrisberry:
@danielpeach or @lwander can you guys pull this down and take a look?

danielpeach:
I don't think it's saving `registry` properly - I checked both Docker triggers and the Kubernetes Run Job stage. Also, if I save a trigger or a run job stage, then reload the page, I have to click the refresh wheel before I see anything in the dropdowns. Based on the logic I see in the `$onInit`, the two issues seem related.

danielpeach:
@icfantv I just fully read your comment about the registry binding. Do you want me to write a fix on top of this?

icfantv:
@danielpeach lemme ping you on slack so we can chat about solutions?

icfantv:
@danielpeach Ok pushed new changes based on our conversation.  PTAL && LMK.  Thanks.

icfantv:
Final update on this.  Loading existing docker pipeline triggers was only getting the one registry for the images of the specified registry resulting in the inability to change the registry.  Retrieving the list of registry types asynchronously and setting on the docker image selector widget was resulting in the form erroneously being marked as dirty with no way to reestablish pristine-ness (pristine events do not propagate upwards in form hierarchy) and this was probably a hack anyway.

The final solution implemented was to modify the docker trigger to retrieve the list of registry names to populate the the select box (selecting the first one in the event of a new or unspecified account) and then updating the images once the account was selected.  Additionally, the list of account/registry-specific images is retrieved and processed every time the account/registry is changed.

The fields are disabled until the images are finished being processed and the form is disabled until at least the org and image are selected.

screen shot 2016-10-28 at 2 41 51 pm
screen shot 2016-10-28 at 2 42 21 pm

icfantv:
@zanthrash @danielpeach @anotherchrisberry PTAL

danielpeach:
I tested this for Docker triggers and for the Kubernetes Run Job stage. It works as expected. LGTM.

icfantv:
Thanks @danielpeach for the GCP/Kubernetes testing.  Will rebase and then merge after CI completes.

PR #2922, opened 2016-10-31
Surfaces pipeline error message when saving fails.

ttomsu:
This work is to support https://github.com/spinnaker/fiat/issues/36

Fiat auth failure preview
yswc6rbpkjr

Generic preview:
newsfrdwg6n

@danielpeach @anotherchrisberry PTAL

### DIFF
--- a/app/scripts/modules/core/pipeline/config/pipelineConfigurer.html
+++ b/app/scripts/modules/core/pipeline/config/pipelineConfigurer.html
@@ -99,12 +99,12 @@
       </div>
       <div class="col-md-9 text-right">
         <span ng-if="viewState.saveError" class="alert alert-danger">
-          There was an error saving your pipeline.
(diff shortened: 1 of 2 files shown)


danielpeach:
LGTM

app/scripts/modules/core/pipeline/config/pipelineConfigurer.html
```diff
@@ -99,12 +99,12 @@
       </div>
       <div class="col-md-9 text-right">
         <span ng-if="viewState.saveError" class="alert alert-danger">
-          There was an error saving your pipeline.
+          <span ng-bind-html="pipelineConfigurerCtrl.getErrorMessage()"></span>
```
ttomsu: This, among **many** other reasons, is why I leave the Angular stuff to the professionals.

PR #2950, opened 2016-11-03
Adds 'Run As User' dropdown + service account service to trigger config

ttomsu:
This is not an indicator that I am actually capable of making _real_ contributions to Deck. As I've said before, I leave that to the professionals.

@danielpeach @anotherchrisberry PTAL

dt24jtkatts

tag: https://github.com/spinnaker/fiat/issues/36


app/scripts/modules/core/serviceAccount/serviceAccount.service.js
```diff
@@ -0,0 +1,24 @@
...
+  require('../config/settings.js'),
+])
+  .factory('serviceAccountService', function(settings, API, $q) {
+
+    let getServiceAccounts = function() {
+      if (!settings.feature.fiatEnabled) {
+        return $q.resolve([]);
+      }
+
+      return API.one('auth/user/serviceAccounts').get();
```
danielpeach: I think the use in Deck is more typically `API.one('auth').one('user').one('serviceAccounts').get()`, but not a big deal.

app/scripts/modules/core/pipeline/config/triggers/cron/cronTrigger.html
```diff
@@ -27,4 +27,19 @@
       {{vm.validationMessages.error}}
     </div>
   </div>
+
+  <div class="form-group" ng-if="vm.fiatEnabled">
+    <div class="col-md-3 sm-label-right">
+      Run As User
+      <help-field key="pipeline.config.cron.trigger.runAsUser"></help-field>
```
danielpeach: I expected to see corresponding additions to `helpContents.js`.

app/scripts/modules/core/pipeline/config/triggers/cron/cronTrigger.html
```diff
@@ -27,4 +27,19 @@
       {{vm.validationMessages.error}}
     </div>
   </div>
+
+  <div class="form-group" ng-if="vm.fiatEnabled">
```
danielpeach: It might be worth refactoring this template into a component (it could be a controller-less component).

ttomsu:
Awesome, thanks Dan!!

PR #2952, opened 2016-11-03
removes fail pipeline checkbox

tomaslin:
Removes the `Fail Pipeline` checkbox since this functionality is replicated by the `If stage fails` radio buttons in the same stage

### DIFF
--- a/app/scripts/modules/core/pipeline/config/stages/manualJudgment/manualJudgmentStage.html
+++ b/app/scripts/modules/core/pipeline/config/stages/manualJudgment/manualJudgmentStage.html
@@ -16,13 +16,6 @@
       </div>
     </stage-config-field>
   </div>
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/pipeline/config/stages/manualJudgment/manualJudgmentStage.html
```diff
@@ -16,13 +16,6 @@
       </div>
     </stage-config-field>
   </div>
-  <stage-config-field label="Fail Pipeline" help-key="pipeline.config.manualJudgment.failPipeline">
```
ajordens: should this help key also be removed?

tomaslin: yup, and the one for pipeline config too... done

ajordens:
LGTM

PR #2953, opened 2016-11-03
Remove Expression field for Check Precondition task details

tomaslin:
Suppress expression field from the UI since this is not guaranteed to be pristine in check precondition stage


tomaslin:
fixes https://github.com/spinnaker/spinnaker/issues/1235

PR #2975, opened 2016-11-10
AcaTask stage correctly fail stage when ACA fails

zanthrash:
The ACA Task trasnsformer was not considering the overall result state of the canary
when setting the state of the sage.

This also sets the stage as SUCCEDED when the canary status is marked as competed AND
the overallResult of the canary are SUCCESS.


icfantv:
LGTM

PR #2976, opened 2016-11-10
(core) account select fix

danielpeach:
@icfantv please review


danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/1294

PR #2978, opened 2016-11-10
(Openstack) Show all security groups on create server group

varikin:
When creating a server group (or a load balancer), the list of security groups to choose from should be all security groups in that account and region. This allows having common security groups platform services like logging servers, service discovery servers, etc.

To accomplish this, the `appCacheBackedMultiSelect` directive had to be updated to allow passing
in an explicit cache instead of having the directive read the cache via cache key from the application list.

### DIFF
--- a/app/scripts/modules/openstack/loadBalancer/configure/wizard/upsert.controller.js
+++ b/app/scripts/modules/openstack/loadBalancer/configure/wizard/upsert.controller.js
@@ -19,15 +19,14 @@ module.exports = angular.module('spinnaker.loadBalancer.openstack.create.control
     require('../../../subnet/subnetSelectField.directive.js'),
     require('../../../network/networkSelectField.directive.js'),
     require('../../../common/isolateForm.directive.js'),
+    require('core/securityGroup/securityGroup.read.service.js')
   ])
   .controller('openstackUpsertLoadBalancerController', function($scope, $uibModalInstance, $state,
                                                                 application, loadBalancer, isNew, loadBalancerReader,
                                                                 accountService, openstackLoadBalancerTransformer,
-                                                                loadBalancerWriter, taskMonitorService) {
+                                                                loadBalancerWriter, taskMonitorService, securityGroupReader) {
     var ctrl = this;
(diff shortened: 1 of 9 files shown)


varikin:
@spinnaker/reviewers PTAL

app/scripts/modules/openstack/loadBalancer/configure/wizard/upsert.controller.js
```diff
@@ -133,6 +137,19 @@ module.exports = angular.module('spinnaker.loadBalancer.openstack.create.control
       $scope.existingLoadBalancerNames = _.flatten(_.map(allLoadBalancerNames[$scope.loadBalancer.account || ''] || []));
     }
 
+    function updateSecurityGroups() {
+      var account = _.get($scope, ['loadBalancer', 'account']);
+      var region = _.get($scope, ['loadBalancer', 'region']);
+      if (!account || !region) {
+        $scope.allSecurityGroups = [];
+      }
+      var securityGroups = securityGroupReader.getAllSecurityGroups()
```
danielpeach: Looks like `securityGroups` isn't being used here (it's also shadowing `securityGroups` in the `.then`).

app/scripts/modules/openstack/loadBalancer/configure/wizard/upsert.controller.js
```diff
@@ -133,6 +137,19 @@ module.exports = angular.module('spinnaker.loadBalancer.openstack.create.control
...
+        .then(function(securityGroups) {
+          var filtered = _.get(securityGroups, [account, 'openstack', region], []);
+          $scope.allSecurityGroups = _.chain(filtered).sortBy('name').value();
```
danielpeach: I think this could just be `_.sortBy(filtered, 'name')`.

app/scripts/modules/openstack/loadBalancer/configure/wizard/upsert.controller.js
varikin: Ah, yes.

PR #2987, opened 2016-11-11
(core): convert retry service to TS

icfantv:


icfantv:
apparently i forgot to submit a PR for this...  @anotherchrisberry @zanthrash PTAL

app/scripts/modules/core/retry/retry.service.ts
```diff
@@ -0,0 +1,35 @@
...
+    const call: any = func();
+    const promise: ng.IPromise<any> = call.then ? call : this.$q.resolve(call);
+    if (limit === 0) {
+      return promise;
+    } else {
+      return promise.then((result: any) => {
+        if (stopCondition(result)) {
+          return result;
+        } else {
+          return this.$timeout(interval).then(() => this.buildRetrySequence(call, stopCondition, limit - 1, interval));
```
danielpeach: I believe this should be `this.buildRetrySequence(func, stopCondition, limit - 1, interval)`. I am also suspicious that the tests didn't fail.

icfantv: i'll take a look.

icfantv: @danielpeach this was a good catch.  i restored the original code as a sanity check and the `$timeout` function is never executing - in either the TS or original JS version.

icfantv: the expects aren't executing either...

danielpeach: Huh. The expects run and pass if you set the timeout interval to 0, though I'm not sure why. You could just do that, since I don't think it invalidates the tests.

icfantv:
@danielpeach what do you want to do here?  Are you comfortable leaving the test at 0?

danielpeach:
Yes, I feel comfortable with that.

icfantv:
Updated interval timing in tests to 0 to force execution of expects.

PR #2988, opened 2016-11-14
(Openstack) Adding self referencial security group rules.

varikin:


varikin:
@spinnaker/reviewers PTAL

PR #2989, opened 2016-11-14
(Openstack) Clear floatingNetworkId correctly.

varikin:
The radio button to enable adding a fip to each instance or not did not
clear out the selection. Clouddriver decides to associate a fip based on
that selection being set or not. So, clearing it when choosing not have
one.


varikin:
@spinnaker/reviewers PTAL

PR #3004, opened 2016-11-15
(cf) Add support to cloning server groups in stages

gregturn:


gregturn:
@spinnaker/reviewers PTAL

PR #3007, opened 2016-11-16
(k8s) add ui support for security context

danielpeach:
@lwander


danielpeach:
security_context

Also included in the run job stage.

lwander:
LGTM (verified)

PR #3024, opened 2016-11-22
(google) ssl load balancer wizard

danielpeach:
@duftler or @jtk54 please review.


jtk54:
To get around deleting/inserting new frontend listeners, make the `certificate` read-only after initial creation. I'll file a bug internally and ask for a workaround. Other than that, LGTM.

PR #3055, opened 2016-12-05
(appengine) create lb message

danielpeach:

screen shot 2016-12-05 at 12 26 15 pm


lwander:
The example is good, but I would supply a link to appengine's official documentation as well.

duftler:
I like it.

danielpeach:
Added a link to the app.yaml documentation.

PR #3070, opened 2016-12-10
(nflx): update dockerfile to remove postinstall

icfantv:
* with the conversion to at-types, the npm postinstall script is no
longer needed since typings are installed via project dependencies.


icfantv:
@anotherchrisberry pretty sure this is all we need to fix Quay.  PTAL.

PR #3080, opened 2016-12-14
Use newlines so that group memberships doesn't get trailing whitespaces

amanya:
Before this patch, group memberships were stored in front50 with a trailing whitespace at the end. cc. @ttomsu


ttomsu:
I recall someone else having this issue too. Thanks for fixing!

PR #3089, opened 2016-12-15
(gce) Make L7 listeners readonly except for certs.

jtk54:
This makes all the forwarding rule attributes read-only after creation for L7 upsert. Certs can still be changed to support rotation.

editable


app/scripts/modules/google/loadBalancer/configure/http/listeners/listener.component.html
```diff
@@ -1,7 +1,7 @@
 <hr ng-if="$ctrl.index > 0">
 <div class="container-fluid form-horizontal">
   <ng-form name="listener">
-    <div class="modal-body">
+    <div class="modal-body" >
```
danielpeach: Not sure this extra space was intentional.

PR #3101, opened 2016-12-16
(aws) fixing regression where ssl cert name was not show when editing a ELB

zanthrash:
@icfantv PTAL


icfantv:
This is actually correct as we won't attempt to reformat the cert if it's already formatted.  I.e., if it already contains `arn:aws:iam::` or `arn:aws:acm:`.

PR #3109, opened 2016-12-19
(gce) Warn about IP change if forwarding rule protocol or port range

jtk54:
changed. 

image


jtk54:
The warning appears if you change either the protocol or the port number in the input forms.

PR #3130, opened 2016-12-29
(appengine) enable stage

danielpeach:
@duftler or @lwander please review.

### DIFF
--- a/app/scripts/modules/appengine/pipeline/stages/enableAsg/appengineEnableAsgStage.ts
+++ b/app/scripts/modules/appengine/pipeline/stages/enableAsg/appengineEnableAsgStage.ts
@@ -4,7 +4,7 @@ import {IAppengineStageScope} from 'appengine/domain/index';
 import {ACCOUNT_SERVICE, AccountService} from 'core/account/account.service';
 import {AppengineStageCtrl} from '../appengineStage.controller';
 
-class AppengineDisableAsgStageCtrl extends AppengineStageCtrl {
+class AppengineEnableAsgStageCtrl extends AppengineStageCtrl {
   static get $inject() { return ['$scope', 'accountService']; }
 
   constructor(protected $scope: IAppengineStageScope, protected accountService: AccountService) {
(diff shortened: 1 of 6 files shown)


app/scripts/modules/appengine/pipeline/stages/enableAsg/appengineEnableAsgStage.ts
```diff
@@ -0,0 +1,49 @@
+import {module} from 'angular';
+
+import {IAppengineStageScope} from 'appengine/domain/index';
+import {ACCOUNT_SERVICE, AccountService} from 'core/account/account.service';
+import {AppengineStageCtrl} from '../appengineStage.controller';
+
+class AppengineEnableAsgStageCtrl extends AppengineStageCtrl {
+  static get $inject() { return ['$scope', 'accountService']; }
+
+  constructor(public $scope: IAppengineStageScope, accountService: AccountService) {
```
icfantv: I think these should these be `private`?

icfantv:
Just the ctor access modifier question and this LGTM.

app/scripts/modules/appengine/pipeline/stages/enableAsg/appengineEnableAsgStage.ts
danielpeach: Please correct me if I'm wrong, but I think they should be `protected` -- I'm calling a base class below, and the compiler complains if these are `private`, regardless of the access modifiers on the base class.

icfantv: For posterity, the presence of access modifiers tells the compiler to create a local reference of the ctor arg.  As expected, the complied code makes no distinction with the access modifiers, thus the modifiers are purely for compile-time checking.  Omitting the access modifier will _NOT_ create a class-level field reference.

In NG2, the access modifier of class properties referenced in component templates must be non-`private` or the compiler will complain as the component templates are essentially turned into TS classes.  This is not an issue in NG1, but for easier migration, any class-level fields referenced in templates should not be `private`.

In general, the most restrictive access for fields, methods, classes should be chosen - even though this same level of security will not carry over to the transpiled code.

PR #3131, opened 2016-12-29
(appengine) start and stop tasks

danielpeach:
@duftler or @lwander please review. Depends on https://github.com/spinnaker/orca/pull/1103 and https://github.com/spinnaker/clouddriver/pull/1345.

I can't see any use case for stopping a server group that is serving traffic, but it's consistent with what the cloud console UI and gcloud allow. I've included a warning message.

screen shot 2016-12-29 at 3 24 59 pm

### DIFF
--- a/app/scripts/modules/appengine/serverGroup/writer/serverGroup.write.service.ts
+++ b/app/scripts/modules/appengine/serverGroup/writer/serverGroup.write.service.ts
@@ -1,0 +1,59 @@
+import {module} from 'angular';
+
+import {TASK_EXECUTOR, ITaskCommand, TaskExecutor, IJob} from 'core/task/taskExecutor';
+import {ITask} from 'core/task/task.read.service';
+import {Application} from 'core/application/application.model';
+import {IAppengineServerGroup} from 'appengine/domain/index';
+
+interface IAppengineServerGroupWriteJob extends IJob {
+  serverGroupName: string;
+  region: string;
+  credentials: string;
+  cloudProvider: string;
+}
+
+export class AppengineServerGroupWriter {
+  static get $inject() { return ['taskExecutor']; }
+
+  constructor(private taskExecutor: TaskExecutor) {}
+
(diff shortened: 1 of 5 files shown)


app/scripts/modules/appengine/serverGroup/writer/serverGroup.write.service.ts
```diff
@@ -0,0 +1,60 @@
...
+      application: application,
+      description: `Stop Server Group: ${serverGroup.name}`,
+    };
+
+    return this.taskExecutor.executeTask(command);
+  }
+
+  private buildJob(serverGroup: IAppengineServerGroup, application: Application, type: string): IAppengineServerGroupWriteJob {
+    return {
+      type: type,
```
icfantv: can just be `type`.

app/scripts/modules/appengine/serverGroup/writer/serverGroup.write.service.ts
```diff
@@ -0,0 +1,60 @@
...
+
+    return this.taskExecutor.executeTask(command);
+  }
+
+  public stopServerGroup(serverGroup: IAppengineServerGroup, application: Application): IPromise<ITask> {
+    let job = this.buildJob(serverGroup, application, 'stopAppEngineServerGroup');
+
+    let command: ITaskCommand = {
+      job: [job],
+      application: application,
```
icfantv: can just be `application,`

app/scripts/modules/appengine/serverGroup/writer/serverGroup.write.service.ts
```diff
@@ -0,0 +1,60 @@
...
+  static get $inject() { return ['taskExecutor']; }
+
+  constructor(private taskExecutor: TaskExecutor) {}
+
+  public startServerGroup(serverGroup: IAppengineServerGroup, application: Application): IPromise<ITask> {
+    let job = this.buildJob(serverGroup, application, 'startAppEngineServerGroup');
+
+    let command: ITaskCommand = {
+      job: [job],
+      application: application,
```
icfantv: ditto

PR #3154, opened 2017-01-10
(appengine) dynamic branch from webhook

danielpeach:
@lwander or @duftler please review. Depends on https://github.com/spinnaker/orca/pull/1112

screen shot 2017-01-10 at 10 34 40 am

### DIFF
--- a/app/scripts/modules/appengine/serverGroup/configure/wizard/basicSettings.html
+++ b/app/scripts/modules/appengine/serverGroup/configure/wizard/basicSettings.html
@@ -75,11 +75,34 @@
         Branch <help-field key="appengine.serverGroup.branch"></help-field>
       </div>
       <div class="col-md-7">
-        <input type="text"
+        <input ng-if="!command.fromTrigger"
+               type="text"
                required
                class="form-control input-sm"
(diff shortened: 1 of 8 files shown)


app/scripts/modules/appengine/serverGroup/configure/wizard/basicSettings.html
```diff
@@ -113,15 +138,7 @@ <h5 class="text-center">
...
-                There is already a server group in this cluster. Do you want to clone it?
-              </p>
-              <p>
-                Cloning copies the entire configuration from the selected server group, allowing you
-                to modify whichever fields (e.g. image) you need to change in the new server group.
-              </p>
-              <p>
-                To clone a server group, select "Clone" from the "Server Group Actions" menu in the details view of the
-                server group.
+                There is already a server group in this cluster.
```
lwander: Why remove that help-text?

danielpeach: I don't think there will be a clone operation for App Engine, which is why I removed these lines. But I think that would be best done in a different PR.

PR #3190, opened 2017-01-17
(aws) params should have default value in security group upsert

danielpeach:
@icfantv some of the places where `upsertSecurityGroup` is called do not call it with `params`, which causes an error. 

Also, many of the places that do call it with `params` don't match the `ISecurityGroupJob` interface... not sure if you'd prefer I massage the callsites or leave `params` as `any`.


icfantv:
I think this is fine.  Thanks for catching.

PR #3197, opened 2017-01-18
(bake) Hide empty parens if there is no short description specified f…

duftler:
…or a given base os option.

### DIFF
--- a/app/scripts/modules/azure/pipeline/stages/bake/bakeStage.html
+++ b/app/scripts/modules/azure/pipeline/stages/bake/bakeStage.html
@@ -18,7 +18,7 @@
                ng-model="stage.baseOs"
                ng-value="baseOsOption.id"
                ng-change="bakeStageCtrl.baseOsChanged()"/>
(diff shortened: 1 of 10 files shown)


duftler:
@danielpeach please review.

app/scripts/modules/azure/pipeline/stages/bake/bakeStage.html
```diff
@@ -18,7 +18,7 @@ <h4 ng-if="viewState.loading" class="text-center">
                ng-model="stage.baseOs"
                ng-value="baseOsOption.id"
                ng-change="bakeStageCtrl.baseOsChanged()"/>
-        {{baseOsOption.id}} ({{baseOsOption.shortDescription}}) <help-field content="{{baseOsOption.detailedDescription}}" />
+        {{bakeStageCtrl.getBaseOsDescription(baseOsOption)}} <help-field content="{{baseOsOption.detailedDescription}}" />
```
danielpeach: I think there should be a corresponding change in `azureBakeStage.js`.

duftler: Good catch, thanks.

PR #3211, opened 2017-01-23
(appengine) edit load balancer stage

danielpeach:
@jtk54 or @lwander or @duftler please review.

This is a first pass at an App Engine Edit Load Balancer stage. This PR does not include JS/HTML for the execution details.

edit_lb_stage_config
lb_select_modal
edit_wizard

### DIFF
--- a/app/scripts/modules/appengine/domain/IAppengineLoadBalancer.ts
+++ b/app/scripts/modules/appengine/domain/IAppengineLoadBalancer.ts
@@ -1,6 +1,7 @@
 import {ILoadBalancer} from 'core/domain/index';
 
 export interface IAppengineLoadBalancer extends ILoadBalancer {
+  credentials?: string;
   split: IAppengineTrafficSplit;
   migrateTraffic: boolean;
 }
--- a/app/scripts/modules/appengine/domain/IAppengineServerGroup.ts
+++ b/app/scripts/modules/appengine/domain/IAppengineServerGroup.ts
@@ -5,6 +5,7 @@ export interface IAppengineServerGroup extends ServerGroup {
   env: 'FLEXIBLE' | 'STANDARD';
   scalingPolicy: IAppengineScalingPolicy;
   servingStatus: 'SERVING' | 'STOPPED';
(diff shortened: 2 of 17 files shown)


app/scripts/modules/appengine/domain/IAppengineLoadBalancer.ts
```diff
@@ -1,6 +1,7 @@
 import {ILoadBalancer} from 'core/domain/index';
 
 export interface IAppengineLoadBalancer extends ILoadBalancer {
+  credentials?: string;
   split: IAppengineTrafficSplit;
   migrateTraffic: boolean;
```
jtk54: I'd add a warning banner in the wizard if the server group setup doesn't match the setup required for `migrateTraffic`.

lwander:
This should probably be dependent on https://github.com/spinnaker/deck/blob/master/settings.js#L114

danielpeach:
@lwander The "Create Load Balancer" stage doesn't work for GCE or Kubernetes (not sure if it ever has), so it would be unfortunate to open up a whole set of broken views for a user that's using App Engine with GCE/Kubernetes. I could add a separate flag under `providers.appengine.defaults`.

lwander:
Makes sense - what are your thoughts on breaking down this distinction between lbs & server groups @duftler?

danielpeach:
I guess the only other thing to consider is that, as a pipeline stage, editing an App Engine load balancer is more akin to an Enable Server Group/Disable Server Group/Disable Cluster stage than an upsert load balancer operation for other providers - it's just giving control over which server groups are enabled.

danielpeach:
I'm going to put this behind an App Engine only feature flag unless anyone objects.

jtk54:
SGTM

app/scripts/modules/appengine/domain/IAppengineLoadBalancer.ts
danielpeach: The `migrateTraffic` checkbox is now disabled in the cases where it would be invalid. From the help text, it should be clear to the user why this is so.

PR #3216, opened 2017-01-24
(core): pull angular-cron-gen library internally

icfantv:
* pull angular-cron-gen library internally to deck to fix several issues with the library.
* the latest version of the library will not allow angular expressions for the new `template-url` property as it uses the `templateUrl` function which executes before the scope is initialized so any scoped property value is not accessible.  this means that our template value set via a controller property is not accessible.  this is a long-standing issue [in the core library](https://github.com/angular/angular.js/issues/2895) and it doesn't look to be fixed anytime soon.
* an hourly cron trigger set in the UI to execute start at a specific time actually generates a daily cron expression so the portion of the hourly cron trigger that allows starting at a specific time was removed from deck's cron picker template.
* the regex used by the library to match hourly cron expressions was incorrectly matching daily cron expressions which was resulting the UI displaying `hourly` instead of `daily` in the select so the regex for hourly was fixed to properly test for matching hourly cron expressions.

### DIFF
--- a/app/scripts/modules/core/pipeline/config/triggers/cron/cronPicker.html
+++ b/app/scripts/modules/core/pipeline/config/triggers/cron/cronPicker.html
@@ -2,96 +2,76 @@
   <div class="row">
     <div class="col-md-12">
       <select class="form-control input-sm"
-              ng-model="state.activeTab"
-              ng-change="regenerateCron(state)"
+              ng-model="$ctrl.activeTab"
+              ng-change="$ctrl.regenerateCron()"
               ng-options="tab for tab in ['minutes', 'hourly', 'daily', 'weekly', 'monthly', 'advanced']"></select>
     </div>
   </div>
   <div class="cron-gen-container">
-    <div ng-if="state.activeTab === 'minutes'">
-      <div class="row" ng-init="regenerateCron(state)">
+    <div ng-if="$ctrl.activeTab === 'minutes'">
+      <div class="row" ng-init="$ctrl.regenerateCron()">
         <div class="col-md-12">
           <input class="form-control input-sm"
                  type="number"
                  min="1"
                  max="59"
-                 ng-change="regenerateCron(state)"
-                 ng-model="state.minutes.minutes"
-                 ng-required="state.activeTab === 'minutes'">
-          minute<span ng-if="state.minutes.minutes > 1">s</span>
+                 ng-change="$ctrl.regenerateCron()"
+                 ng-model="$ctrl.state.minutes.minutes"
+                 ng-required="$ctrl.activeTab === 'minutes'">
+          minute<span ng-if="$ctrl.state.minutes.minutes > 1">s</span>
         </div>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


icfantv:
@anotherchrisberry @danielpeach @zanthrash @cfieber PTAL

anotherchrisberry:
If the `templateUrl` can't be dynamically assigned in the controller, we could probably do it by putting the custom template into the `$templateCache`, then referencing it as a string in the directive, e.g.

```javascript
}).run($templateCache => {
    $templateCache.put('spinnaker-custom-cron-picker-template', $templateCache.get(require('./cronPicker.html')));
  });
```

```html
<cron-gen ng-model="vm.trigger.cronExpression"
                template-url="spinnaker-custom-cron-picker-template"></cron-gen>
```

app/scripts/modules/core/pipeline/config/triggers/cron/cronPicker.html
```diff
@@ -41,30 +41,6 @@
...
-                  ng-options="hour as padNumber(hour) for hour in selectOptions.hours"
-                  ng-required="state.activeTab === 'hourly' && state.hourly.subTab === 'specific'">
-          </select>
-          :
-          <select class="form-control input-sm"
-                  ng-change="regenerateCron(state)"
-                  ng-model="state.hourly.specific.minutes"
-                  ng-options="minute as padNumber(minute) for minute in selectOptions.minutes"
-                  ng-required="state.activeTab === 'hourly' && state.hourly.subTab === 'specific'">
-          </select>
```
anotherchrisberry: It looks like the newer version of this plugin allows users to specify the minutes, e.g. "Every 2 hour(s) on minute 30"; thinking we should do the same

icfantv:
@anotherchrisberry yea, i think that would work RE the template cache.  we'll also need to update our custom template as the internal API has changed.

anotherchrisberry:
When bringing an external library into the build, it's a good idea to very explicitly annotate with comments which lines have been changed, so we are better able to track the drift between our code and the source library. Looking at the source here, I can't tell what's different from the original library.

icfantv:
good idea, re lib documenting.  comments alone could be confusing.

icfantv:
i'll try my hand at using the latest version and updating the internal template.  the latest code has the bug fix around hourly cron expressions.

ajordens:
I'd also suggest a first commit that imports the library, and a second commit that introduces the changes so that we can follow the changes through source control.

icfantv:
@ajordens yep.  all good ideas.

icfantv:
@anotherchrisberry updated our template to match new API and add the exciting new features.  PTAL.

icfantv:
Adding screenshots
screen shot 2017-01-24 at 4 35 58 pm
screen shot 2017-01-24 at 4 36 23 pm

anotherchrisberry:
Nice, I really like the new options.

icfantv:
Yea, definitely pretty cool.

PR #3219, opened 2017-01-24
(core) adds logout button

danielpeach:
@anotherchrisberry @icfantv adds log out button.

Looks like this:
logout

### DIFF
--- a/app/scripts/modules/core/authentication/authentication.initializer.service.ts
+++ b/app/scripts/modules/core/authentication/authentication.initializer.service.ts
@@ -47,11 +47,7 @@ export class AuthenticationInitializer {
   private loginNotification(): void {
     this.authenticationService.authenticationExpired();
     this.userLoggedOut = true;
-    this.$uibModal.open({
-      templateUrl: require('./loggedOut.modal.html'),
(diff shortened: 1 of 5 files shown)


icfantv:
LGTM, but since NF doesn't use this functionality, I'm going to defer to @anotherchrisberry.

As far as L&F goes, if you envision more account-level menu items then I would say a menu is fine.  But if it's only ever going to be Log Out, you might consider just putting button next to the user name to keep it simple and visible.  Just my two cents.

anotherchrisberry:
Agree with what @icfantv said. I imagine we'll include more in the menu someday, so this is fine by me.

ttomsu:
Can you also confirm it will appear when the window width is smaller? I usually see an online of a head, with a pointer cursor when hovering that doesn't do anything on click.

re: NFLX doesn't use this feature: You may want it when Fiat comes online.

danielpeach:
@ttomsu updated / fixed the user menu for small window widths.

small_user_menu

ttomsu:
Awesome! Thank you!

danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/1363

PR #3221, opened 2017-01-25
Adding SSL docker configuration

emjburns:
Set env variable `DECK_CERT=/pat/to/cert.crt` to configure apache with SSL. Must also set `DECK_KEY=/path/to/key.key`. Without these env vars set, apache starts up normally.

My custom settings.js wasn't getting picked up properly when I started my docker container, so moving settings.js to the directory that the ui is now served from (/opt/deck/html) if a custom file exists in /opt/spinnaker/config.

@lwander take a look please?

### DIFF
--- a/docker/run-apache2.sh
+++ b/docker/run-apache2.sh
@@ -3,6 +3,8 @@
 _DECK_HOST=$DECK_HOST
 _DECK_PORT=$DECK_PORT
 _API_HOST=$API_HOST
+_DECK_CERT_PATH=$DECK_CERT
+_DECK_KEY_PATH=$DECK_KEY
 
 if [ -z "$_DECK_HOST" ];
 then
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ WORKDIR deck
 
 RUN docker/setup-apache2.sh
 
-RUN ./gradlew build -PskipTests
+RUN git config --global url.https://github.com/.insteadOf git://github.com/ && ./gradlew build -PskipTests
```
lwander: What is github.com/.insteadOf?

docker/run-apache2.sh
```diff
@@ -3,6 +3,8 @@
 _DECK_HOST=$DECK_HOST
 _DECK_PORT=$DECK_PORT
 _API_HOST=$API_HOST
+_DECK_CERT=$DECK_CERT
```
lwander: These are paths, right? It might be best to name these `DECK_CERT_PATH` & `DECK_KEY_PATH`, in case we want to enable putting the keys directly into the env var in the future.

Dockerfile
emjburns: Oops - it's a workaround for our security policies. We only allow https on corporate network. Removing.

docker/run-apache2.sh
emjburns: Yes, they're paths. From what I understand, these env vars (https://github.com/spinnaker/deck/blob/master/webpack.common.js#L101) are supposed to be paths, because `readFileSync` takes a path as the argument (https://nodejs.org/api/fs.html#fs_fs_readfilesync_file_options). Is this understanding correct?

So, I've renamed the local variables but not the env variables, because I don't want to break existing usages.

lwander: Ah darn, if the existing env vars depend on them being paths there's not much we can do.

PR #3234, opened 2017-01-28
(google) Fix up create/edit security groups.

duftler:


duftler:
Closes https://github.com/spinnaker/spinnaker/issues/1392.

duftler:
@danielpeach please review.

duftler:
The target and source of `Object.assign()` were swapped. Since that resulted in the order of precedence being reversed, the decorated `sourceRanges` was not overwritten by the flattened `sourceRanges`. Had a similar effect on `type` for the Edit operation.

PR #3237, opened 2017-01-30
(core) run as user field indicates when user does not have access to …

danielpeach:
…service account

@ttomsu please review.

Closes https://github.com/spinnaker/fiat/issues/139


ttomsu:
Can you include a warning message if the service account listed in the saved pipeline config is not contained in the user's service account list?

PR #3245, opened 2017-02-01
(netflix): fix spelling/compiler errors

icfantv:
* pin angular mocks version to unbreak test build


icfantv:
@anotherchrisberry PTAL.  should be good to merge once CI build finishes.

PR #3248, opened 2017-02-02
update startup scripts

danielpeach:
@ewiseblatt or @anotherchrisberry 

This should clear up most of the existing issues with running Deck either locally or on our vms.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,19 +1,37 @@
 #!/bin/bash
 
+if [[ -z "$npm_package_engines_node" ]]; then
+  echo "You are probably running this script without NPM. Try running 'npm start' from the Deck directory root."
+  exit
+fi
+
 NODE_VERSION=$(node -v)
 
-if [[ $NODE_VERSION != $npm_package_engines_node ]]; then
-  if [[ -f $HOME/.nvm/nvm.sh ]]; then
-    echo "Updating your node version to $npm_package_engines_node..."
-    . $HOME/.nvm/nvm.sh
-    nvm use $npm_package_engines_node
+function update_node() {
+  echo "Updating your node version to $npm_package_engines_node..."
+  nvm use $npm_package_engines_node
+  if [[ $? != 0 ]]; then
+    echo "Installing node $npm_package_engines_node..."
+    nvm install $npm_package_engines_node
+  fi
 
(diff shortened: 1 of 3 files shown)


start.sh
```diff
@@ -1,19 +1,37 @@
...
+    echo $npm_package_engines_node > $PWD/.nvmrc
+  elif [[ $(cat $PWD/.nvmrc) != $npm_package_engines_node ]]; then
+    echo "WARNING: Your .nvmrc specifies node $(cat $PWD/.nvmrc)."
+  fi
+}
+
 if [[ $NODE_VERSION != $npm_package_engines_node ]]; then
-  if [[ -f $HOME/.nvm/nvm.sh ]]; then
-    echo "Updating your node version to $npm_package_engines_node..."
+  if [[ -f /etc/profile.d/nvm.sh ]]; then
```
ewiseblatt: maybe, but should the precedence be reversed so a user trumps an installation?

start.sh
```diff
@@ -1,19 +1,37 @@
...
+if [[ -z "$npm_package_engines_node" ]]; then
+  echo "You are probably running this script without NPM. Try running 'npm start' from the Deck directory root."
+  exit
+fi
+
 NODE_VERSION=$(node -v)
 
+function update_node() {
+  echo "Updating your node version to $npm_package_engines_node..."
+  nvm use $npm_package_engines_node
```
ewiseblatt: I prefer if ! nvm use ..

start.sh
```diff
@@ -1,19 +1,37 @@
 #!/bin/bash
 
+if [[ -z "$npm_package_engines_node" ]]; then
+  echo "You are probably running this script without NPM. Try running 'npm start' from the Deck directory root."
+  exit
```
ewiseblatt: exit with failure

icfantv:
Just my one question.  I don't know that it's always a good idea to show a warning message to a user if the version they're using is _higher_ than the minimum.

icfantv:
@danielpeach hey, what's the status on this?

danielpeach:
I'm going to close this for now, since it doesn't fully solve what I've been intending to do, which is require that `npm install` runs with the proper version of node.

PR #3249, opened 2017-02-02
(appengine) ad hoc clone server group

danielpeach:
@lwander or @jtk54 or @duftler please review.

Not sure if this is a great idea, but I think there should be a convenient way of deploying into the same cluster for App Engine, even if we can't copy over the rest of the server group's configuration.

clone


jtk54:
Maybe instead of a help text '?', just display the help if the user hovers over 'Clone' to force them to look at it (not just if they're curious). LGTM otherwise. Needs a little rebase.

PR #3256, opened 2017-02-03
(netflix) Add Isolated Testing Target stage

jrsquared:


app/scripts/modules/netflix/pipeline/stage/isolatedTestingTarget/isolatedTestingTargetStage.ts
```diff
@@ -0,0 +1,252 @@
...
+            'authenticationService', 'cloudProviderRegistry', 'serverGroupCommandBuilder',
+            'serverGroupReader', 'awsServerGroupTransformer'];
+  }
+
+  constructor(private $scope: any,
+              private $uibModal: IModalService,
+              private stage: IIsolatedTestingTargetStage,
+              private namingService: NamingService,
+              private providerSelectionService: any,
+              public authenticationService: AuthenticationService,
```
icfantv: i don't see this used outside the ctor, i think we can remove the access modifier completely.

PR #3259, opened 2017-02-04
Adding image name, tag and organization to the docker bake stage

brujoand:
These changes are related to (but not dependent on) this PR https://github.com/spinnaker/rosco/pull/172

The motivation here is to add support for specifying common parameters in the docker bake stage. Not quite sure about the fields in `bakeExecutionDetails.html`, so please comment if someone has strong opinions on this.


lwander:
Sweet - could you add a picture of the new bake UI?

brujoand:
Sure thing 👍 
Bake Configuration
Bake Details

PR #3271, opened 2017-02-07
Slim container & bump spinnaker gradle project version

ttomsu:
@cfieber @duftler @anotherchrisberry FYI


aarondobbing:
The latest deck:master image is broken based on this change in the dockerfile.
run-apache2.sh is deleted, but required to start container. Have rectified in the PR linked below @ttomsu :)

https://github.com/spinnaker/deck/pull/3274

PR #3273, opened 2017-02-08
(gce): Support for setting/updating named ports.

jtk54:
Frontend support for setting named ports. Depends on https://github.com/spinnaker/clouddriver/pull/1420.

### DIFF
--- a/app/scripts/modules/google/loadBalancer/configure/http/backendService/backendService.component.js
+++ b/app/scripts/modules/google/loadBalancer/configure/http/backendService/backendService.component.js
@@ -16,6 +16,9 @@ module.exports = angular.module('spinnaker.deck.gce.httpLoadBalancer.backendServ
     controller: function () {
       this.backingData = this.command.backingData;
       this.loadBalancer = this.command.loadBalancer;
+      if (!this.backendService.namedPort) {
+        this.backendService.namedPort = "http";
+      }
       let servicesByName = this.backingData.backendServicesKeyedByName;
 
       this.onBackendServiceSelect = (selectedBackendService) => {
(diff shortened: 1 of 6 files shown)


app/scripts/modules/google/loadBalancer/configure/http/backendService/backendService.component.js
```diff
@@ -16,6 +16,9 @@ module.exports = angular.module('spinnaker.deck.gce.httpLoadBalancer.backendServ
     controller: function () {
       this.backingData = this.command.backingData;
       this.loadBalancer = this.command.loadBalancer;
+      if (!this.backendService.namedPort) {
+        this.backendService.namedPort = "http";
```
danielpeach: The double quotes here has probably upset the linter.

app/scripts/modules/google/loadBalancer/configure/ssl/gceCreateSslLoadBalancer.controller.ts
```diff
@@ -216,6 +216,7 @@ class SslLoadBalancerCtrl extends CommonGceLoadBalancerCtrl implements ng.ICompo
   private initializeEditMode (): void {
     this.loadBalancer.portRange = this.loadBalancer.portRange.split('-')[0];
     this.loadBalancer.certificate = this.loadBalancer.certificate.split('/').pop();
+    this.loadBalancer.backendService.namedPort = this.loadBalancer.backendService.namedPort ? this.loadBalancer.backendService.namedPort : 'http';
```
danielpeach: In JS, you can use `||` where you would use `?:` in groovy.

danielpeach:
I would hold off on merging this until you have a corresponding PR for configuring the port name mapping in the server group wizard (and in Clouddriver), since someone who created a load balancer with a port name other than `http` would not be able to use it right now.

jtk54:
This feature is on hold for now, closing.

PR #3291, opened 2017-02-10
Fix bad if statement

brujoand:
My JS foo is not great, but I'm guessing this is a typo


duftler:
Closed via cherry-pick in https://github.com/spinnaker/deck/pull/3292.

PR #3294, opened 2017-02-10
(cloudfoundry) Convert to V2ModalWizard

zanthrash:
The original modal wizard directives are being killed.  These files were converted to the V2ModalWizard.

@gregturn 
Please test and merge this PR as soon as you can.


gregturn:
LGTM

PR #3301, opened 2017-02-13
(k8s) adds support for kuberntes config map as volume source

danielpeach:
@lwander please review.

Closes https://github.com/spinnaker/spinnaker/issues/1215

I would like to find a better way to handle the overlap between the Run Job stage config and the Deploy config, but it isn't obvious how to do so since there are a bunch of slight differences (especially in the templates)
configmap
.


jgangemi:
is there an eta on when this will be available as part of a release? i am in need of this functionality.

thanks!!

danielpeach:
@jgangemi looks like it was released in Deck 2.999.0

jgangemi:
@danielpeach i am all up-to-date according to `apt-get` and i don't see this as an option.

also - do you know if it's possible to reference `ConfigMap` values as environment variables yet?

PR #3303, opened 2017-02-13
(netflix) Add back Fast Property key/value search.

zanthrash:
Searching for property keys was prominant in the old UI and got forgotten in the new.

This adds it back via the filter mechinism.

It also cleans up the filter tags when a new search is done on the global view.


icfantv:
LGTM.

PR #3311, opened 2017-02-16
(netflix): set z-index on blesk alerts

icfantv:
* set a `z-index` on the blesk alerts so that the dropdown nav menu will
appear on top of it, instead of behind.


icfantv:
Before:
screen shot 2017-02-15 at 10 25 41 pm

After:
screen shot 2017-02-15 at 10 26 09 pm

anotherchrisberry:
This works, but it would have made more sense to set the z-index here: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/netflix/blesk/bleskOverrides.css#L1

icfantv:
Happy to move it.

PR #3317, opened 2017-02-18
Add feature for user to check their own roles

PerGon:
Allows the possibility for users to check their own roles so it's easier to use the `groupMembership` feature of FIAT.

PTAL @danielpeach @anotherchrisberry @icfantv @jrsquared 
Thanks

Some screenshots:
image

### DIFF
--- a/app/scripts/modules/core/authentication/authentication.initializer.service.ts
+++ b/app/scripts/modules/core/authentication/authentication.initializer.service.ts
@@ -9,6 +9,7 @@ import {AUTHENTICATION_SERVICE, AuthenticationService} from './authentication.se
 
 interface IAuthResponse {
   username: string;
+  roles?: string[];
 }
 
 export class AuthenticationInitializer {
@@ -37,7 +38,11 @@ export class AuthenticationInitializer {
     this.$http.get(this.settings.authEndpoint)
       .then((response: ng.IHttpPromiseCallbackArg<IAuthResponse>) => {
         if (response.data.username) {
-          this.authenticationService.setAuthenticatedUser(response.data.username);
+          this.authenticationService.setAuthenticatedUser({
+            name: response.data.username,
+            authenticated: false,
+            roles: response.data.roles
+          });
           this.$uibModalStack.dismissAll();
           this.visibilityWatch.unsubscribe();
         }
@@ -75,7 +80,11 @@ export class AuthenticationInitializer {
     this.$http.get(this.settings.authEndpoint)
       .then((response: ng.IHttpPromiseCallbackArg<IAuthResponse>) => {
         if (response.data.username) {
-          this.authenticationService.setAuthenticatedUser(response.data.username);
+          this.authenticationService.setAuthenticatedUser({
+            name: response.data.username,
+            authenticated: false,
+            roles: response.data.roles
+          });
           this.$rootScope.authenticating = false;
         } else {
           this.loginRedirect();
@@ -89,7 +98,11 @@ export class AuthenticationInitializer {
       this.$http.get(this.settings.authEndpoint)
         .then((response: ng.IHttpPromiseCallbackArg<IAuthResponse>) => {
           if (response.data.username) {
-            this.authenticationService.setAuthenticatedUser(response.data.username);
+            this.authenticationService.setAuthenticatedUser({
+              name: response.data.username,
+              authenticated: false,
+              roles: response.data.roles
+            });
             this.$rootScope.authenticating = false;
           } else {
             this.loginNotification();
--- a/app/scripts/modules/core/authentication/authentication.service.ts
+++ b/app/scripts/modules/core/authentication/authentication.service.ts
@@ -3,6 +3,7 @@ import {module} from 'angular';
 export interface IUser {
   name: string;
   authenticated: boolean;
+  roles?: string[];
   lastAuthenticated?: number;
 }
 
@@ -10,6 +11,7 @@ export class AuthenticationService {
 
   private user: IUser = {
     name: '[anonymous]',
+    roles: [],
     authenticated: false
   };
 
@@ -19,11 +21,12 @@ export class AuthenticationService {
     return Object.assign({}, this.user);
   }
 
-  public setAuthenticatedUser(authenticatedUser: string): void {
-    if (authenticatedUser) {
-      this.user.name = authenticatedUser;
+  public setAuthenticatedUser(authenticatedUser: IUser): void {
+    if (authenticatedUser && authenticatedUser.name) {
+      this.user.name = authenticatedUser.name;
       this.user.authenticated = true;
       this.user.lastAuthenticated = new Date().getTime();
+      this.user.roles = authenticatedUser.roles;
     }
 
     this.authEvents.forEach((event: Function) => event());
--- a/app/scripts/modules/core/application/modal/groupMembershipConfigurer.component.html
+++ b/app/scripts/modules/core/application/modal/groupMembershipConfigurer.component.html
@@ -8,7 +8,7 @@
                style="width: 200px;"
                ng-model="$ctrl.requiredGroupMembership">
       <ui-select-match>{{$item}}</ui-select-match>
-      <ui-select-choices repeat="choice in []">
+      <ui-select-choices repeat="choice in $ctrl.availableRoles | orderBy">
           {{choice}}
       </ui-select-choices>
     </ui-select>
--- a/app/scripts/modules/core/application/modal/groupMembershipConfigurer.component.js
+++ b/app/scripts/modules/core/application/modal/groupMembershipConfigurer.component.js
@@ -10,7 +10,10 @@ module.exports = angular.module('spinnaker.deck.core.applicationModal.groupMembe
       requiredGroupMembership: '='
     },
     templateUrl: require('./groupMembershipConfigurer.component.html'),
-    controller: function (settings) {
+    controller: function (settings, authenticationService) {
+
+      this.availableRoles = authenticationService.getAuthenticatedUser().roles;
+
       this.fiatEnabled = settings.feature.fiatEnabled;
       if (!this.fiatEnabled) {
         return;
--- a/app/scripts/modules/core/authentication/authentication.service.spec.ts
+++ b/app/scripts/modules/core/authentication/authentication.service.spec.ts
@@ -17,11 +17,15 @@ describe('authenticationService', function() {
       let user: IUser = authenticationService.getAuthenticatedUser();
       expect(user.name).toBe('[anonymous]');
       expect(user.authenticated).toBe(false);
-
-      authenticationService.setAuthenticatedUser('kato@example.com');
+      authenticationService.setAuthenticatedUser({
+        name: 'kato@example.com',
+        authenticated: false,
+        roles: ['roleA', 'RoleB']
+      });
       user = authenticationService.getAuthenticatedUser();
       expect(user.name).toBe('kato@example.com');
       expect(user.authenticated).toBe(true);
+      expect(user.roles).toEqual(['roleA', 'RoleB']);
     });
 
     it('disregards falsy values', function() {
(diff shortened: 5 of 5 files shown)


app/scripts/modules/core/authentication/authentication.initializer.service.ts
```diff
@@ -70,12 +71,23 @@ export class AuthenticationInitializer {
...
+    });
+  }
+
   public authenticateUser() {
     this.$rootScope.authenticating = true;
     this.$http.get(this.settings.authEndpoint)
       .then((response: ng.IHttpPromiseCallbackArg<IAuthResponse>) => {
         if (response.data.username) {
-          this.authenticationService.setAuthenticatedUser(response.data.username);
+          this.authenticationService.setAuthenticatedUser(response.data.username, response.data.roles);
```
icfantv: I think a better solution here is to create a new `interface` type to pass in that encapsulates both pieces of data (since they are tightly-coupled) - such that the `roles` property is optional.

app/scripts/modules/core/authentication/authentication.service.ts
```diff
@@ -3,13 +3,15 @@ import {module} from 'angular';
 export interface IUser {
   name: string;
   authenticated: boolean;
+  roles: string[];
```
icfantv: should be optional, `roles?: string[]` - also, `string` is fine for now, but I could see this being a more complex type at some point.

app/scripts/modules/core/authentication/userMenu/checkUserRolesModal.module.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+let angular = require('angular');
+
+module.exports = angular.module('spinnaker.core.authentication.userMenu.checkUserRolesModal.directive', [])
+  .controller('CheckUserRolesModalCtrl', function($uibModalInstance, userRoles) {
+
+    this.userRoles = userRoles;
+
```
icfantv: whitespace

app/scripts/modules/core/authentication/userMenu/checkUserRolesModal.module.js
```diff
@@ -0,0 +1,12 @@
...
+    this.cancel = $uibModalInstance.dismiss;
+
```
icfantv: ditto

app/scripts/modules/core/authentication/userMenu/checkUserRolesModal.module.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
```
icfantv: This should be a Typescript class.

app/scripts/modules/core/authentication/userMenu/userMenu.directive.js
```diff
@@ -18,8 +18,9 @@ module.exports = angular.module('spinnaker.core.authentication.userMenu.directiv
       link: function(scope) {
         scope.authEnabled = settings.authEnabled;
         scope.user = authenticationService.getAuthenticatedUser();
-        scope.showLogOutDropdown = () => authenticationService.getAuthenticatedUser().authenticated;
+        scope.showLoggedUserMenuDropdown = () => authenticationService.getAuthenticatedUser().authenticated;
```
icfantv: Probably could come up with a better name here.  What do you think about just `showUserMenu`?

icfantv:
@PerGon could you also please elaborate on the business need for such a feature?  Thanks.

PerGon:
Hi @icfantv, thank you for your comments. I'll fix the issues and update the PR asap

Regarding the business need, it goes something like this:
We are soon to roll out fiat and fiat relies on the roles of users to restrict access to applications and accounts.
We use SAML to authenticate our users into Spinnaker, and Spinnaker is fed with the user's roles that are passed via the SAML assertion - but the roles are never exposed to the users. Typically the Identity Providers (in our case, Okta) do not expose the user's groups (in spinnaker we call it roles) to the user, it's something internal for the system administrators to manage.
I don't know how other authentication mechanisms supported by Spinnaker work, but I wonder if it's easy for users to identify their own roles so that they can protect their applications/accounts.

With this feature, users can easily check wich roles they have and, based on that, correctly assign those roles to the applications/accounts. Without it, there's no easy way for users to do it.
Not sure if I exposed it clearly :) let me know if you want more details!

app/scripts/modules/core/authentication/userMenu/userRolesModal.html
```diff
@@ -0,0 +1,18 @@
+<div modal-page class="flex-fill">
+  <modal-close dismiss="$dismiss()"></modal-close>
+  <div class="modal-header">
+    <h3>Roles</h3>
+  </div>
+  <div class="modal-body flex-fill">
+    <div ng-if="!ctrl.userRoles.length" >
+      No roles
+    </div>
+    <div ng-repeat="role in ctrl.userRoles | orderBy:'toString()'">
```
danielpeach: Not sure why you need `toString()` here, since the values are already strings. You can supply '+' or '-' for ascending/descending.

app/scripts/modules/core/authentication/userMenu/userRolesModal.controller.ts
```diff
@@ -0,0 +1,18 @@
+import {module} from 'angular';
+import {IModalServiceInstance} from 'angular-ui-bootstrap';
+
+class UserRolesModalCtrl {
+
+  static get $inject() { return ['$uibModalInstance', 'userRoles']; }
+
+  constructor ( private $uibModalInstance: IModalServiceInstance,
```
danielpeach: Not a big deal, but we generally omit the whitespace you have here (i.e., should be `constructor(private: ...)`

app/scripts/modules/core/authentication/userMenu/userRolesModal.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+                public userRoles: string[]) { }
+
+  public cancel (): void {
```
danielpeach: Same comment as above: `public cancel(): void {`

app/scripts/modules/core/authentication/authentication.initializer.service.ts
```diff
@@ -70,12 +67,23 @@ export class AuthenticationInitializer {
     this.redirectService.redirect(`${this.settings.gateUrl}/auth/redirect?to=${callback}`);
   }
 
+  public openUserRolesModal() {
```
danielpeach: I don't think this really belongs here. I know this service manages the logout modal, but I don't think it needs to be aware of the user roles modal. I would move this into the user menu directive.

app/scripts/modules/core/authentication/authentication.initializer.service.ts
```diff
@@ -5,11 +5,8 @@ import {IModalService, IModalStackService} from 'angular-ui-bootstrap';
 
 import {IDeckRootScope} from '../domain/deckRootScope';
 import {REDIRECT_SERVICE, RedirectService} from './redirect.service';
-import {AUTHENTICATION_SERVICE, AuthenticationService} from './authentication.service';
+import {AUTHENTICATION_SERVICE, AuthenticationService, IAuthResponse} from './authentication.service';
 
-interface IAuthResponse {
```
danielpeach: It feels odd to pull this interface out of this service, since this service interacts with the API directly (whereas the Authentication Service does not). Is there a good reason for moving this interface that I'm not seeing?

app/scripts/modules/core/authentication/userMenu/userRolesModal.controller.ts
danielpeach: Also, it doesn't look like you're using this in the modal's template. Please remove if it's not being used.

app/scripts/modules/core/authentication/userMenu/userRolesModal.controller.ts
danielpeach: Also, it doesn't look like you're using this in the modal's template. Please remove if you don't need it.

PerGon:
Hi @danielpeach 
Thank you for the comments - I'll address them asap! You are right, the modal is a poor solution. I'll put it as an autocomplete feature to the `groupMembership` when editing applications.

app/scripts/modules/core/authentication/authentication.initializer.service.ts
icfantv: Agree with @danielpeach here.  Unless the interface is used application-wide, it should go in the same file where it's primarily used.

PerGon:
I've done the suggested changes (updated the PR description with the new screenshot).

Not sure if it's the best idea, but I wanted to avoid creating yet another interface, so I re-used the `IAuthResponse` - I'm not sure if it's ok for the `authentication.service` to depend on the `authentication.initializer.service`. Thoughts?
@danielpeach @icfantv 
Thanks

danielpeach:
@PerGon looks great - thanks again. I see no issue with the dependency, but will defer to @icfantv.

jrsquared:
Looks much better!

icfantv:
@PerGon i'm not crazy about creating that circular dependency between the `AuthenticationService` and the `AuthenticationInitializer` when `AuthenticationService` had no extraneous dependency before and was agnostic as to how it was being used, which was essentially, as a data container.

What do you think about having the `AuthenticationInitializer` import and construct `IUser` from `AuthenticationService`, passing that object to the `setAuthenticatedUser` method on `AuthenticationService` instead?

PerGon:
Ah, I thought about that, but using the `IUser` interface won't work without additional changes. The auth API expects a parameter called `username` (like in the IAuthResponse), but the "same" field in `IUser` is called `name`. I didn't feel like having two parameters for the same thing or going around to the rest of the code and change all references of `name` to `username`. But I'm ok with doing it if you think it's best :)

In fact, I've tried importing `IUser` to `AuthenticationInitializer` at some point, adding the username variable (yes, being duplicated with name) - but that got lost in the rebasing when I changed my mind :)
I'll be happy to use `IUser` in `AuthenticationInitializer`, just let me know if I should add a new variable `username` or rename `name` altogether.

app/scripts/modules/core/authentication/authentication.initializer.service.ts
```diff
@@ -37,7 +38,7 @@ export class AuthenticationInitializer {
     this.$http.get(this.settings.authEndpoint)
       .then((response: ng.IHttpPromiseCallbackArg<IAuthResponse>) => {
         if (response.data.username) {
-          this.authenticationService.setAuthenticatedUser(response.data.username);
+          this.authenticationService.setAuthenticatedUser(response.data);
```
icfantv: @PerGon E.g., why can't this be:
```
const data: any = response.data;
this.authenticationService.setAuthenticatedUser({
  name: data.username,
  roles: data.roles
});
```

app/scripts/modules/core/authentication/authentication.service.ts
```diff
@@ -19,11 +22,12 @@ export class AuthenticationService {
     return Object.assign({}, this.user);
   }
 
-  public setAuthenticatedUser(authenticatedUser: string): void {
+  public setAuthenticatedUser(authenticatedUser: IAuthResponse): void {
```
icfantv: ...and then change the signature here to be of type `IUser`....

app/scripts/modules/core/authentication/authentication.service.ts
```diff
@@ -19,11 +22,12 @@ export class AuthenticationService {
...
     if (authenticatedUser) {
-      this.user.name = authenticatedUser;
+      this.user.name = authenticatedUser.username;
```
icfantv: ...and then update this to just be `authenticatedUser.name`.

PerGon:
Thanks for all the feedback! It's been a while since I last messed around with JS/TS :)

@icfantv does it look ok?

icfantv:
LGTM.  @anotherchrisberry any additional thoughts?

icfantv:
@PerGon please rebase and we can get this merged.  thanks.

PerGon:
It's done! Thanks @icfantv

PR #3318, opened 2017-02-18
added new stage failure option to allow other branches to complete bu…

robfletcher:
…t stage to fail

See https://github.com/spinnaker/orca/pull/1186


jrsquared:
LGTM

PR #3325, opened 2017-02-23
Add a "force" checkbox to the cancel pipeline modal dialog

jervi:
* Adds a `force=true` parameter to the backend request if checked
* Corresponds with the following PR’s in Gate and Orca:
  * https://github.com/spinnaker/orca/pull/1188
  * https://github.com/spinnaker/gate/pull/345
* The checkbox will only show if the execution engine is `v2`

### DIFF
--- a/app/scripts/modules/core/cancelModal/cancel.html
+++ b/app/scripts/modules/core/cancelModal/cancel.html
@@ -1,0 +1,49 @@
+<div modal-page class="cancel-modal">
+  <form role="form" class="container-fluid no-padding" ng-submit="ctrl.confirm()">
+    <modal-close dismiss="$dismiss()"></modal-close>
+    <div class="modal-header">
+      <h3>{{params.header}}</h3>
+    </div>
+    <div class="modal-body cancel-modal">
+      <div ng-if="params.body" ng-bind-html="params.body"></div>
+      <div ng-if="state.error">
+        <div class="alert alert-danger">
+          <h4>An exception occurred:</h4>
+          <p>
+            {{errorMessage || 'No details provided.'}}
+          </p>
+        </div>
+      </div>
+      <task-reason command="params"></task-reason>
+      <div class="force-cancel" ng-if="params.forceable">
+        <div class="row">
+          <div class="col-md-3 sm-label-right">
+            <label for="force">Force</label>
+          </div>
+          <div class="col-md-7">
+            <input id="force" type="checkbox" ng-model="params.force">
+          </div>
+        </div>
+        <div class="row" ng-if="params.forceable && params.force">
+          <div class="col-md-push-3 col-md-7">
+            <div class="alert alert-info" role="alert">
+              <strong>Note:</strong> This option will force the pipeline to cancel. It should only be used if the task or pipeline hangs.
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+    <div class="modal-footer">
+      <button class="btn btn-default"
+              type="button"
+              ng-click="ctrl.cancel()">{{params.cancelButtonText}}</button>
+      <button class="btn btn-primary"
+              type="submit"
+              ng-disabled="state.submitting || ctrl.formDisabled()"
+              ng-click="ctrl.confirm()">
+        <button-busy-indicator ng-if="state.submitting"></button-busy-indicator>
+        {{params.buttonText}}
+      </button>
+    </div>
+  </form>
+</div>
--- a/app/scripts/modules/core/cancelModal/cancelModal.controller.ts
+++ b/app/scripts/modules/core/cancelModal/cancelModal.controller.ts
@@ -1,0 +1,44 @@
+import {module} from 'angular';
+import {IModalServiceInstance} from 'angular-ui-bootstrap';
+import IScope = ng.IScope;
+
+require('./cancel.less');
+
+interface ICancelModalScope extends IScope {
+  state: any;
+  params: any;
+  errorMessage: string;
+}
+
+export class CancelModalCtrl {
+
+  constructor(public $scope: ICancelModalScope, private $uibModalInstance: IModalServiceInstance, private params: any) {
+    this.$scope.params = params;
+
+    this.$scope.state = {
+      submitting: false
+    };
+  }
+
+  formDisabled = () => this.$scope.state.submitting;
+
+  showError(exception: string): void {
+    this.$scope.state.error = true;
+    this.$scope.state.submitting = false;
+    this.$scope.errorMessage = exception;
+  }
+
+  confirm(): void {
+    if (!this.formDisabled()) {
+      this.$scope.state.submitting = true;
+      this.params.submitMethod(this.params.reason, this.params.force).then(this.$uibModalInstance.close, this.showError);
+    }
+  };
+
+  cancel = () => this.$uibModalInstance.dismiss();
+}
+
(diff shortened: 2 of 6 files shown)


duftler:
Can you please include a screen-grab on the PR?

jervi:
Force checkbox
Checked

anotherchrisberry:
The confirmation modal is meant to be general purpose. I'd much rather see this done by using a custom modal and controller from `execution.directive.js` than wedging it into the confirmation modal.

jervi:
You mean that we should not use the confirmation modal at all for the cancel functionality, but rather replace it with a custom modal?

anotherchrisberry:
Yeah, exactly. Let me know if you need help or have any questions when building it out.

jervi:
Any input on the UX btw?

danielpeach:
"cancel a pipeline using force" sounds a little off to me. What about "This option will force the pipeline to cancel."?

jervi:
Thank you for feedback!
@danielpeach:
New text

@anotherchrisberry I've reverted the changes to the confirmationModal, and implemented a new cancelModal. Any feedback is welcomed :)

app/scripts/modules/core/cancelModal/cancelModal.controller.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
```
icfantv: Can you please convert this to Typescript?

app/scripts/modules/core/cancelModal/cancel.html
```diff
@@ -0,0 +1,51 @@
+<div modal-page class="cancel-modal">
+  <form role="form" class="container-fluid no-padding" ng-submit="ctrl.confirm()">
+    <modal-close (dismiss)="$dismiss()"></modal-close>
```
icfantv: We had to pull ng2 in a subsequent commit for performance reasons so this needs to be flipped back to ng1 syntax, i.e., just drop the `()` around the `dismiss` attribute.

app/scripts/modules/core/cancelModal/cancel.html
```diff
@@ -0,0 +1,51 @@
...
+      <div ng-if="state.error">
+        <div class="alert alert-danger">
+          <h4>An exception occurred:</h4>
+          <p>
+            {{errorMessage || 'No details provided.'}}
+          </p>
+        </div>
+      </div>
+      <task-reason command="params"></task-reason>
+
```
icfantv: whitespace

app/scripts/modules/core/cancelModal/cancel.html
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+        <div class="row" ng-if="params.forceable && params.force">
+          <div class="col-md-push-3 col-md-7">
+            <div class="alert alert-info" role="alert">
+              <strong>Note:</strong> This option will force the pipeline to cancel. It should only be used if the task or pipeline hangs.
+            </div>
+          </div>
+        </div>
+      </div>
+
```
icfantv: whitespace

app/scripts/modules/core/cancelModal/cancelModal.controller.js
```diff
@@ -0,0 +1,34 @@
...
+    this.formDisabled = () => $scope.state.submitting;
+
+    function showError(exception) {
+      $scope.state.error = true;
+      $scope.state.submitting = false;
+      $scope.errorMessage = exception;
+    }
+
+    this.confirm = () => {
+      if (!this.formDisabled()) {
```
icfantv: You will also need to check the submitting state here because it's possible for the UI to not be updated before the user [accidentally] clicks the button a second time - rendering your `ng-disabled` worthless.

app/scripts/modules/core/cancelModal/cancelModal.service.ts
```diff
@@ -0,0 +1,45 @@
+import {module} from 'angular';
+import {IModalService, IModalSettings} from 'angular-ui-bootstrap';
+
+export interface ICancelModalParams {
```
icfantv: Are all these properties really optional?

app/scripts/modules/core/cancelModal/cancelModal.service.ts
```diff
@@ -0,0 +1,45 @@
...
+      templateUrl: require('./cancel.html'),
+      controller: 'CancelModalCtrl as ctrl',
+      resolve: {
+        params: () => extendedParams
+      }
+    };
+
+    return this.$uibModal.open(modalArgs).result;
+  }
+
```
icfantv: whitespace

app/scripts/modules/core/cancelModal/cancelModal.service.ts
```diff
@@ -0,0 +1,45 @@
...
+  public confirm(params: ICancelModalParams): ng.IPromise<any> {
+    const extendedParams: ICancelModalParams = Object.assign({}, this.defaults, params);
+
+    if (extendedParams.body) {
+      extendedParams.body = this.$sce.trustAsHtml(extendedParams.body);
+    }
+
+    const modalArgs: IModalSettings = {
+      templateUrl: require('./cancel.html'),
+      controller: 'CancelModalCtrl as ctrl',
```
icfantv: You can use the `controllerAs` property instead of making this inline.

app/scripts/modules/core/cancelModal/cancelModal.controller.js
jervi: Thanks, but I'm not entirely sure what you mean by this. The `formDisabled` function is just `() => $scope.state.submitting;`. Do I have to check the submitting state another way as well?

jervi:
@icfantv Thank you for all the comments. Please have a look at the updated PR; especially the typescript conversion (I'm a TypeScript newbie :)

anotherchrisberry:
I think this is fine, just needs to be rebased and squashed into a single commit.

app/scripts/modules/core/cancelModal/cancelModal.controller.ts
```diff
@@ -0,0 +1,44 @@
...
+  formDisabled = () => this.$scope.state.submitting;
+
+  showError(exception: string): void {
+    this.$scope.state.error = true;
+    this.$scope.state.submitting = false;
+    this.$scope.errorMessage = exception;
+  }
+
+  confirm(): void {
+    if (!this.formDisabled()) {
```
icfantv: I meant that you should also have a check here for whether or not the form is already being submitted in your if statement.  Currently you are relying on the angular digest cycle to disable the button in your modal and you can't do this in a deterministic way.

jervi: You mean like this: `if (!this.formDisabled() && !this.$scope.state.submitting) {`? I don't understand how that will make any difference, as the `formDisabled()` function just returns the value of `this.$scope.state.submitting`. You might argue that the whole `formDisabled` function could be replaced with submitting state though, both here and in the html...

icfantv: GAH!  I'm sorry.  I totally missed the `()` on the check.  It's good.

jervi:
Thanks for the feedback @icfantv, @anotherchrisberry and @danielpeach ! Squashed and rebased.

PR #3332, opened 2017-02-27
Scope Fast Property to multiple apps

zanthrash:
- Downstream orca change needed to make this work (https://github.com/spinnaker/orca/pull/1192), but this can go in before.
- Made changes to the way scopes are laid out.
- Selected scopes can also be edited to allow for scoping Fast Properties to ASGs that have yet to be created.  This is useful for FP that are needed at startup.

mahe_ _fast_properties_ _fast_property_details__undefined_undefined_

@anotherchrisberry @icfantv @jrsquared PTAL


jrsquared:
LGTM

PR #3334, opened 2017-02-27
(appengine) adds full set of locators for upsert app engine pipeline …

danielpeach:
…stage

When editing an "Edit Load Balancer" stage, you should now see this:

allocation_configuratoin


danielpeach:
Depends on https://github.com/spinnaker/clouddriver/pull/1467 and https://github.com/spinnaker/orca/pull/1194

PR #3342, opened 2017-03-01
 (netflix): update canary UI

icfantv:
* update canary UI to have a more sensible layout
* fixed various linting issues.
* added debug config option for turning off extraneous console noise
* turn of auto reload of application via WDS.

### DIFF
--- a/app/scripts/modules/core/help/helpField.component.ts
+++ b/app/scripts/modules/core/help/helpField.component.ts
@@ -11,6 +11,7 @@ class HelpFieldCtrl implements ng.IComponentController {
 
   public content: string;
   public expand: boolean;
+  public label: string;
   public contents: IHelpFieldContents;
   public displayPopover: boolean;
   public popoverTemplate: string = require('./helpField.popover.html');
@@ -87,21 +88,35 @@ class HelpFieldComponent implements ng.IComponentOptions {
     content: '@',
     placement: '@',
     expand: '=',
+    label: '@'
   };
   public controller: any = HelpFieldCtrl;
   public template = `
-    <div ng-if="$ctrl.expand && $ctrl.contents.content"
-         class="help-contents small"
-         ng-bind-html="$ctrl.contents.content"></div>
-    <a href class="help-field" ng-if="!$ctrl.expand && $ctrl.contents.content"
-            uib-popover-template="$ctrl.popoverTemplate"
-            ng-mouseenter="$ctrl.showPopover()"
-            ng-mouseleave="$ctrl.hidePopover(true)"
-            popover-placement="{{$ctrl.contents.placement}}"
-            popover-is-open="$ctrl.displayPopover"
-            popover-trigger="none">
-      <span class="small glyphicon glyphicon-question-sign"></span>
-    </a>
+    <div class="text-only" ng-if="$ctrl.label">
+      <a href class="help-field" ng-if="!$ctrl.expand && $ctrl.contents.content"
+              uib-popover-template="$ctrl.popoverTemplate"
+              ng-mouseenter="$ctrl.showPopover()"
+              ng-mouseleave="$ctrl.hidePopover(true)"
+              popover-placement="{{$ctrl.contents.placement}}"
+              popover-is-open="$ctrl.displayPopover"
+              popover-trigger="none"
+              ng-bind-html="$ctrl.label">
+      </a>
+    </div>
+    <div ng-if="!$ctrl.label" style="display: inline-block;">
+      <div ng-if="$ctrl.expand && $ctrl.contents.content"
+           class="help-contents small"
+           ng-bind-html="$ctrl.contents.content"></div>
+      <a href class="help-field" ng-if="!$ctrl.expand && $ctrl.contents.content"
+              uib-popover-template="$ctrl.popoverTemplate"
+              ng-mouseenter="$ctrl.showPopover()"
+              ng-mouseleave="$ctrl.hidePopover(true)"
+              popover-placement="{{$ctrl.contents.placement}}"
+              popover-is-open="$ctrl.displayPopover"
+              popover-trigger="none">
+        <span class="small glyphicon glyphicon-question-sign"></span>
+      </a>
+    </div>
   `;
 }
 
--- a/app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
+++ b/app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
@@ -5,11 +5,13 @@ let angular = require('angular');
 import {CLOUD_PROVIDER_REGISTRY} from 'core/cloudProvider/cloudProvider.registry';
 import {SERVER_GROUP_COMMAND_BUILDER_SERVICE} from 'core/serverGroup/configure/common/serverGroupCommandBuilder.service';
 import {LIST_EXTRACTOR_SERVICE} from 'core/application/listExtractor/listExtractor.service';
+import {CANARY_SCORE_CONFIG_COMPONENT} from './canaryScore.component';
 
 module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage', [
   LIST_EXTRACTOR_SERVICE,
   CLOUD_PROVIDER_REGISTRY,
   SERVER_GROUP_COMMAND_BUILDER_SERVICE,
+  CANARY_SCORE_CONFIG_COMPONENT,
   require('core/pipeline/config/pipelineConfigProvider'),
   require('core/config/settings.js'),
 ])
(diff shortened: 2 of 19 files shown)


icfantv:
@anotherchrisberry @jrsquared @danielpeach @zanthrash PTAL

I had to dump the page navigator sections due to an unclear race condition that I was never able to pin down.  Let's talk about how we want the section headers to look as the ones in there now are not great.

Also, please take a look at the field labels and help icon/text values to see if any of the grammar/verbage should be tweaked.

new canary ui

app/scripts/modules/core/pipeline/config/pipelineConfig.less
```diff
@@ -141,10 +141,6 @@ pipeline-configurer {
     margin-bottom: 0;
   }
 
-  .stage-details {
```
icfantv: Crap, I think this was erroneously removed.  Checking now and will re-add.

app/scripts/modules/core/help/helpField.component.ts
```diff
@@ -87,21 +88,37 @@ class HelpFieldComponent implements ng.IComponentOptions {
...
-            uib-popover-template="$ctrl.popoverTemplate"
-            ng-mouseenter="$ctrl.showPopover()"
-            ng-mouseleave="$ctrl.hidePopover(true)"
-            popover-placement="{{$ctrl.contents.placement}}"
-            popover-is-open="$ctrl.displayPopover"
-            popover-trigger="none">
-      <span class="small glyphicon glyphicon-question-sign"></span>
-    </a>
+    <div style="display: inline-block;">
+      <div class="text-only" ng-if="$ctrl.label">
```
jrsquared: `inline-block` wrapped in an empty `inline-block`, one of which is specified with `style`, the other with `class`. Could probably just put the `ng-if` on the other div and remove one layer of divs?

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -33,6 +35,84 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+          {type: 'requiredField', fieldName: 'baseline.account', fieldLabel: 'Account'},
+          {type: 'requiredField', fieldName: 'baseline.cluster', fieldLabel: 'Cluster'},
+          {type: 'requiredField', fieldName: 'clusterPairs', fieldLabel: 'Cluster Pairs'},
+          {type: 'requiredField', fieldName: 'canary.canaryConfig.canaryAnalysisConfig.name', fieldLabel: 'Configuration'},
+          {type: 'custom', fieldLabel: 'Lookback Duration', validate: (_pipeline, stage) => {
+            const cac = stage.canary.canaryConfig.canaryAnalysisConfig,
+              useLookback = cac.useLookback,
+              lookbackMins = cac.lookbackMins;
+            let result = null;
+            if (useLookback && (!lookbackMins || lookbackMins === 0)) {
```
jrsquared: if `lookbackMins === 0`, it would already have failed at `!lookbackMins`

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -33,6 +35,84 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+
+            return result;
+          }},
+          {
+            type: 'custom', validate: (_pipeline, stage) => {
+            let result = null;
+            if (stage.scaleUp.enabled) {
+              const delay = stage.scaleUp.delay;
+              if (!delay || (delay < 0)) {
+                result = 'When a canary scale-up is enabled, the delay value must be non-negative.';
```
jrsquared: Conditional is falsey for `0`, but message implies `0` is valid.

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -33,6 +35,84 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+              result = 'When an analysis type of <strong>Sliding Lookback</strong> is selected, the lookback duration must be positive.';
+            }
+
+            return result;
+          }},
+          {type: 'custom', fieldLabel: 'Report Frequency', validate: (_pipeline, stage) => {
+            const reportFrequency = stage.canary.canaryConfig.canaryAnalysisConfig.canaryAnalysisIntervalMins;
+            let result = null;
+            if (!reportFrequency || (reportFrequency < 1)) {
+              result = 'The <strong>Report Frequency</strong> is required and must be positive.';
```
jrsquared: 0 > x > 1 would fail the test, but would satisfy the message. Should either check for `!(reportFrequency > 0)` or fix message to be clear.

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -33,6 +35,84 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+            }
+
+            return result;
+          }},
+          {
+            type: 'custom', fieldLabel: 'Successful Score', validate: (_pipeline, stage) => {
+            const successfulScore = stage.canary.canaryConfig.canarySuccessCriteria.canaryResultScore;
+            let result = null;
+            if (!successfulScore || (successfulScore < 1)) {
+              result = 'The <strong>Successful Score</strong> is required and must be positive.';
```
jrsquared: same

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -33,6 +35,84 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+            }
+
+            return result;
+          }},
+          {
+            type: 'custom', fieldLabel: 'Unhealthy Score', validate: (_pipeline, stage) => {
+            const unhealthyScore = stage.canary.canaryConfig.canaryHealthCheckHandler.minimumCanaryResultScore;
+            let result = null;
+            if (!unhealthyScore || (unhealthyScore < 1)) {
```
jrsquared: same

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -33,6 +35,84 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+
+            return result;
+          }},
+          {
+            type: 'custom', validate: (_pipeline, stage) => {
+            let result = null;
+            if (stage.scaleUp.enabled) {
+              const capacity = stage.scaleUp.capacity;
+              if (!capacity || (capacity < 1)) {
+                result = 'When a canary scale-up is enabled, the capacity value must be positive.';
```
jrsquared: same

app/scripts/modules/core/help/helpField.component.ts
icfantv: So it actually applied to the outer `div` but it was really only necessary on the second `div` if the `label` property was NOT defined, so I moved it.

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
icfantv: So what I had did work, but your suggestion is certainly cleaner (and also works).  The input is of type `number` with a `min` set to `1`.  Not sure I understand your range inequality, how can something be less than `0` but greater than `1`?  Unless I'm misreading it.

jrsquared: Ha! Correct. Typo on my part. 0 < x < 1. Even with the min set to 1, I still like validators to do a complete validation, just in case someone changes the control, the validation of the underlying field still stands. The message implies that 0.5 is a valid value, so I'd want to validator to let that pass. Or change the message to '1 or greater' to properly capture the necessary value.

icfantv:
Headers updated per JT.

anotherchrisberry:
LGTM

icfantv:
@anotherchrisberry updated to only apply defaults if stage is new.  @csanden PTAL if you wish.

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -42,27 +123,51 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+    stage.canary = stage.canary || {};
+    stage.canary.owner = stage.canary.owner || (user.authenticated ? user.name : null);
+    stage.canary.watchers = stage.canary.watchers || [];
+
+    stage.canary.canaryConfig = stage.canary.canaryConfig || {name: [$scope.pipeline.name, 'Canary'].join(' - ')};
+    stage.canary.canaryConfig.canaryAnalysisConfig = stage.canary.canaryConfig.canaryAnalysisConfig || {};
+
+    if (!stage.scaleUp) {
+      stage.scaleUp.enabled = false;
+    }
```
anotherchrisberry: does this ever get evaluated? seems like #L129 guarantees it won't.

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -42,27 +123,51 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+
+    const cc = stage.canary.canaryConfig;
+    if (cc.actionsForUnhealthyCanary && cc.actionsForUnhealthyCanary.find(action => action.action === 'TERMINATE')) {
```
anotherchrisberry: `some` is probably more semantically correct here than `find`

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -42,27 +123,51 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
+      set(cc, 'canaryHealthCheckHandler.minimumCanaryResultScore', 75);
+      set(cc, 'canarySuccessCriteria.canaryResultScore', 95);
+      cac.notificationHours = [1, 2, 3];
+      cac.canaryAnalysisIntervalMins = 30;
+      cac.useLookback = false;
+      cac.lookbackMins = 0;
+      cac.useGlobalDataset = false;
+      cc.actionsForUnhealthyCanary = [{action: 'DISABLE'}, {action: 'TERMINATE', delayBeforeActionInMins: 60}];
+      this.terminateUnhealthyCanaryEnabled = true;
+      cac.beginCanaryAnalysisAfterMins = 0;
```
anotherchrisberry: It might read easier if the defaults were set as a new object here, rather than a field at a time, e.g.
```javascript
this.terminateUnhealthyCanaryEnabled = true;
scope.stage.canaryConfig = {
  lifetimeHours: 3,
  canaryHealthCheckHandler: {
    minimumCanaryResultScore: 75
  },
  canarySuccessCriteria: {
    canaryResultScore: 95
  },
  actionsForUnhealthyCanary: [
    { action: 'DISABLE' },
    { action: 'TERMINATE, delayBeforeActionInMins: 60 }
  ],
  canaryAnalysisConfig: {
    combinedCanaryResultStrategy: 'AGGREGATE',
    notificationHours: [1, 2, 3],
    canaryAnalysisIntervalMins: 30,
    useLookback: false,
    lookbackMins: 0,
    useGlobalDataset: false,
    beginCanaryAnalysisAfterMins: 0
  }
};
```

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
```diff
@@ -92,28 +196,25 @@ module.exports = angular.module('spinnaker.netflix.pipeline.stage.canaryStage',
...
         $scope.stage.canary.canaryConfig.actionsForUnhealthyCanary = [
           {action: 'DISABLE'}
         ];
       }
 
-      return _.find($scope.stage.canary.canaryConfig.actionsForUnhealthyCanary, function (action) {
-        return action.action === 'TERMINATE';
-      }) !== undefined;
+      return $scope.stage.canary.canaryConfig.actionsForUnhealthyCanary
+          .find((action) => action.action === 'TERMINATE') !== undefined;
```
anotherchrisberry: replace `find(...) !== undefined` with `some(...)`

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
icfantv: Yup, good catch.

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
icfantv: Agreed.  Good call.

app/scripts/modules/netflix/pipeline/stage/canary/canaryStage.js
icfantv: 👍

icfantv:
Updated per recommendations and additional, unnecessary code resulting from implemented recommendations, removed.

PR #3352, opened 2017-03-02
feat(gce) adds support for firewall rule source and target tags

danielpeach:
@duftler please review.

Closes https://github.com/spinnaker/spinnaker/issues/1448

help_text

new_wizard

details_view

### DIFF
--- a/app/scripts/modules/core/help/helpContents.js
+++ b/app/scripts/modules/core/help/helpContents.js
@@ -135,6 +135,9 @@ module.exports = angular.module('spinnaker.core.help.contents', [])
       incoming traffic is directed through a named port (for Spinnaker, the named port is <b>http</b>).
       The mapping from named port to port number is specified per server group
       and can be configured within the server group creation dialogue under <b>Port Name Mapping</b>.`,
+    'gce.securityGroup.sourceCIDRs': 'Traffic is only allowed from sources that match one of these CIDR ranges, or one of the source tags above.',
+    'gce.securityGroup.sourceTags': 'Traffic is only allowed from sources that match one of these tags, or one of the source CIDR ranges below.',
+    'gce.securityGroup.targetTags': 'Traffic is only allowed if the target instance has one of these tags.',
     'gce.serverGroup.resizeWithAutoscalingPolicy': `
       Setting the desired instance count for a server group with an autoscaler is not supported by Spinnaker;
       if the desired instance count differs from the instance count that the autoscaler wants to maintain for its configured metrics,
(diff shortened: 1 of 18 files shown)


danielpeach:
Depends on https://github.com/spinnaker/clouddriver/pull/1483

app/scripts/modules/core/help/helpContents.js
```diff
@@ -135,6 +135,9 @@ module.exports = angular.module('spinnaker.core.help.contents', [])
       incoming traffic is directed through a named port (for Spinnaker, the named port is <b>http</b>).
       The mapping from named port to port number is specified per server group
       and can be configured within the server group creation dialogue under <b>Port Name Mapping</b>.`,
+    'gce.securityGroup.sourceCIDRs': 'Traffic is only allowed from sources that match one of these CIDR ranges, or one of the source tags above.',
+    'gce.securityGroup.sourceTags': 'Traffic is only allowed from sources that match one of these tags, or one of the source CIDR ranges below.',
+    'gce.securityGroup.targetTags': 'Traffic is only allowed only if the target instance has one of these tags. If no tags are provided, all targets are allowed.',
```
duftler: Reads a bit funny with 'only' on each side of 'allowed'.

danielpeach: Whoops.

duftler:
When I have a security group with zero source range rules, the details still indicate there are '(1)': 
image

duftler:
Also, when I clone that security group with zero range rules, the modal shows an empty source range field:
image

app/scripts/modules/core/help/helpContents.js
duftler: Actually, feels like we should do something stronger here.
Before this change, a newly-created security group would have a tag automatically generated (there was no way to specify any target tags). Now, the target tags are optional, and the default behavior is 'allow traffic to all server groups'.
I'm thinking it might be better to add 2 (mutually-exclusive) checkboxes with something like 'auto-generate target tag' and 'allow traffic to all server groups' and have the target tags only permit zero entries if one of those options is selected. This way we take the safer approach for the user and also preserve the existing behavior of allowing a target tag to just be generated on the backend.

duftler:
When Cloning or Editing, it seems like I can have empty Target and Source tags and still submit the modal.

duftler:
LGTM once the 'Auto-generate target tag' option is shown for Clone of a target-tagless security group.
Please squash before merging.

PR #3354, opened 2017-03-03
refactor(CI integration): Rename Jenkins to CI

jervi:
CI integration is confusingly named Jenkins throughout Deck, even though both Jenkins and Travis are supported, and more CI’s will probably be supported in the future. This commit rename Jenkins to CI throughout the UI. There are still some references to Jenkins left in some domain objects (serverGroup.buildInfo.jenkins), but these will be addressed in a separate PR.

BREAKING CHANGE: `jenkins` must be renamed to `ci` in the `triggerTypes` section of `settings.js`. There’s also a new property `feature.convertCiToJenkinsOnPipelineSave` that should be set, but this will default to true to not break things.


app/scripts/modules/appengine/serverGroup/configure/serverGroupCommandBuilder.service.ts
```diff
@@ -66,7 +66,7 @@ export class AppengineServerGroupCommandBuilder {
     let viewState: IViewState = {
       mode: mode,
       submitButtonLabel: this.getSubmitButtonLabel(mode),
-      disableStrategySelection: mode === 'create' ? true : false,
+      disableStrategySelection: mode === 'create',
```
jervi: This change is not related to the rename, just plain old refactoring

jervi:
Some screenshots:

skjermbilde 2017-03-03 kl 11 04 06
skjermbilde 2017-03-03 kl 11 04 39

tomaslin:
Thanks for the PR.

I don't think this is the right solution. We allow Jenkins or Travis ( and other CI systems ) to be added. They should have their own top level labels given that there are separate stage types for each.

The right solution is to move Travis out of the Jenkins umbrella. It should have never been put inside of the Jenkins label and this creates unnecessary confusion. Jenkins should be its own trigger and Travis should be its own trigger. Any new additional CI integration should be its own trigger type.

PR #3360, opened 2017-03-06
fix(core) do not allow free-form text in group membership config

danielpeach:
@ttomsu @PerGon please review.

@PerGon this affects the recent work you did on this component. It prevents users from selecting roles they do not have (though the autocomplete functionality still works).

### DIFF
--- a/app/scripts/modules/core/application/modal/groupMembershipConfigurer.component.html
+++ b/app/scripts/modules/core/application/modal/groupMembershipConfigurer.component.html
@@ -2,8 +2,6 @@
   <div class="col-sm-3 sm-label-right">Group Membership</div>
   <div class="col-sm-2">
     <ui-select multiple
-               tagging
-               tagging-label=""
                class="form-control input-sm"
                style="width: 200px;"
                ng-model="$ctrl.requiredGroupMembership">


ttomsu:
LGTM

icfantv:
LGTM

PerGon:
Thanks for ping me!

The group membership choice does need some improvement - but I think this breaks one important use case.
It's very common that users that have restricted applications sometimes need some help (debugging some issue, setting up more complex pipelines, etc.) - that's the case when the users need to add the Delivery Team group to their application. These users do not belong to the `delivery team` group, but they need to add that group so we can help them.

Is there any other way we can handle such use case? How do you guys feel about it?

Cheers

danielpeach:
@PerGon yep, that seems reasonable. I'm going to revert this. I talked this over with @ttomsu - I think the proper UX is to prevent the user from updating the application if doing so would lock them out of the application (and to explain why they can't update their application).

PerGon:
Thanks @danielpeach 

From the top of my head, a few possibilities could be:

1. Allow the user to write a group to whom (s)he doesn't belong, if there's already a group which the user belongs to in the `group membership` list (preventing lockout)
2. Do not allow the user to freely write groups, but add an "Allow Admin Access" checkbox. Add a property in Deck that specifies which group is the `Admin Group`.

icfantv:
@danielpeach apologies for the quick trigger.

PR #3362, opened 2017-03-07
feat(all): ng2 test, service, and component work

icfantv:
* convert service to ng2
* convert component to ng2
* update component HTML usage to ng2 syntax
* wire in service and component downgrades
* wire in test harness changes that inject downgraded ng2 service into
ng1 test.


icfantv:
@jrsquared @danielpeach PTAL

Basically, this just reintroduces what was previously backed out.  I ran into some issues incorporating the changes from Rob and Misko and will reach out to @petebacondarwin and @gkalpak about moving forward on that front.

app/scripts/modules/core/modal/buttons/modalClose.component.ts
```diff
@@ -1,17 +1,31 @@
...
+  }
 }
 
 export const MODAL_CLOSE_COMPONENT = 'spinnaker.core.modal.modalClose.directive';
-module(MODAL_CLOSE_COMPONENT, []).component('modalClose', new ModalCloseComponent());
+export const MODAL_CLOSE_COMPONENT_DOWNGRADE: IDowngradeItem = {
+  moduleName: MODAL_CLOSE_COMPONENT,
+  injectionName: 'modalClose',
+  moduleClass: ModalCloseComponent,
+  outputs: ['dismiss']
```
jrsquared: Would be nice to have a comment in place on things that are required for upgrade that might not be well documented, and then gives us a reference point for cruft we can rip out once we can do a full ng2 cutover...

icfantv: I just had a thought: ALL of the upgrade/downgrade stuff goes away.  Basically everything associated with `IDowngradeItem`.

PR #3366, opened 2017-03-07
feat(google): Add support for labels.

duftler:


duftler:
@danielpeach please review.

duftler:
Depends on: https://github.com/spinnaker/clouddriver/pull/1501

app/scripts/modules/google/instance/details/instanceDetails.html
```diff
@@ -172,7 +172,18 @@ <h3 select-on-dbl-click>
         </dd>
       </dl>
     </collapsible-section>
-    <collapsible-section heading="SSH">
+    <collapsible-section heading="Labels">
+      <div ng-if="!instance.labels.length">No labels associated with this instance</div>
+      <dl ng-if="instance.labels.length">
+        <dt ng-repeat-start="label in instance.labels">
```
danielpeach: For here (and another template in this PR), could you do an `ng-repeat-start="(key, value) in instance.labels"`? It would save you from doing the translation from map to list in both controllers.

duftler: Done.

app/scripts/modules/google/serverGroup/details/serverGroupDetails.gce.controller.js
```diff
@@ -246,6 +247,12 @@ module.exports = angular.module('spinnaker.serverGroup.details.gce.controller',
       }
     };
 
+    let processLabels = () => {
+      if (!Object.keys(this.serverGroup.instanceTemplateLabels).length) {
```
danielpeach: Can `this.serverGroup.instanceTemplateLabels` be undefined / null here? I think `Object.keys` will throw an error if that's the case.

duftler: Good point. Switched to using `_.size()`.

duftler:
Frontend support for https://github.com/spinnaker/spinnaker/issues/1364.

PR #3369, opened 2017-03-08
fix(pipeline): update pipeline json edit check

icfantv:
* update pipeline validation to check for duplicate `refId` values when
the user edits the raw pipeline JSON.
* update the edit pipeline json controller to typescript
* convert corresponding test to typescript.


icfantv:
@jrsquared @danielpeach PTAL

app/scripts/modules/core/pipeline/config/actions/json/editPipelineJsonModal.controller.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  public $onInit(): void {
+    const copy = cloneDeepWith<IPipeline>(this.pipeline, (value: any) => {
+      if (value && value.$$hashKey) {
+        delete value.$$hashKey;
+      }
+      return undefined; // required for clone operation and typescript happiness
+    });
+    this.removeImmutableFields(copy);
```
icfantv: So, if you return `value` here, the object isn't actually copied and it's essentially a live reference to the original object.  Returning `undefined` tells lodash to actually clone the `value`.

jrsquared:
LGTM

PR #3372, opened 2017-03-09
chore(core): upgrade angular to latest

icfantv:
* upgrade angular 2 to latest RC
* clean up cruft from webpack config file


icfantv:
@danielpeach @jrsquared you'll need to run an `npm run clean && npm install`.

PR #3375, opened 2017-03-10
(netflix) Allow Fast property scope to be updated.

zanthrash:
Once a Fast Property has been set the scope can't be easily updated b/c the key of the
Property is a pipe delimited string of the property key and all the values of its scope.

To get around that limitation we create 2 Persisted property stages.  The first one will
create a new property with the new scope.  Once that completes we follow up with a delete property
stage that cleans up the original property.

When editing the property the user will be shown the original scope and they will be able to select the new scope.  If they choose to update the scope they can remove it and only an upsert property stage will be used.

@anotherchrisberry @icfantv @jrsquared  PTAL


app/scripts/modules/netflix/fastProperties/wizard/propertyScope/propertyScopeUpdatable.component.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  constructor(fastPropertyScopeSearchCategoryService: FastPropertyScopeCategoryService) {
+    fastPropertyScopeSearchCategoryService.getImpactForScope(this.command.originalScope)
+      .then((counts: IImpactCounts) => {
+        this.command.originalScope.instanceCounts = counts;
+        return this.command.originalScope;
+      });
+  }
+
+  public selectScope(scopeOption: Scope) {
```
icfantv: missing return types for methods in this class

PR #3378, opened 2017-03-10
fix(serverGroups): cluster disable warning update

icfantv:
* updated sensitivity around disabling clusters.  specifically, added
checks for corner cases such that a cluster could be disabled and the
user not given added context as to the ramifications of the disable
action.  this change now adds checks for if it's the last active cluster
or if it's the only cluster remaining.


icfantv:
@jrsquared PTAL.

PR #3386, opened 2017-03-14
chore(linter): update linting rules and code

icfantv:
* update linting rules to require access modifier on class fields and
methods for Typescript classes.


icfantv:
@danielpeach @zanthrash @jrsquared - PTAL.  requires access modifier be specified for Typescript class-level fields and methods.  It doesn't look like we can force a return type on class methods via  linting rules.

jrsquared:
👍

PR #3388, opened 2017-03-14
fix(canary): do not break on canary expressions

icfantv:
* add very rudimentary support to indicate the user that they're using
expressions for canary field values but do not allow them to edit it
directly.  rather, we want them to edit via the pipeline JSON editor.
* do not barf on loading canary stages using expressions as canary field
values.

### DIFF
--- a/app/scripts/modules/netflix/pipeline/stage/canary/canaryScore.component.ts
+++ b/app/scripts/modules/netflix/pipeline/stage/canary/canaryScore.component.ts
@@ -1,3 +1,4 @@
+import {isString} from 'lodash';
 import {module} from 'angular';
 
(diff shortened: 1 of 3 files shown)


icfantv:
@jrsquared @zanthrash PTAL.  The canary UI is the gift that keeps on giving.  This PR simply prevents the page from barfing when the canary stage encounters an expression value instead of a number and displays text to the user that they're using an expression that must be modified via the JSON editor as we kinda don't want canary values to be dynamic.

icfantv:
screen shot 2017-03-14 at 3 56 29 pm

jrsquared:
LGTM

PR #3391, opened 2017-03-15
(core, SPIN-2438) Fix Pipeline Details Toggle

zanthrash:
The "Details" toggle in the pipelines tab would open the first time you clicked it and then if
the "Details" link on another pipeline was click all the details sections of all the pipelines would close.
One would then have to re-click the "Details" link to open up the details view.

This fixs the toggle to be inline with expected behavior.


icfantv:
LGTM.

PR #3396, opened 2017-03-16
fix(core) hide capacity in deploy execution details if unnecessary

danielpeach:
@icfantv please review.

App Engine has no notion of capacity at deploy time, so there's nothing to populate here.


icfantv:
LGTM

PR #3397, opened 2017-03-16
fix(gce) prevents race condition in how credentials are set on command

danielpeach:
@ttomsu please review.


ttomsu:
Thank you! LGTM

PR #3400, opened 2017-03-17
fix(core): make org not required in docker trigger

icfantv:
* remove org validation requirement for docker triggers.


icfantv:
@tomaslin @jrsquared PTAL.  should be good to go when CI build finishes.

jrsquared:
👍

PR #3401, opened 2017-03-17
chore(core/help): make help field work in ng2

icfantv:
* add code to make the help field component work in ng2 components.
* create framework for wiring in upgraded ng1 components.


icfantv:
@jrsquared @zanthrash @danielpeach PTAL. (note that this PR splits the work already incorporated into #3399 so that PR will be updated to remove the functionality contained in _this_ PR)

Note that we don't actually touch the ng1 components when "upgrading" them for use in ng2 components.  Rather, we create a new ng2 directive that overloads the selector.

Let's hold off on merging this until #3399 is ready for merging.

jrsquared:
LGTM

PR #3403, opened 2017-03-17
chore(various): cleanup old lodash props

icfantv:
* remove unused lodash properties that weren't removed when the lodash
service was removed.


icfantv:
@jrsquared @zanthrash PTAL, just some cobweb cleanup.

jrsquared:
:shipit:

PR #3406, opened 2017-03-20
feat(gce): fetch disk defaults from clouddriver.

danielpeach:
@duftler please review.

BREAKING CHANGE: (Only for GCE). This PR depends on https://github.com/spinnaker/clouddriver/pull/1522.


app/scripts/modules/google/serverGroup/configure/serverGroupCommandBuilder.service.js
```diff
@@ -134,7 +134,7 @@ module.exports = angular.module('spinnaker.gce.serverGroupCommandBuilder.service
     function calculateOverriddenStorageDescription(instanceTypeDetails, command) {
       if (instanceTypeDetails.storage.localSSDSupported) {
         if (command.localSSDCount !== instanceTypeDetails.storage.count) {
-          command.viewState.overriddenStorageDescription = command.localSSDCount > 0 ? command.localSSDCount + '×375' : '0';
```
danielpeach: The original line results in an odd entry in the instance types table (just a `0`, rather than `0x375` in the screenshot below. I have 1 SSD disk as a default for n2-standard below).

defaults

PR #3409, opened 2017-03-22
fix(kubernetes) update user selected load balancer

tony612:
The issue is that a load balancer can't be updated. Another one will be updated or created.

There're two reasons:
1. `$scope.loadBalancer.account` will be set to the first found account among all accounts as https://github.com/spinnaker/deck/blob/7427282bfea29a779d71b2e609be9a9f2d962f7c/app/scripts/modules/kubernetes/loadBalancer/configure/wizard/upsert.controller.js#L84 shows
2. $scope.loadBalancer is actually `loadBalancer.description`. But the `description.account` in the API response is null. This is fixed in https://github.com/spinnaker/clouddriver/pull/1527


icfantv:
@tony612 please rebase and we'll get this merged.  thanks.

tony612:
@icfantv Done.

PR #3427, opened 2017-03-27
fix(core/pipeline): handle long cluster names

icfantv:
* modify the cluster column to not shove everything off the containing
viewport when the cluster name is super long.


icfantv:
Before:
screen shot 2017-03-27 at 11 52 52 am
After:
screen shot 2017-03-27 at 11 53 32 am

PR #3429, opened 2017-03-27
feat(k8s) lifecycle hooks

danielpeach:
@lwander please review.

Closes https://github.com/spinnaker/spinnaker/issues/1216

lifecyclehooks
hook_details

I'm going to make the rest of the container details have the same style, but that's for a different PR.

### DIFF
--- a/app/scripts/modules/kubernetes/container/lifecycleHook.component.ts
+++ b/app/scripts/modules/kubernetes/container/lifecycleHook.component.ts
@@ -1,0 +1,111 @@
+import {module} from 'angular';
+
+interface IKubernetesHandler {
+  type: 'EXEC' | 'HTTP';
+  execAction?: IKubernetesExecAction;
+  httpGetAction?: IKubernetesHttpGetAction;
+}
+
+interface IKubernetesExecAction {
+  commands: string[];
+}
+
+interface IKubernetesHttpGetAction {
+  path: string;
+  port: number;
+  host: string;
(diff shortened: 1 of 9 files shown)


app/scripts/modules/kubernetes/container/lifecycleHook.component.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (this.handlerEnabled) {
+      this.initializeHandler();
+    } else {
+      this.handler = null;
+      this.execActionCommandsViewValue = null;
+      this.onHandlerChange({handler: null});
+    }
+  }
+
+  public handlerTypeModelToViewValue(handlerTypeModelValue: 'EXEC' | 'HTTP'): 'exec' | 'httpGet' {
```
icfantv: If this is client-side only, I think it's totally ok to use an `enum`.  We haven't talked about an official policy, however.  It reads a lot nicer here and elsewhere in the code.

danielpeach: I agree - `enum`s are nicer to read. Unfortunately, `'EXEC' | 'HTTP'` has to be passed to Clouddriver, and `'exec' | 'httpGet'` have to be rendered.

icfantv: Bummer.  Oh well.

PR #3434, opened 2017-03-28
fix(travis stage): Tabs should be displayed in execution details

jervi:
The new Travis stage was missing tabs in the execution details:
skjermbilde 2017-03-28 kl 12 36 53

After this fix:
skjermbilde 2017-03-28 kl 12 40 49


jervi:
Superb, thanks! I've rebased.

PR #3435, opened 2017-03-28
refactor(gce): refactor http load balancer utils

danielpeach:
@jtk54 please review.

Mostly unifies logic for mapping listener -> url map names (and converts to typescript).


app/scripts/modules/google/loadBalancer/loadBalancer.setTransformer.ts
```diff
@@ -24,21 +27,21 @@ export class GceLoadBalancerSetTransformer {
...
+    return ['gceHttpLoadBalancerUtils'];
   }
 
-  constructor(private elSevenUtils: any) {}
+  constructor(private gceHttpLoadBalancerUtils: GceHttpLoadBalancerUtils) {}
 
-  public normalizeLoadBalancerSet = (loadBalancers: any[]): any[] => {
-    let [elSevenLoadBalancers, otherLoadBalancers] = partition(loadBalancers, this.elSevenUtils.isElSeven);
+  public normalizeLoadBalancerSet = (loadBalancers: IGceLoadBalancer[]): IGceLoadBalancer[] => {
+    let [httpLoadBalancers, otherLoadBalancers] = partition(loadBalancers, (lb) => this.gceHttpLoadBalancerUtils.isHttpLoadBalancer(lb));
```
jtk54: Why did this turn into an anonymous function?

danielpeach: It retains the `this` binding for the method, which doesn't matter here (since the method doesn't depend on `this`), but its probably a better practice to do this consistently.

PR #3439, opened 2017-03-28
fix(azure): Specify cloud provider when deleting lb

EricJizbaMSFT:
Deleting a load balancer was failing because 'loadBalancer.providerType' was undefined and it was defaulting to 'aws'. Also refactor to use the core loadBalancerWriter instead of our own redundant implementation


EricJizbaMSFT:
@rguthriemsft @scotmoor @clguimanmsft PTAL

rguthriemsft:
LGTM

PR #3448, opened 2017-03-29
chore(all): Upgrade angular 1.5 to 1.6

jrsquared:


app/scripts/modules/core/core.module.js
```diff
@@ -199,8 +199,21 @@ module.exports = angular
       'Content-Type': 'application/json;charset=utf-8'
     };
   })
+  .config(function($qProvider) {
+    // Angular 1.6 stops suppressing unhandle rejections on promises. This resets it back to 1.5 behavior.
+    // See https://docs.angularjs.org/guide/migration#migrate1.5to1.6-ng-services-$q
+    $qProvider.errorOnUnhandledRejections(false);
```
icfantv: so basically we're just still running 1.5 but with a 1.6 veneer.  :-)

icfantv:
@jrsquared hey, do you wanna rebase this so we can merge and do a release with just this PR so if we need to roll back, it's clean?  Thanks.

PR #3459, opened 2017-03-30
chore(core): convert search service to TS

danielpeach:
@anotherchrisberry or @icfantv or @jrsquared please review.


app/scripts/modules/core/search/search.service.js
```diff
@@ -1,42 +0,0 @@
...
-
-    function search(params) {
-      var defaultParams = {
-        pageSize: defaultPageSize
-      };
-
-      return $http({
-        url: SETTINGS.gateUrl + '/search',
-        params: angular.extend(defaultParams, params),
-        timeout: SETTINGS.pollSchedule * 2 + 5000, // TODO: replace with apiHost call
```
danielpeach: The API service uses the same timeout, so I just used `API` directly instead of `$http` here.

jrsquared:
LGTM

app/scripts/modules/core/search/search.service.ts
```diff
@@ -0,0 +1,52 @@
...
+  totalMatches?: number;
+}
+
+export class SearchService {
+
+  static get defaultPageSize() { return 500; };
+
+  static get $inject() { return ['$log', 'API']; }
+
+  constructor(private $log: ng.ILogService, private API: Api) { }
```
icfantv: We've started moving to module imports for `ng.*` types, which, I realize now was never officially communicated.  Sorry about that.  Can you please move them all up to the top?  Thanks.

app/scripts/modules/core/securityGroup/securityGroupReader.service.ts
```diff
@@ -416,15 +417,15 @@ export class SecurityGroupReader {
     });
   }
 
-  public loadSecurityGroupsByApplicationName(applicationName: string): any[] {
+  public loadSecurityGroupsByApplicationName(applicationName: string): ng.IPromise<ISecurityGroup[]> {
```
icfantv: Same with this file re `ng.*` types.  Thanks.

app/scripts/modules/core/search/search.service.ts
danielpeach: Sure - sounds good.

PR #3463, opened 2017-03-31
refactor(gce): health check and certificate services

danielpeach:
@jtk54 or @duftler please review.

Refactors these services to depend on the search service. I'm planning on killing off the separate httpHealthCheck service, and this is the first step.

@icfantv or @anotherchrisberry - I added an option into the search service to pass a cache config.


app/scripts/modules/google/certificate/certificate.reader.ts
```diff
@@ -0,0 +1,32 @@
...
+  public listCertificates(): IPromise<IGceCertificate[]> {
+    return this.searchService
+      .search<IGceCertificate>({q: '', type: 'sslCertificates'}, this.infrastructureCaches.get('certificates'))
+      .then((searchResults: ISearchResults<IGceCertificate>) => {
+        if (searchResults && searchResults.results) {
+          return searchResults.results.filter(certificate => certificate.provider === 'gce');
+        } else {
+          return [];
+        }
+      });
```
icfantv: do you want a `.catch` here to return a `[]` as well?

PR #3469, opened 2017-04-03
feat(gce): adds address reader service

danielpeach:
@jtk54 or @duftler please review.

### DIFF
--- a/app/scripts/modules/google/address/address.reader.ts
+++ b/app/scripts/modules/google/address/address.reader.ts
@@ -1,0 +1,62 @@
+import {module, IPromise} from 'angular';
+import {SEARCH_SERVICE, SearchService, ISearchResults} from 'core/search/search.service';
+import {InfrastructureCacheService, INFRASTRUCTURE_CACHE_SERVICE} from 'core/cache/infrastructureCaches.service';
+
+interface IAddressSearchResults {
+  account: string;
+  address: string; // JSON encoded string containing the address. It is missing the account.
+  name: string;
+  provider: string;
+  region: string;
+  type: string;
+}
+
+export interface IGceAddress {
+  account: string;
+  address: string;
+  creationTimestamp: string;
+  description: string;
+  id: number;
+  kind: string;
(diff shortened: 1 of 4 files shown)


app/scripts/modules/google/address/address.reader.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+class GceAddressReader {
+  static get $inject() { return ['searchService', 'infrastructureCaches']; }
+
+  constructor (private searchService: SearchService, private infrastructureCaches: InfrastructureCacheService) {}
+
+  public listAddresses(region?: string): IPromise<IGceAddress[]> {
+    if (region) {
+      return this.listAddresses().then(addresses => addresses.filter(address => address.region === region));
```
jtk54: Can you add an explicit null and a comment in the recursive `listAddresses` to make it a bit clearer? Something like `listAddresses(null /* region */)` to make this a little less subtle.

app/scripts/modules/google/address/address.reader.ts
```diff
@@ -0,0 +1,60 @@
...
+    } else {
+      return this.searchService
+        .search<IAddressSearchResults>({q: '', type: 'addresses'}, this.infrastructureCaches.get('addresses'))
+        .then((searchResults: ISearchResults<IAddressSearchResults>) => {
+          if (searchResults && searchResults.results) {
+            return searchResults.results.map(result => {
+              let address = JSON.parse(result.address) as IGceAddress;
+              address.account = result.account;
+              // An address's selfLink ends with its region.
```
jtk54: Is this the case for global addresses?

app/scripts/modules/google/address/address.reader.ts
danielpeach: Sounds good.

app/scripts/modules/google/address/address.reader.ts
danielpeach: The `selfLink` ends with `'global'` for global addresses.

PR #3483, opened 2017-04-05
fix(kubernetes): Delete Security Group Namespace

ethanfrogers:
Deleting Security Groups in a namespace other than default would error
out becuase it wasn't passing namespace to the securityGroupWriter.


icfantv:
@danielpeach can you PTAL at this and confirm that it's good?  Seems ok.

PR #3497, opened 2017-04-07
fix(core): fix pipeline stage navigation

danielpeach:
@anotherchrisberry or @jrsquared or @icfantv please review.


jrsquared:
LGTM.

jrsquared:
Good catch, thanks @danielpeach

PR #3500, opened 2017-04-07
fix(kubernetes): fixes load balancer annotations details view

danielpeach:
@lwander please review.


danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/1459

PR #3541, opened 2017-04-17
feat(core): show hint if wait for up instances task is not completing

danielpeach:

waitforupinstanceshint

tooltip


jrsquared:
Love this. LGTM.

This has me wondering if we should just be linking the serverGroup to the application object in the data model, then we wouldn't have to pass application around, but I think that's outside the scope of this PR anyway.

danielpeach:
Agreed - having a reference to the application object on a server group would make things a lot easier.

PR #3554, opened 2017-04-19
chore(core): Add JSX listing and cleanup based on rule changes

jrsquared:
The biggest change here is for one of the new rules, `jsx-no-lambda`, which does not allow lambda functions in the render function. This makes sense in that every time the render function gets called, a new function gets instantiated for every lambda function.


app/scripts/modules/core/delivery/executionBuild/ExecutionBuildNumber.tsx
```diff
@@ -15,26 +15,30 @@ export class ExecutionBuildNumber extends React.Component<IExecutionBuildNumberP
     super(props);
   }
 
+  private handleParentPipelineClick = () => {
+      ReactGA.event({category: 'Pipeline', action: 'Execution build number clicked - parent pipeline'});
```
danielpeach: Indentation is off in this block.

app/scripts/modules/core/delivery/executionGroup/execution/Execution.tsx
```diff
@@ -216,18 +216,47 @@ export class Execution extends React.Component<IExecutionProps, IExecutionState>
...
+  }
+
+  private handleDeleteClick = (event: React.MouseEvent<HTMLElement>) => {
+    ReactGA.event({category: 'Pipeline', action: 'Execution delete clicked'});
+    this.deleteExecution();
+    event.stopPropagation();
+  }
+
+  private handleCancelClick = (event: React.MouseEvent<HTMLElement>) => {
+      ReactGA.event({category: 'Pipeline', action: 'Execution cancel clicked'});
```
danielpeach: Indentation is off.

app/scripts/modules/core/delivery/executionBuild/ExecutionBuildNumber.tsx
jrsquared: How embarrassing. A lint PR with lint issues.

jrsquared:
Definitely happy to talk about the best pattern going forward. [Decorators](https://www.npmjs.com/package/core-decorators#autobind) are probably cleaner...I'm still not a huge fan, but that's probably because I'm old and crotchety. I'll keep this PR and we can switch to decorators if we decide it's best.

icfantv:
@jrsquared, @danielpeach FWIW, I like the idea of decorators as well as the entire concept of creating literal function handles on actual class objects feels wrong.

PR #3569, opened 2017-04-21
feat(dcd): pipeline template boilerplate

danielpeach:
@anotherchrisberry or @jrsquared please review - just some boilerplate for working on DCD UI

### DIFF
--- a/app/scripts/modules/core/pipeline/config/templates/ConfigurePipelineTemplateModal.tsx
+++ b/app/scripts/modules/core/pipeline/config/templates/ConfigurePipelineTemplateModal.tsx
@@ -1,0 +1,31 @@
+import * as React from 'react';
+import {Modal, Button} from 'react-bootstrap';
+import {IPipelineTemplate} from './pipelineTemplate.service';
+
+interface IState {}
+
+interface IProps {
+  template: IPipelineTemplate
+  show: boolean;
+  showCallback: (show: boolean) => null;
+}
+
+export class ConfigurePipelineTemplateModal extends React.Component<IProps, IState> {
+
+  private close = (): void => {
+    this.props.showCallback(false);
+  };
+
+  public render() {
+    return (
+      <Modal show={this.props.show} onHide={this.close}>
+        <Modal.Body>
+          <pre>{JSON.stringify(this.props.template || {}, null, 2)}</pre>
+        </Modal.Body>
+        <Modal.Footer>
+          <Button onClick={this.close}>Cancel</Button>
+        </Modal.Footer>
+      </Modal>
+    );
+  }
+}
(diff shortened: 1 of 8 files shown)


app/scripts/modules/core/pipeline/config/templates/ConfigurePipelineTemplateModal.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import {Modal, Button} from 'react-bootstrap';
+import {IPipelineTemplate} from './pipelineTemplate.service';
+const template = require('./template');
```
jrsquared: Is this supposed to be a default template if none is provided to the modal? If so, maybe rename to defaultTemplate.
Or if this is just leftover cruft from testing (I see the same template being passed in to the Modal), remove it.

app/scripts/modules/core/pipeline/config/templates/pipelineTemplate.service.ts
```diff
@@ -0,0 +1,51 @@
...
+  id: string;
+  name: string;
+  type: string;
+}
+
+export class PipelineTemplateService {
+  public getPipelineTemplateFromSourceUrl(/* source: string */): IPromise<IPipelineTemplate> {
+    // Uncomment when the Gate endpoint has been created:
+    // return api.one('pipelineTemplates').withParams({source}).get();
+    return $q.resolve(resolvedTemplate);
```
jrsquared: This one would get removed too.

app/scripts/modules/core/pipeline/config/templates/pipelineTemplate.service.ts
```diff
@@ -0,0 +1,51 @@
+import {module, IPromise} from 'angular';
+import {$q} from 'ngimport';
```
jrsquared: Since you're marking what should be added/uncommented when gate endpoint is created, probably worth marking what should be removed as well? Good for keeping track of cruft for later.

app/scripts/modules/core/pipeline/config/templates/pipelineTemplate.service.ts
```diff
@@ -0,0 +1,51 @@
...
+// import {api} from 'core/api/api.service';
+
+const resolvedTemplate = require('./template.json');
```
jrsquared: This one would get removed too.

app/scripts/modules/core/pipeline/config/templates/ConfigurePipelineTemplateModal.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  private close = (): void => {
+    this.props.showCallback(false);
+  };
+
+  public render() {
+    return (
+      <Modal show={this.props.show} onHide={this.close}>
+        <Modal.Body>
+          <pre>{JSON.stringify(this.props.template || template, null, 2)}</pre>
```
jrsquared: Same comment as above if `template` is leftover from testing vs default.

PR #3580, opened 2017-04-24
Turn on SSLProxyEngine in Deck's apache2 config

viglesiasce:
When Gate is using SSL you need to enable these flags for the '/gate' proxy to work properly.

Please advise on best way to make the verification settings configurable. They are currently disabling verification.


ttomsu:
You should not use the `/gate` proxy with containers. That was added specifically for the all-in-one starter images (and has been a big PITA for adding auth). A more robust installation (like one in a Kubernetes environment) should have Gate traffic going through it's own load balancer to Gate instances, which completely decouples it from the Deck/Apache instances.

viglesiasce:
In my setup I am using the Identity Aware Proxy for authn/authz. I am exposing gate using the HTTP(S) load balancer and do not want (or need to) have Gate exposed publicly. 

This is not a PR saying why the proxying should or should not work. It is enhancing the existing config to work with SSL.

Am I missing something from your comment? You are mentioning the all-in-one configs but this is specifically changing the Dockerfile.

ttomsu:
@viglesiasce and I chatted, and realized that he was trying to reuse the Apache instance as his instance's primary reverse proxy. 

We agreed that the best move is to leave Deck's Apache instance alone to serve it's primary function (serving Deck's static HTML/JS/CSS files) and instead, Vic will setup his own nginx reverse proxy for Deck, Gate, and in his Jenkins server.

PR #3585, opened 2017-04-24
chore(netflix/ci): add  scm reader test suite

icfantv:
* add test suite for scm reader service
* rename reader to be consistent with rest of deck.

### DIFF
--- a/app/scripts/modules/netflix/ci/services/ciBuild.read.service.spec.ts
+++ b/app/scripts/modules/netflix/ci/services/ciBuild.read.service.spec.ts
@@ -2,7 +2,7 @@ import * as moment from 'moment';
 import {IHttpBackendService, mock} from 'angular';
 
 import {SETTINGS} from 'core/config/settings';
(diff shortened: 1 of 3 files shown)


app/scripts/modules/netflix/ci/services/scm.read.service.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+      const type = 'scmType';
+      const limit = 500;
+      $http.expectGET(`${getReposUrl(type, org)}?limit=${limit}`).respond(200, []);
+
+      scmReader.getRepositories(org, type, limit);
+      $http.flush();
+    });
+  });
+
+  describe('commits retrieval', () => {
```
icfantv: These last three `describe` blocks are 90% identical but I couldn't create a workable solution involving iterating over a collection.

icfantv:
@sghill apologies, i forgot to tag you earlier.

PR #3588, opened 2017-04-26
feat(provider/kubernetes): Deploy baked docker images

brujoand:
The objective here is to be able to deploy baked docker containers to Kubernetes, so we don't have to rely on using the docker registry trigger, and we get to keep the traceability. 

I haven't synced with anyone on this approach, so I expect to have to alter worthing and/or some of the flow for this. So feel free to make suggestions for change. 

This PR depends on a change to orca: https://github.com/spinnaker/orca/pull/1299

### DIFF
--- a/app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
+++ b/app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
@@ -137,8 +137,10 @@ module.exports = angular.module('spinnaker.kubernetes.clusterCommandBuilder.serv
     }
 
     function buildImageId(image) {
-      if (image.fromContext) {
+      if (image.fromFindImage) {
         return `${image.cluster} ${image.pattern}`;
+      } else if (image.fromBake) {
+        return `${image.repository} (Baked during execution)`;
       } else if (image.fromTrigger && !image.tag) {
         return `${image.registry}/${image.repository} (Tag resolved at runtime)`;
       } else {
@@ -173,7 +175,7 @@ module.exports = angular.module('spinnaker.kubernetes.clusterCommandBuilder.serv
       return result;
     }
 
-    function findUpstreamImages(current, all, visited = {}) {
+    function findContextImages(current, all, visited = {}) {
       // This actually indicates a loop in the stage dependencies.
       if (visited[current.refId]) {
         return [];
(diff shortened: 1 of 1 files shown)


app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
```diff
@@ -137,8 +137,10 @@ module.exports = angular.module('spinnaker.kubernetes.clusterCommandBuilder.serv
     }
 
     function buildImageId(image) {
-      if (image.fromContext) {
+      if (image.fromContext && image.pattern && image.cluster) {
```
lwander: Does this mean it came from a findImage? I think we should be more explicit here

app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
```diff
@@ -189,11 +191,19 @@ module.exports = angular.module('spinnaker.kubernetes.clusterCommandBuilder.serv
           repository: current.name,
           stageId: current.refId
         });
+      } else if (current.type === 'bake') {
+        result.push({
+          fromContext: true,
+          repository: current.ami_name,
+          organization: current.organization,
+          stageId: current.refId,
+          registry: 'containers.schibsted.io'
```
lwander: Is this meant to be hardcoded here?

app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
lwander: e.g. 

```js
if (image.fromContext) {
  if (image.fromBake) {
  } else if (image.fromFindImage) {
  } else {
  // uh oh
  }
} else if (image.fromTrigger && !image.tag) {
}
```

lwander:
Also could you double check this for me @danielpeach?

app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
brujoand: Oh, no that was me testing. I'll get that out 👍

app/scripts/modules/kubernetes/cluster/configure/CommandBuilder.js
brujoand: Yeah it's meant to cover the findImage case. I agree, it's good to be clear here.

brujoand:
Thanks guys, I've addressed the comments and gotten back in sync with master.

PR #3602, opened 2017-05-01
fix(kubernetes): allow container to be copied if image does not have …

danielpeach:
…registry

Closes https://github.com/spinnaker/spinnaker/issues/1612.


ethanfrogers:
Thanks for this @danielpeach!

PR #3605, opened 2017-05-01
feat(aws): Allow for Deploy/Clone stages to use default block device mappings fo…

ajordens:
…r instance type

Exposes:
- Copy from current server group (default)
- Prefer AMI block device mappings
- Defaults for selected instance type

### DIFF
--- a/app/scripts/modules/amazon/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
+++ b/app/scripts/modules/amazon/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
@@ -55,8 +55,10 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.aws.cloneServerGr
       stage.credentials = $scope.application.defaultCredentials.aws;
     }
 
-    if (stage.isNew && _.get($scope, 'application.attributes.providerSettings.aws.useAmiBlockDeviceMappings')) {
-      stage.useAmiBlockDeviceMappings = $scope.application.attributes.providerSettings.aws.useAmiBlockDeviceMappings;
+    if (stage.isNew) {
+      let useAmiBlockDeviceMappings = _.get($scope, 'application.attributes.providerSettings.aws.useAmiBlockDeviceMappings', false);
+      stage.useAmiBlockDeviceMappings = useAmiBlockDeviceMappings;
+      stage.copySourceCustomBlockDeviceMappings = !useAmiBlockDeviceMappings;
     }
 
     this.targetClusterUpdated = () => {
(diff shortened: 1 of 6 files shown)


ajordens:
---

image

---

image

---

image

ajordens:
PTAL @anotherchrisberry 

Happy to hear suggestions on removing the duplication between the two stages (deploy/clone).

cfieber:
Should this stay on the advanced options area of the deploy dialog instead of jamming it right up above strategy selection (which is arguably significantly more important than this option)?

Also is the default going to be defaults for selected instance type, vs copy from current server group? That is a change in behaviour for anybody that currently has a custom block device mapping

Also would be worth checking on the pipeline stage config if there is a blockDevices block on the cluster and making the right selection in that case

ajordens:
@cfieber That screenshot was from the `Clone` pipeline stage ... for normal deploy operations it's still on the advanced options.

The default option is to `Copy from current server group` which should be unchanged from existing behavior.

image

app/scripts/modules/amazon/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
```diff
@@ -55,8 +55,10 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.aws.cloneServerGr
       stage.credentials = $scope.application.defaultCredentials.aws;
     }
 
-    if (stage.isNew && _.get($scope, 'application.attributes.providerSettings.aws.useAmiBlockDeviceMappings')) {
-      stage.useAmiBlockDeviceMappings = $scope.application.attributes.providerSettings.aws.useAmiBlockDeviceMappings;
+    if (stage.isNew) {
+      let useAmiBlockDeviceMappings = $scope.application.attributes.providerSettings.aws.useAmiBlockDeviceMappings || false;
```
anotherchrisberry: you want to wrap this is a `_.get` so it won't blow up if someone doesn't have that field set in their application, which is fairly likely, e.g.
```javascript
const useAmiBlockDeviceMappings = _.get($scope, 'application.attributes.providerSettings.aws.useAmiBlockDeviceMappings', false);
```
the third argument is the fallback in case the property doesn't exist

app/scripts/modules/amazon/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
```diff
@@ -93,5 +95,30 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.aws.cloneServerGr
     this.processIsSuspended = (process) => {
       return stage.suspendedProcesses && stage.suspendedProcesses.includes(process);
     };
+
+    this.blockDeviceMappingsSource = () => {
```
anotherchrisberry: Should prefix the name with a verb to indicate what it does, e.g. `getBlockDeviceMappingSource`

anotherchrisberry: Generally, in Angular, for something that's going to be called every time a user interacts with the page, you want to manage the state of the field elsewhere, like the `viewState` object.

That would add to the complexity here (you'd have to initialize it, then track changes), and this page (pipeline config) is relatively lightweight, so I'm good with the code as-is.

app/scripts/modules/amazon/serverGroup/configure/serverGroupConfiguration.service.js
```diff
@@ -73,6 +73,31 @@ module.exports = angular.module('spinnaker.aws.serverGroup.configure.service', [
         }
       };
 
+      command.blockDeviceMappingsSource = () => {
```
anotherchrisberry: `getBlockDeviceMappingsSource`

anotherchrisberry: Also want to verify this doesn't leak into the task being sent to Orca - probably isn't, but worth making sure.

ajordens: I didn't see anything leaking into the pipeline json or over the wire ... we've got other functions being added to `command` so I don't anticipate an issue.

PR #3608, opened 2017-05-02
fix(*): Accounts showing up as empty on the Edit Application modal

ajordens:


ajordens:
PTAL @anotherchrisberry / @EricJizbaMSFT

app/scripts/modules/core/account/accountSelectField.directive.js
```diff
@@ -55,8 +55,13 @@ module.exports = angular
           this.mergedAccounts = _.flatten([this.primaryAccounts, this.secondaryAccounts]);
         }
 
-        if (this.component && (!this.mergedAccounts || !this.mergedAccounts.includes(this.component[this.field]))) {
```
ajordens: I believe this was failing because `this.component[this.field]` is an array and `.includes(array)` was always false.

I tried to make this an array-safe equivalent to what was there before but my JS is weak. Also this has not been extensively tested outside of the Edit Application modal ... I'm not sure what prompted the original change @EricJizbaMSFT but please point out anywhere else that should be tested.

ajordens:
Talked with @anotherchrisberry offline and merging this down.

PR #3610, opened 2017-05-02
fix(kubernetes): fixes container copying in server group dialogue

danielpeach:

Yesterday's fix (https://github.com/spinnaker/deck/pull/3602) wasn't quite right.


ethanfrogers:
LGTM 👍

PR #3639, opened 2017-05-05
fix(core/pageTitle): Use promise based API for showing/hiding spinner

christopherthielen:
UI-Router 1.0  `$stateChange*` events polyfill isn't 100% backwards compatible (apparently).  Switch to `TransitionService` promise based API and maintain a reference count of pending transitions.


jrsquared:
Just return types on functions, but otherwise LGTM.

christopherthielen:
Will fix unit tests

PR #3644, opened 2017-05-05
feat(core/visualizer): Allow visualizer toggle from URL query param `?vis=true`

christopherthielen:
The visualizer can be useful to diagnose problems.  By adding a url trigger, we can share links with the visualizer enabled.

### DIFF
--- a/app/scripts/modules/core/core.module.js
+++ b/app/scripts/modules/core/core.module.js
@@ -246,9 +246,7 @@ module.exports = angular
     uiSelectConfig.appendToBody = true;
   })
   .run(function($stateRegistry, $uiRouter) {
-
-    // Type javascript:vis() in the browser url
-    window.vis = function() {
+    function launchVisualizer() {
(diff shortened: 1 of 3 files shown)


jrsquared:
LGTM

anotherchrisberry:
LGTM. While you're in there, can you add similar functionality to enable trace logging on transitions? A surprising number of folks open the console when they're having trouble and get confused by all the messages.

christopherthielen:
I haven't merged this yet because the `?vis=true` is being added to the "Recent Application"

screen shot 2017-05-10 at 11 06 19 am

christopherthielen:
OK, I've updated this, adding support for:
`?trace=true` or `trace=all` or `trace=TRANSITION,UIVIEW,VIEWCONFIG`, etc

I've also adjusted the "save recent places" code to ignore `trace`, `debug`, and `vis` parameter values.

PR #3650, opened 2017-05-08
fix(core): prevent overflow in check preconditions stage config

danielpeach:

longexpression_fix


jrsquared:
LGTM

PR #3714, opened 2017-05-19
fix(kubernetes/runJob): Remove check for enabled on Run Job trigger option

ethanfrogers:
When I put this in at first, I didn't account for all scenarios. It
works for triggers that are always enabled. To support "one click
deployment" scenarios where you trigger a manual execution, then select
an image you wouldn't see the trigger show up in the list of available
triggers. By removing the check, you can select any trigger you define
and use that. If that trigger is used automatically or manually is
outside the scope of this feature.

ping @danielpeach


ethanfrogers:
Thanks @danielpeach. I've rebased on master.

PR #3749, opened 2017-05-25
fix(provider/google): Avoid console errors and mangled instance detai…

duftler:
…ls for standalone instances.


duftler:
@danielpeach please review.

PR #3760, opened 2017-05-26
fix(provider/google): Return proper structure when just the command's…

duftler:
… image is returned as search results.


duftler:
@danielpeach please review.

PR #3787, opened 2017-06-02
feat(pipeline-templates): config view

danielpeach:
@jrsquared or @anotherchrisberry please review.

@robzienert fyi

### DIFF
--- a/app/scripts/modules/core/src/delivery/create/CreatePipelineModal.tsx
+++ b/app/scripts/modules/core/src/delivery/create/CreatePipelineModal.tsx
@@ -1,18 +1,21 @@
 import * as React from 'react';
-import {Button, Modal} from 'react-bootstrap';
+import { Button, Modal } from 'react-bootstrap';
 import * as Select from 'react-select';
 import autoBindMethods from 'class-autobind-decorator';
-import {$log} from 'ngimport';
-import {IHttpPromiseCallbackArg} from 'angular';
-import {cloneDeep, uniqBy} from 'lodash';
-import {Application} from 'core/application/application.model';
-import {IPipeline} from 'core/domain/IPipeline';
-import {SubmitButton} from 'core/modal/buttons/SubmitButton';
-import {ReactInjector, NgReact } from 'core/reactShims';
-import {SETTINGS} from 'core/config/settings';
-import {IPipelineTemplate} from 'core/pipeline/config/templates/pipelineTemplate.service';
-import {TemplateDescription} from './TemplateDescription';
-import {TemplateSelector} from './TemplateSelector';
+import { $log } from 'ngimport';
+import { IHttpPromiseCallbackArg } from 'angular';
+import { cloneDeep, uniqBy } from 'lodash';
+import { Debounce } from 'lodash-decorators';
+
+import { Application } from 'core/application/application.model';
+import { IPipeline } from 'core/domain/IPipeline';
+import { SubmitButton } from 'core/modal/buttons/SubmitButton';
+import { ReactInjector, NgReact } from 'core/reactShims';
+import { SETTINGS } from 'core/config/settings';
+import { IPipelineTemplateConfig, IPipelineTemplate } from 'core/pipeline/config/templates/pipelineTemplate.service';
+
+import { TemplateDescription } from './TemplateDescription';
+import { ManagedTemplateSelector } from './ManagedTemplateSelector';
 
 import './createPipelineModal.less';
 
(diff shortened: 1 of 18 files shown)


app/scripts/modules/core/src/delivery/create/CreatePipelineModal.tsx
```diff
@@ -1,18 +1,19 @@
...
-import {SubmitButton} from 'core/modal/buttons/SubmitButton';
-import {ReactInjector, NgReact } from 'core/reactShims';
-import {SETTINGS} from 'core/config/settings';
-import {IPipelineTemplate} from 'core/pipeline/config/templates/pipelineTemplate.service';
-import {TemplateDescription} from './TemplateDescription';
-import {TemplateSelector} from './TemplateSelector';
+import { $log } from 'ngimport';
+import { IHttpPromiseCallbackArg } from 'angular';
+import { cloneDeep, uniqBy } from 'lodash';
+import { Debounce } from 'lodash-decorators';
```
jrsquared: space between external and internal imports

app/scripts/modules/core/src/delivery/create/CreatePipelineModal.tsx
```diff
@@ -1,18 +1,19 @@
...
+import { Application } from 'core/application/application.model';
+import { IPipeline } from 'core/domain/IPipeline';
+import { SubmitButton } from 'core/modal/buttons/SubmitButton';
+import { ReactInjector, NgReact } from 'core/reactShims';
+import { SETTINGS } from 'core/config/settings';
+import { IPipelineTemplateConfig, IPipelineTemplate } from 'core/pipeline/config/templates/pipelineTemplate.service';
+import { TemplateDescription } from './TemplateDescription';
```
jrsquared: space between local and core imports

app/scripts/modules/core/src/delivery/create/CreatePipelineModal.tsx
```diff
@@ -94,31 +99,62 @@ export class CreatePipelineModal extends React.Component<ICreatePipelineModalPro
...
 
+  private submitPipelineTemplateConfig(): void {
+    const config: Partial<IPipelineTemplateConfig> = {
+      name: this.state.command.name,
+      application: this.props.application.name,
+      type: 'templatedPipeline',
+      limitConcurrent: true,
+      keepWaitingPipelines: false,
+      parallel: true,
+      executionEngine: 'v2',
```
jrsquared: v3 is the new hotness in orca, should we change this to v3?

app/scripts/modules/core/src/delivery/create/CreatePipelineModal.tsx
```diff
@@ -180,6 +217,23 @@ export class CreatePipelineModal extends React.Component<ICreatePipelineModalPro
...
+        loadingTemplateFromSourceError: false,
+        command: Object.assign({}, this.state.command, {template: null}),
+      });
+    };
+  }
+
+  public handleSourceUrlChange(e: React.ChangeEvent<HTMLInputElement>): void {
+    const templateSourceUrl = e.target.value;
+    this.setState({templateSourceUrl});
+    this.loadPipelineTemplateFromSource(templateSourceUrl);
```
jrsquared: Not sure if this _will_ cause any issues, but if the `loadPipelineTemplateFromSource` isn't debounced, you'll be calling setState back to back, which can sometimes have some odd rendering effects...

app/scripts/modules/core/src/delivery/create/CreatePipelineModal.tsx
robzienert: I'd just remove this field entirely. Orca should be the one to decide the default execution engine (which it does).

jrsquared: even better

app/scripts/modules/core/src/delivery/create/CreatePipelineModal.tsx
danielpeach: I'm confused - `loadPipelineTemplateFromSource` is debounced.

jrsquared: Sorry, poor clarity (and also and a poor understanding of the default options debounce) on my part, I was using debounced as a verb at runtime. I meant that if `wait` had expired for the debounced function.  I thought `leading` was the default behavior, not `trailing`. I have much less concern for back-to-back `setState` calls with a `trailing` debounce. So nothing to see here.

danielpeach: Sounds good - thanks.

PR #3788, opened 2017-06-02
chore(core): Update core/tsconfig.json to include all typescript files

christopherthielen:
Make things compile -- export interfaces as necessary

### DIFF
--- a/app/scripts/modules/amazon/tsconfig.json
+++ b/app/scripts/modules/amazon/tsconfig.json
@@ -42,10 +42,12 @@
     "useBabel": true,
     "useCache": true
   },
-  "files": [
-    "src/index.ts"
+  "include": [
+    "src/**/*.ts",
+    "src/**/*.tsx"
   ],
   "exclude": [
-    "./lib"
+    "./lib",
+    "**/*.spec.ts"
   ]
 }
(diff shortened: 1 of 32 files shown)


christopherthielen:
Some Typescript files were reachable only through a `.js` file.  
I switched to a glob `includes`, then exported and renamed all interfaces as necessary.

app/scripts/modules/core/src/delivery/create/TemplateSelector.tsx
```diff
@@ -25,7 +23,7 @@ export class TemplateSelector extends React.Component<IProps, IState> {
           <Select
             options={this.createOptionsFromTemplates()}
             clearable={true}
-            value={selected ? {label: selected.metadata.name, id: selected.id} : null}
+            value={selected ? {label: selected.metadata.name, value: selected.id} : null}
```
christopherthielen: @danielpeach PTAL
I think this was supposed to be ` {label: selected.metadata.name, value: selected.id}`.  My IDE was not complaining, but `tsc` was.

jrsquared:
😍

christopherthielen:
@jrsquared I switched `amazon/tsconfig.json` as well

app/scripts/modules/core/src/delivery/create/TemplateSelector.tsx
danielpeach: You're right, thanks for finding.

PR #3806, opened 2017-06-09
refactor(provider/amazon): Convert load balancer transformer to TS

jrsquared:
Convert loadBalancer.transformer to TS.  This is a little bit of a mess and it was because of this conversion that I want to create three different types for load balancer shapes which will come in the next PR. Since this is 'working' as-is, want to get this in, then clean up the shapes.

For context, we should really have three shapes for load balancers (and for server groups and any other shape coming back from the API):
* **IUpsertLoadBalancerCommand**: The shape that we use to upsert a load balancer
* **ILoadBalancerSourceData**: The raw shape that comes back from the API
* **ILoadBalancer**:  the shape that we transform source data into to use in the UI

Because of this lack of distinction, this PR has some pretty hacky type conversions that I didn't understand what was happening until I got a fully functional conversion. Will fix in next PR.


app/scripts/modules/amazon/src/domain/IAmazonLoadBalancer.ts
```diff
@@ -1,45 +1,89 @@
...
+  detachedInstances?: IInstance[];
+  elb?: any;
+  isInternal?: boolean;
+  regionZones: string[];
+  subnets?: string[];
+  subnetDetails?: ISubnet[];
+  subnetType?: string;
+}
+
+export interface IClassicListener {
```
christopherthielen: haha, that is so like Listener

PR #3809, opened 2017-06-10
fix(core/executions): Show durations in execution details/permalink

christopherthielen:
- Added an observable for execution filter model's viewState


christopherthielen:
Since this viewState data comes from a global service, I added an observable to the service.  Come at me

jrsquared:
LGTM!

anotherchrisberry:
This looks good as a fix for the current functionality. However, the current functionality is not a good pattern: people have to know to go to the executions view to toggle stage durations, and that the field on the executions view affects the rendering of the single execution (i.e. permalink) view.

It would be a lot more intuitive to include the same checkbox in both views.

christopherthielen:
@icfantv @anotherchrisberry @jrsquared ptal

app/scripts/modules/core/src/delivery/details/singleExecutionDetails.component.ts
```diff
@@ -0,0 +1,142 @@
+import { IScope, module } from 'angular';
+import { StateService } from '@uirouter/core';
+import { Subscription } from 'rxjs/Subscription';
+
+import { SCHEDULER_FACTORY } from 'core/scheduler';
+import { Application } from 'core/application';
+import { EXECUTION_SERVICE, ExecutionService } from 'core/delivery';
+import { SchedulerFactory, IScheduler } from 'core/scheduler';
+import { IExecution } from 'core/domain';
+import { ExecutionFilterModel } from 'core/delivery';
```
anotherchrisberry: move to #L7 (already a `core/delivery` import up there)

app/scripts/modules/core/src/delivery/details/singleExecutionDetails.component.ts
```diff
@@ -0,0 +1,142 @@
...
+      return;
+    }
+
+    executionService.getExecution($state.params.executionId).then((execution) => {
+      this.execution = this.execution || execution;
+      executionService.transformExecution(this.app, execution);
+      executionService.synchronizeExecution(this.execution, execution);
+      if (!execution.isActive) {
+        executionScheduler.unsubscribe();
+        executionLoader.unsubscribe();
```
anotherchrisberry: seems like you'd only need one of these, probably `executionLoader.unsubscribe()` - is there a reason to unsubscribe the scheduler, too?

(not a big deal, just trying to remember/understand why this was like this in the first place)

app/scripts/modules/core/src/delivery/details/singleExecutionDetails.component.ts
```diff
@@ -0,0 +1,142 @@
...
+class SingleExecutionDetailsController {
+  private executionScheduler: IScheduler;
+  private executionLoader: Subscription;
+
+  private execution: IExecution;
+  private stateNotFound: boolean;
+  private app: Application;
+
+  // Not sure if/where this is used.
+  public showDetails = () => true;
```
anotherchrisberry: looks like this is the only reference to `showDetails` in the project - let's remove it

anotherchrisberry:
a few minor nits, LGTM otherwise

icfantv:
@christopherthielen if you could also add a screenshot for this, please, that would be helpful.  we _try_ to remember to do this for UI functionality so peeps don't have to pull the branch and run it to see.  thanks.

christopherthielen:
screen shot 2017-06-14 at 11 00 18 am

app/scripts/modules/core/src/delivery/details/singleExecutionDetails.component.ts
christopherthielen: I just looked at the implementation and I have a suspicion that both calls are necessary to avoid leak.  @jrsquared ?

anotherchrisberry:
don't forget to squash your commits before merging

christopherthielen:
I squashed using github UI

https://github.com/spinnaker/deck/commits/master

anotherchrisberry:
fancy

PR #3814, opened 2017-06-11
feat(provider/kubernetes): Add Run Job Pod Labels

ethanfrogers:
Adds the ability to add labels to the Run Job pod definition. This is
useful for pin pointing pods for loggin purposes. I took the route of
createing a Label Configurer that extends the Annotation Configurer.
This is primarily because it performs the same function (creting a map
of labels/annotations) so it does the same work. I can refactor if
necessary.

Part of [#1683](https://github.com/spinnaker/spinnaker/issues/1683)

Related Clouddriver PR: https://github.com/spinnaker/clouddriver/pull/1683

@danielpeach @lwander PTAL

### DIFF
--- a/app/scripts/modules/kubernetes/label/labelConfigurer.component.html
+++ b/app/scripts/modules/kubernetes/label/labelConfigurer.component.html
@@ -1,0 +1,25 @@
+<table class="table table-condensed packed tags" style="border-top: 2px solid white;">
+  <tr ng-repeat="view in $ctrl.annotationViews track by $index">
+    <td class="table-label"><b>Key</b></td>
+    <td>
+      <input class="form-control input-sm" type="text" ng-model="view.key" required ng-change="$ctrl.onKeyOrValueChange($event)">
+    </td>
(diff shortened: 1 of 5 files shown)


app/scripts/modules/kubernetes/label/labelConfigurer.component.html
```diff
@@ -0,0 +1,25 @@
+<table class="table table-condensed packed tags" style="border-top: 2px solid white;">
+  <tr ng-repeat="view in $ctrl.annotationViews track by $index">
```
danielpeach: I would rather see the annotation configurer become a more generic component (rather than have the label configurer extend it) - especially since `$ctrl.annotationViews` ends up here.

ethanfrogers:
@danielpeach I initially went that way but figured I'd give this a try first. I'll refactor. It'll probably be fewer changes anyway! Thanks for looking at this.

ethanfrogers:
Closing this and will resubmit once I refactor. I'm also planning on adding Pod Annotations to the Run Job stage configuration.

PR #3823, opened 2017-06-13
feat(provider/kubernetes): Run Job extended config

ethanfrogers:
Adds the ability to add labels and annotations to the Run Job pod
definition. This is useful for pin pointing pods for logging
purposes. As per request, I've made the annotation configurer more generic
and updated all references to it. I didn't use the `mapEditor` from
`core` because it requires you to define an `onChange` function which
would mean we needed to add the same code handled by the `mapConfigurer`
to other controllers where it's used.

@danielpeach PTAL. Totally cool with renaming it.


app/scripts/modules/kubernetes/mapConfigurer/mapConfigurer.component.ts
```diff
@@ -0,0 +1,63 @@
+import {module} from 'angular';
```
danielpeach: Please change the formatting to `import { module }` here and below. (That's the new style, though it's not well reflected in the codebase right now).

danielpeach: Also, could you please move this component into the `common` directory?

ethanfrogers:
@danielpeach you're totally right. `mapEditor` does the same thing. It's styled a bit differently so it might make sense to move all of them over to it (server group, load balancer, run job). 

_ More generic Label Configurer on top. Map editor on bottom_

app/scripts/modules/core/src/forms/mapEditor/mapEditor.component.js
```diff
@@ -14,6 +16,8 @@ module.exports = angular
       addButtonLabel: '@',
       allowEmpty: '=?',
       onChange: '&',
+      labelsLeft: '=?',
```
danielpeach: This could be a one-way binding - i.e., switch this to `labelsLeft: '<?'`.

ethanfrogers:
#### Map Editor with K/V Labels on the Side

image

#### Map Editor with Label on top and K/V Labels on side

image


#### Map Editor in default state

image

app/scripts/modules/kubernetes/serverGroup/configure/wizard/advancedSettings.html
```diff
@@ -1,13 +1,26 @@
 <ng-form name="advancedSettings">
   <div class="form-group form-horizontal">
-    <kubernetes-annotation-configurer component="command" field="replicaSetAnnotations" label="Replica Set Annotations">
-    </kubernetes-annotation-configurer>
+    <div class="col-md-12">
+      <map-editor model="command.replicaSetAnnotations"
+          label="Replica Set Annotations"
```
danielpeach: Please align these attributes under `model` (same for the `map-editor`s below).

app/scripts/modules/kubernetes/serverGroup/configure/wizard/advancedSettings.html
```diff
@@ -1,13 +1,26 @@
...
+          add-button-label="Add Annotations"
+          labels-left="true"></map-editor>
+    </div>
 
-    <kubernetes-annotation-configurer component="command" field="podAnnotations" label="Pod Annotations">
-    </kubernetes-annotation-configurer>
+    <div class="col-md-12">
+      <map-editor model="command.podAnnotations"
+          label="Pod Annotations"
+          add-label="Add Annotations"
```
danielpeach: Should be `add-button-label`.

app/scripts/modules/core/src/forms/mapEditor/mapEditor.component.js
ethanfrogers: That's my "I don't know Angular showing" 😆

ethanfrogers:
@danielpeach I noticed the styling differences (font size is a touch smaller). The borders also look lighter in the screen shot. I'll see what I can do about those differences.

ethanfrogers:
Closes [#1683](https://github.com/spinnaker/spinnaker/issues/1683)

PR #3835, opened 2017-06-15
fix(permissions): Prevent user from only specifying READ permission, …

ttomsu:
…and thus locking everyone out of that application.
Fixes https://github.com/spinnaker/fiat/issues/132

### DIFF
--- a/app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
+++ b/app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
@@ -133,6 +133,12 @@ export class PermissionsConfigurer extends React.Component<IPermissionsConfigure
     }
   }
 
+  private willApplicationLockoutAllUsers(): boolean {
+    return !!(this.props.permissions) &&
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
```diff
@@ -133,6 +133,12 @@ export class PermissionsConfigurer extends React.Component<IPermissionsConfigure
     }
   }
 
+  private willApplicationLockoutAllUsers(): boolean {
+    return !!(this.props.permissions) &&
+        !!(compact(this.props.permissions.READ).length) &&
```
danielpeach: Would be more consistent with the logic in the modal controllers (and you wouldn't have to coerce 0 into `false`) if you used `compact(this.props.permissions.READ).length > 0`. Not a big deal.

ttomsu: Yep, good call. Done.

PR #3858, opened 2017-06-20
feat(pipeline/RunJob): Allow container name conf

ethanfrogers:
Allow configuring container name in the RunJob stage. The default name
was the current date which makes it difficult to pin point containers for
logging purposes. Also, changes the default container name to `job`
since Clouddriver does that for containers without a name anyway.

Addresses [#1770](https://github.com/spinnaker/spinnaker/issues/1770)

#### Screenshot
image

@danielpeach PTAL!

### DIFF
--- a/app/scripts/modules/kubernetes/pipeline/stages/runJob/runJobStage.html
+++ b/app/scripts/modules/kubernetes/pipeline/stages/runJob/runJobStage.html
@@ -18,6 +18,12 @@
     </namespace-select-field>
   </stage-config-field>
 
+  <stage-config-field label="Container Name">
+    <div>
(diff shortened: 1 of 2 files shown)


app/scripts/modules/kubernetes/pipeline/stages/runJob/runJobStage.html
```diff
@@ -18,6 +18,12 @@
     </namespace-select-field>
   </stage-config-field>
 
+  <stage-config-field label="Container Name">
+    <div>
+        <input type="text" class="form-control input-sm" ng-model="ctrl.stage.container.name">
```
danielpeach: Indentation is a little off.

ethanfrogers:
@danielpeach indent reduced! thanks!

PR #3859, opened 2017-06-20
feat(provider/google): Add ILB listeners.

jtk54:
This should break less things this time around.


danielpeach:
I don't see anything next to "DNS Name" in the load balancer details view:
screen shot 2017-06-20 at 6 42 32 pm

danielpeach:
A few other things:

- I see a console error when I try to edit an ILB: `Cannot read property 'name' of undefined`.
- When I add a listener, I'm seeing some loud browser default validation ("Please fill out this field"). You should be able to turn this off with something like a `novalidate` attribute on the enclosing `<form>` tag for the whole wizard.
- When creating a new ILB, I think the form should pick the first subnet from the set of available options (rather than leaving the field blank).
- Are there any issues editing an ILB that was created by Spinnaker before this change?

app/scripts/modules/google/src/loadBalancer/configure/internal/gceCreateInternalLoadBalancer.controller.ts
```diff
@@ -167,22 +174,25 @@ class InternalLoadBalancerCtrl extends CommonGceLoadBalancerCtrl implements ng.I
         const healthCheckNamesToOmit = this.isNew ? [] : [this.loadBalancer.backendService.healthCheck.name];
         this.existingHealthCheckNamesByAccount = this.gceCommonLoadBalancerCommandBuilder
           .groupHealthCheckNamesByAccount(data.healthChecks as IGceHealthCheck[], healthCheckNamesToOmit);
-      });
+      })
```
danielpeach: Dropped semicolon

app/scripts/modules/google/src/loadBalancer/configure/internal/gceCreateInternalLoadBalancer.controller.ts
```diff
@@ -218,29 +228,41 @@ class InternalLoadBalancerCtrl extends CommonGceLoadBalancerCtrl implements ng.I
...
+    this.taskMonitor.submit(() => this.gceMultipleListenerLoadBalancerWriter.upsertLoadBalancer(serializedLoadBalancers, this.application, descriptor));
   }
 
   private initializeEditMode (): void {
+    this.originalListeners = _.cloneDeep(this.loadBalancer.listeners);
     this.loadBalancer.ports = this.loadBalancer.ports.join(', ');
     this.loadBalancer.subnet = this.gceXpnNamingService.decorateXpnResourceIfNecessary(this.loadBalancer.project, this.loadBalancer.subnet);
     this.loadBalancer.network = this.gceXpnNamingService.decorateXpnResourceIfNecessary(this.loadBalancer.project, this.loadBalancer.network);
+
+    this.loadBalancer.listeners.forEach((listener: any) =>  {
```
danielpeach: s/any/IGceListener

app/scripts/modules/google/src/loadBalancer/configure/internal/transformer.service.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+const angular = require('angular');
+import * as _ from 'lodash';
+
+module.exports = angular.module('spinnaker.gce.deck.internalLoadBalancer.transformer', [])
+  .factory('internalLoadBalancerTransformer', function () {
+    // SERIALIZE
```
danielpeach: Drop comment

app/scripts/modules/google/src/loadBalancer/configure/internal/listeners/listener.component.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+import * as _ from 'lodash';
+import { GCE_ADDRESS_SELECTOR } from 'google/loadBalancer/configure/common/addressSelector.component';
```
danielpeach: Do you need this? I don't see where it's used.

app/scripts/modules/google/src/loadBalancer/configure/internal/listeners/listener.component.js
```diff
@@ -0,0 +1,55 @@
...
+
+module.exports = angular.module('spinnaker.deck.gce.internalLoadBalancer.listener.component', [GCE_ADDRESS_SELECTOR])
+  .component('gceInternalLoadBalancerListener', {
+    bindings: {
+      loadBalancer: '=',
+      subnetOptions: '=',
+      listener: '=',
+      deleteListener: '&',
+      index: '=',
+      application: '=',
```
danielpeach: It's better for these to be `<` rather than `=` if you're not going to be mutating the value passed in (I think that's true of all of the bindings except for `listener`.)

app/scripts/modules/google/src/loadBalancer/configure/internal/gceCreateInternalLoadBalancer.controller.ts
```diff
@@ -218,29 +228,41 @@ class InternalLoadBalancerCtrl extends CommonGceLoadBalancerCtrl implements ng.I
...
+    this.updateSubnetOptions();
+    this.loadBalancer.listeners = this.loadBalancer.listeners || [];
+    this.loadBalancer.listeners.push({} as IGceListener);
+  }
+
+  public removeListener (index: number): void {
+    this.loadBalancer.listeners.splice(index, 1);
+  }
+
+ public setSessionAffinity (viewState: ViewState): void {
```
danielpeach: Looks like this got moved to the left a little bit.

app/scripts/modules/google/src/loadBalancer/configure/internal/listeners/listener.component.html
```diff
@@ -0,0 +1,97 @@
...
+              <th>Protocol</th>
+              <th>Subnet</th>
+              <th>Ports<help-field key="gce.internalLoadBalancer.ports"></help-field></th>
+            </tr>
+          </thead>
+          <tbody>
+          <tr>
+            <td>
+              <select ng-disabled="$ctrl.listener.created" class="form-control input-sm"
+                      ng-change="$ctrl.protocolUpdated()"
```
danielpeach: This is not being called - needs to be passed to the listener component from `createGceInternalLoadBalancerCtrl`.

app/scripts/modules/google/src/loadBalancer/configure/internal/listeners/listener.component.html
```diff
@@ -0,0 +1,97 @@
...
+                      ng-model="$ctrl.loadBalancer.ipProtocol"
+                      ng-options="protocol for protocol in ['TCP', 'UDP']"></select>
```
danielpeach: Is this a per-listener setting? Or is it set for the load balancer? (It's being set on the load balancer itself right now).

app/scripts/modules/google/src/loadBalancer/configure/internal/listeners/listener.component.html
jtk54: Injected the function into the component from the controller, should be wired up now.

app/scripts/modules/google/src/loadBalancer/configure/internal/listeners/listener.component.html
jtk54: Good point, this should be consistent over the whole ILB.

jtk54:
Slowly addressing the rest of the comments, will ping when this is ready for another pass.

jtk54:
ping @danielpeach this is ready

app/scripts/modules/google/src/loadBalancer/configure/internal/listeners/listener.component.html
danielpeach: This should be moved from here, since it's not possible to set it on each listener. Probably should go under 'Basic Settings'.

app/scripts/modules/google/src/loadBalancer/configure/internal/gceCreateInternalLoadBalancer.controller.ts
```diff
@@ -218,29 +228,39 @@ class InternalLoadBalancerCtrl extends CommonGceLoadBalancerCtrl implements ng.I
     this.loadBalancer.loadBalancerName = this.getName(this.loadBalancer, this.application);
   }
 
+  public addListener (): void {
+    this.updateSubnetOptions();
+    this.loadBalancer.listeners = this.loadBalancer.listeners || [];
+    this.loadBalancer.listeners.push({subnet: this.subnetOptions[0], ports: 'port1,port2,etc'} as IGceListener);
```
danielpeach: Not sure if the port hint is necessary (since we have a help field), but if you do want to include that, it should go as a `placeholder` on the `input` field.

jtk54:
Closing since there's an easier way than this PR to have happy customers 👍

PR #3864, opened 2017-06-23
feat(kubernetes): Configure Service Account Name

ethanfrogers:
Allows configurable Service Account Name for Server Groups and Run Job
configuration to enable applications access to the Kubernetes API if
need be. Default behavior is to assign the `default` Service Account if
left blank. I also added it to the Server Group details. Note, if the
service account is `default`, it will not show up as the API doesn't
return it with the ReplicaSet/Deployment.

#### Server Group Config
image

#### Server Group Details
image

#### Run Job Stage
image

@danielpeach PTAL
@lwander FYI

### DIFF
--- a/app/scripts/modules/kubernetes/serverGroup/configure/wizard/advancedSettings.html
+++ b/app/scripts/modules/kubernetes/serverGroup/configure/wizard/advancedSettings.html
@@ -1,5 +1,31 @@
 <ng-form name="advancedSettings">
   <div class="form-group form-horizontal">
+
+    <div class="form-group">
+      <div class="col-md-4 sm-label-right">
+        Service Account Name
+      </div>
+      <div class="col-md-7">
+        <input type="text" class="form-control input-sm" ng-model="command.serviceAccountName">
+      </div>
+    </div>
+
+    <div class="form-group">
+      <div class="col-md-4 sm-label-right">
+        Termination Grace Period
+        <help-field key="kubernetes.serverGroup.terminationGracePeriodSeconds"></help-field>
+      </div>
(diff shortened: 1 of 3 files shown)


app/scripts/modules/kubernetes/serverGroup/configure/wizard/advancedSettings.html
```diff
@@ -1,5 +1,31 @@
...
+
+    <div class="form-group">
+      <div class="col-md-4 sm-label-right">
+        Termination Grace Period
+        <help-field key="kubernetes.serverGroup.terminationGracePeriodSeconds"></help-field>
+      </div>
+      <div class="col-md-3">
+        <div class="input-group">
+          <input type="number" min="0"
+             class="form-control input-sm"
```
danielpeach: Could you please align these attributes with `type`? (thanks, I know it was just copied from below)

app/scripts/modules/kubernetes/serverGroup/configure/wizard/advancedSettings.html
```diff
@@ -1,5 +1,31 @@
 <ng-form name="advancedSettings">
   <div class="form-group form-horizontal">
+
+    <div class="form-group">
```
danielpeach: Super minor, but it feels as if the ordering here should match the ordering in the details panel.

ethanfrogers:
@danielpeach comments addressed. I moved Service Account Name above Termination Grace Period Seconds in the details panel. I tried switching them in Advanced Settings but it didn't look quite right to me.

PR #3869, opened 2017-06-26
feat(provider/amazon): Enforce target group naming convention

jrsquared:
Force target group names to start with load balancer name because reasons.

screen shot 2017-06-26 at 10 42 42 am
screen shot 2017-06-26 at 10 42 50 am

@anotherchrisberry or @christopherthielen PTAL


christopherthielen:
cool

PR #3877, opened 2017-06-27
fix(provider/amazon): Only enforce TG naming to be based on appname

jrsquared:
Also do not allow the user to enter a target group name that will be longer than 32 characters (AWS limit)

@christopherthielen PTAL
@cfieber FYI


christopherthielen:
Yep

PR #3879, opened 2017-06-28
feat(gce): allow ILB port update

danielpeach:
@jtk54 please review. I've manually tested this, and there don't seem to be any problems.


danielpeach:
Yes.

PR #3881, opened 2017-06-29
Load balancer pod -- shouldComponentUpdate

christopherthielen:
apiproxy load balancers with everything turned on

before
screen shot 2017-06-28 at 5 13 53 pm

after
screen shot 2017-06-28 at 5 15 37 pm


jrsquared:
Slick. I was wondering when this would need to be done; was hoping we'd have a redux story before needing to do this...

Looks great!

PR #3884, opened 2017-07-03
feat(core/amazon): add build info to changes

icfantv:
* add build info, if available, to changes dialog.

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/details/serverGroupDetails.aws.controller.js
+++ b/app/scripts/modules/amazon/src/serverGroup/details/serverGroupDetails.aws.controller.js
@@ -154,6 +154,11 @@ module.exports = angular.module('spinnaker.amazon.serverGroup.details.controller
                 this.changeConfig = {
                   metadata: get(this.serverGroup.entityTags, 'creationMetadata')
                 };
+                if (this.serverGroup.buildInfo.jenkins) {
+                  this.changeConfig.buildInfo = {
+                    jenkins: this.serverGroup.buildInfo.jenkins
+                  };
+                }
               } else {
                 autoClose();
               }
--- a/app/scripts/modules/core/src/diffs/changes.html
+++ b/app/scripts/modules/core/src/diffs/changes.html
@@ -3,6 +3,24 @@
   <div class="modal-header has-sticky-headers">
     <h3>Changes to {{$ctrl.nameItem.name}}</h3>
   </div>
+  <div class="modal-body" ng-if="(!$ctrl.previousBuildLink || !$ctrl.currentBuildLink) && $ctrl.buildInfo.ancestor && $ctrl.buildInfo.target">
+    <div class="col-md-6">
+      <strong>Previous: </strong>Build: #{{$ctrl.buildInfo.ancestor}}
+    </div>
+    <div class="col-md-6">
+      <strong>Current: </strong>Build: #{{$ctrl.buildInfo.target}}
+    </div>
+  </div>
+  <div class="modal-body" ng-if="$ctrl.previousBuildLink && $ctrl.currentBuildLink">
+    <div class="col-md-6">
+      <strong>Previous: </strong>
+      <a ng-href="{{$ctrl.previousBuildLink}}" target="_blank">Build: #{{$ctrl.buildInfo.ancestor}}</a>
+    </div>
+    <div class="col-md-6">
+      <strong>Current: </strong>
+      <a ng-href="{{$ctrl.currentBuildLink}}" target="_blank">Build: #{{$ctrl.buildInfo.target}}</a>
+    </div>
+  </div>
   <div class="modal-body" style="max-height: 800px; overflow-y: scroll;" sticky-headers>
     <div ng-if="$ctrl.commits.length">
       <sticky-header>
--- a/app/scripts/modules/core/src/diffs/viewChangesLink.component.ts
+++ b/app/scripts/modules/core/src/diffs/viewChangesLink.component.ts
@@ -1,13 +1,15 @@
+import {extend} from 'lodash';
 import {IComponentController, IComponentOptions, module} from 'angular';
 import {IModalInstanceService, IModalService} from 'angular-ui-bootstrap';
 
-import {ICreationMetadata, ICreationMetadataTag} from 'core/domain';
+import {IBuildDiffInfo, ICreationMetadata, ICreationMetadataTag, IJenkinsInfo} from 'core/domain';
 import {ICommit} from './commitHistory.component';
 import {COMMIT_HISTORY_COMPONENT} from './commitHistory.component';
 import {EXECUTION_SERVICE, ExecutionService} from 'core/delivery/service/execution.service';
 import {JAR_DIFF_COMPONENT, IJarDiff} from './jarDiff.component';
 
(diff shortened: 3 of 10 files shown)


app/scripts/modules/amazon/src/serverGroup/details/serverGroupDetails.aws.controller.js
```diff
@@ -154,6 +154,11 @@ module.exports = angular.module('spinnaker.amazon.serverGroup.details.controller
                 this.changeConfig = {
                   metadata: get(this.serverGroup.entityTags, 'creationMetadata')
                 };
+                if (this.serverGroup.buildInfo.jenkins) {
```
icfantv: TBH, it feels weird to be doing the jenkins check outside of the netflix module, but we're doing it elsewhere so not sure what the best solution here is.  It feels like we should move this into netflix, but not sure how realistic/feasible that is, currently.

app/scripts/modules/core/src/diffs/changes.html
```diff
@@ -3,6 +3,24 @@
   <div class="modal-header has-sticky-headers">
     <h3>Changes to {{$ctrl.nameItem.name}}</h3>
   </div>
+  <div class="modal-body" ng-if="(!$ctrl.previousBuildLink || !$ctrl.currentBuildLink) && $ctrl.buildInfo.ancestor && $ctrl.buildInfo.target">
```
icfantv: make this section fixed and always visible, even if the diff sections are scrollable.

app/scripts/modules/core/src/diffs/viewChangesLink.component.ts
```diff
@@ -1,3 +1,4 @@
+import {extend} from 'lodash';
```
icfantv: I don't love what's happened to this component (and it's all my fault).  I think it needs to be more normalized across the two places it's used such that it's completely agnostic as to the data source.

app/scripts/modules/core/src/pipeline/config/stages/deploy/deployExecutionDetails.controller.js
```diff
@@ -79,9 +82,17 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.deploy.details.co
       $scope.provider = context.cloudProvider || context.providerType || 'aws';
 
       $scope.changeConfig = {
+        buildInfo: context.buildInfo || {},
         commits: $scope.stage.context.commits,
         jarDiffs: $scope.stage.context.jarDiffs
       };
+      const serverGroupName = context['deploy.server.groups'][context.source.region][0];
```
icfantv: This was the CD/Orca fix/feature to add the build information to a single server group retrieval (Clouddriver, https://github.com/spinnaker/clouddriver/pull/1718), and add the previous/current build numbers to the commit diff task (Orca, https://github.com/spinnaker/orca/pull/1429).

See the previous comment about the jenkins reference.

icfantv:
...fixing broken test...

icfantv:
test fixed, here's a screenshot i forgot earlier.
screen shot 2017-07-03 at 12 44 33 pm

app/scripts/modules/amazon/src/serverGroup/details/serverGroupDetails.aws.controller.js
anotherchrisberry: Is the `changeConfig` feature specific to Netflix? `buildInfo.jenkins` is not (check out the end of Part 1 of [this codelab](https://www.spinnaker.io/guides/tutorials/codelabs/gce-source-to-prod/#part-2-validate-test)), so I'd guess it would work for anyone using Rosco and having the entity tags feature enabled.

app/scripts/modules/core/src/diffs/changes.html
```diff
@@ -3,6 +3,24 @@
...
+    <div class="col-md-6">
+      <strong>Previous Build:&nbsp;</strong>{{$ctrl.buildInfo.ancestor}}
+    </div>
+    <div class="col-md-6">
+      <strong>Current Build:&nbsp;</strong>{{$ctrl.buildInfo.target}}
```
anotherchrisberry: For consistency, we should include a hash before the build numbers.

Also, is the `&nbsp;` necessary?

app/scripts/modules/core/src/diffs/changes.html
```diff
@@ -3,6 +3,24 @@
...
+  <div class="modal-body" ng-if="$ctrl.previousBuildLink && $ctrl.currentBuildLink">
+    <div class="col-md-6">
+      <strong>Previous Build:&nbsp;</strong>
+      <a ng-href="{{$ctrl.previousBuildLink}}" target="_blank">{{$ctrl.buildInfo.ancestor}}</a>
+    </div>
+    <div class="col-md-6">
+      <strong>Current Build:&nbsp;</strong>
+      <a ng-href="{{$ctrl.currentBuildLink}}" target="_blank">{{$ctrl.buildInfo.target}}</a>
+    </div>
+  </div>
```
anotherchrisberry: Same comment/question as above re: hash/nbsp

app/scripts/modules/core/src/diffs/viewChangesLink.component.ts
```diff
@@ -7,7 +8,20 @@ import {COMMIT_HISTORY_COMPONENT} from './commitHistory.component';
...
+  host: string;
+  name: string;
+  number: string;
+}
+
+export interface IBuildInfo {
+  ancestor: string,
+  jenkins?: IJenkinsInfo;
+  target: string,
+}
```
anotherchrisberry: Feels like this is really a domain class associated with `IServerGroup` and could move in there, although the `ancestor` field is maybe specific to an entity tag on the server group.

app/scripts/modules/amazon/src/serverGroup/details/serverGroupDetails.aws.controller.js
icfantv: The `changeConfig` is just a means of getting data into the changes link component.  Which is why I want to normalize the input.  Appreciate the additional info on jenkins, that's helpful.

app/scripts/modules/core/src/diffs/changes.html
icfantv: Yea, I went both ways on the actual text, but makes sense to match server groups.  I didn't like the spacing look w/o the non-breaking space.  Let me play with the text a little more after adding `Build` and see if that makes any difference.

app/scripts/modules/core/src/diffs/viewChangesLink.component.ts
icfantv: That's a good point.  This actually belongs on `IStage` since `ancestor` and `target` are the new fields I added to the new `buildInfo` property on the stage context in orca.  I'll move to there.

app/scripts/modules/core/src/diffs/changes.html
icfantv: Decided to match the existing pattern on ASGs which do not use the non-breaking space and have a `Version: Build: #xx` syntax.

icfantv:
Rebased and changes made per discussion.

anotherchrisberry:
LGTM

PR #3894, opened 2017-07-07
feat(provider/google): Support TCP Proxy Load Balancing

sayap:
Mostly copy-paste from SSL Proxy code.

### DIFF
--- a/app/scripts/modules/google/src/loadBalancer/configure/ssl/listener.html
+++ b/app/scripts/modules/google/src/loadBalancer/configure/ssl/listener.html
@@ -23,9 +23,11 @@
               </ui-select>
             </td>
             <td>
-              <input class="form-control input-sm"
-                     readonly
-                     required ng-model="ctrl.loadBalancer.portRange"/>
+              <select class="form-control input-sm"
+                      ng-options="port for port in ctrl.portOptions"
+                      ng-disabled="!ctrl.isNew"
+                      ng-model="ctrl.loadBalancer.portRange">
+              </select>
             </td>
           </tr>
           </tbody>
--- a/app/scripts/modules/google/src/loadBalancer/configure/tcp/gceCreateTcpLoadBalancer.controller.ts
+++ b/app/scripts/modules/google/src/loadBalancer/configure/tcp/gceCreateTcpLoadBalancer.controller.ts
@@ -1,0 +1,219 @@
+import { module, IScope } from 'angular';
+import { IModalInstanceService } from 'angular-ui-bootstrap';
+import { StateService } from '@uirouter/angularjs';
+import * as _ from 'lodash';
+
+import {
+  ACCOUNT_SERVICE,
+  AccountService,
+  Application,
+  IAccount,
+  ICredentials,
+  ILoadBalancerUpsertCommand,
+  IInstance,
+  INFRASTRUCTURE_CACHE_SERVICE,
+  InfrastructureCacheService,
+  IRegion,
+  LOAD_BALANCER_WRITE_SERVICE,
+  LoadBalancerWriter,
+  TASK_MONITOR_BUILDER,
+  TaskMonitorBuilder
+} from '@spinnaker/core';
+
+import { IGceBackendService, IGceHealthCheck, IGceLoadBalancer } from 'google/domain/index';
+import { GCEProviderSettings } from 'google/gce.settings';
+import { CommonGceLoadBalancerCtrl } from '../common/commonLoadBalancer.controller';
+import {
+  GCE_COMMON_LOAD_BALANCER_COMMAND_BUILDER,
+  GceCommonLoadBalancerCommandBuilder
+} from '../common/commonLoadBalancerCommandBuilder.service';
+import { GCE_HEALTH_CHECK_SELECTOR_COMPONENT } from '../common/healthCheck.component';
+
+class ViewState {
+  constructor(public sessionAffinity: string) {}
+}
+
+interface IListKeyedByAccount {
+  [account: string]: string[];
+}
+
+interface ITcpLoadBalancerUpsertDescription extends ILoadBalancerUpsertCommand {
+  backendService: IGceBackendService;
+  loadBalancerName: string;
+  instances: IInstance[];
+}
+
+class TcpLoadBalancer implements IGceLoadBalancer {
+  public stack: string;
+  public detail: string;
+  public loadBalancerName: string;
+  public portRange = '443';
+  public ipProtocol = 'TCP';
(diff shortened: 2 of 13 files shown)


sayap:
PR for clouddriver: https://github.com/spinnaker/clouddriver/pull/1727

app/scripts/modules/google/src/loadBalancer/configure/tcp/listener.html
```diff
@@ -0,0 +1,28 @@
...
+                     readonly
+                     required ng-model="ctrl.loadBalancer.portRange"/>
+            </td>
+          </tr>
+          </tbody>
+        </table>
+      </div>
+    </div>
+  </ng-form>
+  <div class="form-group small" style="margin-top: 20px;">
```
danielpeach: This looks extraneous.

app/scripts/modules/google/src/loadBalancer/configure/tcp/listener.html
```diff
@@ -0,0 +1,28 @@
...
+          <thead>
+          <tr>
+            <th>Port</th>
+          </tr>
+          </thead>
+          <tbody>
+          <tr>
+            <td>
+              <input class="form-control input-sm"
+                     readonly
```
danielpeach: Should this be `readonly`?

sayap: This should actually be a dropdown list, for both SSL and TCP Proxy. Will change

jtk54:
Backend was merged: https://github.com/spinnaker/clouddriver/pull/1727. I think we should focus on getting this functionality in before worrying too much about the de-duplication between TCP and SSL LBs.

app/scripts/modules/google/src/loadBalancer/configure/tcp/gceCreateTcpLoadBalancer.controller.ts
```diff
@@ -0,0 +1,220 @@
...
+  backendService: IGceBackendService;
+  loadBalancerName: string;
+  instances: IInstance[];
+}
+
+class TcpLoadBalancer implements IGceLoadBalancer {
+  public stack: string;
+  public detail: string;
+  public loadBalancerName: string;
+  public portRange = '443';
```
danielpeach: Any reason not to match the default in cloud console (25)?

app/scripts/modules/google/src/loadBalancer/configure/ssl/listener.html
```diff
@@ -23,9 +23,11 @@
               </ui-select>
             </td>
             <td>
-              <input class="form-control input-sm"
-                     readonly
-                     required ng-model="ctrl.loadBalancer.portRange"/>
+              <select class="form-control input-sm"
```
danielpeach: This doesn't seem to work properly - if regardless of the value I select here, the resulting frontend's port is 443. This is not a UI bug (I verified that the payload leaving Deck is sending the right port), but this should be resolved before merging this PR.

app/scripts/modules/google/src/serverGroup/configure/wizard/cloneServerGroup.gce.controller.js
```diff
@@ -277,7 +279,16 @@ module.exports = angular.module('spinnaker.serverGroup.configure.gce.cloneServer
             : [])
         .value();
 
-      return loadBalancerNames.concat(selectedSslLoadBalancerNames);
+      var selectedTcpLoadBalancerNames = _.chain(loadBalancerIndex)
```
danielpeach: s/var/let

app/scripts/modules/google/src/serverGroup/configure/wizard/loadBalancingPolicy/loadBalancingPolicySelector.component.ts
```diff
@@ -41,19 +41,20 @@ class GceLoadBalancingPolicySelectorController implements ng.IComponentControlle
...
     if (has(this, 'command.backingData.filtered.loadBalancerIndex')) {
       const index = this.command.backingData.filtered.loadBalancerIndex;
       const selected = this.command.loadBalancers;
 
       const hasSsl = selected.find((loadBalancer: any) => get(index[loadBalancer], 'loadBalancerType') === 'SSL');
+      const hasTcp = selected.find((loadBalancer: any) => get(index[loadBalancer], 'loadBalancerType') === 'TCP');
       const hasHttp = selected.find((loadBalancer: any) => get(index[loadBalancer], 'loadBalancerType') === 'HTTP');
 
-      if ((hasSsl && hasHttp)) {
+      if (((hasSsl || hasTcp) && hasHttp)) {
```
danielpeach: Please remove the extra set of parentheses here (the fault is in the original code, but no reason to duplicate it).

sayap:
Thanks for the review. Good catch on the issues 😅 

I have created PR for issue 1 and issue 3, but can't really reproduce issue 2 yet. I guess it could be related to issue 1?

danielpeach:
I consistently get an NPE here on `collectEntries`: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/GCEUtil.groovy#L1275

Actually, I get this for SSL load balancers also. I'm guessing that's because I don't have any URL maps in my project right now.

@jtk54 can you look? I think we just need 
```
projectUrlMaps?.collectEntries { ... } ?: []
```

danielpeach:
Could you please center this field? I also don't think it should be a table - could just be a standalone input field.
tcp-port

Please also resolve the remaining comments in this PR, then we can get it merged. Thanks.

sayap:
Just pushed a commit that addresses the comments.

Regarding the request to center the port field, I make the field aligns with the others instead.

Regarding the port default of 443, I think it is more important to be consistent between SSL Proxy and TCP Proxy in spinnaker, compared to being consistent between cloud console and spinnaker. Also, our use case for TCP Proxy is to workaround the lack of mutual TLS support in the HTTPS Proxy, so port 443 works for me :)

jtk54:
This PR depends on https://github.com/spinnaker/clouddriver/pull/1748 -- without it, creating a server group attached to TCP/SSL fails if not urlMaps exist. Looks like that bug has been dormant for a while. Otherwise this PR tests out fine for me.

jtk54:
Dependent PR merged, this one needs a rebase then LGTM.

sayap:
Thanks, rebased.

jtk54:
Thanks for your contribution.

PR #3899, opened 2017-07-10
chore(nightly build): Add cloudbuild-tagged.yaml for nightly build

ttomsu:


cloudbuild-tagged.yaml
```diff
@@ -0,0 +1,14 @@
+steps:
+- name: 'java:8'
```
jtk54: I bet there's a way to wire the yarn cache into a starting image... Something to look at for another day.

PR #3930, opened 2017-07-18
refactor(*): Update typescript to 2.4 and fix breaking changes

christopherthielen:
Supersedes #3910 

I cherry picked the commit by @jrsquared into master, updated `@types/angular` and `@types/angular-mocks`, removed the `$onInit()` methods, and fixed an error or two.  I also removed some more `ng.*` stuff.  

Seems to work on my machine, but it would be safest if ya'll pulled it down to check if it works with retina.

@jrsquared @icfantv @anotherchrisberry


jrsquared:
What the heck. Totally works. I'm not going to try to figure out why, I'm just totally happy to get this behind me. MERGE IT NOW.

PR #3972, opened 2017-08-01
feat(CI/Jenkins): Add parameter type checking/mapped elements to Jenkins task UI.

Atchuu:
Changes were made to the Jenkins task UI used by Deck to support
specific parameter types already provided by Igor. The default was left as
input:text to keep the functionality for unknown/unspecified parameter types.

Recommendations/changes welcome.

[https://github.com/spinnaker/spinnaker/issues/1872](Issue)

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/jenkins/jenkinsStage.html
+++ b/app/scripts/modules/core/src/pipeline/config/stages/jenkins/jenkinsStage.html
@@ -65,9 +65,38 @@
         <b class="break-word">{{parameter.name}}</b>
         <help-field content="{{parameter.description}}" ng-if="parameter.description"></help-field>
       </div>
-      <div class="col-md-5">
-        <input disabled ng-if="useDefaultParameters[parameter.name]" type="text" class="form-control input-sm" value="{{parameter.defaultValue}}"/>
-        <input ng-if="!useDefaultParameters[parameter.name]" type="text" class="form-control input-sm" ng-model="userSuppliedParameters[parameter.name]" ng-change="updateParam(parameter.name)"/>
+      <div class="col-md-5" ng-switch on="parameter.type">
+        <div ng-switch-when="PasswordParameterDefinition">
+          <!-- Passwords do not expose their defaults, for obvious reasons, so we can skip this -->
+          <input ng-if="!useDefaultParameters[parameter.name]" type="password" class="form-control input-sm" ng-model="userSuppliedParameters[parameter.name]" ng-change="updateParam(parameter.name)"/>
+        </div>
+        <div ng-switch-when="BooleanParameterDefinition">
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/jenkins/jenkinsStage.html
```diff
@@ -65,9 +65,38 @@ <h4 class="text-left" >Job Parameters</h4>
...
-        <input disabled ng-if="useDefaultParameters[parameter.name]" type="text" class="form-control input-sm" value="{{parameter.defaultValue}}"/>
-        <input ng-if="!useDefaultParameters[parameter.name]" type="text" class="form-control input-sm" ng-model="userSuppliedParameters[parameter.name]" ng-change="updateParam(parameter.name)"/>
+      <div class="col-md-5" ng-switch on="parameter.type">
+        <div ng-switch-when="PasswordParameterDefinition">
+          <!-- Passwords do not expose their defaults, for obvious reasons, so we can skip this -->
+          <input ng-if="!useDefaultParameters[parameter.name]" type="password" class="form-control input-sm" ng-model="userSuppliedParameters[parameter.name]" ng-change="updateParam(parameter.name)"/>
+        </div>
+        <div ng-switch-when="BooleanParameterDefinition">
+          <input disabled ng-if="useDefaultParameters[parameter.name] && parameter.defaultValue == 'true'" type="checkbox" class="input-sm" checked/>
+          <input disabled ng-if="useDefaultParameters[parameter.name] && parameter.defaultValue == 'false'" type="checkbox" class="input-sm"/>
```
anotherchrisberry: Can these be collapsed into a single input?
```html
<input disabled ng-if="useDefaultParameters[parameter.name] type="checkbox" ng-check="parameter.defaultValue === 'true'"/>
```
?

Atchuu: I will test it and find out. Give me one second.

Atchuu: It doesn't appear to like it. I seem to recall trying this when I was first setting the boolean section up.

image

anotherchrisberry: Cool, thanks for checking

Atchuu: There was a syntax error that I didn't notice (missing "). Fixing it allows the checkbox to appear. The problem now is that default value seems to be ignored. (it isn't auto-checking the checkbox) I can investigate further if you think it's useful, but at this point, it's one extra if/then check so I'm not sure if it's a good use of time :). Thoughts?

PR #3987, opened 2017-08-04
fix(k8s): allow containers with similar image names to be selected

danielpeach:

Closes https://github.com/spinnaker/spinnaker/issues/1880.


ethanfrogers:
I'll give it a quick test!

PR #3992, opened 2017-08-07
style(core/styleguide) Cataloging Spinnaker styles in to a style guide

archana-s:
Adding a consolidated styleguide for Spinnaker based on existing patterns. 

The style guide includes the following:
1. colors
2. type
3. buttons/links
4. icons
5. tabs 
6. layouts
7. layers
8. pods
9. pills 
10. icons
It uses postcss for CSS processing and style guide generation. The minified CSS and style guide are present in public directory. All of these styles are currently scoped at body with ".styleguide" selector. It is currently intended to be used for plugins going in to spinnaker.


archana-s:
@anotherchrisberry I placed styleguide inside src. Does this belong more to /lib?

PR #3997, opened 2017-08-08
fix(core/cluster): Fix sort order of accounts in clusters view

christopherthielen:


app/scripts/modules/core/src/cluster/filter/clusterFilter.service.ts
```diff
@@ -118,7 +118,7 @@ export class ClusterFilterService {
     this.clusterFilterModel.asFilterModel.addTags();
```
christopherthielen: the line above this one `this.sortGroupsByHeading` sorts the `asFilterModel.groups` array in place using the local variable `groups` as a reference

PR #4006, opened 2017-08-10
feat(openstack): Add advance options to openstack bake configuration

edwinavalos:
@danielpeach this new one should be GTG

### DIFF
--- a/app/scripts/modules/openstack/pipeline/stages/bake/openstackBakeStage.js
+++ b/app/scripts/modules/openstack/pipeline/stages/bake/openstackBakeStage.js
@@ -3,7 +3,13 @@
 const angular = require('angular');
 import _ from 'lodash';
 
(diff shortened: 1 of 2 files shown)


app/scripts/modules/openstack/pipeline/stages/bake/openstackBakeStage.js
```diff
@@ -85,11 +97,40 @@ module.exports = angular.module('spinnaker.openstack.pipeline.stage.bakeStage',
...
+
     this.showTemplateFileName = function() {
       return $scope.viewState.roscoMode || $scope.stage.templateFileName;
     };
 
     $scope.$watch('stage', deleteEmptyProperties, true);
 
     initialize();
-  });
+  });:console.warn();
```
danielpeach: So close - we can merge after this gets dropped.

edwinavalos: UGH MY EDITOR IS THE WORST

PR #4072, opened 2017-09-05
feat(search): add advanced search/filtering

icfantv:
* adds a new endpoint `infrastructureNew` for testing the new search API and flushing out any issues.
* outstanding issues remaining:
  * do we allow more than one tag of the same type to be entered?  searches are AND, not OR.
  * when a search returns too many results, we should display some sort of message, but the current design does not really accommodate a message bar somewhere.
  * performance on certain types of searches are _SLOW_ so we should decide how we want to handle this.
  * due to how application searches were being performed, searches without a keyword will return all applications.  https://github.com/spinnaker/clouddriver/pull/1875 was entered to fix this.
  * support for other cloud providers will be needed in clouddriver if they want to be able to take advantage of this feature.
* the UI is expected to change based on internal feedback.
* the original search page should still function normally.


app/scripts/modules/core/src/application/ApplicationDisplayRenderer.tsx
```diff
@@ -0,0 +1,36 @@
+import { IApplication } from 'core/domain';
```
icfantv: All the application renderer implementing classes have the same structure so if there's an issue with one, there's probably the same issue with all of them.

app/scripts/modules/core/src/application/ApplicationSearchResultFormatter.ts
```diff
@@ -18,6 +16,9 @@ export class ApplicationSearchResultFormatter implements ISearchResultFormatter
   public displayFormatter(searchResult: IApplicationSearchResult): IPromise<string> {
     return $q.when(searchResult.application);
   }
+  public get displayRenderer(): IResultRenderer {
```
icfantv: All of the formatters will have this change for their corresponding renderer.

app/scripts/modules/core/src/domain/IIndexed.ts
```diff
@@ -0,0 +1,3 @@
+export interface IIndexed {
```
icfantv: This was needed to make typescript happy in the abstract render class for the default cell renderer.  I don't love it.

app/scripts/modules/core/src/projects/ProjectSearchResultFormatter.ts
```diff
@@ -16,13 +14,18 @@ export class ProjectSearchResultFormatter implements ISearchResultFormatter {
 
   public get displayName() { return 'Projects'; }
   public get order() { return 0; }
+  public get icon() { return 'hand-spock-o'; }
```
icfantv: We need to decide on a proper Project icon....as cool as _live long and prosper_ is...

app/scripts/modules/core/src/search/externalSearch.registry.ts
```diff
@@ -20,7 +20,7 @@ export interface IExternalSearchConfig {
    * Method to fetch search results
    * @param query
    */
-  search: (query: string) => IPromise<ISearchResult[]>;
+  search: (query: string | { [key: string]: string }) => IPromise<ISearchResult[]>;
```
icfantv: Since the new search API can take a map of key/value filter pairs, we need to update the search API in deck to handle this and pass along the filters accordingly.

app/scripts/modules/core/src/search/infrastructure/infrastructure.controller.js
```diff
@@ -75,13 +72,13 @@ module.exports = angular.module('spinnaker.search.infrastructure.controller', [
     var autoNavigate = false;
 
     if (angular.isDefined($location.search().q)) {
-      this.query = $location.search().q;
+      $scope.query = $location.search().q;
```
icfantv: I broke the original search page such that the query string was no longer a query parameter in the URL.  Changes to this file unbreak it.

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
```diff
@@ -0,0 +1,185 @@
...
+
+    ctrl.viewState = {
+      status: SearchStatus.INITIAL,
+      minCharactersToSearch: 3,
+    };
+
+    this.handleFilterChange = function (filters) {
+
+      const params = {};
+      filters.slice(0).reverse().forEach(filter => params[SearchFilterTypeRegistry.getFilterByModifier(filter.modifier).key] = filter.text);
```
icfantv: If the user specifies more than one filter of the same kind, we want the first one to win.  See _issues remaining_ to decide on whether or not we want to allow multiple at all...

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
```diff
@@ -0,0 +1,185 @@
...
+      $location.search(params);
+      $location.replace();
+      loadNewQuery(params);
+    };
+
+    this.clearFilters = r => clusterFilterService.overrideFiltersForUrl(r);
```
icfantv: I don't think this is actually used/necessary anymore...

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
```diff
@@ -0,0 +1,185 @@
...
+
+    $scope.pageSize = searchService.defaultPageSize;
+    var autoNavigate = false;
```
icfantv: or this

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
```diff
@@ -0,0 +1,185 @@
...
+
+    if (angular.isDefined($location.search().q)) {
```
icfantv: or this

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
```diff
@@ -0,0 +1,185 @@
...
+        action: this.createApplication
+      },
+      {
+        displayName: 'Create Project',
+        action: this.createProject
+      },
+      refreshMenuItem,
+    ];
+
+    this.hasResults = () => {
```
icfantv: I suspect this isn't used in this page.

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
```diff
@@ -0,0 +1,185 @@
...
+      return $scope.categories.length || $scope.projects.length;
+    };
+
+    this.noMatches = () => !this.hasResults() && ctrl.query && Object.keys(ctrl.query).length > 0;
```
icfantv: or this

anotherchrisberry:
Overall it looks nice (haven't looked at the code yet). 

Some issues I'm seeing:
* I can't get any search query to return a project
* The page title always says `search results for "[object Object]"`
* I keep typing some text, then trying to click one of the field options, e.g. "name" or "key", but clicking on it just closes the menu without selecting an option, so then I have to delete a character or type another one to get the menu to reappear, then use the keyboard to select a field. I've done this at least fifteen times now.
* There is no cursor distinction between enabled search result categories and disabled ones, so you have to guess to find out you can click on the tabs on the left
* There's a pretty big gap between the name of a thing and the other fields in the table, so I'm finding it difficult to distinguish things like security groups, where there are several with similar names in different regions/accounts - scanning across the table is not great.

icfantv:
@anotherchrisberry I think I've addressed all of your bullets.  I'll need to come up with a better tab title than just the JSON.

Adding screenshots:
screen shot 2017-09-05 at 10 00 02 pm
screen shot 2017-09-05 at 10 02 16 pm
screen shot 2017-09-05 at 10 03 24 pm
screen shot 2017-09-05 at 10 06 38 pm
screen shot 2017-09-05 at 10 07 35 pm

anotherchrisberry:
Hmm, every time I click on one of the field options (region, account, name, key), it always sets it to "Stack".

I'm also seeing weird behavior when doing a name search for "Deck". It eventually comes back, but it's showing every application. I think you mentioned you were going to look into this.

(Will dig into the actual code tomorrow)

icfantv:
Whoops, I know what I did wrong for the click support, I'll get a fix in this morning.  We'll need to roll out a new clouddriver to get my application search change in.  I'll confirm with Tomás that this is good to go.

app/scripts/modules/core/src/domain/IIndexed.ts
jrsquared: What about just using `get<string>(item, key)` in the cell renderers so we don't litter generic keys across all of our main types?

app/scripts/modules/core/src/domain/IProject.ts
```diff
@@ -15,10 +17,17 @@ export interface IProjectConfig {
   clusters: IProjectCluster[];
 }
 
-export interface IProject {
+export interface IProject extends IIndexed {
+  config: IProjectConfig;
+  createTs?: number;
+  displayName?: string;
+  email: string;
+  href?: string;
```
jrsquared: `href` doesn't actually exist on the types until they return in a search result; it feels a little dirty to mutate objects as a result of searching.

app/scripts/modules/core/src/application/application.less
```diff
@@ -164,3 +164,22 @@ input[type="checkbox"], input[type="radio"] {
   cursor: pointer;
 }
 
+@table-name: application;
+.generate-table(@table-name);
+
+.@{table-name}-table {
+  .table-header, .@{table-name}-row {
+    .@{table-name}-application {
+      left: 0;
```
anotherchrisberry: I don't think the `left` is doing anything here or any of the other columns

app/scripts/modules/core/src/projects/ProjectDisplayRenderer.tsx
```diff
@@ -0,0 +1,35 @@
+import { IProject } from 'core/domain';
+import {
+  AbstractBaseResultRenderer,
+  ITableColumnConfigEntry
+} from 'core/search/searchResult/AbstractBaseResultRenderer';
+
+import './project.less';
+
+export class ProjectDisplayRenderer extends AbstractBaseResultRenderer<IProject> {
```
jrsquared: Since these renderers are effectively static, can't we just eschew the whole `extends AbstractBase` thing and just have this configuration be a const passed into a `ResultRenderer<T>.new(config)` or something (didn't spend too much time thinking of architecture)? Seems weird having a static instance with a public accessor (The Singleton pattern feels weird in JavaScript)

app/scripts/modules/core/src/projects/ProjectSearchResultFormatter.ts
jrsquared: It's not.

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
```diff
@@ -0,0 +1,157 @@
+'use strict';
```
jrsquared: Why is this not typescript?

app/scripts/modules/core/src/search/searchResult/AbstractBaseResultRenderer.tsx
```diff
@@ -0,0 +1,120 @@
...
+      );
+    } else {
+      result = (<div key="unknown" className={`${this.getRendererClass()}-account`}>-</div>);
+    }
+
+    return result;
+  };
+
+  public valueOrDefaultCellRenderer(item: T, key: string, defaultValue = ''): JSX.Element {
+    return (<div className={`${this.getRendererClass()}-${key}`} key={key}>{item[key] || defaultValue}</div>);
```
jrsquared: Single lines of JSX don't need the parentheses. I thought lint would break on it, but apparently not. I don't care either way, I should probably just delete this comment.

app/scripts/modules/core/src/domain/IApplication.ts
```diff
@@ -0,0 +1,22 @@
+import { IIndexed } from './IIndexed';
+
+export interface IApplication extends IIndexed {
```
anotherchrisberry: I feel like we're conflating search results with actual domain items here (same with cluster, instance, etc.).

Can we rename these if they're new, or create separate classes to be more clear what they are instead of combining them with the existing domain classes?

app/scripts/modules/core/src/domain/ISecurityGroup.ts
```diff
@@ -14,21 +16,28 @@ export interface IUsages {
...
 }
 
-export interface ISecurityGroup {
+export interface ISecurityGroup extends IIndexed {
   account?: string;
   accountId?: string;
   accountName?: string;
   application?: string;
   cloudProvider?: string;
+  cluster?: string;
```
anotherchrisberry: What's the `cluster` field on a security group? Same with `sequence`, `serverGroup`...

app/scripts/modules/core/src/domain/IServerGroup.ts
```diff
@@ -11,9 +12,10 @@ export interface IAsg {
   desiredCapacity: number;
 }
 
-export interface IServerGroup {
+export interface IServerGroup extends IIndexed {
   account: string;
   app?: string;
+  application?: string;
```
anotherchrisberry: why both `app` and `application`?

app/scripts/modules/core/src/loadBalancer/LoadBalancerSearchResultFormatter.ts
```diff
@@ -16,11 +14,14 @@ export class LoadBalancerSearchResultFormatter implements ISearchResultFormatter
 
   public get displayName() { return 'Load Balancers'; }
   public get order() { return 5; }
-  public get iconClass() { return 'icon icon-elb'; }
+  public get iconClass() { return 'icon icon-elb'; }  // TODO: ANG replace w/ fa-sitemap
```
anotherchrisberry: any reason not to do this now?

app/scripts/modules/core/src/search/widgets/Search.tsx
```diff
@@ -256,7 +276,16 @@ export class Search extends React.Component<ISearchProps, ISearchState> {
     this.inputElement.focus();
   }
 
-  private handleFilterChange(key: Key): void {
+  private filterClicked(filter: Filter): void {
+    this.setState({
+      activeFilter: filter.props.filterType
+    }, () => {
```
jrsquared: `setState` callbacks are discouraged. Is there a reason you are using a callback instead of just putting this in `componentDidUpdate()`?

It is also odd that after this setState resolves setting `activeFilter`, you (may) immediately call another `setState` inside `handleFilterSelection()` to set `activeFilter` again. Is there benefit in doing these so quickly in succession or can this be refactored?

app/scripts/modules/core/src/search/widgets/Search.tsx
```diff
@@ -267,13 +296,18 @@ export class Search extends React.Component<ISearchProps, ISearchState> {
...
+  private handleMouseDown(): void {
+    this.mouseDownFired = true;
+  }
+
   private handleClearClick(): void {
     this.setState({
       isOpen: false,
       tags: []
     });
     this.inputElement.value = '';
```
jrsquared: Did you try to actually bind `inputElement`s `value` to state and then modify the state `onChange` instead of stepping outside of React for this? (https://facebook.github.io/react/docs/forms.html)

app/scripts/modules/core/src/search/infrastructure/infrastructure.states.ts
```diff
@@ -25,4 +25,22 @@ module(INFRASTRUCTURE_STATES, [
   });
   stateConfigProvider.addRewriteRule('/', '/infrastructure');
   stateConfigProvider.addRewriteRule('', '/infrastructure');
+}).config((stateConfigProvider: StateConfigProvider) => {
+  stateConfigProvider.addToRootState({
+    name: 'infrastructureNew',
+    url: '/infrastructureNew?q',
```
anotherchrisberry: maybe `/infrastructure/v2`?

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
anotherchrisberry: If we only allow a single filter by kind, we shouldn't allow users to select more than one of each kind - it's not intuitive that other selections will be ignored...

app/scripts/modules/core/src/search/infrastructure/infrastructureNew.controller.js
anotherchrisberry: Agreed, this should be in TS

icfantv: because i hate freedom.

app/scripts/modules/core/src/search/searchResult/baseTable.less
```diff
@@ -0,0 +1,65 @@
...
+      font-weight: 600;
+      background-color: @lightest_grey;
+      z-index: 2;
+      .fa {
+        padding-left: 2px;
+        &.disabled {
+          color: @mid_light_grey;
+        }
+      }
+      .column-picker {
```
anotherchrisberry: I don't think this is used anywhere

app/scripts/modules/core/src/search/widgets/Filters.tsx
```diff
@@ -17,34 +17,60 @@ export interface IFiltersProps {
...
   };
 
+  private filters: Filter[] = [];
+
+  private refCallback(filter: Filter): void {
+    this.filters.push(filter);
+  }
+
+  private handleClick(key: string): void {
+    console.log('key', key);
```
anotherchrisberry: console

app/scripts/modules/core/src/search/searchResult/AbstractBaseResultRenderer.tsx
anotherchrisberry: I personally don't mind wrapping something like this in parens, and would be very angry with lint if it broke on this.

anotherchrisberry:
Okay, all comments added to the code

app/scripts/modules/core/src/application/application.less
icfantv: you're right.  good catch.

app/scripts/modules/core/src/search/searchResult/AbstractBaseResultRenderer.tsx
icfantv: i'm beginning to think you're the one who hates freedom.

app/scripts/modules/core/src/projects/ProjectDisplayRenderer.tsx
icfantv: @christopherthielen and I have talked about a better way to architect this and the persisted property stuff in the -nflx branch.  He doesn't think this should keep us from committing this currently and I want to get this in front of our stakeholders.  I can work on it as soon as it's released as they'll have some time to play.

anotherchrisberry: That makes sense.

app/scripts/modules/core/src/domain/ISecurityGroup.ts
icfantv: I'm not sure where that came from, it's not even on the model in the search results.  Probably an erroneous addition.  It's removed by reverting the domain object changes and adding new ones specific to search results.

app/scripts/modules/core/src/domain/IServerGroup.ts
icfantv: Issue resolved by moving to search result specific domain objects.

app/scripts/modules/core/src/search/widgets/Search.tsx
icfantv: Stop reviewing old code, mister....
I didn't, because I was completely unaware about that - I'll flag that for investigation.

jrsquared: I reviewed this at 10am... it was fresh then!

app/scripts/modules/core/src/loadBalancer/LoadBalancerSearchResultFormatter.ts
icfantv: nope.  done.  there's gobs of other usages in the code for icon/icon-elb we should pull as well.

app/scripts/modules/core/src/search/widgets/Search.tsx
icfantv: Good catch.

PR #4093, opened 2017-09-08
style(infrastructure): Add new spinner to infrastructure search

archana-s:
Added the new three bar spinner to infrastructure search. In the process, cleaned up inputs.css from styleguide since we do not need anything from it today unless we make some modifications. 

This diff includes the following three changes: 
1. three bar spinner for infrastructure search
2. cleaned up inputs.css from styleguide 
3. fixed presence of scroller when not necessary in containers in infrastructure area.


app/scripts/modules/core/src/search/infrastructure/infrastructure.controller.js
```diff
@@ -10,6 +10,7 @@ import { OVERRIDE_REGISTRY } from 'core/overrideRegistry/override.registry';
 import { RECENT_HISTORY_SERVICE } from 'core/history/recentHistory.service';
 import { PAGE_TITLE_SERVICE } from 'core/pageTitle/pageTitle.service';
 import { INFRASTRUCTURE_SEARCH_SERVICE } from 'core/search/infrastructure/infrastructureSearch.service';
+import { SPINNER_COMPONENT } from 'core/widgets/spinners/spinner.component';
```
archana-s: Should this be loaded at the bootstrap module level instead of at each component using it?

PR #4095, opened 2017-09-08
fix(quay): Un-break quay build

christopherthielen:


tomaslin:
it's fixing docker builds

PR #4113, opened 2017-09-19
chore(spinner): fix react warning for missing key

icfantv:


app/scripts/modules/core/src/widgets/spinners/Spinner.tsx
```diff
@@ -24,7 +24,7 @@ export class Spinner extends React.Component<ISpinnerProps> {
     const rows = [];
     let i: number;
     for (i = 0; i < count; i++) {
-      rows.push(<div className="bar" />)
+      rows.push(<div key={i} className="bar" />)
```
archana-s: Thank you :)

icfantv: np

PR #4118, opened 2017-09-20
feat(pipeline): Add pipeline config section for artifacts.

jtk54:
Adds a section to pipeline config for declaring expected artifacts from trigger sources, and adds support in pub/sub triggers to select artifacts the trigger can produce.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
@@ -1,0 +1,35 @@
+import { IComponentController, IComponentOptions, module } from 'angular';
+
+import { PIPELINE_CONFIG_PROVIDER } from 'core/pipeline/config/pipelineConfigProvider';
+import { IExpectedArtifact, MissingArtifactPolicy } from 'core/domain/IExpectedArtifact';
+import { IPipeline } from 'core/domain/IPipeline';
+
+class ArtifactController implements IComponentController {
+  public artifact: IExpectedArtifact;
+  public pipeline: IPipeline;
+  public missingPolicies: string[] = Object.keys(MissingArtifactPolicy).map(key => MissingArtifactPolicy[key as any]);
+
+  public removeArtifact(): void {
+    const artifactIndex = this.pipeline.expectedArtifacts.indexOf(this.artifact);
+    this.pipeline.expectedArtifacts.splice(artifactIndex, 1);
+
+    this.pipeline.triggers
+      .forEach(t => t.expectedArtifacts = t.expectedArtifacts.filter(a => !this.isEqual(a, this.artifact)));
+  }
+
+  private isEqual(first: IExpectedArtifact, other: IExpectedArtifact): boolean {
+    return first.name === other.name
+      && first.type === other.type;
+  }
+}
+
+class ArtifactComponent implements IComponentOptions {
+  public bindings = { artifact: '<', pipeline: '<' };
+  public templateUrl = require('./artifact.html');
+  public controller = ArtifactController;
+}
+
+export const ARTIFACT = 'spinnaker.core.pipeline.trigger.artifact';
+module(ARTIFACT, [
+  PIPELINE_CONFIG_PROVIDER,
+]).component('artifact', new ArtifactComponent());
--- a/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.html
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.html
@@ -1,0 +1,65 @@
+<div class="row">
+  <div class="col-md-12">
+    <div class="form-horizontal panel-pipeline-phase">
+      <div class="form-group row">
+        <div class="col-md-10">
+          <div class="row">
+            <label class="col-md-3 sm-label-right">
+              Type
+              <help-field key="pipeline.config.artifact.type"></help-field>
+            </label>
+            <div class="col-md-9">
+              <input type="text"
+                     required
+                     class="form-control input-sm"
+                     ng-model="$ctrl.artifact.type"/>
+            </div>
+          </div>
+        </div>
+
+        <div class="col-md-2 text-right">
+          <button class="btn btn-sm btn-default" ng-click="$ctrl.removeArtifact()">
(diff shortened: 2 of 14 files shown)


jtk54:
image


image

app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
```diff
@@ -18,6 +18,21 @@
       </div>
     </div>
   </page-section>
+  <page-section key="artifacts" label="Artifacts" badge="pipeline.expectedArtifacts.length" no-wrapper="true">
```
danielpeach: I'd call this section `Expected Artifacts`.

danielpeach: I'd also consider moving this down on the page (probably below parameters or notifications), especially since we don't have much support for artifacts yet. I'd also feature flag this.

app/scripts/modules/core/src/pipeline/config/triggers/triggers.directive.js
```diff
@@ -33,5 +34,13 @@ module.exports = angular.module('spinnaker.core.pipeline.config.trigger.triggers
       $scope.pipeline.triggers.push(newTrigger);
     };
 
+    this.addArtifact = function() {
```
danielpeach: Use an arrow function here.

app/scripts/modules/core/src/pipeline/config/triggers/trigger.html
```diff
@@ -37,6 +37,46 @@
...
+          <div class="form-group">
+            <label class=" col-md-3 sm-label-right">Expected Artifacts</label>
+            <div class="col-md-9">
+              <ui-select ng-if="pipeline.expectedArtifacts.length"
+                         multiple
+                         ng-model="trigger.expectedArtifacts"
+                         class="form-control input-sm">
+                <ui-select-match>{{$item.name}}</ui-select-match>
+                <ui-select-choices repeat="artifact in pipeline.expectedArtifacts">
+                  <span ng-bind-html="artifact.name"></span>
```
danielpeach: If the artifact doesn't have a name, I see an empty option here.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.html
```diff
@@ -0,0 +1,49 @@
+<div class="row">
+  <div class="col-md-12">
+    <div class="form-horizontal panel-pipeline-phase">
+      <div class="form-group row">
+        <div class="col-md-10">
+          <div class="row">
+            <label class="col-md-3 sm-label-right">Type</label>
```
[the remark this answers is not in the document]
jtk54: Added tooltips describing the forms and added the `required` and validation.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.html
```diff
@@ -0,0 +1,49 @@
...
+          <div class="row">
+            <label class="col-md-3 sm-label-right">Name</label>
+            <div class="col-md-9">
+              <input type="text" class="form-control input-sm" ng-model="$ctrl.artifact.name"/>
+            </div>
+          </div>
+        </div>
+      </div>
+
+      <div class="form-group row">
```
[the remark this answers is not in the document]
jtk54: Added `Ignore` as another option.

lwander:
spinnaker/spinnaker#1915

anotherchrisberry:
It's probably worth adding some explanatory text to the top of the section. It's not clear to me what these fields are or why I'd need or want to use them.

jtk54:
I fixed up the empty artifact suggestion in the ui-select and added explanatory text to the section with another help field. PTAL @danielpeach and @anotherchrisberry and let me know what you think.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.pipeline.expectedArtifacts.splice(artifactIndex, 1);
+
+    this.pipeline.triggers.map(t => {
+      const found = t.expectedArtifacts.find(a => {
+        return this.isEqual(a, this.artifact);
+      });
+      if (found) {
+        t.expectedArtifacts = t.expectedArtifacts.filter(a => !this.isEqual(a, found));
+      }
+    });
```
anotherchrisberry: You don't need `map` since you're not doing anything with it, so should probably use `forEach` (just to be more clear that you're just iterating through the items).

Seems like this could be simplified quite a bit (unless I'm missing something):
```javascript
this.pipeline.triggers
  .forEach(t => t.expectedArtifacts = t.expectedArtifacts.filter(a => !this.isEqual(a, this.artifact)));
```

app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
```diff
@@ -36,6 +36,27 @@
   <page-section key="parameters" label="Parameters" no-wrapper="true" badge="pipeline.parameterConfig.length">
     <parameters pipeline="pipeline"></parameters>
   </page-section>
+  <page-section key="artifacts" label="Expected Artifacts" badge="pipeline.expectedArtifacts.length" no-wrapper="true">
```
anotherchrisberry: Is this only a thing for pub/sub, or are there other triggers that will include artifacts?

jtk54: We'll have other triggers include artifacts. Ideally eventually most/all triggers would be able to do this. We have a few starting use cases with pub/sub -> gae and docker -> k8s v2.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
jtk54: Yep good call -- there was a conundrum with `isEqual` causing weird behavior since angular was add a `hashCode` field. I had to do some contortions to figure out what was going on and didn't revert it.

jtk54:
Feature flagged this with Dan's help, take a look.

anotherchrisberry:
LGTM

PR #4155, opened 2017-09-26
fix(react): Do not suppress unhandled rejections in promises.

christopherthielen:
When react components chain off promises (angular $q promises) for callbacks (`setState()`, etc), any thrown errors would be swallowed by $qProvider.errorOnUnhandledRejections(true)

This change will cause more noise in the console, but I think this is a necessary change.  When other kinds of rejections (which aren't actually errors) occur and spams the console, we should address these cases individually.  Generally, this means adding a catch block on some promise chain.


jrsquared:
😍 😍 😍 😍 😍 😍 😍

BrunoCarrier:
@christopherthielen you are my hero

PR #4156, opened 2017-09-26
Update Tooltip: Pipeline Config -> Property File

dalvizu:
Add docs from the 'pipeline expressions' guide about where
the contents of the 'Property File' are usable, as well
as a link back to the doc:

https://www.spinnaker.io/guides/user/pipeline-expressions/


christopherthielen:
lgtm

PR #4165, opened 2017-09-27
style(development only): Added linting for colors

archana-s:
Added postcss-colorfix which watches for color regressions and offers guidance based on our existing palette

postcss-colorfix relies on a color map to know what spinnaker's palette is and then uses nearest-color to recommend one from the palette. 

Does not affect any UI/product. This linting happens only during development.


package.json
```diff
@@ -142,8 +142,10 @@
     "loader-utils": "^1.1.0",
     "ngtemplate-loader": "^1.3.1",
     "node-libs-browser": "^2.0.0",
+    "postcss-colorfix": "^0.0.3",
     "postcss-extend": "^1.0.5",
     "postcss-import": "^10.0.0",
+    "postcss-less": "^1.1.0",
```
christopherthielen: do we need `postcss-less`? It looks like postcss is processing plain CSS from `less-loader`

christopherthielen:
lgtm

PR #4167, opened 2017-09-28
feat(provider/gce): Support for connection draining in LBs.

jtk54:
Affects TCP, SSL, and L7.


jtk54:
SSL/TCP: 
ssl

HTTP: 
http

jtk54:
https://github.com/spinnaker/spinnaker/issues/1964 FYI @edwinavalos

PR #4186, opened 2017-09-29
fix(stage): fix default timeout for deploy stage

emjburns:
Was incorrectly set to 1 hour.
PTAL @anotherchrisberry @jrsquared


emjburns:
@jrsquared thanks for noticing all my effort

emjburns:
@anotherchrisberry It's true, I can't spell. You'll see it over and over...

PR #4187, opened 2017-09-29
fix(pipeline_template): Cast numeric variable types during plan

robzienert:
Sending int/float variables as strings cause validation errors on the server side.


app/scripts/modules/core/src/pipeline/config/templates/configurePipelineTemplateModal.controller.ts
```diff
@@ -119,6 +119,10 @@ export class ConfigurePipelineTemplateModalController implements IController {
       .map(v => {
         if (v.type === 'object') {
           v.value = load(v.value);
+        } else if (v.type === 'int') {
+          return [v.name, parseInt(v.value)];
```
danielpeach: The base should be explicit here `parseInt(..., 10)`, since apparently the default is browser-specific.

robzienert: TIL thanks!

PR #4202, opened 2017-10-02
feat(pipeline): Change Artifact UI to use ExpectedArtifact model.

jtk54:


app/scripts/modules/core/src/pipeline/config/triggers/trigger.directive.js
```diff
@@ -40,6 +41,15 @@ module.exports = angular.module('spinnaker.core.pipeline.config.trigger.triggerD
       return artifact.type && artifact.name;
     };
 
+    this.summarizeExpectedArtifacts = function() {
```
danielpeach: Does `expected.summary` end up in the pipeline config? Instead of mutating the artifacts, you could just a define a function that formats the artifact:

```
<ui-select-choices ...> 
   <span>{{ triggerCtrl.myExpectedArtifactFormatter(expected) }}</span>
</ui-select-choices/>
```

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
```diff
@@ -1,25 +1,59 @@
...
     const artifactIndex = this.pipeline.expectedArtifacts.indexOf(this.artifact);
     this.pipeline.expectedArtifacts.splice(artifactIndex, 1);
 
     this.pipeline.triggers
-      .forEach(t => t.expectedArtifacts = t.expectedArtifacts.filter(a => !this.isEqual(a, this.artifact)));
+      .forEach(t => t.expectedArtifacts = t.expectedArtifacts.filter(a => !this.equals(a, this.artifact)));
+  }
+
+  public addField(): void {
+    const newField = {
```
danielpeach: You should be able to just type this as `const newField: IArtifactField` instead of using `as IArtifactField` below.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
```diff
@@ -1,25 +1,59 @@
...
+        && firstField.value === otherField.value
+        && firstField.missingPolicy === otherField.missingPolicy
+        && firstField.expression === otherField.expression;
+    };
+    if (first.fields.length !== other.fields.length) {
+      return false;
+    }
+
+    return first.fields.every(firstField => {
+      return find(other.fields, otherField => fieldIsEqual(firstField, otherField)) != null;
```
danielpeach: I would just use `other.fields.some(otherField => fieldIsEqual(firstField, otherField))`

app/scripts/modules/core/src/help/help.contents.ts
```diff
@@ -85,7 +85,17 @@ module(HELP_CONTENTS, [])
...
         <p>You can always edit the cluster configuration after you've created it.</p>`,
-
+    'pipeline.config.expectedArtifact.fieldName': `
+        <p>The Artifact field name.</p>`,
+    'pipeline.config.expectedArtifact.fieldType': `
+        <p>The Artifact field type. This must be one of 'MUST_MATCH' or 'FIND_IF_MISSING'.</p>
+        <p><strong>MUST_MATCH</strong> means the incoming Artifact's field value must match exactly with the expected value.</p>
+        <p><strong>FIND_IF_MISSING</strong> means the field is optional and a 'Missing Policy' will fire if the value is not present.</p>`,
+    'pipeline.config.expectedArtifact.missingPolicy': `
+        <p>The policy if the field type is 'FIND_IF_MISSING' and the field value is not present.</p>
```
danielpeach: This can be an incremental improvement for another PR, but it would be better if people didn't have to look at enums. Take a look at https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/stages/stageConstants.ts.

jtk54: I'll take this as improvements on artifact UI v2 -- I'll create a tracking bug for these.

PR #4208, opened 2017-10-03
fix(provider/gce) Update max disk number for local-ssd to 8 from 4

edwinavalos:
From: https://cloud.google.com/compute/docs/disks/local-ssd#create_local_ssd
"Each local SSD is 375 GB in size, but you can attach up to eight local SSD devices for 3 TB of total local SSD storage space per instance"

4 was the old limit

Fixes https://github.com/spinnaker/spinnaker/issues/1945


danielpeach:
Uh sorry. Premature lgtm: can you please update this help text line? https://github.com/spinnaker/deck/blob/master/app/scripts/modules/google/src/help/gce.help.ts#L25

PR #4223, opened 2017-10-04
feat(redblack): Expose `delayBeforeDisableSec`

ajordens:
This allows a user to specify the amount of time that Spinnaker should
wait after launching new instances and disabling old server groups or
instances.


ajordens:
image

image

Verified that `orca` does all the right things when passed a `delayBeforeDisableSec`.

app/scripts/modules/core/src/deploymentStrategy/strategies/rollingredblack/additionalFields.html
```diff
@@ -1,9 +1,21 @@
...
+<div class="form-group" style="margin-top: 15px">
   <div class="col-md-6">
     <label>
       Percentages
       <help-field key="strategy.rollingRedBlack.targetPercentages"></help-field>
     </label>
     <number-list model="$ctrl.command.targetPercentages" label="percentage"/>
   </div>
+  <div class="col-md-12 form-inline">
+    <label>Wait Before Disable
```
ajordens: This will possibly change to `Wait Before Termination` depending on the outcome of a couple conversations around RRB.

app/scripts/modules/core/src/deploymentStrategy/strategies/rollingredblack/additionalFields.html
```diff
@@ -1,9 +1,21 @@
-<div class="form-group">
+<div class="form-group" style="margin-top: 15px">
```
ajordens: We do this in some of the other strategies ... The label looked a little squishy otherwise. 

Not necessarily a fan of the inline style but 'twas there previously.

.gitignore
```diff
@@ -28,6 +28,7 @@ dist/
 lib/
 transpiled/
 app/scripts/modules/core/src/styleguide/src/public/
+coverage
```
anotherchrisberry: what's this guy?

app/scripts/modules/core/src/deploymentStrategy/strategies/rollingredblack/additionalFields.html
anotherchrisberry: We do this in a lot of places, nothing to worry about

app/scripts/modules/core/src/deploymentStrategy/strategies/redblack/additionalFields.html
```diff
@@ -6,9 +6,22 @@
...
+    <input class="form-control input-sm" style="width: 50px" type="number" ng-model="$ctrl.command.maxRemainingAsgs" min="2">
+  </div>
+  <div class="col-md-12 form-inline">
+    <label>Wait Before Disable
+      <help-field
+        content="Time to wait before disabling old server group"></help-field>
+    </label>
+    <input class="form-control input-sm"
+           style="width: 60px"
+           min="0"
```
anotherchrisberry: Not sure if the `min` will do anything without `type="number"` on the input.

app/scripts/modules/core/src/deploymentStrategy/strategies/rollingredblack/additionalFields.html
```diff
@@ -1,9 +1,21 @@
...
     <number-list model="$ctrl.command.targetPercentages" label="percentage"/>
   </div>
+  <div class="col-md-12 form-inline">
+    <label>Wait Before Disable
+      <help-field
+        content="Time to wait before disabling instances in old server group"></help-field>
+    </label>
+    <input class="form-control input-sm"
+           style="width: 60px"
+           min="0"
```
anotherchrisberry: `type="number"`

.gitignore
ajordens: Somehow I ended with code coverage, no idea where it came from?

app/scripts/modules/core/src/deploymentStrategy/strategies/redblack/additionalFields.html
ajordens: Yer probably right, threw `type="number"` back in there.

PR #4235, opened 2017-10-09
feat(provider/kubernetes): Enable annotations, labels and secrets for security groups (aka ingress resources) (spinnaker/spinnaker#2000)

mdirkse:
This enables the setting of annotations, labels and TLS secrets on k8s Security groups (aka ingress resources) from the UI.

It is one of 2 PR's that fix spinnaker/spinnaker#2000 (the other being spinnaker/clouddriver#2005)

### DIFF
--- a/app/scripts/modules/kubernetes/help/kubernetes.help.ts
+++ b/app/scripts/modules/kubernetes/help/kubernetes.help.ts
@@ -127,6 +127,8 @@ const helpContents: {[key: string]: string} = {
   'kubernetes.ingress.rules.host': 'The fully qualified domain name of a network host. Any traffic routed to this host matches this rule. May not be an IP address, or contain port information.',
   'kubernetes.ingress.rules.path': 'POSIX regex (IEE Std 1003.1) matched against the path of an incoming request.',
   'kubernetes.ingress.rules.port': 'The port on the specifed load balancer to route traffic to.',
+  'kubernetes.ingress.tls.host': 'The fully qualified domain name of a network host. Any traffic routed to this host can be secured with TLS. May not be an IP address, or contain port information.',
+  'kubernetes.ingress.tls.secret': '(Optional) Name of the Kubernetes secret that will be used to secure TLS connections to the security group. Note that Spinnaker will not create any secrets, they are assumed to exist.',
 };
 
 export const KUBERNETES_HELP = 'spinnaker.kubernetes.help.contents';
--- a/app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js
+++ b/app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js
@@ -34,6 +34,8 @@ module.exports = angular.module('spinnaker.securityGroup.kubernetes.create.contr
       basicSettings: require('./basicSettings.html'),
       backend: require('./backend.html'),
       rules: require('./rules.html'),
+      tls: require('./tls.html'),
+      advancedSettings: require('./advancedSettings.html'),
     };
 
     $scope.state = {
@@ -181,6 +183,14 @@ module.exports = angular.module('spinnaker.securityGroup.kubernetes.create.contr
             cloudProvider: 'kubernetes',
             region: $scope.securityGroup.namespace,
           };
+
+          // Change TLS hosts from string to array for Clouddriver (if it isn't already an array)
+          for (let idx in $scope.securityGroup.tls) {
+            if (!Array.isArray($scope.securityGroup.tls[idx].hosts)) {
+              $scope.securityGroup.tls[idx].hosts = [$scope.securityGroup.tls[idx].hosts];
+            }
+          }
+
           return securityGroupWriter.upsertSecurityGroup($scope.securityGroup, application, descriptor, params);
         }
       );
--- a/app/scripts/modules/kubernetes/securityGroup/details/details.html
+++ b/app/scripts/modules/kubernetes/securityGroup/details/details.html
@@ -75,11 +75,34 @@
         No rules created
       </dl>
     </collapsible-section>
+    <collapsible-section heading="TLS">
+      <dl ng-class="insightCtrl.vm.filtersExpanded ? '' : 'dl-horizontal dl-medium'"
+          ng-repeat="tls in securityGroup.ingress.spec.tls"
+          ng-if="securityGroup.ingress.spec.tls.length">
+        <dt>Host</dt>
+        <dd>{{tls.hosts[0]}}</dd>
+      </dl>
+      <dl ng-if="!securityGroup.ingress.spec.tls.length">
+        No TLS entries created
+      </dl>
+    </collapsible-section>
     <collapsible-section heading="Status">
       <dl ng-class="insightCtrl.vm.filtersExpanded ? '' : 'dl-horizontal dl-medium'"
           ng-repeat="ingress in securityGroup.ingress.status.loadBalancer.ingress">
         <dt>IP</dt>
         <dd>{{ingress.ip}}</dd>
+      </dl>
+    </collapsible-section>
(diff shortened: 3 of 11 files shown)


lwander:
LG2M2

ethanfrogers:
@mdirkse do you have a PR for Clouddriver as well?

mdirkse:
@ethanfrogers not yet, hence the WIP tag ;)
I'll get that done and remove the WIP tag when everything is set to go.

mdirkse:
Turns out I wasn't quite finished yet. I needed to add some logic to properly store the TLS entries in a format that Clouddriver gets, otherwise they weren't getting picked up (see https://github.com/bolcom/deck/blob/2e8cb9352ac2c3d99c35d19b5d8960d32856b749/app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js#L187)

mdirkse:
ok, everything works, ready to be merged

danielpeach:
Could you please add screenshots that show the changes for both the server group wizard and the details panel?

app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js
```diff
@@ -181,6 +182,16 @@ module.exports = angular.module('spinnaker.securityGroup.kubernetes.create.contr
             cloudProvider: 'kubernetes',
             region: $scope.securityGroup.namespace,
           };
+
+          // Write TLS property so clouddriver understands it and our certs get configured
+          $scope.securityGroup.tls = [];
+          for (var idx in $scope.securityGroup.rules) {
```
danielpeach: Please use `let` here instead of `var`. Variables defined with`var` are not block scoped - i.e., `idx` is defined for the remainder of the function after this loop.

app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js
```diff
@@ -181,6 +182,16 @@ module.exports = angular.module('spinnaker.securityGroup.kubernetes.create.contr
...
+            let rule = $scope.securityGroup.rules[idx];
```
danielpeach: Please use `const` here instead of `let`.

app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js
```diff
@@ -181,6 +182,16 @@ module.exports = angular.module('spinnaker.securityGroup.kubernetes.create.contr
...
+            if (typeof rule.secret !== 'undefined' && rule.secret !== '') {
```
danielpeach: Reading this makes me think that other JS false values might be acceptable here. If that's the case, I would make that explicit, e.g., 
```
if (rule.secret || rule.secret === null || rule.secret === 0 || rule.secret === NaN)  { ... }
```
otherwise, I would just stick with `if (rule.secret) { ... }`.

mdirkse:
Ok so my initial approach to the secret field was kinda dumb, as k8s allows you to specify a tls object with just a host field and no secret. I'll put the tls stuff in advanced as well.

mdirkse:
Screenshots:
screenshot from 2017-10-20 13-36-08
screenshot from 2017-10-20 13-38-03

mdirkse:
Alright, all done on my side.

app/scripts/modules/kubernetes/help/kubernetes.help.ts
```diff
@@ -127,6 +127,8 @@ const helpContents: {[key: string]: string} = {
   'kubernetes.ingress.rules.host': 'The fully qualified domain name of a network host. Any traffic routed to this host matches this rule. May not be an IP address, or contain port information.',
   'kubernetes.ingress.rules.path': 'POSIX regex (IEE Std 1003.1) matched against the path of an incoming request.',
   'kubernetes.ingress.rules.port': 'The port on the specifed load balancer to route traffic to.',
+  'kubernetes.ingress.tls.host': 'The fully qualified domain name of a network host. Any traffic routed to this host can be secured with TLS. May not be an IP address, or contain port information. ',
```
danielpeach: Extra space at the end of the string.

app/scripts/modules/kubernetes/securityGroup/details/details.html
```diff
@@ -75,12 +75,35 @@ <h3 select-on-dbl-click>
...
       </dl>
     </collapsible-section>
+    <collapsible-section heading="TLS">
+      <dl ng-class="insightCtrl.vm.filtersExpanded ? '' : 'dl-horizontal dl-medium'"
+          ng-repeat="tls in securityGroup.ingress.spec.tls"
+          ng-if="securityGroup.ingress.spec.tls && securityGroup.ingress.spec.tls.length > 0">
+        <dt>Host</dt>
+        <dd>{{tls.hosts[0]}}</dd>
+      </dl>
+      <dl ng-if="!securityGroup.ingress.spec.rules || securityGroup.ingress.spec.rules.length == 0">
```
danielpeach: s/==/===

danielpeach: Actually, you can just make this a single condition, since you won't hit any NPEs in a template:
`ng-if="!securityGroup.ingress.spec.rules.length"`

app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js
```diff
@@ -181,6 +183,14 @@ module.exports = angular.module('spinnaker.securityGroup.kubernetes.create.contr
             cloudProvider: 'kubernetes',
             region: $scope.securityGroup.namespace,
           };
+
+          // Turn it into an array so clouddriver can work with it (if it isn't already an array)
+          for (let idx in $scope.securityGroup.tls) {
+            if ($scope.securityGroup.tls[idx].hosts.constructor !== Array) {
```
danielpeach: Please use `Array.isArray` here.

app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js
```diff
@@ -181,6 +183,14 @@ module.exports = angular.module('spinnaker.securityGroup.kubernetes.create.contr
             cloudProvider: 'kubernetes',
             region: $scope.securityGroup.namespace,
           };
+
+          // Turn it into an array so clouddriver can work with it (if it isn't already an array)
```
danielpeach: Please make this comment a little more specific here (e.g., the inverse of your comment below - `Change TLS hosts from string to array for Clouddriver`).

mdirkse:
@danielpeach thanks, fixed!

app/scripts/modules/kubernetes/securityGroup/configure/wizard/upsert.controller.js
```diff
@@ -181,6 +183,14 @@ module.exports = angular.module('spinnaker.securityGroup.kubernetes.create.contr
             cloudProvider: 'kubernetes',
             region: $scope.securityGroup.namespace,
           };
+
+          // Change TLS hosts from string to array for Clouddriver (if it isn't already an array)
+          for (let idx in $scope.securityGroup.tls) {
+            if (!Array.isArray($scope.securityGroup.tls[idx].hosts.constructor)) {
```
danielpeach: I should have been more clear. I think this is always false -- `Array.isArray([]) === true`, but `Array.isArray([].constructor]) === false`.

Should be `!Array.isArray($scope.securityGroup.tls[idx].hosts)`.

mdirkse: ah sorry, should've caught that

PR #4256, opened 2017-10-12
fix(kubernetes): fix template loading

danielpeach:
@lwander


danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/2025

danielpeach:
This will run through the templates twice, but since there's no top-level kubernetes module (and the existing templates aren't nested under the `v1` module, that feels fine.

PR #4271, opened 2017-10-16
fix(artifact): s/ul/ol

danielpeach:
@lwander 
`ol` will count for you and you won't get a bullet point and a number.


lwander:
Oh awesome - thanks

PR #4284, opened 2017-10-18
feat(entitytags): Show replaced server group details in popover

ajordens:
As of spinnaker/orca#1705, `orca` is now adding previous server group
details to the `spinnaker:metadata` tag.

These details include:
- previous server group name
- image name
- image id

image

### DIFF
--- a/app/scripts/modules/core/src/entityTag/entitySource.popover.html
+++ b/app/scripts/modules/core/src/entityTag/entitySource.popover.html
@@ -1,16 +1,29 @@
 <dl>
   <dt>{{$ctrl.executionType}}</dt>
   <dd>{{$ctrl.metadata.value.description}}</dd>
-  <div ng-if="$ctrl.execution">
+  <div ng-if="$ctrl.execution" style="margin-top: 10px">
     <dt>Triggered by</dt>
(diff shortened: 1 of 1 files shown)


ajordens:
Might be nice to figure out a way to better squeeze this into the sidebar (without popover) but image name is soooo long and image id is perhaps not descriptive enough.

Debated storing another attribute in the tag for the build # (`#191`) in the above screenshot, but that is a little different for containers.

app/scripts/modules/core/src/entityTag/entitySource.popover.html
```diff
@@ -1,16 +1,29 @@
...
     <dd>{{$ctrl.metadata.value.user}}</dd>
   </div>
-  <div ng-if="$ctrl.metadata.value.comments">
+
+  <div ng-if="$ctrl.metadata.value.previousServerGroup" style="margin-top: 10px">
+    <dt>Replaced</dt>
+    <dd>
+      {{$ctrl.metadata.value.previousServerGroup.name}}<br/>
+      {{$ctrl.metadata.value.previousServerGroup.imageName}}
+      <span ng-if="$ctrl.metadata.value.previousServerGroup.imageId !== $ctrl.metadata.value.previousServerGroup.imageName">
```
ajordens: titus image id == image name so no sense doubling up

PR #4286, opened 2017-10-18
fix(provider/openstack): load balancer network and sg creation bugfix…

ewiseblatt:
…es (#4281)

fix(provider/openstack): Fixed network so it is filtered by account for load balancer creation.
fix(provider/openstack): Fixed network so it refreshes when a new account is selected.
fix(provider/openstack): Fixed network so it resets to default when a new account is selected and the selected network becomes invalid.
fix(provider/openstack): Fixed security groups so it clears and refreshes when a new account is selected.
fix(provider/openstack): Fixed security groups refresh button.
fix(provider/openstack): Removed security group refresh tooltip that wasn't showing any data.

@edwinavalos can you verify this in deck as well?

### DIFF
--- a/app/scripts/modules/openstack/common/cacheBackedMultiSelect.template.html
+++ b/app/scripts/modules/openstack/common/cacheBackedMultiSelect.template.html
@@ -10,7 +10,7 @@
     </ui-select>
   </div>
   <div class="col-md-1 sm-label-left">
-    <a href="" ng-click="forceRefreshCache()" uib-tooltip-template="refreshTooltipTemplate">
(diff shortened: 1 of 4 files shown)


ewiseblatt:
@danielpeach please review

danielpeach:
LGTM

johnrengelman:
@ewiseblatt We are testing out some of these changes and what I've noticed is that you are currently forced to select both an Interface network and Security Groups for the LB.
Both of these are optional in our OST installation. Specifically giving the LB a FIP causes issues in certain configurations.

Should the network drop down have an option for "none" and should the security group selector be optional?

PR #4287, opened 2017-10-18
feat(provider/kubernetes): Enable setting of labels on k8s Service objects #2035

mdirkse:
fixes spinnaker/spinnaker#2035

### DIFF
--- a/app/scripts/modules/kubernetes/loadBalancer/configure/wizard/advancedSettings.html
+++ b/app/scripts/modules/kubernetes/loadBalancer/configure/wizard/advancedSettings.html
@@ -76,4 +76,8 @@
     <map-editor model="loadBalancer.serviceAnnotations" label="Service Annotations" add-button-label="Add Annotations" labels-left="true">
     </map-editor>
   </div>
+  <div class="col-md-12">
+    <map-editor model="loadBalancer.serviceLabels" label="Service Labels" add-button-label="Add Label" labels-left="true">
+    </map-editor>
+  </div>
 </form>
--- a/app/scripts/modules/kubernetes/loadBalancer/details/details.html
+++ b/app/scripts/modules/kubernetes/loadBalancer/details/details.html
@@ -171,5 +171,11 @@
         </kubernetes-key-value-details>
       </dl>
     </collapsible-section>
+    <collapsible-section heading="Service Labels" ng-if="loadBalancer.description.serviceLabels">
+      <dl class="dl-horizontal dl-flex">
+        <kubernetes-key-value-details map="loadBalancer.service.metadata.labels">
+        </kubernetes-key-value-details>
+      </dl>
+    </collapsible-section>
   </div>
 </div>


app/scripts/modules/kubernetes/loadBalancer/configure/wizard/advancedSettings.html
```diff
@@ -76,4 +76,8 @@
     <map-editor model="loadBalancer.serviceAnnotations" label="Service Annotations" add-button-label="Add Annotations" labels-left="true">
     </map-editor>
   </div>
+  <div class="col-md-12">
+    <map-editor model="loadBalancer.serviceLabels" label="Service Labels" add-button-label="Add Label" labels-left="true">
```
danielpeach: Related to your corresponding Clouddriver changes - if I clear all the service labels here while editing a load balancer, what does Clouddriver do? (The way it's implemented now, I think you'll inherit the existing labels.)

mdirkse: That's correct. Hmmm, guess I should implement it so that the labels get cleared.

mdirkse: I'm not sure, I think it takes the empty set. In any case, the logic is the same as it currently is for annotations.

mdirkse: From Clouddriver:

```groovy
    def labels = description.serviceLabels ?: existingService?.metadata?.labels

    serviceBuilder = serviceBuilder.withLabels(labels)

    def annotations = description.serviceAnnotations ?: existingService?.metadata?.annotations

    serviceBuilder = serviceBuilder.withAnnotations(annotations)
```

danielpeach: Yeah, I think passing an empty object from Deck is the right behavior. But the Clouddriver implementation is wrong - should be ternary on `description.serviceLabels == null`.

If you could also fix this for annotations (in Clouddriver), that would be great.

mdirkse: You mean like this?

```groovy
 def annotations = description.serviceAnnotations == null ? existingService?.metadata?.annotations : description.serviceAnnotations
```

danielpeach: Yes, exactly. Thank you.

mdirkse: Done: https://github.com/spinnaker/clouddriver/pull/2014/commits/37a9dfbbc481b1460484a4bc3e7fcf616b4d0361#diff-6d44078ab20702f2e88f8d27785bfc33

app/scripts/modules/kubernetes/loadBalancer/details/details.html
```diff
@@ -171,5 +171,11 @@ <h3 select-on-dbl-click>
         </kubernetes-key-value-details>
       </dl>
     </collapsible-section>
+    <collapsible-section heading="Service Labels" ng-if="loadBalancer.description.serviceLabels">
```
danielpeach: Can I get in a situation where `loadBalancer.description.serviceLabels` is an empty object (in that case this section would be rendered but empty)?

mdirkse: Not sure tbh, but everything works the same as with the annotations so the behavior would be consistent ;)

mdirkse: Is there anything else you expect from me?

PR #4288, opened 2017-10-19
feat(pipeline_template): Better support for templated pipelines with dynamic sources

jervi:
* ~Add "Edit configuration as JSON" link for templated pipelines (this kind of reverts #4192)~
  skjermbilde 2017-10-19 kl 15 47 54
  Context: #4192 by @anotherchrisberry makes it impossible to change the source of a templated pipeline through Deck. If you select "Edit as JSON" you will see the rendered pipeline, not the configuration. I've added another choice in the drop down that has the previous behaviour of "Edit as JSON", so that you can actually change the configuration of a templated pipeline
  * Update: I have instead added a new tab inside the "Edit as JSON" modal. See screenshot below.
* Don't display error message when editing templated pipelines with dynamic source - real fix in Orca
* Render execution graph when editing templated pipelines with dynamic source - real fix in Orca

Orca PR: https://github.com/spinnaker/orca/pull/1718
Gate PR: https://github.com/spinnaker/gate/pull/471


robzienert:
+1 to @danielpeach.

jervi:
I've discussed this briefly with @danielpeach, and I will try to address his comments by doing the following:
- Implement an error/info message if no executions have been run
- Make it obvious which execution is used to render the configuration
- Look at implementing a drop down where you can select which execution to use
- Edit as JSON: Create a tab in the JSON modal that switches between the config and the rendered pipeline

jervi:
Ok, it took a bit longer than anticipated, but I think the result became pretty good. I have addressed all the stuff discussed above. Some updated screenshots:

No executions run:
No executions run

Only one execution run:
Only one execution run

Multiple executions run (selects the latest by default):
Multiple executions run (selects the latest by default)

Drop down for selecting execution:
Drop down for selecting execution

New tab in "Edit as JSON" modal:
New tab in 'Edit as JSON' modal
New tab in 'Edit as JSON' modal

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -365,6 +368,19 @@ module.exports = angular.module('spinnaker.core.pipeline.config.pipelineConfigur
       return msg;
     };
 
+    this.getPipelineExecutions = () => {
+      executionService.getExecutionsForConfigIds($scope.pipeline.application, $scope.pipeline.id, 5)
```
jervi: This will fetch the five newest executions. Is that a sensible number, or should I use something else?

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
```diff
@@ -35,7 +35,27 @@
     <div class="band band-info" ng-if="isTemplatedPipeline && !pipeline.locked">
       <span class="glyphicon glyphicon small glyphicon-lock"></span> Manual edits are not allowed on templated pipelines
     </div>
-
+    <div class="band band-info" ng-if="hasDynamicSource && pipelineExecutions && pipelineExecutions.length">
+      <span class="glyphicon glyphicon small glyphicon-wrench"></span> This template has a dynamic source. The
+      configuration is currently rendered using
+      <span ng-if="pipelineExecutions.length > 1" class="dropdown" uib-dropdown>
+        &nbsp;&nbsp;<button type="button" class="btn btn-xs btn-default dropdown-toggle" uib-dropdown-toggle>
+          {{ renderablePipeline.trigger.parentPipelineName || "Build #" + renderablePipeline.trigger.buildNumber }} <span class="caret"></span>
```
danielpeach: I understand that it might be useful to use `parentPipelineName` or `buildNumber` to differentiate between executions, but there are executions that will not have either of these properties. I think the default has to be the timestamp of the execution.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -365,6 +368,19 @@ module.exports = angular.module('spinnaker.core.pipeline.config.pipelineConfigur
...
+        .then(executions => {
+          $scope.pipelineExecutions = executions;
+        })
+        .catch(() => $scope.pipelineExecutions = []);
+      return $scope.pipelineExecutions;
+    };
+
+    $scope.changePipelineExecution = executionId => {
+      $location.search('executionId', executionId);
```
danielpeach: This should either be a `$state.go` or if possible you should be able to include a `ui-sref` in your template that updates the query parameter.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
```diff
@@ -35,7 +35,27 @@
...
+        </button>
+        <ul class="dropdown-menu" role="menu" uib-dropdown-menu>
+          <li ng-repeat="p in pipelineExecutions" ng-class="{disabled: p.id == renderablePipeline.executionId}">
+            <a title="{{p.startTime | date:'medium'}}" ng-click="(p.id == renderablePipeline.executionId) ? false : changePipelineExecution(p.id); true;">{{ p.trigger.parentPipelineName || "Build #" + p.trigger.buildNumber }}
+              <span class="text-muted small">({{p.startTime | relativeTime}})</span>
+              <span class="glyphicon small glyphicon-ok" ng-if="p.id == renderablePipeline.executionId"></span></a>
```
danielpeach: Please move the closing anchor tag down a line to match the opening anchor tag.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
```diff
@@ -35,7 +35,27 @@
...
+    <div class="band band-info" ng-if="hasDynamicSource && pipelineExecutions && pipelineExecutions.length">
+      <span class="glyphicon glyphicon small glyphicon-wrench"></span> This template has a dynamic source. The
+      configuration is currently rendered using
+      <span ng-if="pipelineExecutions.length > 1" class="dropdown" uib-dropdown>
+        &nbsp;&nbsp;<button type="button" class="btn btn-xs btn-default dropdown-toggle" uib-dropdown-toggle>
+          {{ renderablePipeline.trigger.parentPipelineName || "Build #" + renderablePipeline.trigger.buildNumber }} <span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu" role="menu" uib-dropdown-menu>
+          <li ng-repeat="p in pipelineExecutions" ng-class="{disabled: p.id == renderablePipeline.executionId}">
+            <a title="{{p.startTime | date:'medium'}}" ng-click="(p.id == renderablePipeline.executionId) ? false : changePipelineExecution(p.id); true;">{{ p.trigger.parentPipelineName || "Build #" + p.trigger.buildNumber }}
```
danielpeach: Please move the click handler into a real function on the controller.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -173,7 +174,8 @@ module.exports = angular.module('spinnaker.core.pipeline.config.pipelineConfigur
         controllerAs: '$ctrl',
         size: 'lg modal-fullscreen',
         resolve: {
-          pipeline: () => $scope.renderablePipeline,
+          pipeline: () => $scope.pipeline,
```
danielpeach: This seems to revert the changes in https://github.com/spinnaker/deck/pull/4192, but I don't see how this PR addresses the bug that PR fixed. Can you verify that the behavior won't be reintroduced by this change?

jervi: But did that PR really fix any bugs? It just changed the behavior of the edit JSON feature, and actually made it much harder to use templated pipelines. If you’d like to change the template after the change in #4192, you can’t do it in Deck anymore.
With this change, you can edit the configuration, and you can see (read only) the tendered pipeline. Best of both worlds!
But I’d like @chrisberry’s opinion on this, it was his PR.

app/scripts/modules/core/src/styleguide/public/styleguide.html
```diff
@@ -1,5 +1,5 @@
...
...,.sp-badge.info *{background-color:var(--color-accent-g2);color:var(--color-primary);font-weight:700}.pill.category>:first-child,.sp-badge.category>:first-child{border-radius:var(--pill-border-radius) 0 0 var(--pill-border-radius);background-color:var(--color-primary);color:var(--color-text-on-dark);text-transform:uppercase;font-weight:400;position:relative}.pill.success,.sp-badge.success{background-color:var(--color-success)}.pill.danger,.sp-badge.danger{background-color:var(--color-danger)}.pill.warn,.sp-badge.warn{background-color:var(--color-warning)}.pill.alert,.sp-badge.alert{background-color:var(--color-alert)}.sp-badge.large,.sp-badge.large *{font-size:var(--pill-large-font-size)}.sp-badge.large i{margin-top:4px}i.button{color:var(--color-accent);cursor:pointer}i.button:hover{color:var(--color-primary)}i.disabled{color:var(--color-accessory)}:root{--is-phone:false;--is-tablet:false;--is-desktop:false}@media only screen and (min-device-width:320px) and (max-device-width:480px){:root{--is-phone:true}}@media only screen and (min-device-width:768px) and (max-device-width:1024px){:root{--is-tablet:true}}@media screen and (min-device-width:992px){:root{--is-desktop:true}}</style>
```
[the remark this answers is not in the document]
jervi: This file seems to be auto generated. The change is the css change in `tabs.css`. It should probably be added to `.gitignore`.

jervi:
Thank you for feedback, @danielpeach! I've addressed most of them.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
jervi: I'm using this quite a few places. And I guess it could be useful in other places as well. If I want to create a component, even if it end up just being used here, should I go the React route and create a React component and include it in angular using `react2angular`?

danielpeach: Yes, that sounds good.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
danielpeach: Yes - before that PR, changes in the pipeline JSON weren't reflected in the UI until you saved and refreshed.

I think you need `pipeline: () => $scope.isTemplatedPipeline ? $scope.pipeline : $scope.renderablePipeline`.

jervi: But `$scope.renderablePipeline = $scope.plan || $scope.pipeline;`, so for non templated pipelines, this should not matter (because `$scope.plan` will be undefined). And for templated pipelines, it is just wrong. I don't want the renderable pipeline, I want the configuration, which is in `$scope.pipeline`.
I have tested, and the UI is updated when you do config changes, so I would argue that the change I introduce here is the correct behaviour.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
jervi: Done 👍

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
```diff
@@ -35,7 +35,32 @@
...
+    <div class="band band-info" ng-if="hasDynamicSource && pipelineExecutions.length">
+      <span class="glyphicon glyphicon small glyphicon-wrench"></span> This template has a dynamic source. The
+      configuration is currently rendered using
+      <span ng-if="pipelineExecutions.length > 1" class="dropdown" uib-dropdown>
+        &nbsp;&nbsp;<button type="button" class="btn btn-xs btn-default dropdown-toggle" uib-dropdown-toggle>
+          <execution-build-title execution="currentExecution" default-to-timestamp="true" ></execution-build-title> <span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu" role="menu" uib-dropdown-menu>
+          <li ng-repeat="p in pipelineExecutions" ng-class="{disabled: p.id == currentExecution.id}">
+            <a title="{{p.startTime | date:'medium'}}" ui-sref="{ executionId: p.id }"><execution-build-title execution="p" default-to-timestamp="true" ></execution-build-title>
```
anotherchrisberry: replace `date:'medium'` with `timestamp`

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -24,13 +26,16 @@ module.exports = angular.module('spinnaker.core.pipeline.config.pipelineConfigur
...
         plan: '<',
         isTemplatedPipeline: '<',
+        hasDynamicSource: '<',
+        templateError: '<',
       },
       controller: 'PipelineConfigurerCtrl as pipelineConfigurerCtrl',
       templateUrl: require('./pipelineConfigurer.html'),
     };
   })
+  .component('executionBuildTitle', react2angular(ExecutionBuildTitle, ['execution', 'defaultToTimestamp']))
```
anotherchrisberry: Let's keep this in a separate file. It feels hidden in here, tucked between a directive declaration and the directive's controller declaration.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
anotherchrisberry: I've also tested the UI and it seems to work with this change, so I'm okay with it, even if I don't fully understand it (we are not using templates for the day-to-day management of Spinnaker internally).

app/scripts/modules/core/src/delivery/executionBuild/ExecutionBuildTitle.tsx
```diff
@@ -0,0 +1,58 @@
...
+    return (
+      <span>
+        { this.props.execution.trigger.parentPipelineName &&
+          <span>{this.props.execution.trigger.parentPipelineName}</span>
+        }
+        { this.props.execution.buildInfo && this.props.execution.buildInfo.number &&
+          <span><span className="build-label">Build</span> #{this.props.execution.buildInfo.number}</span>
+        }
+        { this.props.defaultToTimestamp && !this.props.execution.trigger.parentPipelineName && !this.props.execution.buildInfo &&
+          <span>{new Date(this.props.execution.startTime).toLocaleString()}</span>
```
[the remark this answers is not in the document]
jervi: Fixed

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
jervi: Fixed

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
jervi: Fixed

jervi:
Thanks for reviewing, guys! I have updated the PR and rebased.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
```diff
@@ -35,7 +35,32 @@
...
+          </li>
+        </ul>
+      </span>
+      <span ng-if="pipelineExecutions.length === 1"><execution-build-title execution="currentExecution" default-to-timestamp="true" ></execution-build-title></span>
+    </div>
+    <div class="band band-warning" ng-if="pipelineExecutions && !pipelineExecutions.length">
+      <span class="glyphicon glyphicon small glyphicon-alert"></span> This template has a dynamic source. The
+      configuration cannot be rendered before the pipeline has executed at least once.
+    </div>
+    <div class="band band-warning" ng-if="templateError && pipelineExecutions.length && !pipeline.isNew">
```
danielpeach: The conditions here aren't quite complete, because you can hit a template planning error even if you're not using dynamic sources - and now if you're not using dynamic sources, those errors are going to be silent.

In the past, template planning errors were surfaced in the configuration modal (and if you hit a planning error when loading the pipeline config view, the template configuration modal opened automatically). I'd like to stick to that model.

app/scripts/modules/core/src/delivery/executionBuild/ExecutionBuildTitle.tsx
```diff
@@ -0,0 +1,67 @@
+/*
```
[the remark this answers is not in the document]
jervi: Fixed in new PR

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
jervi: Nothing has really changed if you're not configuring a pipeline with dynamic source. This will always be false if you're not using dynamic sources; and the modal will open like it did before before.
The modal is not of much help when an error occurs while using dynamic sources, because it looks like this:
Configuration modal

PR #4291, opened 2017-10-19
feat(provider/amazon): Rollback support for PREVIOUS_IMAGE strategy

ajordens:
Rolling back to a previous image is available iff:
- a `spinnaker:metadata` entity tag exists with `previousServerGroup`
details
- there are no other server groups in the cluster/region

image

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
+++ b/app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
@@ -2,6 +2,7 @@
 
 const angular = require('angular');
 
+import { get } from 'lodash';
 import { SERVER_GROUP_WRITER, TASK_MONITOR_BUILDER } from '@spinnaker/core';
 
 module.exports = angular.module('spinnaker.amazon.serverGroup.details.rollback.controller', [
@@ -18,6 +19,28 @@ module.exports = angular.module('spinnaker.amazon.serverGroup.details.rollback.c
 
       var desired = serverGroup.capacity.desired;
 
+      var rollbackType = 'EXPLICIT';
+
+      if (allServerGroups.length === 0 && serverGroup.entityTags) {
+        const previousServerGroup = get(serverGroup, 'entityTags.creationMetadata.value.previousServerGroup');
+        if (previousServerGroup) {
+          rollbackType = 'PREVIOUS_IMAGE';
+          $scope.previousServerGroup = {
+            name: previousServerGroup.name,
+            imageName: previousServerGroup.imageName
+          };
+
+          if (previousServerGroup.imageId && previousServerGroup.imageId !== previousServerGroup.imageName) {
+            $scope.previousServerGroup.imageId = previousServerGroup.imageId;
+          }
+
+          const buildNumber = get(previousServerGroup, 'buildInfo.jenkins.number');
+          if (buildNumber) {
+            $scope.previousServerGroup.buildNumber = buildNumber;
+          }
+        }
+      }
+
       if (desired < 10) {
         var healthyPercent = 100;
       } else if (desired < 20) {
@@ -28,7 +51,7 @@ module.exports = angular.module('spinnaker.amazon.serverGroup.details.rollback.c
       }
 
       $scope.command = {
-        rollbackType: 'EXPLICIT',
+        rollbackType: rollbackType,
         rollbackContext: {
           rollbackServerGroupName: serverGroup.name,
           targetHealthyRollbackPercentage: healthyPercent
@@ -51,6 +74,11 @@ module.exports = angular.module('spinnaker.amazon.serverGroup.details.rollback.c
         var command = $scope.command;
         if (!$scope.verification.verified) {
           return false;
+        }
+
+        if (rollbackType === 'PREVIOUS_IMAGE') {
+          // no need to validate when using an explicit image
+          return true;
         }
 
         return command.rollbackContext.restoreServerGroupName !== undefined;
--- a/app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
+++ b/app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
@@ -10,13 +10,25 @@
         <div class="col-sm-3 sm-label-right">
           Restore to
         </div>
-        <div class="col-sm-6">
-          <ui-select ng-model="command.rollbackContext.restoreServerGroupName" class="form-control input-sm">
+        <div class="col-sm-8">
+          <ui-select ng-model="command.rollbackContext.restoreServerGroupName"
+                     class="form-control input-sm"
+                     ng-if="command.rollbackType === 'EXPLICIT'">
             <ui-select-match placeholder="Select...">{{ctrl.label($select.selected)}}</ui-select-match>
             <ui-select-choices group-by="ctrl.group" repeat="serverGroup.name as serverGroup in allServerGroups">
               <span ng-bind-html="ctrl.label(serverGroup)"></span>
             </ui-select-choices>
           </ui-select>
+          <div ng-if="command.rollbackType === 'PREVIOUS_IMAGE'" style="margin-top: 5px">
+            {{ previousServerGroup.name }} <span class="small">(no longer deployed)</span><br/>
+            <span class="small">
+              <strong>Image</strong>: {{ previousServerGroup.imageName}}
+              <span ng-if="previousServerGroup.imageId">({{ previousServerGroup.imageId }})</span><br/>
+            </span>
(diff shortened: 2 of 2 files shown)


app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
```diff
@@ -18,6 +18,31 @@ module.exports = angular.module('spinnaker.amazon.serverGroup.details.rollback.c
 
       var desired = serverGroup.capacity.desired;
 
+      var rollbackType = 'EXPLICIT';
+
+      if (allServerGroups.length === 0 && serverGroup.entityTags) {
```
anotherchrisberry: I'm not familiar with all the use cases, but do we really want to restrict this capability to clusters where there are no other server groups? What if someone wants to roll back a couple of releases?

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
```diff
@@ -18,6 +18,31 @@ module.exports = angular.module('spinnaker.amazon.serverGroup.details.rollback.c
...
+        var creationMetadata = serverGroup.entityTags.creationMetadata;
+
+        if (creationMetadata.value.previousServerGroup) {
```
anotherchrisberry: worth adding a guard here to verify `creationMetadata` exists

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
```diff
@@ -18,6 +18,31 @@ module.exports = angular.module('spinnaker.amazon.serverGroup.details.rollback.c
...
+          };
+
+          if (previousServerGroup.imageId && previousServerGroup.imageId !== previousServerGroup.imageName) {
+            $scope.previousServerGroup.imageId = previousServerGroup.imageId;
+          }
+
+          var buildInfo = previousServerGroup.buildInfo;
+          if (buildInfo && buildInfo.jenkins && buildInfo.jenkins.number) {
+            $scope.previousServerGroup.buildNumber = previousServerGroup.buildInfo.jenkins.number;
+          }
```
anotherchrisberry: might as well bring lodash in here (just the `get` method):

```javascript
import { get } from 'lodash';
...
const buildNumber = get(previousServerGroup, 'buildInfo.jenkins.number');
if (buildNumber) {
  $scope.previousServerGroup.buildNumber = buildNumber;
}
```

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
```diff
@@ -10,13 +10,26 @@
...
+                     ng-if="command.rollbackType === 'EXPLICIT'">
             <ui-select-match placeholder="Select...">{{ctrl.label($select.selected)}}</ui-select-match>
             <ui-select-choices group-by="ctrl.group" repeat="serverGroup.name as serverGroup in allServerGroups">
               <span ng-bind-html="ctrl.label(serverGroup)"></span>
             </ui-select-choices>
           </ui-select>
+          <div ng-if="command.rollbackType === 'PREVIOUS_IMAGE'" style="margin-top: 5px">
+            {{ previousServerGroup.name }}<br/>
+            <span class="small" ng-if="previousServerGroup.buildNumber">
+              <strong>Build</strong> #{{ previousServerGroup.buildNumber }} (no longer deployed)<br/>
```
anotherchrisberry: Should probably put `(no longer deployed)` next to the previous server group name. It's potentially misleading, as this build might still be deployed in a different cluster or region?

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
```diff
@@ -76,6 +89,29 @@
...
+            </li>
+            </li>
+            <li>Disable {{ serverGroup.name }}</li>
+            <li>Restore minimum capacity of <em>new server group</em> [
+              <strong>min</strong>: {{ serverGroup.capacity.min }}
+              ]</li>
+          </ol>
+          <p>
+            This rollback will affect server groups in {{ serverGroup.account }} ({{ serverGroup.region }}).
+          </p>
```
anotherchrisberry: Why is this sentence only being displayed if we're rolling back to the previous image and not explicitly selecting an image?

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
ajordens: That's fair.

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
```diff
@@ -76,6 +89,29 @@
...
+              <strong>max</strong>: {{ serverGroup.capacity.max }},
+              <strong>desired</strong>: {{ serverGroup.capacity.desired }}
+              ]<br/>(minimum capacity pinned at {{serverGroup.capacity.desired}} to prevent autoscaling down during rollback)
+            <li ng-if="command.rollbackContext.targetHealthyRollbackPercentage < 100">
+              Wait for at least {{minHealthy(command.rollbackContext.targetHealthyRollbackPercentage)}}
+              instances to report as healthy
+            </li>
+            </li>
+            <li>Disable {{ serverGroup.name }}</li>
+            <li>Restore minimum capacity of <em>new server group</em> [
```
anotherchrisberry: Is `Set` more appropriate than `Restore` here, since we're creating a new server group and the minimum capacity was never set the the `min` before?

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
ajordens: The dominant use case (at least what edge/GregB thinks is 99%) is you've destroyed the previous server group.

I was looking to avoid asking the user to make a choice on what or how they wanted to rollback. I'll reach out to Ludo and GregB once it's live and get some feedback.

At least now we've got the metadata on what the previous versions were, so if someone did want to rollback a few separate server groups in the same region/cluster, they would at least be able to fairly easily determine what the previous versions were.

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
ajordens: Gotcha. Incoming.

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
ajordens: We're initially pinning the min capacity to something other than what it should be ... `Restore` still feels like reasonable wording if you ask me. 

FWIW it jives with the task name in orca as well `restoreMinCapacityFromSnapshot` which shows up when you do a normal red/black.

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
ajordens: The steps are a little different depending on the strategy so not _too_ much duplication. 

It would seem more confusing to me to try and `ng-if` each descriptive step?

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
anotherchrisberry: 🌮

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
anotherchrisberry: Sorry, not suggesting interleaving it with `ng-if`s. Duplication makes sense here - keeps it readable.

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
anotherchrisberry: Sounds good.

app/scripts/modules/amazon/src/serverGroup/details/rollback/rollbackServerGroup.html
ajordens: Good ... I thought you were talking some crazy right there.

ajordens:
Revised Screenshot

image

PR #4308, opened 2017-10-23
feat(rrb): Allow for specifying pipelines to run before disable

ajordens:
image


anotherchrisberry:
I feel like we're basically at a custom strategy here. Your PR screenshot isn't showing parameters, but it looks like that's another set of fields that we'll surface.

ajordens:
The UI was borrowed from the `Custom` strategy, pipeline parameters are displayed.

image

I'm not completely sold on the UI (or perhaps it's implementation) given how similar but different it is from the `Custom` strategy.

Primarily want something at this point for early adopters to poke at.

ajordens:
I have goals of making a more generic pipeline/strategy selection component while delegating how we store the data to a controller.

PR #4310, opened 2017-10-23
bugfix(aws): don't show copy capacity options for clone dialog either

tomaslin:


app/scripts/modules/amazon/src/serverGroup/configure/wizard/capacity/capacitySelector.component.html
```diff
@@ -43,7 +44,7 @@
       </label>
     </div>
   </div>
-  <div ng-if="!$ctrl.command.useSourceCapacity">
+  <div ng-if="!$ctrl.command.useSourceCapacity || $ctrl.command.viewState.mode !== 'editPipeline'">
```
tomaslin: without this the clone dialog just shows up as empty so yeah

PR #4327, opened 2017-10-25
fix(core/pipeline): Fix corner case where masterStage was not set

jervi:
In a few cases, where a group only contains one groupStage, the masterStage can be undefined. This will lead to a Deck crash when rendering the executions page. Fixed by copying the masterStage from the groupStage.
@jrsquared PTAL


jervi:
Duplicated by proper fix in #4332

PR #4328, opened 2017-10-25
feat(trigger/pubsub): Suggest subscriptions from echo configuration.

jtk54:


app/scripts/modules/core/src/pubsub/pubsubSubscription.service.ts
```diff
@@ -0,0 +1,15 @@
+import { module } from 'angular';
+
+import { API_SERVICE, Api } from 'core/api/api.service';
+
+export class PubsubSubscriptionService {
+  constructor(private API: Api) { 'ngInject'; }
+
+  public getPubsubSubscriptions(): ng.IPromise<string[]> {
```
danielpeach: You should drop the `ng` namespacing by importing `IPromise` from `angular`.

app/scripts/modules/core/src/pipeline/config/triggers/pubsub/pubsub.trigger.ts
```diff
@@ -1,19 +1,29 @@
...
 
 class PubsubTriggerController implements IController {
   public pubsubSystems = ['kafka', 'google'];
+  public pubsubSubscriptions: string[];
 
-  constructor(public trigger: IPubsubTrigger) {
+  constructor(public trigger: IPubsubTrigger,
+              public pubsubSubscriptionService: PubsubSubscriptionService) {
     'ngInject';
+    pubsubSubscriptionService.getPubsubSubscriptions().then(subscriptions => {
```
danielpeach: Ideally the full set of states of this call would be surfaced to the user. We generally don't end up surfacing failure, but you should show a spinner that indicates that the subscriptions are still loading. See https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/triggers/jenkins/jenkinsTrigger.module.js#L89-L98

For a failed call, you can add `.catch(() => this.pubsubSubscriptions = []);`

app/scripts/modules/core/src/pipeline/config/triggers/pubsub/pubsub.trigger.ts
```diff
@@ -1,19 +1,29 @@
...
+  PubsubSubscriptionService,
+} from '@spinnaker/core';
 
 class PubsubTriggerController implements IController {
   public pubsubSystems = ['kafka', 'google'];
+  public pubsubSubscriptions: string[];
 
-  constructor(public trigger: IPubsubTrigger) {
+  constructor(public trigger: IPubsubTrigger,
+              public pubsubSubscriptionService: PubsubSubscriptionService) {
```
danielpeach: Should be `private`.

app/scripts/modules/core/src/pipeline/config/triggers/pubsub/pubsub.trigger.ts
```diff
@@ -1,19 +1,29 @@
 import { IController, module } from 'angular';
 
 import { PIPELINE_CONFIG_PROVIDER, PipelineConfigProvider } from 'core/pipeline/config/pipelineConfigProvider';
-import { IPubsubTrigger } from '@spinnaker/core';
+import {
+  IPubsubTrigger,
+  PUBSUB_SUBSCRIPTION_SERVICE,
+  PubsubSubscriptionService,
+} from '@spinnaker/core';
```
danielpeach: Don't use `@spinnaker/core` from inside `core`. Use `core/pubsub` instead.

jtk54: Done.

app/scripts/modules/core/src/pipeline/config/triggers/pubsub/pubsub.trigger.ts
jtk54: `private` throws a linter error, I just removed `public`.

app/scripts/modules/core/src/pipeline/config/triggers/pubsub/pubsub.trigger.ts
jtk54: Done.

app/scripts/modules/core/src/pubsub/pubsubSubscription.service.ts
jtk54: Done.

app/scripts/modules/core/src/pipeline/config/triggers/pubsub/pubsub.trigger.ts
```diff
@@ -1,19 +1,35 @@
...
-  constructor(public trigger: IPubsubTrigger) {
+  constructor(public trigger: IPubsubTrigger,
+              pubsubSubscriptionService: PubsubSubscriptionService) {
     'ngInject';
+
+    this.subscriptionsLoaded = false;
+    pubsubSubscriptionService.getPubsubSubscriptions().then(subscriptions => {
+      this.subscriptionsLoaded = true;
+      this.pubsubSubscriptions = subscriptions;
+    }).catch(() => this.pubsubSubscriptions = []);
```
danielpeach: Should flip `subscriptionsLoaded` even in the case where the call failed (the spinner should stop spinning and you should see an empty dropdown).

You can use `.finally` here:

```
pubsubSubscriptionService.getPubsubSubscriptions()
  .then(subscriptions => this.pubsubSubscriptions = subscriptions)
  .catch(() => this.subscriptions = [])
  .finally(() => this.subscriptionsLoaded = true);
```

PR #4340, opened 2017-10-30
feat(gce): UHC support for l7 load balancers

danielpeach:
@jtk54 please review.


app/scripts/modules/google/src/loadBalancer/configure/http/commandBuilder.service.js
```diff
@@ -136,7 +136,7 @@ module.exports = angular.module('spinnaker.deck.gce.httpLoadBalancer.backing.ser
     }
 
     function getHealthChecks () {
-      return gceHealthCheckReader.listHealthChecks('HTTP');
+      return gceHealthCheckReader.listHealthChecks();
```
jtk54: Would we still need the `listHealthChecks` that takes a string argument?

danielpeach: Good question - looks like autohealing policies require (or maybe required) HTTP health checks. I'll check if they still do.

danielpeach:
Depends on https://github.com/spinnaker/clouddriver/pull/2030

PR #4337, opened 2017-10-27
refactor(kubernetes): kubernetes is now a module

kevinawoo:
This converts kubernetes so that it can be built as a module. I followed the conventions laid out in `docker`, `google`, and `amazon` modules.

Here's an included script to help test it out locally.
https://gist.github.com/kevinawoo/1fb8db8d54918ad05dc2b328481fd77c

to run it it: 
```
curl -O https://gist.githubusercontent.com/kevinawoo/1fb8db8d54918ad05dc2b328481fd77c/raw/26edbadec7dceda1aecddb375111c12cd3084dc2/test-kubernetes-module.sh
chmod +x test-kubernetes-module.sh
./test-kubernetes-module.sh
```

PR #4345, opened 2017-10-31
refactor(openstack): openstack is now a module

kevinawoo:
This converts openstack so that it can be built and imported as a module. I followed the conventions laid out in [`kubernetes`](https://github.com/spinnaker/deck/pull/4337) `docker`, `google`, and `amazon` modules.


app/scripts/modules/openstack/src/pipeline/stages/cloneServerGroup/cloneServerGroupExecutionDetails.controller.js
```diff
@@ -0,0 +1,63 @@
+'use strict';
```
jrsquared: This has since been refactored out. Remove this file.

app/scripts/modules/openstack/src/pipeline/stages/cloneServerGroup/cloneServerGroupExecutionDetails.html
```diff
@@ -0,0 +1,41 @@
+<div ng-controller="openstackCloneServerGroupExecutionDetailsCtrl">
```
jrsquared: remove this file.

jrsquared:
A couple files that need to be removed, but otherwise LGTM

edwinavalos:
@johnrengelman FYI

PR #4377, opened 2017-11-06
feat(moniker): Use server group moniker in multi-select

andrewbackes:
When using multi-select on the cluster view page, monikers are not carried through with the server group information. This PR will fix it. I'm not exactly sure why only certain fields are mapped through from the server group model. It seems like IServerGroup should be used here, but I think that is a separate issue.


andrewbackes:
@lwander FYI

PR #4379, opened 2017-11-06
chore(halconfig): Allow canary to be enabled

rebalag:
Add canary feature flag to enable halyard to manage canary feature for deployment

Update halconfig/settings.js with feature from settings.js for canary configuration.

### DIFF
--- a/halconfig/settings.js
+++ b/halconfig/settings.js
@@ -10,6 +10,7 @@ var fiatEnabled = '{%features.fiat%}' === 'true';
 var jobsEnabled = '{%features.jobs%}' === 'true';
 var pipelineTemplatesEnabled = '{%features.pipelineTemplates%}' === 'true';
 var artifactsEnabled = '{%features.artifacts%}' === 'true';
+var canaryEnabled = '{%features.canary%}' === 'true';
 var timezone = '{%timezone%}';
 var version = '{%version%}';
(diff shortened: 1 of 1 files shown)


halconfig/settings.js
```diff
@@ -10,6 +10,7 @@ var fiatEnabled = '{%features.fiat%}' === 'true';
 var jobsEnabled = '{%features.jobs%}' === 'true';
 var pipelineTemplatesEnabled = '{%features.pipelineTemplates%}' === 'true';
 var artifactsEnabled = '{%features.artifacts%}' === 'true';
+var canaryEnabled = '{%features.canary%}' === 'true';
```
danielpeach: Not sure what you're planning on naming the corresponding Halyard flags, but I'd like to refer to that system as `legacyCanary` or something. In the near future there will be a fully open-source canary system for Spinnaker, and I'd like to avoid calling that one `canaryV2`.

rebalag: Will update feature name exposed by hal as legacyCanary. This is to get halconfig in sync with feature present under deck/settings.js. Feature name under feature section will remain "canary" as that is the name under which the feature is implemented in deck. Is that acceptable?

danielpeach: Yes, that's totally fine. Thanks.

PR #4381, opened 2017-11-07
feat(core): Link to failing synthetic stage rather than "No reason provided."

ajordens:
image

### DIFF
--- a/app/scripts/modules/amazon/src/pipeline/stages/resizeAsg/resizeAsgStepLabel.html
+++ b/app/scripts/modules/amazon/src/pipeline/stages/resizeAsg/resizeAsgStepLabel.html
@@ -1,5 +1,8 @@
-<span class="task-label">
+<span class="task-label" ng-if="!step.context.useNameAsLabel">
   Resize Server Group:
   {{step.context.serverGroupName}}
   ({{step.context.region}})
+</span>
+<span class="task-label" ng-if="step.context.useNameAsLabel">
+  {{step.name}}
 </span>
--- a/app/scripts/modules/core/src/delivery/details/StageFailureMessage.tsx
+++ b/app/scripts/modules/core/src/delivery/details/StageFailureMessage.tsx
@@ -3,6 +3,10 @@ import * as React from 'react';
 import { IExecutionStage, ITaskStep } from 'core/domain';
 
 import { robotToHuman } from 'core/presentation/robotToHumanFilter/robotToHuman.filter';
+import { UISref } from '@uirouter/react';
+import { UIRouterContext } from '@uirouter/react-hybrid';
+
+import { ReactInjector } from 'core';
 
 export interface IStageFailureMessageProps {
   message?: string;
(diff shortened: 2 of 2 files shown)


app/scripts/modules/amazon/src/pipeline/stages/resizeAsg/resizeAsgStepLabel.html
```diff
@@ -1,5 +1,8 @@
-<span class="task-label">
+<span class="task-label" ng-if="!step.context.useNameAsLabel">
   Resize Server Group:
   {{step.context.serverGroupName}}
   ({{step.context.region}})
 </span>
+<span class="task-label" ng-if="step.context.useNameAsLabel">
```
ajordens: I don't really like having to send an attribute back from `orca` indicating how I want a UI label to render, but I also don't really like that it disregards the specified stage name for this particular stage.

image

In RRB, we'll get `Grow to X% of Desired Size` rather than `Resize Server Group: ...`.

Thoughts?

anotherchrisberry: I don't love doing it, either, but we do similar things with Manual Judgment stages and some other places I can't think of right now, so it's fine in my opinion.

app/scripts/modules/core/src/delivery/details/StageFailureMessage.tsx
```diff
@@ -28,7 +35,15 @@ export class StageFailureMessage extends React.Component<IStageFailureMessagePro
   private getState(props: IStageFailureMessageProps): IStageFailureMessageState {
     const { stage } = props;
     if (stage && (stage.isFailed || stage.isStopped)) {
-      return { isFailed: true, failedTask: (stage.tasks || []).find(t => t.status === 'TERMINAL' || t.status === 'STOPPED') };
+      let failedTask = (stage.tasks || []).find(t => t.status === 'TERMINAL' || t.status === 'STOPPED');
```
anotherchrisberry: make this a `const` (it doesn't look like you're reassigning a value anywhere)

app/scripts/modules/core/src/delivery/details/StageFailureMessage.tsx
```diff
@@ -28,7 +35,15 @@ export class StageFailureMessage extends React.Component<IStageFailureMessagePro
...
+      let failedStage = stage;
+      let failedStageIndex = -1;
+      if (!failedTask && stage.after) {
+        failedStage = stage.after.find(s => s.status === 'TERMINAL' || s.status === 'STOPPED');
+        failedStageIndex = stage.after.indexOf(failedStage);
+      }
+
+      return { isFailed: true, failedTask: failedTask, failedStage: failedStage, failedStageIndex: failedStageIndex };
```
anotherchrisberry: you can simplify this a bit, since the variables and the fields have the same name:
```javascript
return { isFailed: true, failedTask, failedStage, failedStageIndex };
```

jrsquared:
Very nice

app/scripts/modules/amazon/src/pipeline/stages/resizeAsg/resizeAsgStepLabel.html
ajordens: If such a thing comes up in the future (or we ever wish to retrofit this), I believe we should be able to accomplish something similar in `orca` using expressions to grab the server group name once it appears in the context.

Ideally such business doesn't not belong in `deck`.

PR #4385, opened 2017-11-07
feat(provider/kubernetes): runJob context images

ethanfrogers:
add support for using images found from context in the runJob stage. i
refactored the `Resolve from Trigger` checkbox, and included Context and
Trigger images in the same dropdown to bring it more in line with the
Server Group configuration.

spinnaker/spinnaker#1961
spinnaker/spinnaker#1960

@danielpeach can I get some early feedback on this? I had to copy a bunch of code over from the Server Group module. I feel like there may be a better way to handle that, I just don't know what it is.

### DIFF
--- a/app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.html
+++ b/app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.html
@@ -18,6 +18,20 @@
     </namespace-select-field>
   </stage-config-field>
 
+  <stage-config-field label="Image">
+    <ui-select name="containerSelect" ng-model="ctrl.stage.container" class="form-control input-sm" required>
+        <ui-select-match>{{ ctrl.stage.container.imageDescription | kubernetesImageId }}</ui-select-match>
+        <ui-select-choices
+          repeat="container in ctrl.containers | orderBy: 'imageDescription.imageId'"
+          group-by="ctrl.groupByRegistry"
+          refresh-delay="0"
+          ui-disable-choice="container.message"
+          refresh="ctrl.searchImages($select.search)">
+          <span ng-bind-html="container.imageDescription | kubernetesImageId | highlight: $select.search"></span>
+        </ui-select-choices>
+      </ui-select>
+  </stage-config-field>
+
   <stage-config-field label="Container Name">
     <div>
       <input type="text" class="form-control input-sm" ng-model="ctrl.stage.container.name">
@@ -38,45 +52,7 @@
         </select>
     </div>
   </stage-config-field>
-
-  <stage-config-field label="Image">
-    <div class="form-group" ng-if="ctrl.hasDockerPipelineTriggers()">
-      <div>
-        <label class="sm-label">
-          Resolve from Trigger
-        </label>
-        <input type="checkbox"
-               style="margin-left: 20px"
-               ng-model="ctrl.stage.container.imageDescription.fromTrigger"/>
-      </div>
-    </div>
-    <div ng-if="!ctrl.stage.container.imageDescription.fromTrigger">
-      <docker-image-and-tag-selector
-        specify-tag-by-regex="false"
-        account="ctrl.stage.container.imageDescription.account"
-        organization="ctrl.stage.container.imageDescription.organization"
-        registry="ctrl.stage.container.imageDescription.registry"
-        repository="ctrl.stage.container.imageDescription.repository"
-        tag="ctrl.stage.container.imageDescription.tag"
-        show-registry="true"
-        on-change="ctrl.onChange"></docker-image-and-tag-selector>
-    </div>
-    <div ng-if="ctrl.stage.container.imageDescription.fromTrigger" class="form-group">
-        <div class="sm-label-right col-md-3">
-        Trigger
-        </div>
-        <div class="col-md-9">
-          <ui-select ng-model="ctrl.container"
-                    class="form-control input-sm"
-                    ng-change="ctrl.updateContainerImage()" required>
-            <ui-select-match>{{ctrl.container}}</ui-select-match>
-            <ui-select-choices repeat="image in ctrl.triggerImages() | filter: $select.search | orderBy">
-              <span ng-bind-html="image | highlight: $select.search"></span>
-            </ui-select-choices>
-          </ui-select>
-        </div>
-    </div>
-  </stage-config-field>
+  
   <stage-config-field label="Commands" help-key="kubernetes.containers.command">
     <kubernetes-container-commands commands="ctrl.stage.container.command">
     </kubernetes-container-commands>
--- a/app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
+++ b/app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
@@ -1,6 +1,8 @@
 'use strict';
 
 import _ from 'lodash';
+import { PIPELINE_CONFIG_SERVICE } from '@spinnaker/core';
+import { KUBERNETES_IMAGE_ID_FILTER } from 'kubernetes/presentation/imageId.filter';
 
 const angular = require('angular');
 
@@ -9,7 +11,10 @@ module.exports = angular.module('spinnaker.kubernetes.pipeline.stage.runJobStage
   require('kubernetes/container/arguments.component.js').name,
   require('kubernetes/container/environmentVariables.component.js').name,
   require('kubernetes/container/volumes.component.js').name,
-  require('./runJobExecutionDetails.controller.js').name
+  require('kubernetes/image/image.reader.js').name,
+  require('./runJobExecutionDetails.controller.js').name,
+  PIPELINE_CONFIG_SERVICE,
+  KUBERNETES_IMAGE_ID_FILTER,
 ])
   .config(function(pipelineConfigProvider) {
     pipelineConfigProvider.registerStage({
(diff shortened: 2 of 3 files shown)


app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
```diff
@@ -76,47 +101,96 @@ module.exports = angular.module('spinnaker.kubernetes.pipeline.stage.runJobStage
         this.accounts = accounts;
       });
 
-    this.stage.cloudProvider = 'kubernetes';
-    this.stage.application = $scope.application.name;
 
-    if (!this.stage.credentials && $scope.application.defaultCredentials.kubernetes) {
-      this.stage.credentials = $scope.application.defaultCredentials.kubernetes;
+    function buildImageId(image) {
```
danielpeach: This is just a label for the UI, right? I wouldn't want to maintain this logic in two places.

It feels like you could have an Angular filter that takes an image and returns the string, and then use that filter in the server group configuration dialogue and here.

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
```diff
@@ -62,10 +45,52 @@ module.exports = angular.module('spinnaker.kubernetes.pipeline.stage.runJobStage
...
-      this.container = buildImageDescriptor(this.stage.container.imageDescription);
+    if (!this.stage.credentials && $scope.application.defaultCredentials.kubernetes) {
+      this.stage.credentials = $scope.application.defaultCredentials.kubernetes;
     }
 
+    if(this.stage.container) {
+      this.stage.container.imageDescription.imageId = buildImageId(this.stage.container.imageDescription);
+    }
+
+    function findContextImages(current, all, visited = {}) {
```
danielpeach: I don't know if this logic should totally be shared between the run job stage and the server group dialogue..but it does feel as if you should be able to write some function that makes this a little prettier in both places. Like

```
upstreamStagesForEach(this.stage, stages, current => {
  if (current.type === 'findImage') { ... }
  if (current.type === 'bake') { ... }
})
```

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
ethanfrogers: Can you point me toward an example of something like this?

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
ethanfrogers: Ah, so a shared function that handles visiting each pre-req stage and implementing the result creation as a callback?

ethanfrogers: Also, were would I need to put something like that? Is there a file where we can put shared function like this?

ethanfrogers:
@danielpeach I changed the `findContextImages` stuff to use `pipelineConfigService.getAllUpstreamDependencies`.

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
danielpeach: Yes - I was thinking you could define a custom filter named `kubernetesImageId` (or something) and include it inside your `ui-select`:

`<span ng-bind-html="container.imageDescription | kubernetesImageId | highlight: $select.search"></span>`

You can define a filter like this: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/robotToHumanFilter/robotToHuman.filter.ts

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.html
```diff
@@ -18,6 +18,20 @@
     </namespace-select-field>
   </stage-config-field>
 
+  <stage-config-field label="Image">
+    <ui-select name="containerSelect" ng-model="ctrl.stage.container" class="form-control input-sm" required>
+        <ui-select-match>{{ ctrl.stage.container.imageDescription | kubernetesImageId }}</ui-select-match>
+        <ui-select-choices
+          repeat="container in ctrl.containers | orderBy:'imageDescription.imageId'"
```
danielpeach: Hmm really didn't see this `orderBy` when I made that `imageId` suggestion. What order do they end up in? The ones that come from triggers/pipeline context should be sorted to the top.

app/scripts/modules/kubernetes/src/presentation/imageId.filter.ts
```diff
@@ -0,0 +1,37 @@
+import { module } from 'angular';
+
+export interface IDockerImage {
+    fromFindImage: boolean;
```
danielpeach: Please indent with 2 spaces in this file.

app/scripts/modules/kubernetes/src/presentation/imageId.filter.ts
```diff
@@ -0,0 +1,37 @@
...
+    registry: string;
+    cluster: string;
+    pattern: string;
+    fromTrigger: boolean;
+}
+
+export function imageId(image: IDockerImage): string {
+    if (image.fromFindImage) {
+        return `${image.cluster} ${image.pattern}`;
+        } else if (image.fromBake) {
```
danielpeach: Align these with the `if`

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.html
danielpeach: Maybe that's completely determined by the `group-by` function. In that case, omit this line.

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
```diff
@@ -62,10 +49,36 @@ module.exports = angular.module('spinnaker.kubernetes.pipeline.stage.runJobStage
...
+      }
+
+      return {
+        fromContext: true,
+        fromBake: true,
+        repository: stage.ami_name,
+        organization: stage.organization,
+        stageId: stage.refId
+      };
+    });
```
danielpeach: I think you need to filter out `undefined` from this list (in the case where you had an upstream stage whose type wasn't `findImage` or `bake`). Just tag a `.filter(image => !!image)` to the end.

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
```diff
@@ -62,10 +49,36 @@ module.exports = angular.module('spinnaker.kubernetes.pipeline.stage.runJobStage
       this.stage.dnsPolicy = 'ClusterFirst';
     }
 
-    if (this.stage.container.imageDescription.fromTrigger === true) {
```
danielpeach: What makes this unnecessary now?

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
```diff
@@ -76,47 +89,79 @@ module.exports = angular.module('spinnaker.kubernetes.pipeline.stage.runJobStage
...
+            requests: {
+              memory: null,
+              cpu: null,
+            },
+            limits: {
+              memory: null,
+              cpu: null,
+            },
+            ports: [
+            {
```
danielpeach: Weird formatting here (probably taken from something I did). I would move the opening and closing brackets and braces together:
```
ports: [{
  ...
}],
```

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
ethanfrogers: `fromTrigger` is now set when the image is selected. It's populated as part of the `searchImages` function.

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
ethanfrogers: 👍

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.html
ethanfrogers: Yeah, they end up in a weird order. There's no sorting logic in the `group-by` function either so I need to figure out another solution.

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/runJobStage.js
```diff
@@ -76,47 +89,87 @@ module.exports = angular.module('spinnaker.kubernetes.pipeline.stage.runJobStage
...
+              hostIp: null,
+            }],
+            livenessProbe: null,
+            readinessProbe: null,
+            envVars: [],
+            command: [],
+            args: [],
+            volumeMounts: [],
+          };
+        }).sort((a,b) => {
```
ethanfrogers: @danielpeach we can discuss this, but this is what I can up with to get around the sorting without the imageId property on each container.

app/scripts/modules/kubernetes/src/presentation/imageId.filter.ts
```diff
@@ -0,0 +1,37 @@
...
+export function imageId(image: IDockerImage): string {
+  if (image.fromFindImage) {
+    return `${image.cluster} ${image.pattern}`;
+  } else if (image.fromBake) {
+    return `${image.repository} (Baked during execution)`;
+  } else if (image.fromTrigger && !image.tag) {
+    return `${image.registry}/${image.repository} (Tag resolved at runtime)`;
+  } else {
+    if (image.registry) {
+        return `${image.registry}/${image.repository}:${image.tag}`;
```
danielpeach: Back two spaces here and below.

ethanfrogers:
Before:
image

image


After:
image

PR #4427, opened 2017-11-13
fix(provider/kubernetes): fix destroy multiple

ethanfrogers:
when destroying multiple servergroups, the namespace key wasn't being
populated properly. `serverGroup.region` is the correct key to use for
namespace.

ping @danielpeach or @lwander 

Fixes spinnaker/spinnaker#2027

### DIFF
--- a/app/scripts/modules/kubernetes/src/serverGroup/paramsMixin.js
+++ b/app/scripts/modules/kubernetes/src/serverGroup/paramsMixin.js
@@ -8,21 +8,21 @@ module.exports = angular
 
     function destroyServerGroup(serverGroup) {
       return {
-        namespace: serverGroup.namespace,
+        namespace: serverGroup.region || serverGroup.namespace,
         interestingHealthProviderNames: ['KubernetesService']
       };
     }
 
     function enableServerGroup(serverGroup) {
       return {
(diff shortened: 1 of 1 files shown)


app/scripts/modules/kubernetes/src/serverGroup/paramsMixin.js
```diff
@@ -8,7 +8,7 @@ module.exports = angular
 
     function destroyServerGroup(serverGroup) {
       return {
-        namespace: serverGroup.namespace,
+        namespace: serverGroup.region,
```
danielpeach: This definitely worked at some point under some conditions. Could you please make this 
```
namespace: serverGroup.region || serverGroup.namespace,
```
instead?

ethanfrogers: Yeah, I started to do that at first, but the other 2 are just `serverGroup.region` so I went with this. Should I go ahead and change all 3?

danielpeach: Hmm. Yes? I guess that feels best. Thanks.

ethanfrogers:
@danielpeach comments addressed.

PR #4449, opened 2017-11-15
fix(provider/openstack): Update securityGroups after account change

noahgorka:
Fixes bug in server group creation where selecting an account with multiple OpenStack accounts did not load Security Groups associated with that specific account.


edwinavalos:
@johnrengelman FYI

johnrengelman:
LGTM

PR #4465, opened 2017-11-16
fix(provider/kubernetes): fix lint issues

ethanfrogers:
fix lint issues from the last few round of runjob prs

ping @danielpeach


danielpeach:
Thanks!

ethanfrogers:
there's like one more but I need to make sure removing it won't break anything.

PR #4492, opened 2017-11-20
feat(provider/kubernetes): v1 pod logs

ethanfrogers:
add access to pod logs in v1 provider

Clouddriver PR: https://github.com/spinnaker/clouddriver/pull/2162

@lwander @danielpeach PTAL.


lwander:
@skim1420 fyi

PR #4493, opened 2017-11-20
fix(provider/gce): Break out spinnaker-member-of into individual labels.

duftler:


duftler:
Relates to: https://github.com/spinnaker/clouddriver/pull/2163

app/scripts/modules/google/src/serverGroup/configure/serverGroupCommandBuilder.service.js
```diff
@@ -247,8 +247,12 @@ module.exports = angular.module('spinnaker.gce.serverGroupCommandBuilder.service
       if (instanceTemplateLabels) {
         Object.assign(command.labels, instanceTemplateLabels);
 
-        if (command.labels['spinnaker-member-of']) {
-          delete command.labels['spinnaker-member-of'];
+        if (command.labels['spinnaker-region']) {
```
ewiseblatt: I'm assuming this is ok based on the change and what was before, but I'm not familiar with this in absolute terms.

duftler: This logic just ensures that a clone server group dialog is not pre-populated with labels that a) may not make sense, and b) will get overwritten by clouddriver anyway.

PR #4511, opened 2017-11-22
feat(core): console output jump to end

ethanfrogers:
constrain modal height and add shortcut to end of the output by jumping
`pre` content to end.

@anotherchrisberry PTAL. I pulled the `800px` from the changes modal so it's not _totally_ arbitrary.

### DIFF
--- a/app/scripts/modules/core/src/instance/details/console/consoleOutput.modal.controller.js
+++ b/app/scripts/modules/core/src/instance/details/console/consoleOutput.modal.controller.js
@@ -26,5 +26,10 @@ module.exports = angular.module('spinnaker.core.instance.details.console.control
     );
 
     $scope.close = $uibModalInstance.dismiss;
+    
+    $scope.jumpToEnd = () => {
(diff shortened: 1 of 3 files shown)


ethanfrogers:
image

anotherchrisberry:
I really like the idea here; let's use some existing styles to handle the modal size:
* We can add `modal-fullscreen` to the `size` in the `consoleOutputLink.directive.js` file. The modal will then adapt its size to the height of the window
* in the `consoleOutput.modal.html`, add `class="flex-fill"` to the top `div` to allow the whole thing to grow/shrink
* change the link to a button and put it in the footer, e.g. `<button class="btn btn-link" ng-click="jumpToEnd()">Jump to end</button>`

It's probably just as much (or more?) code changed, but should be a little more responsive.

ethanfrogers:
@anotherchrisberry comments addressed. that looks much better, thanks for the feedback. i updated the screenshot as well.

PR #4523, opened 2017-11-27
feat(kubernetes): export v2 provider

danielpeach:


danielpeach:
If you feel like it, but the only way anyone will really remember is if they need something in the NPM package. I wouldn't worry about it.

PR #4529, opened 2017-11-28
chore(core): Fix lint

jrsquared:


jrsquared:
It _should_ be setup to show errors in the console... webpack caught it for me. So... not sure? Haha

PR #4545, opened 2017-11-29
feat(amazon): Support passing a capacity constraint during resize

ajordens:
`clouddriver` can enforce that capacity of the server group being
resized has not changed in the background.

This addresses situations where autoscaling has occurred in the
background and subsequently been "undone" by a concurrent resize operation.

image

image


ajordens:
The checkboxes on this modal look pretty janky.

PR #4546, opened 2017-11-30
config(provider/openstack) - change default protocol for load balance…

johnrengelman:
…r health check to HTTP.


edwinavalos:
LGTM will merge when the test passes.

edwinavalos:
I think this is where it's failing, https://github.com/spinnaker/deck/blob/d0cc46fa3a21b374aed53c80ea593ca6b4f0a640/app/scripts/modules/openstack/src/loadBalancer/configure/wizard/upsert.controller.spec.js#L33

johnrengelman:
thanks @edwinavalos - that shows me for just going for it rather than running the tests. :(
Pushed a fix on the PR.

PR #4548, opened 2017-11-30
feat(gce): UI for zone selection.

jtk54:
Depends on https://github.com/spinnaker/clouddriver/pull/2201.

### DIFF
--- a/app/scripts/modules/google/src/serverGroup/configure/wizard/zones/regionalSelector.directive.html
+++ b/app/scripts/modules/google/src/serverGroup/configure/wizard/zones/regionalSelector.directive.html
@@ -4,5 +4,8 @@
     <label><input type="checkbox" ng-model="vm.command.regional"/>
       Distribute instances across multiple zones
     </label>
+    <label ng-if="vm.command.regional"><input type="checkbox" ng-model="vm.command.selectZones"/>
+      Select zones explicitly
+    </label>
   </div>
 </div>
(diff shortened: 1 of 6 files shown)


jtk54:
image

jtk54:
image

jtk54:
image

danielpeach:
Instead of the `Select explicit zones` checkbox, couldn't you just always have that `ui-select` available, but prepopulate it?

Is there a limit to the number of zones you can select?

jtk54:
We want to preserve the 'just use the default' option, which is why there are two views with a checkbox.

The limit is just the number of zones we get from Clouddriver for that region.

app/scripts/modules/google/src/serverGroup/configure/wizard/zones/zoneSelector.directive.html
```diff
@@ -22,4 +22,14 @@
       </li>
     </ul>
   </div>
+  <div class="col-md-7" ng-if="vm.command.selectZones">
+    <ui-select multiple
```
danielpeach: Should have a minimum of 1 zone selected I think - from the Clouddriver PR, if you select 0 zones, I think you'll end up with the default 3.

app/scripts/modules/google/src/serverGroup/configure/wizard/zones/regionalSelector.directive.html
```diff
@@ -4,5 +4,8 @@
     <label><input type="checkbox" ng-model="vm.command.regional"/>
       Distribute instances across multiple zones
     </label>
+    <label ng-if="vm.command.regional"><input type="checkbox" ng-model="vm.command.selectZones"/>
+      Select explicit zones
```
danielpeach: `Select zones explicitly`, maybe? `explicit zones` sounds a little different...

app/scripts/modules/google/src/serverGroup/configure/wizard/zones/zoneSelector.directive.html
jtk54: Agreed, added validation on this.

PR #4550, opened 2017-11-30
fix(provider/k8sv2): Move delete manifest form into component

andrewbackes:
Moving this form into it's own component will allow it to be re-used in the 'Delete Manifest' pipeline stage.

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/delete/delete.controller.ts
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/delete/delete.controller.ts
@@ -10,8 +10,9 @@ import {
   TaskMonitorBuilder
 } from '@spinnaker/core';
 import { IManifestCoordinates } from '../IManifestCoordinates';
+import { KUBERNETES_DELETE_MANIFEST_OPTIONS_FORM } from './deleteOptionsForm.component';
 
-interface IDeleteCommand {
+export interface IDeleteCommand {
   name: string;
   location: string;
   account: string;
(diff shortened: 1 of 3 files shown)


andrewbackes:
@lwander fyi

app/scripts/modules/kubernetes/src/v2/manifest/delete/deleteEntry.component.ts
```diff
@@ -0,0 +1,49 @@
+import { IComponentOptions, IController, module } from 'angular';
+
+import { IDeleteCommand } from './delete.controller';
+
+class KubernetesDeleteManifestCtrl implements IController {
+  public command: IDeleteCommand;
+}
+
+class KubernetesDeletManifestEntryComponent implements IComponentOptions {
```
lwander: What's the intended name here? I don't know where Entry comes from, and I thought we were trying to avoid the use of Manifest?

andrewbackes: I was mimicking the naming you used here: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/kubernetes/src/v2/manifest/wizard/manifestEntry.component.ts

Trying to keep it consistent.

andrewbackes:
Let me rename some things for clarity.

andrewbackes:
@lwander I renamed the components. Do you think their intentions are described better?

andrewbackes:
@lwander I renamed `Object` to `Entity`. What do you think?

andrewbackes:
@lwander okay, back to calling it `Manifest` =P

PR #4567, opened 2017-12-08
feat(provider/k8sv2): Delete/Resize manifest stage

andrewbackes:
Adds delete and resize manifest stages:
Resize Stage

Delete Stage

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/selector/labelEditor/labelEditor.component.ts
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/selector/labelEditor/labelEditor.component.ts
@@ -1,32 +1,30 @@
 import { IComponentOptions, IController, module } from 'angular';
 
-import { ILabelSelector, LabelKinds } from '../IManifestLabelSelector';
+import { IManifestLabelSelector, LABEL_KINDS } from '../IManifestLabelSelector';
 
 class KubernetesManifestLabelEditorCtrl implements IController {
+  public selectors: IManifestLabelSelector[];
+  public valueMetadata: string[] = []; // the index of 'selectors' and 'valueMetadata' align.
+  public kinds: string[] = LABEL_KINDS;
 
-  public selectors: ILabelSelector[];
-  public valueMetadata: string[]; // the index of 'selectors' and 'valueMetadata' align.
-  public kinds: string[] = LabelKinds;
-
-  constructor() {
-    this.valueMetadata = [];
+  public $onInit() {
     this.selectors.forEach(e => {
       this.valueMetadata.push(e.values.join(', '));
     });
   }
 
-  public addField() {
-    this.selectors.push({ key: '', kind: '', values: [] });
+  public addField(): void {
+    this.selectors.push({ key: '', kind: 'EQUALS', values: [] });
     this.valueMetadata.push('');
   }
 
-  public removeField(index: number) {
+  public removeField(index: number): void {
     this.selectors.splice(index, 1);
     this.valueMetadata.splice(index, 1);
   }
 
-  public convertValueStringToArray(index: number) {
-    this.selectors[index].values = this.valueMetadata[index].split(',').map( e => e.trim() );
+  public convertValueStringToArray(index: number): void {
+    this.selectors[index].values = this.valueMetadata[index].split(',').map(e => e.trim());
   }
 }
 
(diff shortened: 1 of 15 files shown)


andrewbackes:
@danielpeach - Would you mind looking this over for me?
@lwander - fyi

lwander:
Looks good! Can you add a screenshot of what the label selector looks like?

Also - the use of name & labels is mutually exclusive, and the UI should probably reflect that.

app/scripts/modules/kubernetes/src/v2/manifest/selector/labelEditor/labelEditor.component.ts
```diff
@@ -0,0 +1,90 @@
...
+  public kinds: string[] = LabelKinds;
+
+  constructor() {
+    this.valueMetadata = [];
+    this.selectors.forEach(e => {
+      this.valueMetadata.push(e.values.join(', '));
+    });
+  }
+
+  public addField() {
```
danielpeach: Please add return types to these methods (`void` for these three).

app/scripts/modules/kubernetes/src/v2/manifest/selector/labelEditor/labelEditor.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.valueMetadata.push('');
+  }
+
+  public removeField(index: number) {
+    this.selectors.splice(index, 1);
+    this.valueMetadata.splice(index, 1);
+  }
+
+  public convertValueStringToArray(index: number) {
+    this.selectors[index].values = this.valueMetadata[index].split(',').map( e => e.trim() );
```
danielpeach: Extraneous space around the arrow function passed to `.map`.

app/scripts/modules/kubernetes/src/v2/manifest/selector/labelEditor/labelEditor.component.ts
```diff
@@ -0,0 +1,90 @@
...
+          <th>Kind</th>
+          <th>Values</th>
+          <th></th>
+        </tr>
+        </thead>
+        <tbody>
+        <tr ng-repeat="selector in ctrl.selectors">
+          <td>
+            <input class="form-control input input-sm" type="text"
+                    name="{{$index}}"
```
danielpeach: Please align these attributes with `class`.

app/scripts/modules/kubernetes/src/v2/manifest/selector/selector.component.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  constructor(private $q: IQService, private accountService: AccountService) {
+    'ngInject';
+    const dataToFetch = {
+      accounts: this.accountService.getAllAccountDetailsForProvider('kubernetes', 'v2'),
+    };
+    this.$q.all(dataToFetch)
+      .then((fetchedData: any) => {
+        this.accounts = fetchedData.accounts;
+    });
```
danielpeach: Please indent

app/scripts/modules/kubernetes/src/v2/manifest/selector/selector.component.ts
```diff
@@ -0,0 +1,66 @@
...
+    this.kindsMetadata = this.selector.kinds.join(', ');
+  }
+
+  public stringToArray() {
```
danielpeach: Please add return type.

app/scripts/modules/kubernetes/src/v2/manifest/selector/selector.component.ts
danielpeach: These stages usually pick an account for you (first try to use the default account in `settings.js` if it's included in the list of accounts, then just the first account in the list.

app/scripts/modules/kubernetes/src/v2/manifest/selector/selector.component.ts
```diff
@@ -0,0 +1,66 @@
...
+import { IManifestSelector } from './IManifestSelector';
+
+class KubernetesManifestSelectorCtrl implements IController {
+  public selector: IManifestSelector;
+  public accounts: IAccountDetails;
+  public kindsMetadata: string;
+
+  constructor(private $q: IQService, private accountService: AccountService) {
+    'ngInject';
+    const dataToFetch = {
```
danielpeach: Would only this pattern if you're planning on making other async calls in the future in this controller. If not, this can just be
```
this.accountService.getAllAccountDetailsForProvider('kubernetes', v2').then(accounts => {
  this.accounts = accounts;
});
```

andrewbackes:
Added a toggle between 'Name' and 'Lables':
Name
Labels

Example of what the Label Editor looks like:
Label Editor

andrewbackes:
@danielpeach I think that I have addressed all of the improvements you suggested. Would you mind looking it over once more to double check for me?

@lwander - FYI

PR #4570, opened 2017-12-08
fix(moniker): stop pre-constructing frigga-style cluster names

asher:


app/scripts/modules/core/src/naming/naming.service.ts
```diff
@@ -103,8 +103,7 @@ export class NamingService {
   }
 
   public getMoniker(app: string, stack: string, detail: string): IMoniker {
-    const cluster = this.getClusterName(app, stack, detail);
-    return { app, stack, detail, cluster };
+    return { app, stack, detail };
```
asher: With this, a frigga-style cluster name will still be displayed in the UI:

screen shot 2017-12-08 at 3 03 41 pm

But the resulting `moniker` in the pipeline.json will be:
```
      "moniker": {
        "app": "spintest",
        "detail": "test",
        "stack": "asher"
      },
```

PR #4571, opened 2017-12-09
feat(core/pipeline): Expose "Override stage timeout" UI for Pipeline Stage

christopherthielen:
The UI for overriding stage timeouts was not enabled for the Pipeline Stage.  Some pipelines may take longer than 12 hrs to run, so users should be allowed to choose the timeout.

screen shot 2017-12-08 at 6 52 40 pm

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/pipeline/pipelineStage.js
+++ b/app/scripts/modules/core/src/pipeline/config/stages/pipeline/pipelineStage.js
@@ -18,6 +18,7 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.pipelineStage', [
       controllerAs: 'pipelineStageCtrl',
       templateUrl: require('./pipelineStage.html'),
       executionDetailsUrl: require('./pipelineExecutionDetails.html'),
+      defaultTimeoutMs: 12 * 60 * 60 * 1000, // 12 hours
       validators: [
         { type: 'requiredField', fieldName: 'pipeline', },
       ],


christopherthielen:
lgtm

app/scripts/modules/core/src/pipeline/config/stages/pipeline/pipelineStage.js
```diff
@@ -18,6 +18,7 @@ module.exports = angular.module('spinnaker.core.pipeline.stage.pipelineStage', [
       controllerAs: 'pipelineStageCtrl',
       templateUrl: require('./pipelineStage.html'),
       executionDetailsUrl: require('./pipelineExecutionDetails.html'),
+      defaultTimeoutMs: 12 * 60 * 60 * 1000, // 12 hours
```
jrsquared: I know we have to put this here... but I hate that we have to put this here.

jrsquared:
lgtm

app/scripts/modules/core/src/pipeline/config/stages/pipeline/pipelineStage.js
ajordens: I hate so too @jrsquared ... 

Once upon a time we had talked about sourcing some of these details from orca.

christopherthielen: I was thinking the same thing (pulling from orca)

PR #4580, opened 2017-12-12
fix(provider/k8s-v2): Use 'manifestName' in place of 'name' in adhoc operations

andrewbackes:
https://github.com/spinnaker/orca/pull/1852


andrewbackes:
@lwander

andrewbackes:
Ah, good catch. Should be done now.

PR #4582, opened 2017-12-12
K8s init containers

ethanfrogers:
Adds support for initContainers. This diff is big but it reads pretty well. I don't mind splitting it up if necessary.

* Split out the container detail for Server Groups and Instances so that I could reuse them.
* Support initContainers wherever a container is supported
* Selectively show certain fields based on `initContainer == true`

Screen shots incoming.

@danielpeach @lwander 

### Server Group - Container Detail
image

### Instance - Container Detail
image

### Init Container Configuration
image


app/scripts/modules/kubernetes/src/serverGroup/configure/configuration.service.js
```diff
@@ -172,12 +170,20 @@ module.exports = angular.module('spinnaker.serverGroup.configure.kubernetes.conf
...
           } else {
-            result.dirty.containers = result.dirty.containers || [];
-            result.dirty.containers.push(container.image);
+            result = result || [];
+            result.push(container.image);
           }
         }
       });
-      command.containers = validContainers;
+      return validContainers;
```
ethanfrogers: the diff here is kind of crappy. i'm not _totally_ happy with this function since it mutates and returns something. i broke it so that it could be used for both container sets. suggestions welcome.

ethanfrogers:
related clouddriver PR: https://github.com/spinnaker/clouddriver/pull/2219

app/scripts/modules/kubernetes/src/container/configurer.directive.js
```diff
@@ -16,6 +16,7 @@ module.exports = angular.module('spinnaker.kubernetes.container.configurer.direc
         container: '=',
         index: '=',
         command: '=',
+        initContainer: '@',
```
danielpeach: Would use `<` here rather than `@`.

app/scripts/modules/kubernetes/src/serverGroup/configure/configuration.service.js
```diff
@@ -16,7 +16,8 @@ module.exports = angular.module('spinnaker.serverGroup.configure.kubernetes.conf
     function configureCommand(application, command, query = '') {
 
       // this ensures we get the images we need when cloning or copying a server group template.
-      let queries = command.containers
+      const containers = command.containers.concat(command.initContainers);
```
danielpeach: `command.containers.concat(command.initContainers || []);` would make me feel a little better, since I'm not 100% sure that `initContainers` will always be initialized as an empty array e.g., for a clone operation.

app/scripts/modules/kubernetes/src/serverGroup/configure/configuration.service.js
danielpeach: Could return `[validContainers, invalidContainers]` as a tuple (not so bad if you use array destructuring) and then mutate `result.dirty` separately.

app/scripts/modules/kubernetes/src/instance/details/containerDetail.component.ts
```diff
@@ -0,0 +1,86 @@
+import { module } from 'angular';
+
+class KubernetesInstanceContainerDetail implements ng.IComponentOptions {
```
danielpeach: Please import `IComponentOptions` from `'angular'` and drop the `ng` namespace.

app/scripts/modules/kubernetes/src/serverGroup/details/containerDetail.component.ts
```diff
@@ -0,0 +1,83 @@
+import { module } from 'angular';
+
+class KubernetesServerGroupContainerDetail implements ng.IComponentOptions {
```
danielpeach: Please import `IComponentOptions` from `'angular'` and drop the `ng` namespace.

ethanfrogers:
@danielpeach thanks! will address!

ethanfrogers:
@danielpeach comments addressed. thanks for the desctucturing tip!

andrewbackes:
@ethanfrogers - Ever since this PR was merged, we've been seeing a spinner icon when editing an existing k8s deployment stage.

Spinner

Rolling back to right before this PR fixes the problem.

kevinawoo:
Here's the stack trace
```
vendor.bundle.js:96401 TypeError: Cannot read property 'forEach' of undefined
    at getValidContainers (app.js:26352)
    at configureContainers (app.js:26389)
    at configureNamespaces (app.js:26425)
    at configureAccount (app.js:26441)
    at app.js:26270
    at processQueue (vendor.bundle.js:98708)
    at vendor.bundle.js:98752
    at Scope.$digest (vendor.bundle.js:99847)
    at Scope.$apply (vendor.bundle.js:100145)
    at done (vendor.bundle.js:94263) "Possibly unhandled rejection: {}"
```

for this function
```js
function getValidContainers(command, containers) {
    var validContainers = [];
    var invalidContainers = [];
    containers.forEach(function (container) {
      if (container.imageDescription.fromContext || container.imageDescription.fromTrigger) {
        validContainers.push(container);
      } else {
        var matchingContainers = command.backingData.filtered.containers.filter(function (test) {
          if (container.imageDescription.registry) {
            return test.imageDescription.imageId === container.imageDescription.imageId;
          } else {
            return _lodash2.default.endsWith(test.imageDescription.imageId, container.imageDescription.imageId);
          }
        });

        if (matchingContainers.length === 1) {
          validContainers.push(container);
        } else {
          invalidContainers.push(container.image);
        }
      }
    });
```

ethanfrogers:
whoops! sorry about that guys. will get a patch submitted ASAP!

ethanfrogers:
@andrewbackes @kevinawoo this has been resolved in #4587.

PR #4590, opened 2017-12-15
feat(provider/aws): Pipeline support for rolling back a cluster

ajordens:
The underlying `RollbackClusterStage` supports rolling back a cluster
across 1..* regions.

This is an initial implementation and will undergo a round of feedback
with internal stake holders.


ajordens:
image

ajordens:
image

ajordens:
Relates to spinnaker/orca#1866

This is what I would consider the minimal necessary to expose this stage in the UI.

I'd like to run it past a few internal stakeholders and do at least one more iteration before more broadly communicating. 

Ultimately there should not be much that is cloudprovider-specific (beyond cluster selection) ... maybe there is a better implementation that wouldn't require an `awsRollbackClusterStage` for example. Much of that was borrowed from existing stage implementations.

app/scripts/modules/amazon/src/pipeline/stages/rollbackCluster/awsRollbackClusterStage.js
```diff
@@ -0,0 +1,59 @@
+'use strict';
+
+const angular = require('angular');
+
+module.exports = angular.module('spinnaker.amazon.pipeline.stage.rollbackClusterStage', [
```
ajordens: Straight up borrowed ... what's the better way of doing this @anotherchrisberry? 

At first glance I didn't find any Reactified aws pipeline stages ... despite there being a fair bit of overlap.

app/scripts/modules/amazon/src/pipeline/stages/rollbackCluster/awsRollbackClusterStage.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    ctrl.reset = () => {
+      ctrl.accountUpdated();
+      ctrl.resetSelectedCluster();
+    };
+
+    stage.regions = stage.regions || [];
+    stage.cloudProvider = 'aws';
+    stage.targetHealthyRollbackPercentage = stage.targetHealthyRollbackPercentage || 100;
```
ajordens: only change

app/scripts/modules/amazon/src/pipeline/stages/rollbackCluster/rollbackClusterStage.html
```diff
@@ -0,0 +1,35 @@
...
+      component="stage"
+      accounts="accounts">
+    </account-region-cluster-selector>
+  </div>
+  <stage-platform-health-override application="application"
+                                  stage="stage"
+                                  platform-health-type="'Amazon'">
+  </stage-platform-health-override>
+
+  <div class="row">
```
ajordens: Also borrowed save for this block.

app/scripts/modules/core/src/pipeline/config/stages/rollbackCluster/RollbackClusterExecutionDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+export function RollbackClusterExecutionDetails(props: IExecutionDetailsSectionProps) {
+  const { AccountTag } = NgReact;
+  const { stage } = props;
+
+  const imagesToRestore = stage.context.imagesToRestore && stage.context.imagesToRestore.map((entry: any) => {
+    const imageDetails = entry.buildNumber ? '#' + entry.buildNumber : entry.image;
+    const method = (entry.rollbackMethod === 'EXPLICIT') ? 'previous server group' : 'new server group';
+    const helpId = 'cluster.rollback.' + entry.rollbackMethod.toLowerCase();
+    return (
+      <div key={entry.region}>
```
ajordens: Had a lint error without the `key` ... not sure if I'm doing something wrong.

app/scripts/modules/amazon/src/pipeline/stages/rollbackCluster/awsRollbackClusterStage.js
jrsquared: We haven't gotten around to React-ing pipeline stages yet...

app/scripts/modules/core/src/pipeline/config/stages/rollbackCluster/RollbackClusterExecutionDetails.tsx
jrsquared: The `key` is to identify the uniqueness of the element in the array created by `stage.context.imagesToRestore.map((entry: any) => {}`. As long as the key you choose is guaranteed to be unique across that entire map, then you're doing something right!

jrsquared:
Looks good to me! I was going to complain that you're adding some `.js` but since it's a copy, I'll let it slide _this time_. Hopefully soon we'll get to React-ing the stage configs.

app/scripts/modules/core/src/pipeline/config/stages/rollbackCluster/RollbackClusterExecutionDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { StageExecutionLogs, StageFailureMessage } from 'core/pipeline/details';
+
+export function RollbackClusterExecutionDetails(props: IExecutionDetailsSectionProps) {
+  const { AccountTag } = NgReact;
+  const { stage } = props;
+
+  const imagesToRestore = stage.context.imagesToRestore && stage.context.imagesToRestore.map((entry: any) => {
+    const imageDetails = entry.buildNumber ? '#' + entry.buildNumber : entry.image;
+    const method = (entry.rollbackMethod === 'EXPLICIT') ? 'previous server group' : 'new server group';
+    const helpId = 'cluster.rollback.' + entry.rollbackMethod.toLowerCase();
```
jrsquared: template string.
```
const helpId = `cluster.rollback.${entry.rollbackMethod.toLowerCase()}`
```

app/scripts/modules/amazon/src/pipeline/stages/rollbackCluster/awsRollbackClusterStage.js
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    if (!stage.credentials && $scope.application.defaultCredentials.aws) {
+      stage.credentials = $scope.application.defaultCredentials.aws;
+    }
+    if (!stage.regions.length && $scope.application.defaultRegions.aws) {
+      stage.regions.push($scope.application.defaultRegions.aws);
+    }
+
+    ctrl.pluralize = function(str, val) {
```
anotherchrisberry: doesn't look like this is being used anywhere

app/scripts/modules/core/src/pipeline/config/stages/rollbackCluster/RollbackClusterExecutionDetails.tsx
```diff
@@ -0,0 +1,60 @@
...
+    return (
+      <div key={entry.region}>
+        <dt>
```
anotherchrisberry: These should probably be wrapped in a `<dl>` - not sure how the browser handles dd/dt tags sitting outside a dl...

app/scripts/modules/core/src/pipeline/config/stages/rollbackCluster/RollbackClusterExecutionDetails.tsx
ajordens: 👍

app/scripts/modules/amazon/src/pipeline/stages/rollbackCluster/awsRollbackClusterStage.js
ajordens: 👍

app/scripts/modules/core/src/pipeline/config/stages/rollbackCluster/RollbackClusterExecutionDetails.tsx
ajordens: 👍

PR #4628, opened 2018-01-05
feat(core): import kayenta

sbwsg:
Requires us to temporarily downgrade react, prop-types, and react-dom in deck's package.json.  With higher versions of those packages yarn considers them in conflict with kayenta and doesn't hoist them to the root node_modules directory.  Webpack is then unable to find them during build.

### DIFF
--- a/app/scripts/app.ts
+++ b/app/scripts/app.ts
@@ -8,6 +8,7 @@ import { GOOGLE_MODULE } from '@spinnaker/google';
 import { OPENSTACK_MODULE } from '@spinnaker/openstack';
 import { CANARY_MODULE } from './modules/canary/canary.module';
 import { KUBERNETES_V1_MODULE, KUBERNETES_V2_MODULE } from '@spinnaker/kubernetes';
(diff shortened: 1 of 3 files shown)


danielpeach:
The commit message should probably be 'feat(core):` since it's really pulling in `@spinnaker/kayenta`.

sbwsg:
The issue seems to be that yarn is placing some dependencies of kayenta and react-bootstrap into their own subdirectories.  Some deps of Kayenta end up in `node_modules/@spinnaker/kayenta/node_modules` and some of react-bootstrap end up in `node_modules/react-bootstrap/node_modules`.  Because of this, webpack is unable to find those dependencies when it comes time to compile.

In order to work around this I've to added these paths to webpack's resolve plugin.

This workaround feels like a hack.  I am not clear why yarn is placing certain dependencies of these two packages into their own node_modules directories and not into the root node_modules.  Ultimately that seems to be causing this problem.

@anotherchrisberry do you have any insight here into yarn's behaviour?  Or a better way of tackling this issue?

PR #4646, opened 2018-01-09
feat(provider/ecs): Added Disable ASG stage

BrunoCarrier:
feat(provider/ecs): Added Disable ASG stage


BrunoCarrier:
I am not sure why I didn't push that a while ago.  I hope that this format of doing things is still valid, which it seems to be

PR #4658, opened 2018-01-11
feat(provider/ecs): Updated the ECS Module with all its content

BrunoCarrier:
feat(provider/ecs): Updated the ECS Module with all its content


BrunoCarrier:
This should be merged only after https://github.com/spinnaker/deck/pull/4650 is merged

PR #4667, opened 2018-01-15
Alternative application view in the clusters tab

jervi:
Adds a new, alternative way of looking at the clusters. In the new "Application view", the clusters are primarily sorted by cluster name instead of environment. It will also display a warning if more than one server group is active at the same time in the same region.

Please let me know what you think, I'm open to suggestions!

skjermbilde 2018-01-15 kl 10 24 45

skjermbilde 2018-01-15 kl 10 24 23

skjermbilde 2018-01-15 kl 10 25 38

skjermbilde 2018-01-15 kl 10 37 18

Tip: it might be easier to read the PR if you [ignore whitespace](https://github.com/spinnaker/deck/pull/4667/files?w=1).

### DIFF
--- a/app/scripts/modules/core/src/cluster/ClusterPod.tsx
+++ b/app/scripts/modules/core/src/cluster/ClusterPod.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
+import * as classNames from 'classnames';
 import { BindAll } from 'lodash-decorators';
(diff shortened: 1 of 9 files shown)


jrsquared:
Talked about this offline. Closing in favor of a yet-to-be-built larger effort around regrouping the cluster view.

PR #4669, opened 2018-01-16
feat(webhooks): preconfigured webhook params

ethanfrogers:
adds support for preconfigured webhook parameters. these parameters can
be set in orca's config and are then exposed to users. instead of
expecting users to fillout an entire payload, these parameters can be
used to expose the important parts and abstract the other configuration
away.

related orca pr: spinnaker/orca#1887

@anotherchrisberry @ajordens PTAL

### Configuration
image

### Execution
image


ethanfrogers:
spinnaker/spinnaker#2224

PR #4673, opened 2018-01-16
Add/update Travis Slack notifications

erikmunson:
Our original Travis notification key for the `#builds` Slack channel seems to have broken/gotten invalidated. This re-adds an updated (encrypted) key for `#builds` *and* adds notifications for the new `#builds-deck` channel which should theoretically only include build updates for deck.

This is the first repo I'm getting working with Travis/Slack again so may need to come back and debug, but this *should* work.


erikmunson:
FYI this appears to have worked as intended.

christopherthielen:
lgtm

PR #4686, opened 2018-01-19
feat(provider/appengine): Container deploy w/ free url or artifact

sbwsg:
appengine_pipeline_artifact

### DIFF
--- a/app/scripts/modules/appengine/serverGroup/configure/wizard/basicSettings.html
+++ b/app/scripts/modules/appengine/serverGroup/configure/wizard/basicSettings.html
@@ -209,6 +209,39 @@
 
     <div ng-if="basicSettingsCtrl.isContainerImageSource()">
       <div class="form-group">
+        <label class="col-md-3 sm-label-right">Resolve URL</label>
+        <div class="col-md-7">
+          <div class="radio radio-inline">
+            <label>
+              <input type="radio"
+                     ng-model="command.fromArtifact"
+                     ng-value="false"> via text input
+            </label>
+          </div>
+          <div class="radio radio-inline">
+            <label>
+              <input type="radio"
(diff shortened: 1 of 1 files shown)


app/scripts/modules/appengine/serverGroup/configure/wizard/basicSettings.controller.ts
```diff
@@ -53,6 +53,9 @@ class AppengineServerGroupBasicSettingsCtrl implements IController {
     }
 
     const artifact = copy(expected.matchArtifact);
+    if (artifact == null) {
+      return '';
+    }
     return Object.keys(artifact)
```
lwander: I think there is a filter registered for this, something like summarize-expected-artifacts

sbwsg:
:+1: screenshot added.

app/scripts/modules/appengine/serverGroup/configure/wizard/basicSettings.controller.ts
sbwsg: I'll take another look here.  I saw exceptions in this function around fetching keys for null/undefined.  They occurred after deleting an Expected Artifact from a Configuration stage that was in use in a Deploy stage.

lwander:
Looking good!

What happens when GCS is selected, and they want to deploy a docker container from an artifact?

sbwsg:
Fair question!  Currently that isn't supported.  I'll update this PR once work is done to support separate `app.yaml` and container artifacts in clouddriver.

sbwsg:
@lwander I've removed the `if (artifact == null)` checks and will follow up on multiple-artifact appengine deploys in a new PR.  Otherwise ok to merge?

lwander:
SGTM!

PR #4696, opened 2018-01-21
fix(provider/kubernetes): Disable editing of managed ReplicaSets

wjoel:
ReplicaSets owned by a Kubernetes controller should be scaled and edited by using the server group manager to change the controller.

Otherwise the controller that owns the ReplicaSet will immediately undo the scaling/editing.

Disable scaling and editing if the manifest is owned by a controller, and include links to the server group manager if there is one.

Allow ReplicaSets to be deleted, but show a warning explaining that the controller that owns the ReplicaSet may recreate it.

Fixes https://github.com/spinnaker/spinnaker/issues/2211

We prefer small, well tested pull requests.

Please refer to [Contributing to Spinnaker](https://spinnaker.io/community/contributing/).

When filling out a pull request, please consider the following:

* Follow the commit message conventions [found here](http://www.spinnaker.io/v1.0/docs/how-to-submit-a-patch).
* Provide a descriptive summary for your changes.
* If it fixes a bug or resolves a feature request, be sure to link to that issue.
* Add inline code comments to changes that might not be obvious.
* Squash your commits as you keep adding changes.
* Add a comment to @spinnaker/reviewers for review if your issue has been outstanding for more than 3 days.

Note that we are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.


lwander:
@sbwsg could you take a peek as well?

lwander:
paging @sbwsg

PR #4713, opened 2018-01-23
fix(provider/openstack): Subnet list now updates on region change

ajn0592:
Previously, when changing the region when creating a new openstack server group the subnet list did not update. This is because there wasn't a watch on the region changing. This has now been fixed.


ajn0592:
@edwinavalos Should be good to go! Thank You!

PR #4714, opened 2018-01-23
chore(core): Ignore styleguide.html changes since it is generated

jrsquared:
@archana-s PTAL


archana-s:
Thank you. Looks good.

PR #4726, opened 2018-01-25
fix(provider/gce): Fix search endpoint calls.

jtk54:
We previously used /search to query the infrastructure caches and
suggest several types of components in the UI. There was a recently
added constraint that requires /search queries to provide a string
`>=` length 3, which broke our cache queries. This change is a temporary
fix/workaround to fix the contract breakage.

### DIFF
--- a/app/scripts/modules/google/src/certificate/certificate.reader.ts
+++ b/app/scripts/modules/google/src/certificate/certificate.reader.ts
@@ -20,7 +20,7 @@ export class GceCertificateReader {
 
   public listCertificates(): IPromise<IGceCertificate[]> {
     return this.searchService
-      .search<IGceCertificate>({ q: '', type: 'sslCertificates' }, this.infrastructureCaches.get('certificates'))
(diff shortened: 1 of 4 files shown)


app/scripts/modules/google/src/certificate/certificate.reader.ts
```diff
@@ -20,9 +20,10 @@ export class GceCertificateReader {
 
   public listCertificates(): IPromise<IGceCertificate[]> {
     return this.searchService
-      .search<IGceCertificate>({ q: '', type: 'sslCertificates' }, this.infrastructureCaches.get('certificates'))
+      .search<IGceCertificate>({ q: '***', type: 'sslCertificates' }, this.infrastructureCaches.get('certificates'))
       .then((searchResults: ISearchResults<IGceCertificate>) => {
         if (searchResults && searchResults.results) {
+          debugger;
```
ajordens: I don't know what this is?

jtk54: Pauses execution and opens a debugger in chrome if you have the console open -- need to remove :)

PR #4728, opened 2018-01-25
feat(core): Indicate templated pipelines

sbwsg:
Add a small table icon to the group header / execution title of pipelines that are based on templates.

Open to suggestions on better icon choice.

templated_pipeline_name

templated_pipeline_group

Includes a tooltip on hover:

screenshot from 2018-01-25 16-30-34


christopherthielen:
nice

PR #4739, opened 2018-01-30
fix(provider/kubernetes): add error message when manifest yaml isn't valid

sbwsg:
931a5aa2-e668-4879-9273-2f3efa7173e6

A small change to add an indicator when entered manifest YAML isn't valid.


lwander:
woah

PR #4746, opened 2018-01-31
style(pipelines): Remove explicit use of $q.defer

ezimanyi:
Small fix to replace explicit use of $q.defer with directly
returning a promise.


ezimanyi:
@danielpeach: Small fix in response to the comment you raised in #4743.

PR #4750, opened 2018-02-01
fix(core/search): Export the new GlobalSearch react component

jervi:
To be able to use it from e.g. the customized deck


jervi:
@erikmunson

erikmunson:
❤️

PR #4756, opened 2018-02-01
fix(provider/kubernetes): Remove a trailing 'for' from title in show manifest modal

sbwsg:
before
before

after
after

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/showManifestYaml.component.ts
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/showManifestYaml.component.ts
@@ -2,8 +2,6 @@ import { copy, IComponentOptions, IController, IRootScopeService, module } from 
 import { IModalService } from 'angular-ui-bootstrap';
 import { dump } from 'js-yaml';
 
-import { ServerGroupTemplates } from '@spinnaker/core';
-
 class KubernetesShowManifestYaml implements IController {
   public manifest: any;
   public text: string;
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/serverGroup/templates/index.ts
```diff
@@ -1,3 +1,4 @@
 export const ServerGroupTemplates = {
   userData: require('../details/userData.html'),
+
```
lwander: Necessary?

app/scripts/modules/core/src/pipeline/executions/execution/Execution.tsx
```diff
@@ -289,14 +289,14 @@ export class Execution extends React.Component<IExecutionProps, IExecutionState>
           { this.props.title && (
             <h4 className="execution-name">
               {this.props.showAccountLabels && accountLabels}
-              {this.props.execution.fromTemplate && <i className="from-template fa fa-table" title="Template Pipeline" />}
+              {this.props.execution.fromTemplate && <i className="from-template fa fa-table" title="Pipeline from template" />}
```
danielpeach: I think these changes are from an earlier change.

app/scripts/modules/core/src/serverGroup/templates/index.ts
sbwsg: most important line in the whole PR

PR #4763, opened 2018-02-02
fix(core/search): Remove angular2react bridging of native react component

jervi:


jervi:
@erikmunson One more thing :)

PR #4787, opened 2018-02-05
Add form validator helpers and better amazon target group validation

jrsquared:


app/scripts/modules/core/src/presentation/Tooltip.tsx
```diff
@@ -1,35 +1,57 @@
...
-      tooltip = <BSTooltip id={this.props.value}>{this.props.template}</BSTooltip>;
+    const { delayShow, id, placement, template, value } = this.props;
+    const useId = id || value || 'tooltip';
+
+    let tooltip = <BSTooltip id={useId}>{value}</BSTooltip>;
+    if (template) {
+      tooltip = <BSTooltip id={useId}>{template}</BSTooltip>;
     }
 
+    const hasValue = (value && value.length > 0) || template;
```
jrsquared: Allowing no value, and hiding when there is no value, allows the element and child element to not have to be re-rendered if switching between a tooltip and no tooltip. For this specific PR, allows focus to be kept on the input element.

app/scripts/modules/core/src/presentation/FormElements.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
+
+import { Tooltip } from 'core/presentation';
+
+export const SpInput = (props: React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLInputElement>, HTMLInputElement> & { error?: string, name: string }) => {
```
jrsquared: Will happily accept name suggestions.

PR #4802, opened 2018-02-06
style(core): Styling react-select to match spinnaker select

archana-s:
Added a few overrides for react select to match spinnaker select design specs.


jrsquared:
v.nice

PR #4809, opened 2018-02-07
feat(dryrun): distinguish dry run pipelines in execution view

robfletcher:


robfletcher:
@christopherthielen showed me what to do. I just typed it.

PR #4813, opened 2018-02-07
feat(provider/kubernetes): List trigger artifacts in execution status

sbwsg:
https://github.com/spinnaker/spinnaker/issues/2299

artifacts

### DIFF
--- a/app/scripts/modules/core/src/pipeline/status/ArtifactList.tsx
+++ b/app/scripts/modules/core/src/pipeline/status/ArtifactList.tsx
@@ -1,0 +1,79 @@
+import * as React from 'react';
+import { IArtifact } from 'core/domain';
+
+import './artifactList.less';
+
+export interface IArtifactListProps {
+  artifacts: IArtifact[],
+};
+
+export interface IArtifactListState {}
+
+export class ArtifactList extends React.Component<IArtifactListProps, IArtifactListState> {
+  constructor(props: IArtifactListProps) {
+    super(props);
+  }
+
+  private tooltip(artifact: IArtifact): string {
+    const tooltipEntries = [];
+    if (artifact.name) {
+      tooltipEntries.push(`Name: ${artifact.name}`);
+    }
+    if (artifact.type) {
+      tooltipEntries.push(`Type: ${artifact.type}`);
+    }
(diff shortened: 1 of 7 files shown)


app/scripts/modules/core/src/pipeline/status/ArtifactList.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export class ArtifactList extends React.Component<IArtifactListProps, IArtifactListState> {
+  constructor(props: IArtifactListProps) {
+    super(props);
+  }
+
+  private tooltip(artifact: IArtifact): string {
+    const tooltipEntries = [];
+    if (artifact.name) {
+      tooltipEntries.push(`Artifact: ${artifact.name}`);
```
lwander: I think it might be safer to reference this as `Name: ${artifact.name}`, or `Artifact name: ...` if that feels better

app/scripts/modules/core/src/pipeline/status/ArtifactList.tsx
```diff
@@ -0,0 +1,71 @@
...
+      return null;
+    }
+    return (
+      <ul className="trigger-details artifacts">
+        {artifacts.map((artifact: IArtifact, i: number) => {
+          const { name, version } = artifact;
+          return (
+            <li key={`${i}-${name}`} className="break-word" title={this.tooltip(artifact)}>
+              <dl>
+                <div>
```
lwander: I think the "type" should appear here as well, probably closer to the front

lwander: Maybe it could even take the place of the keyword "Artifact"

sbwsg: :+1: I've added a `Type: <type>` row.  Replacing the keyword "Artifact" with, for example, "docker/image" pushed the artifact's name a bit too far to the right.

app/scripts/modules/core/src/pipeline/status/ArtifactList.tsx
sbwsg: :+1:

PR #4814, opened 2018-02-07
chore(halconfig): Add ecs to defaultProviders list

nearlyfreeapps:
Add ecs to the defaultProviders list in halconfig/settings.js

This should make it easier for anyone using halyard to test the new ecs provider.


nearlyfreeapps:
That makes sense! ECS support is pretty early days so this was an attempt to ease the testing process (installing via `hal config version edit --version master-latest-unvalidated`). Will revisit with proper halyard support.

PR #4818, opened 2018-02-07
fix(core/serverGroup): Fallback to no providerVersion if user does not have account permission

jrsquared:
@lwander FYI since your name is on this bit...


jrsquared:
In some cases, we don't have permission to perform any actions in the account, but we can still see server groups that were created.

PR #4819, opened 2018-02-07
feat(core/styleguide): Add underline-only style for a/button

erikmunson:
Ran into a case where I'd love to have the `.nostyle` behavior for color/size/etc., but wanted to retain the underline-on-hover behavior. My guess is that we're going to run into this case more than once (and probably significantly more than once), so adding it here.


app/scripts/modules/core/src/styleguide/src/styles/buttons.css
```diff
@@ -10,8 +10,9 @@ These are three categories for buttons based on function:
     3. passive(default) - Buttons that are used in the page which do not immediately call for action. They are not the primary action for the page.
     4. danger - Buttons that are used to indicate operations that can have consequences.
     5. nostyle - No styles attached to the button. Relies on the user to set the styles based on parent element.
-    6. buttons with icons - Buttons with icons to the left and right of the buttons
-    7. zombie - Buttons that have a dashed border and is typically used with configurations.
+    6. underline only - Same as nostyle, but adds a text underline on hover to suggest interactivity.
```
archana-s: can you please also add an example for the underline-only?

app/scripts/modules/core/src/styleguide/src/styles/buttons.css
```diff
@@ -86,6 +87,7 @@ Here are the variations:
 1. a - Default link style.
 2. a.button(.primary/.passive/.danger) - Links can also look like buttons.
 3. a.nostyle - There is also a variation of link with no styles attached.
+4. a.underline-only - same as nostyle, but adds a text underline on hover to suggest interactivity.
 
 <a>Link</a>
```
archana-s: an example here too <a class="underline-only">Something</a>

archana-s:
just a note to add examples so we all know when we can use it.

erikmunson:
After talking more about this, I'm actually going to keep it out of the style guide in the short term while we figure out a better way to handle the use case we have in mind.

PR #4825, opened 2018-02-08
fix(provider/kubernetes): exclude stages when k8s v2 is the only provider

sbwsg:
~Providers are passed around as strings when assessing whether stages support them or not.  Both the kubernetes v1 and the kubernetes v2 provider share a provider string of `kubernetes`.  Therefore the stage filtering assumes that any stage supported by v1 is also supported by v2 but this is not the case.~

~This workaround allows a field to be added to a stage that lists specific provider versions that it doesn't support.  After the initial set of stages is filtered via provider strings, a second round of filtering is performed to remove any stage where the set of accounts for this application are entirely excluded by the stage's type.~

~This solution doesn't feel quite right but I can't see a way past this problem without a large refactor such that everywhere we pass a provider string we also pass around a version.  Maybe that would be a better solution though.  Open to ideas here.~

Fixed.

#### Before: (A Deploy stage appears in an app with only a k8s v2 account configured)

before

#### After: (The Deploy stage is filtered out because its excludedCloudProviders field includes k8s v2)

after

https://github.com/spinnaker/spinnaker/issues/2277

### DIFF
--- a/app/scripts/modules/core/src/account/account.service.ts
+++ b/app/scripts/modules/core/src/account/account.service.ts
@@ -161,6 +161,14 @@ export class AccountService {
       .then((accounts) => accounts.filter((account) => account.authorized !== false));
   }
 
+  public applicationAccounts(application: Application = null): Promise<IAccountDetails[]> {
+    return Promise.all([this.listProviders(application), this.listAccounts()]).then(([providers, accounts]) => {
+      return providers.reduce((memo, p) => {
+        return memo.concat(accounts.filter(acc => acc.cloudProvider === p));
+      }, [] as IAccountDetails[]);
+    });
+  }
+
   public listProviders(application: Application = null): IPromise<string[]> {
     return this.listAllAccounts()
       .then((accounts: IAccount[]) => {
--- a/app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
+++ b/app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
@@ -52,9 +52,9 @@ module.exports = angular.module('spinnaker.core.pipeline.config.stage', [
       selectedStageType: null,
     };
 
-    accountService.listProviders($scope.application).then((providers) => {
-      $scope.options.stageTypes = pipelineConfig.getConfigurableStageTypes(providers);
-      $scope.showProviders = providers.length > 1;
+    accountService.applicationAccounts($scope.application).then(accounts => {
+      $scope.options.stageTypes = pipelineConfig.getConfigurableStageTypes(accounts);
+      $scope.showProviders = new Set(accounts.map(a => a.cloudProvider)).size > 1;
     });
 
     if ($scope.pipeline.strategy) {
--- a/app/scripts/modules/core/src/domain/IStageTypeConfig.ts
+++ b/app/scripts/modules/core/src/domain/IStageTypeConfig.ts
@@ -15,6 +15,7 @@ export interface IStageTypeConfig extends IStageOrTriggerTypeConfig {
   configuration?: any;
   defaultTimeoutMs?: number;
   disableNotifications?: boolean;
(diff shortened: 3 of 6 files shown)


lwander:
Nice!

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
```diff
@@ -52,8 +53,12 @@ module.exports = angular.module('spinnaker.core.pipeline.config.stage', [
       selectedStageType: null,
     };
 
-    accountService.listProviders($scope.application).then((providers) => {
-      $scope.options.stageTypes = pipelineConfig.getConfigurableStageTypes(providers);
+    accountService.providerAccounts($scope.application).then(providerAccounts => {
+      const providers = Object.keys(providerAccounts);
+      const accounts = providers.reduce((memo, provider) => memo.concat(providerAccounts[provider]), []);
+      let stageTypes = pipelineConfig.getConfigurableStageTypes(providers);
```
danielpeach: My slight preference would be that `getConfigurableStageTypes` returned the exact set of `stageTypes` to render (i.e., this logic should be encapsulated in that method, probably by passing it a list `{ provider: string, providerVersion: string}[]`).

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
```diff
@@ -197,6 +202,20 @@ module.exports = angular.module('spinnaker.core.pipeline.config.stage', [
       }
     }
 
+    function stageExcludesAllAccounts(stageType, accounts) {
+      if (stageType.excludedCloudProviders) {
+        const anyNotExcluded = some(accounts, account => {
```
danielpeach: Would use native `[].some` here and below.

danielpeach: Would also be a little more readable if the bool were swapped such that this was an `[].every`. i.e., every `account` is excluded by at least one of the `excludedCloudProviders`.

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
```diff
@@ -197,6 +202,20 @@ module.exports = angular.module('spinnaker.core.pipeline.config.stage', [
...
+          return some(stageType.excludedCloudProviders, excluded => (
```
danielpeach: Could make this `stageType.excludedCloudProviders || []` and remove the entire enclosing `if { ... } else { ... }`.

app/scripts/modules/core/src/account/account.service.ts
```diff
@@ -158,6 +162,16 @@ export class AccountService {
     return result;
   }
 
+  public providerAccounts(application: Application = null): Promise<IProviderAccounts> {
+    const result: Promise<IProviderAccounts> = Promise.all([this.listProviders(application), this.listAccounts()]).then(([providers, accounts]) => {
```
danielpeach: Would return directly on this line.

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
sbwsg: :+1: done, thanks, def the better approach.

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
sbwsg: done

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
sbwsg: done

sbwsg:
@danielpeach I've fixed the issues you highlighted with multiple provider versions and removed the custom interfaces in favor of an array of `IAccountDetails`.

PR #4829, opened 2018-02-09
fix(core/*): Fix viewing of account badges, and servergroup/instance/foo details for accounts you don't have access to

christopherthielen:
- Add `getAllAccounts` to return `IAccountDetails` metadata for all accounts
- Passthrough other APIs such as `getAccountDetails` through getAllAccounts
- `getAccounts` continues to return only accounts the user is authorized for


christopherthielen:
@jrsquared plz
@anotherchrisberry fyi

jrsquared:
v.nice

PR #4831, opened 2018-02-12
fix(core/stages): Hide excluded providers

sbwsg:
fixes https://github.com/spinnaker/spinnaker/issues/2277

- Hide excluded provider icons in stage selection dropdown
- Hide excluded providers from provider selection modal

## Before Icon Fix

before_icon_fix

## After Icon Fix

after_icon_fix

## Before Modal Fix

Excluded provider appears as an Add Server Group option.

before_provider_choice_fix

## After Modal Fix

No Modal Appears

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/pipelineConfigProvider.ts
+++ b/app/scripts/modules/core/src/pipeline/config/pipelineConfigProvider.ts
@@ -1,9 +1,10 @@
 import { auto, IServiceProvider, module } from 'angular';
-import { isNil, cloneDeep, intersection, memoize } from 'lodash';
+import { uniq, isNil, cloneDeep, intersection, memoize } from 'lodash';
 import { $log } from 'ngimport';
 
 import { Application } from 'core/application/application.model';
 import { IExecution, IStage, ITriggerTypeConfig, IStageTypeConfig, IArtifactKindConfig, IStageOrTriggerTypeConfig } from 'core/domain';
+import { CLOUD_PROVIDER_REGISTRY, CloudProviderRegistry, ICloudProviderConfig } from 'core/cloudProvider/cloudProvider.registry';
 import { SETTINGS } from 'core/config/settings';
 
 import { IAccountDetails } from 'core/account/account.service';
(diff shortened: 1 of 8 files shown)


sbwsg:
working on the tests now

app/scripts/modules/core/src/pipeline/config/pipelineConfigProvider.ts
```diff
@@ -107,7 +109,12 @@ export class PipelineConfigProvider implements IServiceProvider {
         }
       });
     } else {
-      cloudProviders = providers;
+      cloudProviders = Array.from(new Set(accounts.reduce((memo, acc) => {
```
jrsquared: what's the benefit of converting an array to a set and back to an array again?

app/scripts/modules/core/src/pipeline/config/stages/deploy/deployStage.js
```diff
@@ -6,6 +6,8 @@ import {StageConstants} from 'core/pipeline/config/stages/stageConstants';
 
 const angular = require('angular');
 
+const excludedCloudProviders = [{ cloudProvider: 'kubernetes', providerVersion: 'v2' }];
```
jrsquared: I know you just refactored this bit, but why is there anything referencing kubernetes in the core module?

app/scripts/modules/core/src/pipeline/config/pipelineConfigProvider.ts
sbwsg: The intention is to ensure each element in the array is unique.

jrsquared: Ah, got it.

app/scripts/modules/core/src/pipeline/config/stages/deploy/deployStage.js
sbwsg: Ah, tbh I spent a bunch of time trying to figure out how to exclude a provider version from a stage type but didn't actually give any consideration to the fact I was polluting the core module with provider-specific code.  I'll go back and see if there's a different way to approach this to keep the kubernetes exclusion limited to the k8s module.

sbwsg: I've refactored so that providers now specify stages to exclude in their own module rather than stages excluding providers in the core module.

app/scripts/modules/core/src/pipeline/config/pipelineConfigProvider.ts
sbwsg: updated to use the more explicit `uniq` lodash helper.

jrsquared:
nice. lgtm.

PR #4834, opened 2018-02-12
fix(amazon/vpc): Fix initial render of vpc

christopherthielen:


app/scripts/modules/amazon/src/vpc/VpcTag.tsx
```diff
@@ -20,20 +20,20 @@ export class VpcTag extends React.Component<IVpcTagProps, IVpcTagState> {
   }
 
   private updateState(props: IVpcTagProps): void {
```
erikmunson: fwiw, ideally the initial call to `updateState` should be in a `componentDidMount`, not the constructor.

PR #4847, opened 2018-02-13
fix(provider/appengine): server group wizard: gcs "storage account" -> "artifact account"

sbwsg:
Before:

storage_account

After:

artifact_account


sbwsg:
:+1: https://github.com/spinnaker/deck/pull/4848

PR #4851, opened 2018-02-14
core(styleguide): Added more css vars and a css vars only file

archana-s:
Converted more values to CSS custom variables. These are all now available as json.

We have been getting requests for all CSS properties in spinnaker to be reused in other areas and products. The purpose of this PR is to expose the existing CSS vars as JSON that others can use. 
This one handles only vars specific to colors, type and buttons that will be tried out soon.


app/scripts/modules/core/src/styleguide/src/styles/typography.css
```diff
@@ -166,14 +166,24 @@ Here's the styleguide template for this project: ../styleguide_template/template
...
-  /* alter the display start size to appropriate value for desktop and mobile and all the
-   * display sizes will be adjusted appropriately */
-  --display-start-font-size: 16px;
-
-  --display-4-font-size: calc(var(--display-start-font-size) * 1.8);
-  --display-3-font-size: calc(var(--display-start-font-size) * 1.5);
-  --display-2-font-size: calc(var(--display-start-font-size) * 1.32);
-  --display-1-font-size: var(--display-start-font-size);
+  --heading-font-weight: 600;
+  --heading-1-font-size: 22px;
```
erikmunson: totes-optional nit: might be worth breaking up this block of definitions a bit. a new line between the `heading-*-font-[size|height]` vars and the `*-font-weight` perhaps?

erikmunson:
👍 love this

fashion

PR #4863, opened 2018-02-16
fix(provider/kubernetes): exclude manifest stages from v1

sbwsg:
Stages available in an application configured only with a V1 Provider, Before & After:

before_fix after_fix


app/scripts/modules/core/src/pipeline/config/pipelineConfigProvider.ts
```diff
@@ -136,7 +136,6 @@ export class PipelineConfigProvider implements IServiceProvider {
       return configurableStageTypes;
     }
     configurableStageTypes.forEach(type => type.cloudProviders = this.getCloudProvidersForStage(type, allStageTypes, accounts));
-    // remove stage types where all given provider accounts are explicitly excluded by that type
```
sbwsg: no longer true

PR #4885, opened 2018-02-20
fix(core/pipeline): Fix pipeline config - add stage that uses BaseProviderStageCtrl

christopherthielen:
@jrsquared @ethanfrogers @lwander 

This (hopefully) fixes an issue when adding a stage to a pipeline.  It seems there was some `$q` trickery going previously, which was lost when wrapping a native promise as a `$q` promise `$q.when(observable$.toPromise())`.  I suspect this was due to the `$q` promise participating in the existing AngularJS digest cycle, while the wrapped promise was executing after an event loop tick.

This switches the `BaseProviderStageCtrl` to use an observable API which executes synchronously, when possible.

### DIFF
--- a/app/scripts/modules/core/src/account/account.service.ts
+++ b/app/scripts/modules/core/src/account/account.service.ts
@@ -1,4 +1,4 @@
-import { chain, intersection, zipObject } from 'lodash';
+import { chain, intersection, zipObject, uniq } from 'lodash';
 import { ILogService, IPromise, IQResolveReject, IQService, module } from 'angular';
 import { Observable } from 'rxjs';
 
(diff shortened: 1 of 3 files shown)


ethanfrogers:
spinnaker/spinnaker#2358

benjaminws:
It looks like this was reverted. Am I misunderstanding?

lwander:
It was reverted in the release branch, where I mistakenly applied this. The release branch was cut a few days before the target of this fix was introduced, so it's not needed there.

benjaminws:
Ahh, thanks for the follow-up! Wanted to make sure I didn't miss a
conversation about this.

lwander:
No worries, sorry that looks confusing

PR #4901, opened 2018-02-22
fix(provider/google): add resize stage timeout override

sbwsg:
image


sbwsg:
indeed.  it's more accurately 3h21m and 30 seconds but i figured it wasn't worth including the seconds since it wont surface in the UI.

PR #4906, opened 2018-02-23
chore(*): Upgrade React to 16

jrsquared:
Upgrade React to 16 and all update all related dependencies.

Newer react-bootstrap rebuilt the pagination, so this also includes a new pagination for the applications view.

A few console warnings for uirouter bits, but doesn't seem to actually break anything? (@christopherthielen)


sbwsg:
I pulled down this branch and tried it out after deleting node_modules and doing a fresh `yarn`.  Everything seems to work about the same as before - an App Engine deploy and a Compute Engine scale-up pipeline both went through the motions without problem.  I did see a lot of router warning / error spam as you mention.

One issue I spotted was that clicking on the `Configure` link of a pipeline and then clicking the `Pipelines` tab to go back to the list view results in the following errors appearing in console:
remove_child_error

This error didn't break anything but on a repeat attempt at the same user journey, I got the following transition error and wasn't able to get back to the pipelines list:

transition_err

I'm not convinced this is a react16-specific issue though.

jrsquared:
Thanks for pulling it down and testing! Yeah, those are the uirouter warnings i was mentioning we have to look at. I'll have to look at the transition error more closely too. Thanks again!

PR #4910, opened 2018-02-25
feat(core/overrideRegistry): Pass the original component to the overriding component as a prop

christopherthielen:
Should enable something like this:
```
@Overrides('other.component')
class Custom extends React.Component {
  render() {
    const { OriginalComponent, ...props } = this.props;
    return (
      <h3>Custom header</h3>
      <OriginalComponent {...props}/>
    )
  }
}
```


jrsquared:
Nice

PR #4916, opened 2018-02-28
 feat(amazon/loadBalancers): Disable create button until security groups are loaded

jrsquared:
@tomaslin

### DIFF
--- a/app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
+++ b/app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
@@ -37,6 +37,7 @@ export interface IWizardModalState {
   errorPages: Set<string>;
   formInvalid: boolean;
   pages: string[];
+  waiting: Set<string>;
 }
 
 @BindAll()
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
```diff
@@ -153,9 +155,15 @@ export class WizardModal<T extends FormikValues> extends React.Component<IWizard
     setErrors(this.validate(values));
   }
 
+  private setWaiting(section: string, isWaiting: boolean): void {
+    const waiting = new Set(this.state.waiting);
+    isWaiting ? waiting.add(section) : waiting.delete(section);
+    this.setState({ waiting });
```
erikmunson: is the intention to only ever have a max of 1 waiting section? not clear on why there's a `Set` if it's only ever going to have 1 thing in it.

erikmunson: nvm cannot read, disregard.

PR #4918, opened 2018-02-28
feat(pubsub): specify run-as-user for pubsub triggers

danielpeach:


danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/2479

PR #4921, opened 2018-02-28
fix(provider/kubernetes): manifests array nests itself on save

sbwsg:
fixes https://github.com/spinnaker/spinnaker/issues/2470


lwander:
LGTM, thank you for fixing this!

PR #4924, opened 2018-02-28
feat(provider/kubernetes): delineate default artifacts in executions

sbwsg:
Adds an asterisk to indicate default artifacts used in an execution, and includes the phrase "Default Artifact" in the artifact's tooltip.

asterisk
---
tooltip

fixes https://github.com/spinnaker/spinnaker/issues/2329


lwander:
LGTM

PR #4926, opened 2018-02-28
fix(artifacts): Fix no expected artifacts message

ezimanyi:
The "You don't have any expected artifacts" message currently shows
when there are in fact artifacts.


ezimanyi:
Looks like I accidentally created a merge commit by choosing 'Merge' instead of 'Squash and Merge'. (It seems the other spinnaker repos disable creating a merge commit, so I had not run into this before.)

lwander:
Oh no!

PR #4930, opened 2018-02-28
feat(core/notification): Add a quick one-off slack channel affordance

erikmunson:
We wanted to add a little `#` channel indicator in the common Slack notification input we use across the UI. I've intentionally done it in a less-than-reusable way since I'd like to get inputs with before/after annotations into the style guide shortly and didn't want to duplicate work.

### DIFF
--- a/app/scripts/modules/core/src/notification/modal/editNotification.less
+++ b/app/scripts/modules/core/src/notification/modal/editNotification.less
@@ -1,0 +1,13 @@
+.slack-channel-form-control {
+  .slack-channel-annotation {
+    display: block;
+    position: absolute;
+    top: 5px;
+    left: 25px;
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/notification/modal/editNotification.less
```diff
@@ -0,0 +1,13 @@
+.slack-channel-form-control {
+  .slack-channel-annotation {
```
archana-s: if you use a <div> since we are anyway positioning it absolute, we dont need the display: block.

erikmunson: yeah you're right, i don't even remember why i picked a span in the beginning.

PR #4934, opened 2018-03-01
fix(package): Manually bump lockfile entry for fsevents@^1.0.0

erikmunson:
People (myself included) have been getting errors installing `fsevents`. Turns out the version of `fsevents` that's resolved for `fsevents@^1.0.0` in the lockfile is `v1.1.1`, which is exactly one patch version before the package started including binaries for node 8.

Since the packages depending on `fsevents` don't need to be updated, I just went in and manually resolved `fsevents@^1.0.0` to `v1.1.3`, given that we're already using that version for some other packages without issue.

Seems to work great upon basic testing.


christopherthielen:
wcgw?

PR #4937, opened 2018-03-01
fix(core/search): Get rid of useless prompt, hide 'Recently Viewed' header if no history

erikmunson:
Couple of things:

1) I got rid of the "Please enter a search query to get started" text, because it was kind of useless and mostly just added visual noise.

2) Now when there's no recent history, we won't show the awkwardly orphaned "Recently Viewed" header above... nothing at all.

### DIFF
--- a/app/scripts/modules/core/src/search/searchResult/SearchResultGrid.tsx
+++ b/app/scripts/modules/core/src/search/searchResult/SearchResultGrid.tsx
@@ -10,12 +10,6 @@ export interface ISearchResultGridProps {
   resultSet: ISearchResultSet;
 }
 
-const NoQuery = () => (
-  <div className="flex-grow vertical center middle">
-    <h2>Please enter a search query to get started</h2>
-  </div>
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/search/searchResult/SearchResultGrid.tsx
```diff
@@ -57,7 +51,7 @@ export class SearchResultGrid extends React.Component<ISearchResultGridProps> {
     switch (status) {
       case SearchStatus.INITIAL:
       case undefined:
-        return <NoQuery/>;
```
erikmunson: I can't for the life of me figure out where this ever would've gotten rendered, given that its parent component doesn't render unless there *is* a search query. Regardless, I got rid of it.

christopherthielen: this might have been a vestige of a previous version

PR #4938, opened 2018-03-01
fix(provider/gce): Fix call to list backends

ezimanyi:
The call to get current backends is returning no results
as the query string needs to be at least 3 characters. Use the
allowShortQuery parameter to override this behavior.


ezimanyi:
@sbwsg : Yes, it's deliberately a string 'true' as it's a string being passed directly to a query string rather than something we're evaluating in JavaScript.

ezimanyi:
Looks like I accidentally created a merge commit by choosing 'Merge' instead of 'Squash and Merge'.  (It seems the other spinnaker repos disable creating a merge commit, so I had not run into this before.)

PR #4945, opened 2018-03-02
feat(provider/kubernetes): link to resources next to YAML in deploy pipeline

sbwsg:
This is a first pass at https://github.com/spinnaker/spinnaker/issues/2327

Deployments and ReplicaSets now get linked from deploy pipelines to their respective details panels in the clusters view.

From here:

q8mg0vvphtz

To here:

ff9vfltnop2

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/showManifestDetails.component.ts
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/showManifestDetails.component.ts
@@ -1,0 +1,80 @@
+import { module, IComponentOptions, IController } from 'angular';
+import { trim } from 'lodash';
+import { ReactInjector } from 'core/reactShims';
+
+const supportedKinds = ['deployment', 'replicaset'];
+
+class KubernetesShowManifestDetails implements IController {
+  public manifest: any;
+  public accountId: string;
+
+  public canOpen(): boolean {
+    return !!(
+      this.manifest &&
+      this.manifest.kind &&
+      this.manifest.metadata &&
+      this.manifest.metadata.annotations &&
+      supportedKinds.includes(this.manifest.kind.toLowerCase())
+    );
+  }
+
+  public openDetails() {
+    const kind = this.manifest.kind.toLowerCase();
+    if (kind === 'deployment') {
+      this.openDeploymentDetails();
+    } else if (kind === 'replicaset') {
+      this.openReplicaSetDetails();
+    }
+  }
+
+  private buildParams(annotations: any): any {
+    return {
+      accountId: this.accountId,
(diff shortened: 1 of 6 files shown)


app/scripts/modules/kubernetes/src/v2/manifest/showManifestDetails.component.ts
```diff
@@ -0,0 +1,81 @@
...
+
+class KubernetesShowManifestDetails implements IController {
+  public manifest: any;
+  public accountId: string;
+
+  public canOpen() {
+    return !!(
+      this.manifest &&
+      this.manifest.kind &&
+      kinds.indexOf(this.manifest.kind.toLowerCase()) > -1
```
danielpeach: Can just be `kinds.includes(this.manifest.kind.toLowerCase())`

app/scripts/modules/kubernetes/src/v2/manifest/showManifestDetails.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  };
+}
+
+const kinds = ['deployment', 'replicaset'];
+
+class KubernetesShowManifestDetails implements IController {
+  public manifest: any;
+  public accountId: string;
+
+  public canOpen() {
```
danielpeach: Annotate return types when possible: `public canOpen(): boolean`

app/scripts/modules/kubernetes/src/v2/manifest/showManifestDetails.component.ts
```diff
@@ -0,0 +1,81 @@
+import { module, IComponentOptions, IController } from 'angular';
+import { trim } from 'lodash';
+import { ReactInjector } from 'core/reactShims';
+
+function stripQuotes(str: string): string {
```
danielpeach: These functions seem slightly odd to me in the sense that you're not exporting them and they're not private methods on the controller class.

app/scripts/modules/kubernetes/src/v2/manifest/showManifestDetails.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  return trim(str, '"');
+}
+
+function extractAnnotations(annotations?: any): any {
+  if (annotations == null) {
```
danielpeach: Not sure if it's possible that this condition could be `true`, but if it was, it looks like you would still render the link, but it wouldn't work.

sbwsg: i've disabled the link when annotations are not present in the metadata and i've removed this condition.

PR #4954, opened 2018-03-03
fix(core): Fix dismissal of new server group template modal

ezimanyi:
The template selection modal that appears when adding a new server
group in a 'Deploy' stage should dismiss when clicking the 'x'.


ezimanyi:
@sbwsg @danielpeach : This bug was bothering me as I was working on another task related to server groups. Open to feedback if there's a better way of achieving this in Angular.  I wish I didn't need to separately update the template for each provider, but as far as I can tell this is the correct way of making this change.

PR #4958, opened 2018-03-05
Scoped import instead of direct

mikerowehl:
Import fix for a recently added change. The import the way it was written works fine when running under yarn directly, but errors out if you try to build kubernetes as a module.


danielpeach:
Whoops, made + merged this change independently. Thanks.

PR #4965, opened 2018-03-07
style(core): Fixed a bad spinner

archana-s:
Spinner that shows up next to "Start manual execution" link button in the executions page was off. Fixed it.


erikmunson:
👍 random thought — i bet we can do something kind of nice for this use case that is more standardized in the style guide — added something to my own notes to think more about this use case whenever we do more work on higher order style guide patterns.

PR #4966, opened 2018-03-07
fix(core/application): fix scrolling of applications list

christopherthielen:
Closes https://github.com/spinnaker/spinnaker/issues/2508

### DIFF
--- a/app/scripts/modules/core/src/application/search/Applications.tsx
+++ b/app/scripts/modules/core/src/application/search/Applications.tsx
@@ -143,7 +143,7 @@ export class Applications extends React.Component<{}, IApplicationsState> {
     return (
       <div className="infrastructure">
         <div className="infrastructure-section search-header">
-          <div className="container">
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/application/search/Applications.tsx
```diff
@@ -143,7 +143,7 @@ export class Applications extends React.Component<{}, IApplicationsState> {
     return (
       <div className="infrastructure">
         <div className="infrastructure-section search-header">
-          <div className="container">
+          <div className="container" style={{ overflowY: 'auto' }}>
```
erikmunson: do we have a util class that does just this? seems like maybe something worth having.

erikmunson: i suppose probably in the style guide is the best place thinking about it, perhaps we should do a pass of more granular utilities for this kind of thing sometime(TM)

PR #4969, opened 2018-03-07
fix(core/overrideRegistry): Copy static class properties to Overridable component

christopherthielen:
Static methods were missing from Overridable components

### DIFF
--- a/app/scripts/modules/core/src/overrideRegistry/Overridable.tsx
+++ b/app/scripts/modules/core/src/overrideRegistry/Overridable.tsx
@@ -49,6 +49,8 @@ export function Overridable(key: string) {
  */
 export function overridableComponent <P extends IOverridableProps, T extends React.ComponentClass<P>> (OriginalComponent: T, key: string): T {
   class OverridableComponent extends React.Component<P, { Component: T }> {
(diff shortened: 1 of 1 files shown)


jrsquared:
👍

erikmunson:
Can you help me understand the use case here? I imagine static methods work fine if they're only being ref'd inside the overridable or overriding component itself things work ok — is this to make referencing static methods work from outside the component?

christopherthielen:
The use case is all usage of static methods anywhere, including inside the class itself.   

`@Overridable` replaces the class reference with a different class, so any attempt to access static members results in something like this:  `Uncaught TypeError: CapacityDetailsSection_1.resizeServerGroup is not a function`

This class:

```js
@Overridable()
export class Foo {
  static bar = "bar";
  constructor() {
    console.log(Foo.bar)
  }
}
```

roughly translates to:

```js
var Foo = overridableComponent((function () {
  function RawFoo() {
    console.log(Foo.bar);  // Foo references the variable above
  }
  RawFoo.bar = "bar";
  return RawFoo;
})());
```

Even within the class, Foo is dereferenced from it's final definition.

erikmunson:
aha, that's the piece I was missing. thanks.

PR #4975, opened 2018-03-08
fix(pipeline_templates): prevent [object Object] when object variable's defaultValue is provided

sbwsg:


app/scripts/modules/core/src/pipeline/config/templates/configurePipelineTemplateModal.controller.ts
```diff
@@ -2,7 +2,7 @@ import { module, IController, IScope, IHttpPromiseCallbackArg, IPromise } from '
 import { IModalInstanceService } from 'angular-ui-bootstrap';
 import { load, dump } from 'js-yaml';
 import { BindAll } from 'lodash-decorators';
-import { without, chain, has } from 'lodash';
+import { without, chain, has, isString, trim } from 'lodash';
```
sbwsg: remove

PR #4978, opened 2018-03-08
feat(core/presentation): Add margin selector for flex-container-h layout

christopherthielen:


christopherthielen:
@archana-s I'm punting on refactoring the flex layout selectors for now.  I'd like to spend an hour with you and @erikmunson to identify the common layouts and choose some better class names.

PR #4989, opened 2018-03-13
feat(core/pipeline): Allow overriding of default timeout for "Script stage" from UI

christopherthielen:
@emjburns


emjburns:
LGTM

PR #5001, opened 2018-03-14
feat(amazon/serverGroup): Change default capacity constraint to off

erikmunson:
Since adding a capacity constraint option for server group resizes, we've realized that for the majority of ad-hoc resizes capacity constraints are not a helpful protection. In the worst case, this can lead to urgent resizes being delayed due to the initial attempt failing and the user needing to go back and retry with capacity constraints turned off.

For now, the fastest path to fixing this problem is to change the default state so that capacity constraints are opt-in. Going forward, there may be more we can do to simplify along the lines of better messaging and/or allowing users to choose a single capacity setting they'd like to affect (i.e. only modify `desired` while leaving other processes free to modify `min`/`max`).

cc @emjburns


emjburns:
@erikmunson Thanks!

PR #5014, opened 2018-03-16
feat(provider/kubernetes): trim runJob logs from executions until explicitly requested

sbwsg:
Run Jobs that generate a large amount of log output can dramatically slow down loading of the pipeline executions screen in deck.

This PR sends an `expand` query param when fetching executions history and, by default, sets it not to fetch logs.  Fetching console output is now done when the user clicks on the `Console Output` link in the RubJob stage details.

see related PRs for gate https://github.com/spinnaker/gate/pull/515 and orca https://github.com/spinnaker/orca/pull/2061

Here's what a loading runJob log looks like:

load_console_output

### DIFF
--- a/app/scripts/modules/core/src/pipeline/service/execution.service.ts
+++ b/app/scripts/modules/core/src/pipeline/service/execution.service.ts
@@ -33,11 +33,11 @@ export class ExecutionService {
       return this.getFilteredExecutions(applicationName, this.activeStatuses, this.runningLimit);
     }
 
-    private getFilteredExecutions(applicationName: string, statuses: string[], limit: number, pipelineConfigIds: string[] = null): IPromise<IExecution[]> {
+    private getFilteredExecutions(applicationName: string, statuses: string[], limit: number, pipelineConfigIds: string[] = null, expand = false): IPromise<IExecution[]> {
         const statusString = statuses.map((status) => status.toUpperCase()).join(',') || null;
         const call = pipelineConfigIds ?
           this.API.all('executions').getList({ limit, pipelineConfigIds, statuses }) :
-          this.API.one('applications', applicationName).all('pipelines').getList({ limit, statuses: statusString, pipelineConfigIds });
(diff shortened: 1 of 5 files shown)


ethanfrogers:
that makes so much more sense! thanks @sbwsg.

app/scripts/modules/core/src/pipeline/service/execution.service.ts
```diff
@@ -33,11 +33,11 @@ export class ExecutionService {
       return this.getFilteredExecutions(applicationName, this.activeStatuses, this.runningLimit);
     }
 
-    private getFilteredExecutions(applicationName: string, statuses: string[], limit: number, pipelineConfigIds: string[] = null): IPromise<IExecution[]> {
+    private getFilteredExecutions(applicationName: string, statuses: string[], limit: number, pipelineConfigIds: string[] = null, expand = false): IPromise<IExecution[]> {
```
ezimanyi: Is there a reason the new parameter doesn't have an explicit type?

app/scripts/modules/core/src/pipeline/service/execution.service.ts
```diff
@@ -55,17 +55,17 @@ export class ExecutionService {
    * application will be used to correlate and filter the retrieved executions to only include those pipelines
    * @return {<IExecution[]>}
    */
-    public getExecutions(applicationName: string, application: Application = null): IPromise<IExecution[]> {
+    public getExecutions(applicationName: string, application: Application = null, expand = false): IPromise<IExecution[]> {
```
ezimanyi: Same question as above about whether this should have an explicit type

app/scripts/modules/core/src/pipeline/service/execution.service.ts
sbwsg: typescript infers the type when it's trivial to do so from the default value (and our linter will complain if the annotation is added needlessly).

ezimanyi: 👍

PR #5021, opened 2018-03-19
feat(artifacts): Support embedded/base64 artifact type

sbwsg:
fixes https://github.com/spinnaker/spinnaker/issues/2526

default_artifact_base64

With invalid input:

base64_latin1_error

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
@@ -52,7 +52,7 @@ class ArtifactCtrl implements IController {
 
     if (artifactKindConfig.length) {
       const config = artifactKindConfig[0];
-      const template: Element = config.template as any;
+      const template: string = config.template;
       this.description = config.description;
 
       const ctrl = config.controller;
--- a/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/base64/base64.artifact.ts
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/base64/base64.artifact.ts
@@ -1,0 +1,46 @@
+import { module } from 'angular';
+
+import { PIPELINE_CONFIG_PROVIDER } from 'core/pipeline/config/pipelineConfigProvider';
+import { IArtifact } from 'core/domain/IArtifact';
+import { PipelineConfigProvider } from 'core/pipeline';
+
+import './base64.artifact.less';
+
+export const BASE64_ARTIFACT = 'spinnaker.core.pipeline.trigger.artifact.base64';
+module(BASE64_ARTIFACT, [
+  PIPELINE_CONFIG_PROVIDER,
+]).config((pipelineConfigProvider: PipelineConfigProvider) => {
+  pipelineConfigProvider.registerArtifactKind({
+    label: 'Base64',
+    description: 'An artifact that includes its referenced resource as part of its payload.',
+    key: 'base64',
+    isDefault: false,
+    isMatch: true,
+    controller(artifact: IArtifact) {
+      'ngInject';
+      this.artifact = artifact;
+      this.artifact.type = 'embedded/base64';
+    },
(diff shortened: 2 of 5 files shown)


app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
```diff
@@ -52,7 +52,7 @@ class ArtifactCtrl implements IController {
 
     if (artifactKindConfig.length) {
       const config = artifactKindConfig[0];
-      const template: Element = config.template as any;
+      const template: string = config.template as any;
```
danielpeach: Isn't the type of `template` on `IArtifactKindConfig` a `string`? Should be able to drop `as any`.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/base64/base64.artifact.ts
```diff
@@ -0,0 +1,87 @@
...
+      Name
+    </label>
+    <div class="col-md-8">
+      <input type="text"
+             placeholder="base64-artifact"
+             class="form-control input-sm"
+             ng-model="ctrl.artifact.name" />
+    </div>
+  </div>
+  <div ng-if="ctrl.encodeDecodeError != null" class="form-group row">
```
danielpeach: In most of the spots where we surface this kind of error (see the Pipeline JSON edit modal), we put the error below the text box.

danielpeach:
Nice. 

- Is there any reasonable way to drop the "Failed to execute 'btoa' on 'Window'" part of the error message without parsing the error message? Users probably don't care how we converted their string to base64.
- Someone can disagree with me, but I wonder how useful the Base64 text box is - it's probably just noise in most cases, and users might be better served by a tooltip and/or a copy-to-clipboard icon.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/base64/base64.artifact.ts
```diff
@@ -0,0 +1,87 @@
+import { module } from 'angular';
+
+import { PIPELINE_CONFIG_PROVIDER } from 'core/pipeline/config/pipelineConfigProvider';
+import { IArtifact } from 'core/domain/IArtifact';
+import { PipelineConfigProvider } from 'core/pipeline';
+
+import './base64.artifact.less';
+
+export const BASE64_ARTIFACT = 'spinnaker.core.pipeline.trigger.github.artifact';
```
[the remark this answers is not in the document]
sbwsg: whoops, thankyou!

sbwsg:
@ezimanyi yes indeed, that work was already done here: https://github.com/spinnaker/clouddriver/pull/2416/files

@dpeach for 1) I've added an explicit error message for when a char is outside the latin1 range so that it wont have the "Failed to executed 'btoa' on 'window'" part.  I expect that will be the most common error when converting between base64 / decoded content.  All other errors will be printed to screen with the entire error message.  Short of parsing it I can't think of a way to remove that string prefix.  And for 2) I agree about the lack of utility of the base64 text content being in the UI.  I've replaced the `<pre>` with a copy-to-clipboard icon.

sbwsg:
@lwander I've enabled the artifact as a matchArtifact.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/base64/base64.artifact.ts
```diff
@@ -0,0 +1,46 @@
...
+    <div class="col-md-8">
+      <input type="text"
+             placeholder="base64-artifact"
+             class="form-control input-sm"
+             ng-model="ctrl.artifact.name" />
+    </div>
+  </div>
+  <div ng-if="ctrl.encodeDecodeError" class="form-group row">
+    <div class="col-md-12 error-message">
+      Error encoding/decoding artifact content: {{ ctrl.encodeDecodeError }}
```
lwander: Why does the match artifact have encoding errors?

PR #5022, opened 2018-03-19
 fix(amazon/loadBalancers): Disable editing target group names and disallow duplicate names

jrsquared:


christopherthielen:
seems fine

PR #5048, opened 2018-03-21
fix(core/entityTag): Fix notifications popover re-mounting during each render.

christopherthielen:
Put the SFC as an instance variable on the class so it doesn't get re-created during each render.


erikmunson:
👍looks great

PR #5054, opened 2018-03-23
feat(titus): Support specifying a percentage of instances to relaunch

ajordens:


ajordens:
image

app/scripts/modules/titus/src/migration/titusMigrationConfigurer.component.html
```diff
@@ -10,11 +10,21 @@
     <div class="config-field">
       <label>
         Batch size
-        <help-field key="strategy.rollingPush.concurrentRelaunches"></help-field>
+        <help-field key="strategy.rollingPush.concurrentRelaunches.migration"></help-field>
       </label>
       <input class="form-control input-sm" type="number" min="1"
              ng-model="$ctrl.config.config.concurrentRelaunches"
              ng-change="$ctrl.configChanged()" />
+      <ui-select ng-model="$ctrl.config.config.concurrentRelaunchesAsPercentage"
```
ajordens: Almost certainly a more elegant way of doing this

app/scripts/modules/titus/src/migration/titusMigrationConfigurer.component.html
```diff
@@ -23,7 +33,7 @@
         </label>
         <input class="form-control input-sm" type="number" min="0"
                ng-model="$ctrl.config.config.waitTime"
-               ng-change="$ctrl.configChanged()"/> seconds between updates
+               ng-change="$ctrl.configChanged()"/> <span style="font-size: 12px">seconds between updates</span>
```
ajordens: image

Quick hack to fix this wrapping?

sargun:
Does it make sense to add a validator ensuring the percentage is > 0 and =< 100

ajordens:
Fair point @sargun.

@anotherchrisberry -- You probably have something better than `max="{{ $ctrl.config.config.concurrentRelaunchesAsPercentage ? 100 : 9999 }}"`

image

JavaScript++

PR #5060, opened 2018-03-24
fix(core-package): Add explicit do-not-ignores to .npmignore so yalc doesn't exclude them

christopherthielen:
Yalc ignores a certain set of files like LICENSE or HISTORY (I'm not sure why) and this PR explicitly unignores them.

See: https://github.com/whitecolor/yalc/issues/8


erikmunson:
🤔 weird. lgtm

PR #5077, opened 2018-03-27
feat(aws): Support rollback of a disabled server group

ajordens:
A disabled server group can be rolled back iff at least one enabled
server group exists in the same cluster.

The largest enabled server group in the cluster will will be selected as
the rollback source.

depends on spinnaker/clouddriver#2455

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/details/AmazonServerGroupActions.tsx
+++ b/app/scripts/modules/amazon/src/serverGroup/details/AmazonServerGroupActions.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { BindAll } from 'lodash-decorators';
 import { Dropdown, Tooltip } from 'react-bootstrap';
-import { get, find, filter } from 'lodash';
+import { get, find, filter, orderBy } from 'lodash';
 
 import { IOwnerOption, IServerGroupActionsProps, IServerGroupJob, NgReact, ReactInjector, SETTINGS } from '@spinnaker/core';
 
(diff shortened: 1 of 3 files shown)


ajordens:
There has been a request for rollback to be more prominently featured on both the enabled and disabled server groups.

Also considering other ways to more prominently feature rollbacks. There's been some discussion of a "button" on the cluster view ... _or_ maybe we pull the `Rollback` action out of the drop-down and feature it more explicitly on the sidebar. 

Thoughts @anotherchrisberry?

app/scripts/modules/amazon/src/serverGroup/details/AmazonServerGroupActions.tsx
```diff
@@ -26,6 +26,24 @@ export class AmazonServerGroupActions extends React.Component<IAmazonServerGroup
...
+  private isRollbackEnabled(): boolean {
+    const { app, serverGroup } = this.props;
+
+    if (!serverGroup.isDisabled) {
+      // enabled server groups are always a candidate for rollback
+      return true;
+    }
+
+    // if the server group selected for rollback is disabled, ensure that at least one enabled server group exists
+    return app.getDataSource('serverGroups').data.filter((g: IAmazonServerGroup) =>
```
anotherchrisberry: instead of `.filter`, use `.some` - then you can skip the `length` check below

app/scripts/modules/amazon/src/serverGroup/details/AmazonServerGroupActions.tsx
```diff
@@ -140,23 +158,45 @@ export class AmazonServerGroupActions extends React.Component<IAmazonServerGroup
   };
 
   private rollbackServerGroup(): void {
-    const { app, serverGroup } = this.props;
+    const app = this.props.app;
```
anotherchrisberry: It's okay to use destructuring to get a single property out of another property (not necessary but totally fine)

anotherchrisberry:
chuck_good

PR #5080, opened 2018-03-27
feat(core): move from provider version UI implementation to skins

danielpeach:
Decouples a UI implementation (a "skin") from the provider version returned by an account.

### DIFF
--- a/app/scripts/modules/core/src/cloudProvider/cloudProvider.registry.ts
+++ b/app/scripts/modules/core/src/cloudProvider/cloudProvider.registry.ts
@@ -12,7 +12,8 @@ export interface ICloudProviderConfig {
   name: string;
   logo?: ICloudProviderLogo;
   providerVersion?: string;
-  defaultVersion?: boolean;
+  skin?: string;
+  defaultSkin?: boolean;
   [attribute: string]: any;
 }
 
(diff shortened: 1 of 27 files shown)


danielpeach:
Depends on https://github.com/spinnaker/gate/pull/522 and https://github.com/spinnaker/clouddriver/pull/2459

app/scripts/modules/core/src/cloudProvider/cloudProvider.registry.ts
```diff
@@ -25,30 +26,30 @@ class Providers {
...
+      this.providers.find(p => p.cloudProvider === cloudProvider && p.config.skin === config.skin)
     ).concat([{ cloudProvider, config }]);
   }
 
-  public get(cloudProvider: string, providerVersion?: string): ICloudProviderConfig {
-    if (providerVersion) {
-      const provider = this.providers.find(p => p.cloudProvider === cloudProvider && p.config.providerVersion === providerVersion);
-      return provider ? provider.config : this.getDefaultConfig(cloudProvider);
+  public get(cloudProvider: string, skin?: string): ICloudProviderConfig {
+    if (skin) {
```
lwander: Why are we looking things up by their `skin`? Isn't that a property of what you want to find? Or is it overloading `providerVersion`?

app/scripts/modules/kubernetes/src/v1/kubernetes.v1.module.ts
```diff
@@ -59,7 +59,8 @@ module(KUBERNETES_V1_MODULE, [
     cloudProviderRegistryProvider.registerProvider('kubernetes', {
       name: 'Kubernetes',
       providerVersion: 'v1',
-      defaultVersion: true,
+      skin: 'v1',
```
lwander: So skin and provider version are the same thing now? Why have both?

PR #5083, opened 2018-03-27
fix(artifacts): Set the name field on the default github artifacts

ezimanyi:


ezimanyi:
Intended to address spinnaker/spinnaker#2572

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/github/defaultGithub.artifact.ts
```diff
@@ -18,6 +18,14 @@ module(DEFAULT_GITHUB_ARTIFACT, [
       'ngInject';
       this.artifact = artifact;
       this.artifact.type = 'github/file';
+      const pathRegex = new RegExp('/repos/[^/]*/[^/]*/contents/(.*)$');
```
sbwsg: minor nit: this would currently match `/repos///contents/' and give a name of an empty string.  should the user be warned if they enter something technically correct but practically invalid?

ezimanyi: Agreed it's not ideal, though I'm conflicted about how much to surface validation issues to the user.  There's a more in-depth discussion on the issue linked above around how much the name should be tied to the reference that I think we should think about more before making user-facing changes.

As an alternative to validation, we could also consider just showing the user the artifact name that's extracted and letting them change it if desired.  Whichever approach we use we should probably think about having it apply to all artifact types.

So I'm inclined to leave this as is as this is silently fixing a but that effectively makes default Github artifacts never work---but to consider the input validation question in the context of that broader discussion (which should continue beyond this quick fix).

sbwsg: totally agree, let's tackle the broader question in future PRs.  :shipit:

PR #5085, opened 2018-03-28
feat(kubernetes): use v2 load balancer and security group transformers

danielpeach:
The v2 provider was implicitly using the v1 provider's transformers, which was leading to some odd behavior.

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/loadBalancer/transformer.ts
+++ b/app/scripts/modules/kubernetes/src/v2/loadBalancer/transformer.ts
@@ -1,0 +1,53 @@
+import { module, IQService, IPromise } from 'angular';
+import { chain, camelCase } from 'lodash';
+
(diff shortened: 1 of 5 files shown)


app/scripts/modules/kubernetes/src/v2/loadBalancer/transformer.ts
```diff
@@ -0,0 +1,53 @@
...
+    return this.$q.resolve(loadBalancer);
+  }
+
+  private buildInstanceCounts(serverGroups: IServerGroup[]): IInstanceCounts {
+    const instanceCounts = chain(serverGroups)
+      .map(serverGroup => serverGroup.instances)
+      .flatten()
+      .reduce(
+        (acc: IInstanceCounts, instance: any) => {
+          acc[camelCase(instance.health.state) as keyof IInstanceCounts]++;
```
lwander: heh

PR #5086, opened 2018-03-28
style(amazon,appengine,azure,cloudfoundry,core,dcos,google,kubernetes,openstack,titus): IcoMoon updates 

archana-s:
Moved to icomoon for autogenerating the json file and CSS for custom icons we use.

The two custom icons we use are the one for load balancer and a special kind of spinner that is used for inplace edits. I removed the SVGs for these icons and use the selection.json from icomoon. I used icomoon to autogenerate the icon fonts for these two icons. If we add more custom icons, we need to go icomoon.app and the selection.json thats in this PR can be used to open all our existing icons and add new ones. I will update documentation on how to add new custom icons soon. 

Reference: https://icomoon.io/app/#/select


anotherchrisberry:
chuck_good

ajordens:
@archana-s Not a big deal for this PR ... but in the future I suspect you could get away with `style(*): IcoMoon updates`.

PR #5098, opened 2018-03-29
Manually filter out removed artifact IDs from stages

sbwsg:
When an expected artifact is deleted any references to it continue to live on in stages' JSON.
When the UI attempts to render artifact info based on those references it's unable to find the correlating artifact objects (since they're deleted) and so surfaces the error to the DOM in odd ways.

See https://github.com/spinnaker/spinnaker/issues/2562

This is a crummy fix because it hard-codes the locations of all the places artifacts might live in the `removeExpectedArtifact` method.  It will need to be updated every time we add a new field on a stage that will store reference to an artifact.  I'm not sure what we could do as a better fix for this though.  Ideas welcome.


lwander:
Ahh this will be very helpful

PR #5107, opened 2018-04-02
fix(artifacts): deleting expected artifact removes stale artifact references from stage

sbwsg:
This PR adds a small service that allows stages to register walker functions that know how to traverse that stage looking for artifact references.

The k8s deploy manifest stage and GCE clone stage both use this new service to register walker functions for themselves.  The trigger directive's `removeExpectedArtifact` method then utilizes these walkers to visit and remove stale artifact references from those stages when an expected artifact is deleted from a pipeline.

This approach is general, feels a bit convoluted, but works.  Very open to better approaches here.

Fixes https://github.com/spinnaker/spinnaker/issues/2562

### Before:

Clicking the trash can icon to delete an expected artifact:

Resulted in this side effect on deploy manifest stages:

37721390-b151420e-2cff-11e8-80a5-1df99a14197d

### After

Deleting an expected artifact now manifests as this:

### DIFF
--- a/app/scripts/modules/core/src/artifact/ArtifactReferenceService.ts
+++ b/app/scripts/modules/core/src/artifact/ArtifactReferenceService.ts
@@ -1,0 +1,57 @@
+import { module } from 'angular';
+import { IStage } from 'core/domain'
+import { isEmpty, get } from 'lodash';
+
+type SupportedStage = 'stage';
+
+interface IWalker {
+  (refContainer: any): (string|number)[][];
+}
+
+interface IReference {
+  category: SupportedStage;
+  walker: IWalker;
+}
+
+export class ArtifactReferenceServiceProvider {
+  private references: IReference[] = [];
+
+  public $get() {
+    return this;
+  }
+
+  public registerReference(category: SupportedStage, walker: any) {
+    this.references.push({ category, walker });
+  }
+
+  public removeReferenceFromStages(reference: string, stages: IStage[]) {
+    (stages || []).forEach(stage => {
+      this.references.forEach(ref => {
+        const paths: (string|number)[][] = ref.walker(stage).filter(path => !isEmpty(path));
+        paths.map(p => p.slice(0)).forEach(path => {
+          let tail = path.pop();
+          let obj = stage;
+          if (path.length > 0) {
+            obj = get(stage, path);
+          }
+          if (Array.isArray(obj[tail])) {
+            obj = obj[tail];
+            tail = obj.indexOf(reference);
+          }
+          if (obj[tail] !== reference) {
+            return;
+          }
+          if (Array.isArray(obj)) {
+            obj.splice(tail as number, 1);
+          } else {
+            delete obj[tail];
+          }
+        });
+      });
+    });
+  }
+}
+
+export const ARTIFACT_REFERENCE_SERVICE_PROVIDER = 'spinnaker.core.artifacts.referenceServiceProvider';
+module(ARTIFACT_REFERENCE_SERVICE_PROVIDER, [])
+  .provider('artifactReferenceService', ArtifactReferenceServiceProvider);
--- a/app/scripts/modules/core/src/pipeline/config/triggers/triggers.directive.js
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/triggers.directive.js
@@ -2,11 +2,13 @@
 
 import { PIPELINE_CONFIG_PROVIDER } from 'core/pipeline/config/pipelineConfigProvider';
 import { UUIDGenerator } from 'core/utils/uuid.service';
+import { ARTIFACT_REFERENCE_SERVICE_PROVIDER } from 'core/artifact/ArtifactReferenceService';
 
 const angular = require('angular');
 
 module.exports = angular.module('spinnaker.core.pipeline.config.trigger.triggersDirective', [
   PIPELINE_CONFIG_PROVIDER,
+  ARTIFACT_REFERENCE_SERVICE_PROVIDER,
 ])
   .directive('triggers', function() {
     return {
@@ -20,7 +22,7 @@ module.exports = angular.module('spinnaker.core.pipeline.config.trigger.triggers
       templateUrl: require('./triggers.html')
     };
   })
-  .controller('triggersCtrl', function($scope, pipelineConfig) {
+  .controller('triggersCtrl', function($scope, pipelineConfig, artifactReferenceService) {
     this.addTrigger = function() {
       var triggerTypes = pipelineConfig.getTriggerTypes(),
           newTrigger = {enabled: true};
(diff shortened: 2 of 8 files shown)


app/scripts/modules/core/src/pipeline/config/triggers/triggers.directive.js
```diff
@@ -2,11 +2,13 @@
 
 import { PIPELINE_CONFIG_PROVIDER } from 'core/pipeline/config/pipelineConfigProvider';
 import { UUIDGenerator } from 'core/utils/uuid.service';
+import { ARTIFACT_REFERENCE_SERVICE_PROVIDER } from '@spinnaker/core';
```
danielpeach: use `core/...` instead of `@spinnaker/core` from inside the core module.

app/scripts/modules/core/src/artifact/ArtifactReferenceService.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  public $get() {
+    return this;
+  }
+
+  public registerReference(category: string, walker: any) {
+    this.references.push({ category, walker });
+  }
+
+  public walkReferences(category: string, instance: any, visitor: any) {
```
danielpeach: I don't see a super strong reason for each `visitor` to be dynamic, rather than statically defined on the `ArtifactReferenceService` itself (especially since you only have one `visitor` right now), but it's up to you. You could have each `Walker` return a `[object, key]` tuple, and a method on `ArtifactReferenceService` could pull the key from the object.

lwander:
Does this have to be registered everywhere expected artifacts are used?

sbwsg:
I should double-check.  Which other places are you thinking of here?

app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestStage.ts
```diff
@@ -35,5 +38,18 @@ module(KUBERNETES_DEPLOY_MANIFEST_STAGE, [
         { type: 'requiredField', fieldName: 'moniker.cluster', fieldLabel: 'Cluster' }
       ],
     });
+
+    artifactReferenceServiceProvider.registerReference('stage', (obj: IStage, fn: any) => {
+      if (obj.type === 'deployManifest') {
```
danielpeach: I wonder if `Walker` should retain the `stage` or `trigger` type or name. You could run a `Walker`s callback only if the stage or trigger matches the type of the `Walker`, rather than relying on these kinds of `if` blocks to decide if the callback is relevant to the stage.

app/scripts/modules/core/src/artifact/ArtifactReferenceService.ts
```diff
@@ -0,0 +1,32 @@
+import { module } from 'angular';
+
+interface Walker {
```
danielpeach: Please use `IWalker` and `IReference` - just convention for Deck.

lwander:
Any stage that declares "produces artifacts", like "run job"

sbwsg:
Yeah I'll need to take care of those as well, made slightly tricky by the fact that the produces artifact component doesn't seem to have access to the pipeline's complete list of stages.  I'll see if I can find a way to fix that as part of this PR.

app/scripts/modules/core/src/pipeline/config/triggers/triggers.directive.js
```diff
@@ -51,8 +53,25 @@ module.exports = angular.module('spinnaker.core.pipeline.config.trigger.triggers
...
+      if (!pipeline.stages || pipeline.stages.length === 0) {
+        return;
+      }
+
+      pipeline.stages.forEach(stage => {
+        artifactReferenceService.walkReferences('stage', stage, (obj, key) => {
+          if (Array.isArray(obj)) {
+            obj.splice(key, 1);
+          } else {
+            delete obj[key];
```
ezimanyi: This looks like it will delete the selected artifact regardless of whether it's the artifact we deleted

sbwsg: oy.  that'll teach me for only testing with a single expected artifact.  thankyou.

sbwsg:
(1) The registration only happens once for each stage type on application start rather than every time a stage is visited in the UI.  The `.config()` angular module method is used for this.  At least this is my understanding from some quick `console.log`s strewn about the code.

(2) A view-driven approach for this was one that I initially considered but rejected for the reasons you identify.  Relying on specific UI components to perform data cleanup seems like a potential footgun as the UI around artifacts becomes richer.

(3) I hadn't thought of this and will take a shot at adopting that approach instead.  Cheers!

sbwsg:
@ezimanyi regarding (3) - I don't think the lifecycle of the controller classes that call `getExpectedArtifactsAvailableToStage()` fit with the requirements of artifact reference removal.  They are instantiated and destroyed in lockstep with the views they control.  More concretely, here is the control flow if we introduced observables to the controllers as they stand now:

1. user defines expected artifact in config view
2. user creates deploy stage
4. user opens deploy stage for editing
3. deploy stage controller is instantiated, calls `getExpectedArtifactsAvailableToStage` and begins observing changes on artifacts
4. user goes back to config screen
5. the deploy stage controller is destroyed along with its view component, firing the `$destroy` event on the controller's scope.
6. at this point we could a) stop observing in response to the event or b) continue observing and ignore the event.

If we do (a) then the stage is no longer observing changes in artifacts so cannot respond to deletion when the user edits the config screen.  If we do (b) then we can continue listening for artifact deletion but we've created a reference that we'll need to somehow manually clean up or risk creating a memory leak as the observer will live on in perpetuity.

So if we go down the road of manually cleaning up the references then we'll need to figure out the machinery of exactly when and how to do that.  I'm not sure how much we've gained over the approach we've taken in this PR so far.

I might be over-thinking this though.  Is the approach still preferable?  What are your thoughts on the above?

app/scripts/modules/core/src/artifact/ArtifactReferenceService.ts
sbwsg: `IWalker`s now return a list of paths into the stage and the dynamic visitor has been removed.  The register API feels less complicated in the trivial case.  The google clone stage still has to do a bunch of work to construct its list of paths though.

sbwsg:
@lwander this PR now takes into account the produces-artifacts component as well

ezimanyi:
@sbwsg : I see from your explanation how making an observable doesn't actually gain us that much, so wouldn't really be any less complex than the current implementation.  This somehow still feels more complex than it needs to be (as you noted above) but I'm not sure I can think of something better after thinking about it for a while so this may be necessary complexity.

sbwsg:
It definitely feels like this problem is screaming out for some normalization in the database sense - a lookup table from stage objects to artifact references that can then have those references trivially removed.

This would require refactoring every point in all the services that touch artifact references though.  Not sure if that's a worthy tradeoff either.

PR #5126, opened 2018-04-06
feat(provider/appengine): add halyard feature flag for container image url deployments

sbwsg:


sbwsg:
I think so.  We don't have any integration test coverage for this feature and it's arguably not been exercised enough to shake out issues.

PR #5140, opened 2018-04-09
chore(core): bump package to 0.0.187

danielpeach:


danielpeach:
fix(core): skin selection (#5139)

PR #5145, opened 2018-04-10
feat(tagging): Select which upstream stages to include in image search

jervi:
When tagging an image, the default behaviour is to look for images in all ancestor stages. In some cases, this is not the desired behaviour. This commit makes it possible to optionally define a list of stages that will be considered in the tagging stage.
This commit corresponds with the following orca PR: https://github.com/spinnaker/orca/pull/2138

image


jervi:
@spinnaker/reviewers PTAL

anotherchrisberry:
A couple of thoughts:
* I don't think `name` is the best field to use here, as it's a field that anyone can change. It's probably better to use `refId`.
* I think the help text is really good. Can we display it under label, so it's more clear that the checkboxes are optional? Something like
```
Stages (optional)
(small text) Limit which upstream stages will be considered when locating AMI's to tag. If left unchecked, AMIs generated by any upstream stage will be tagged.
          [ ] stage 1
          [ ] stage 2
          [ ] stage 3
```

Also, do we know which types of stages might generate an image? I suspect there are just a handful of stage types that would make sense to show here, and if there is just one, it might make sense to not show this section at all...

jervi:
I agree on the `name` field. The reason I went with `name` was that I couldn't figure out a way to have different values for the displayed label and the checkbox value in the `<checklist>` directive.
I'll try to fix it anyway, and will go through the other suggestions as well.

jervi:
Sorry about the slow progress, I'm on paternity leave, so I've had some other priorities 👶 
The first commit now enhances the `checklist` directive to support key/values in the form of a `Map`, instead of just plain `string` arrays. The implementation is backward compatible, meaning `items` support both array and map.
The second commit addresses the rest of the feedback:
- Changed the model to use `refId` instead of stage name
- Help text is always visible (I had to implement a small change in the stageConfigField directive to support this)
- Only displaying stages that actually can produce AMI/GCE image
skjermbilde 2018-05-02 kl 23 01 37

jervi:
@spinnaker/reviewers PTA(nother)L

jervi:
@anotherchrisberry I've updated this and the corresponding Orca PR now. Finally back from paternity leave :)

jrsquared:
LGTM

jrsquared:
@jervi Actually, since you have two distinct commits in here, I'll let you update the PR so we can rebase and merge the two in separately.

PR #5160, opened 2018-04-11
fix(kubernetes) remove upper limit for CPU Target on resize

kevinawoo:
same as https://github.com/spinnaker/deck/pull/5159, but for `release-1.6.x` branch.


lwander:
If this is intended as a patch, it needs to go into `spinnaker:release-1.6.x`, right now it's still going into `spinnaker:master`.

kevinawoo:
derp derp. I fixed it! Thanks for the 👀s

lwander:
Sorry one more thing, can you tag this as a "fix" given that it's technically fixing a bug, and going into a patch release?

lwander:
Thanks!

PR #5176, opened 2018-04-14
chore(core) fixed imports

kevinawoo:
webpack adds an alias for `@spinnaker/core` so that each module can be build separately and imported


jrsquared:
good find

PR #5181, opened 2018-04-16
feat(provider/kubernetes): run job node selector

ethanfrogers:
add support for run job node selector. looks like clouddriver already
supports this so we just need to expose it via the UI

fixes spinnaker/spinnaker#2553


ethanfrogers:
@sbwsg i don't mind porting this over to the 1.7 branch as well. just let me know!

sbwsg:
great!  feel free to open a PR against the `release-1.7.x` branch with this change

PR #5182, opened 2018-04-16
chore(google): bump npm package version

danielpeach:


app/scripts/modules/core/src/securityGroup/filter/SecurityGroupFilterService.ts
```diff
@@ -1,4 +1,3 @@
-import { module } from 'angular';
```
danielpeach: Linter error.

PR #5190, opened 2018-04-17
feat(deploy/ecs): add ecs support

melbit-michaelw:
Adding ECS functionality into Deck. There will also be a halyard Pull request associated with this work.

### DIFF
--- a/halconfig/settings.js
+++ b/halconfig/settings.js
@@ -74,6 +74,11 @@ var dcos = {
     account: '{%dcos.default.account%}'
   }
 };
+var ecs = {
+  defaults: {
+    account: '{%ecs.default.account%}'
+  }
+};
 var entityTagsEnabled = false;
 var netflixMode = false;
 var notificationsEnabled = '{%notifications.enabled%}' === 'true';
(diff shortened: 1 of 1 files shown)


halconfig/settings.js
```diff
@@ -110,6 +110,7 @@ window.spinnakerSettings = {
       },
       useAmiBlockDeviceMappings: false,
     },
+    ecs: ecs,
```
sbwsg: Where is the `ecs` variable defined?  This appears to be a lint error.

BrunoCarrier: @melbit-michaelw other providers define defaults.  You probably want to do something like this:

```    ecs: {
      defaults: {
        account: 'test',
        region: 'us-east-1',
        iamRole: 'BaseIAMRole',
      },```

I am not sure where these defaults are used though

melbit-michaelw: @sbwsg , @BrunoCarrier Added the var definition for ecs, and changed the commit message to be fix based rather than feature.

sbwsg:
ok great ping me when it's readdy and i'll cherry-pick it into the 1.7 release branch.

melbit-michaelw:
It appears the git hooks were reformatting the code.. I've re-committed without enabling them, and now it looks as I expected. I'm not sure why they were a problem..

@sbwsg Ready for your review.

sbwsg:
lgtm

PR #5191, opened 2018-04-17
feat(provider/kubernetes): run job node selector

ethanfrogers:
add support for run job node selector. looks like clouddriver already
supports this so we just need to expose it via the UI

fixes spinnaker/spinnaker#2553


ethanfrogers:
@sbwsg PTAL

PR #5192, opened 2018-04-17
feat(artifacts) Let pipeline stages emit artifacts

avram:
See also https://github.com/spinnaker/orca/pull/2152


avram:
Wrong change.

PR #5195, opened 2018-04-17
feat(provider/kubernetes): v1 suport tolerations

ethanfrogers:
adds support for Pod tolerations for V1 server groups and run job.

### DIFF
--- a/app/scripts/modules/kubernetes/src/common/tolerations/tolerations.component.ts
+++ b/app/scripts/modules/kubernetes/src/common/tolerations/tolerations.component.ts
@@ -1,0 +1,114 @@
+import { IComponentOptions, IController, module } from 'angular';
+
+class KubernetesTolerationsController implements IController {
+  public tolerations: any;
+
+  public addToleration() {
+    if (!this.tolerations) {
+      this.tolerations = [];
+    }
+
+    this.tolerations.push({
+      key: '',
+      operator: 'Equal',
+      effect: 'NoSchedule',
+      value: '',
+    });
+  }
+
+  public removeToleration(index: any) {
+    this.tolerations.splice(index, 1);
+  }
+
+  public operators(): string[] {
+    return ['Exists', 'Equal'];
+  }
+
+  public effects(): string[] {
+    return ['NoSchedule', 'PreferNoSchedule', 'NoExecute'];
+  }
+}
+
+class KubernetesTolerationsComponent implements IComponentOptions {
+  public bindings: any = { tolerations: '=' };
+  public controller: any = KubernetesTolerationsController;
+  public controllerAs: any = 'ctrl';
+  public template = `
+  <form name="tolerations">
+    <div class="sm-label-left">Tolerations</div>
+    <table class="table table-condensed">
+      <tbody>
+        <tr ng-repeat="toleration in ctrl.tolerations">
+          <td>
+            <div class="form-group">
+                <div class="col-md-3 sm-label-right">
+                  Key
+                  <help-field key="kubernetes.serverGroup.tolerations.key"></help-field>
+                </div>
+                <div class="col-md-3">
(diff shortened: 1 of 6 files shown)


ethanfrogers:
### Add Toleration
image

### Details
image

ethanfrogers:
@sbwsg @danielpeach i want to tread `operators` and `effects` as constants but `eslint` is complaining that those aren't used in my terminal. any suggestions on how to work around that?

ethanfrogers:
@benjaminws FYI

sbwsg:
Given that the fields are only used inside the template string I think the only way to workaround would be to annotate the `const` declarations with lint comments targeting the no-unused-* rules https://eslint.org/docs/rules/no-unused-vars or https://palantir.github.io/tslint/rules/no-unused-variable/ (will vary depending on which rule's warning you're seeing in the terminal).

jrsquared:
If you convert it all to React, then you won't have to hack around it being used only in the template! 😉

app/scripts/modules/kubernetes/src/common/tolerations/tolerations.component.ts
```diff
@@ -0,0 +1,108 @@
+import { IComponentOptions, IController, module } from 'angular';
+
```
danielpeach: Would add 

```
type IOperator = 'Exists' | 'Equal';
type IEffect = 'NoSchedule' | 'PreferNoSchedule' | 'NoExecute';

interface IToleration {
   key: string;
   operator: IOperator;
   effect: IEffect;
   value: string;
}
```

then in the class:

```
public tolerations: ITolerations[];
public operatiors: IOperator[];
public effects: IEffect[];
```

app/scripts/modules/kubernetes/src/common/tolerations/tolerations.component.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.tolerations.push({
+      key: '',
+      operator: 'Equal',
+      effect: 'NoSchedule',
+      value: '',
+    });
+  }
+
+  public removeToleration(index: any) {
```
danielpeach: s/`index: any`/`index: number`

app/scripts/modules/kubernetes/src/common/tolerations/tolerations.component.ts
```diff
@@ -0,0 +1,108 @@
...
+    this.tolerations.splice(index, 1);
+  }
+}
+
+class KubernetesTolerationsComponent implements IComponentOptions {
+  public bindings: any = { tolerations: '=' };
+  public controller: any = KubernetesTolerationsController;
+  public controllerAs: any = 'ctrl';
```
danielpeach: Can omit `any`, `controllerAs` will be implicitly typed as a `string` from the assignment.

app/scripts/modules/kubernetes/src/help/kubernetes.help.ts
```diff
@@ -32,6 +32,11 @@ const helpContents: { [key: string]: string } = {
     '(Optional) Either a number or a percentage (1 vs. 10%) representing the maximum number of new pods to deploy during a rollout.',
   'kubernetes.serverGroup.deployment.maxUnavailable':
     '(Optional) Either a number or a percentage (1 vs. 10%) representing the maximum number of pods to destroy during a rollout.',
+  'kubernetes.serverGroup.tolerations.key': `Key is the taint key that the toleration applies to. Empty means match all taint keys. If the key is empty, operator must be Exists; this combination means to match all values and all keys.`,
```
danielpeach: Don't need the backtick syntax, but it's also not hurting anything.

ethanfrogers:
@jrsquared i mean...you aren't wrong. 🤣

ethanfrogers:
@sbwsg @jrsquared @danielpeach totally fair. i'll close and rewrite in react!

PR #5197, opened 2018-04-17
feat(provider/kubernetes): Surface specially named annotations in details UI

sbwsg:
Annotations on k8s resources with a key matching `(section-title).details.(html)?.spinnaker.io/(key-name)` will be used to populate a section in the details panel for that resource.

Hyphens in the section title and key name are replaced with spaces.

The content of the annotation is interpolated using the resource (ala instance links) so that information can be dynamically set with `{{ }}` template literals.

Finally, prior to reaching the DOM, the content is pushed through DOMPurify.sanitize.

Fixes https://github.com/spinnaker/spinnaker/issues/2529 and, by extension, provides support for https://github.com/spinnaker/spinnaker/issues/2528

screen shot 2018-04-17 at 1 59 07 pm

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/annotationCustomSections.component.ts
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/annotationCustomSections.component.ts
@@ -1,0 +1,110 @@
+import { IComponentOptions, IController, ISCEService, IInterpolateService, module } from 'angular';
+import { get, partition } from 'lodash';
+import * as DOMPurify from 'dompurify';
+
+interface IAnnotationsMap {
+  [key: string]: string;
+}
+
+interface ICustomSection {
+  title: string;
+  key: string;
+  isHtml: boolean;
+  content: string;
+}
+
+interface ICustomSectionMap {
+  [key: string]: ICustomSection[];
+}
+
+class KubernetesAnnotationCustomSections implements IController {
+  private resource: any;
+  public manifest: any;
+  public customSections: ICustomSectionMap;
+
+  constructor(private $sce: ISCEService, private $interpolate: IInterpolateService) {
+    const annotations: IAnnotationsMap = get(this, ['manifest', 'metadata', 'annotations']);
+    if (annotations == null) {
+      return;
+    }
+    this.populateCustomSections(annotations);
+  }
+
+  private populateCustomSections(annotations: IAnnotationsMap) {
+    const customSections: ICustomSectionMap = Object.keys(annotations).reduce(
+      (memo: ICustomSectionMap, annotationKey: string) => {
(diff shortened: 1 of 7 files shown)


app/scripts/modules/kubernetes/src/v2/serverGroupManager/details/details.html
```diff
@@ -97,6 +97,8 @@ <h3 class="horizontal middle space-between flex-1" select-on-dbl-click>
       </ul>
     </collapsible-section>
 
+    <kubernetes-annotation-custom-sections manifest="ctrl.serverGroupManager.manifest" resource="ctrl.serverGroupManager"></kubernetes-annotation-custom-sections>
```
lwander: Can you add this to the instance details as well?

lwander:
Nice!

Before closing those issues could you create a new Guide in spinnaker.io?

image

We should probably cover how to write those annotations, and what set of variables are available to `{{ }}` in the templates

app/scripts/modules/kubernetes/src/v2/serverGroupManager/details/details.html
sbwsg: What `instance` are you referring to here?  I have added it to app/scripts/modules/kubernetes/src/v2/instance/details/details.html as part of this PR.  is there somewhere else to include it?

gardleopard:
This looks great!

app/scripts/modules/kubernetes/src/v2/manifest/annotationCustomSections.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    // Sort section contents such that text entries appear before HTML entries.
+    this.customSections = Object.keys(customSections).reduce((memo: ICustomSectionMap, sectionTitle: string) => {
+      const entriesHtmlText = partition(customSections[sectionTitle], section => section.isHtml);
+      memo[sectionTitle] = entriesHtmlText[1].concat(entriesHtmlText[0]);
+      return memo;
+    }, {});
+  }
+
+  private annotationToEntry(content: string, annotationKey: string): ICustomSection {
+    const keyParts = /([^.]+).details\.(html\.)?spinnaker\.io\/(.*)/.exec(annotationKey);
```
ezimanyi: The first period need to be escaped

sbwsg:
Thanks, fixed!

sbwsg:
Documentation for this feature in PR here: https://github.com/spinnaker/spinnaker.github.io/pull/686

app/scripts/modules/kubernetes/src/v2/serverGroupManager/details/details.html
sbwsg: ping

PR #5203, opened 2018-04-18
feat(provider/kubernetes): v1 suport tolerations

ethanfrogers:
adds support for Pod tolerations for V1 server groups and run job.


ethanfrogers:
@jrsquared @danielpeach PTAL. I converted the component to React. It's my first React component so there's a chance that i may be being overly verbose. Let me know if there's anything I could do better!

ethanfrogers:
#### Edit
image

#### Server Group Detail
image

app/scripts/modules/kubernetes/src/common/tolerations/TolerationsComponent.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+  private removeToleration(index: number): () => void {
+    return () => {
+      const tolerations = cloneDeep(this.state.tolerations);
+      tolerations.splice(index, 1);
+      this.props.onTolerationChange(tolerations);
+    };
+  }
+
+  private handleKeyChange(index: number): (event: any) => void {
```
danielpeach: These methods can be replaced by a single method in a TS-friendly way:

```
private handleChange(index: number, tolerationProperty: keyof IToleration): (event: any) => void {
  return (event: any) => {
    const tolerations = cloneDeep(this.state.tolerations);
    tolerations[index][tolerationProperty] = event.target.value;
    this.props.onTolerationChange(tolerations);
  };
}
```

And called below as `this.handleChange(i, 'effect')`.

app/scripts/modules/kubernetes/src/common/tolerations/TolerationsComponent.tsx
```diff
@@ -0,0 +1,210 @@
...
+
+  public render() {
+    return (
+      <div>
+        <div className="sm-label-left">Tolerations</div>
+        <table className="table table-condensed">
+          <tbody>
+            {this.state.tolerations.map((row, i) => {
+              return (
+                <tr key={i}>
```
danielpeach: Using an array index as a key has never bitten me, but it is supposedly bad practice. You should try to use the tolerations's `key` if it exists.

app/scripts/modules/kubernetes/src/common/tolerations/TolerationsComponent.tsx
```diff
@@ -0,0 +1,210 @@
...
+  onTolerationChange: (tolerations: IToleration[]) => void;
+}
+
+export interface IKubernetesTolerationsState {
+  tolerations: IToleration[];
+}
+
+@BindAll()
+export class KubernetesTolerations extends React.Component<IKubernetesTolerationsProps, IKubernetesTolerationsState> {
+  private static operators: Option[] = [{ value: 'Exists', label: 'Exists' }, { value: 'Equal', label: 'Equal' }];
```
danielpeach: Would be nice `Option` had a type parameter, but I doubt it does.

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/configureJob.controller.js
```diff
@@ -168,6 +168,11 @@ module.exports = angular
       this.stage.containers[index].lifecycle.preStop = handler;
     };
 
+    this.onTolerationChange = tolerations => {
+      this.stage.tolerations = tolerations;
+      $scope.$digest();
```
danielpeach: I think it is nicer to use `$applyAsync` here, but who knows.

app/scripts/modules/kubernetes/src/common/tolerations/tolerations.component.ts
```diff
@@ -0,0 +1,9 @@
+import { module } from 'angular';
+import { KubernetesTolerations } from './TolerationsComponent';
```
danielpeach: Please sort local dependencies below external dependencies.

ethanfrogers: 👍

app/scripts/modules/kubernetes/src/common/tolerations/TolerationsComponent.tsx
ethanfrogers: That's one of the changes i thought about making. Just wanted to get some feedback first! Agreed!

app/scripts/modules/kubernetes/src/common/tolerations/TolerationsComponent.tsx
ethanfrogers: So, default to `key` where non-empty and fall back to index when it is?

danielpeach: Yep, exactly.

app/scripts/modules/kubernetes/src/pipeline/stages/runJob/configureJob.controller.js
ethanfrogers: 👍

app/scripts/modules/kubernetes/src/common/tolerations/TolerationsComponent.tsx
ethanfrogers: I attempted to use this here but ran into some weird behavior when updating the `key` property.

app/scripts/modules/kubernetes/src/common/tolerations/TolerationsComponent.tsx
```diff
@@ -0,0 +1,194 @@
...
+  ];
+
+  public tolerations: IToleration[];
+
+  constructor(props: IKubernetesTolerationsProps) {
+    super(props);
+    this.state = this.getState(props);
+  }
+
+  private getState(props: IKubernetesTolerationsProps): IKubernetesTolerationsState {
```
jrsquared: Since state will always just be `props.tolerations`, you probably don't even need state at all, then you can get rid of `componentWillReceiveProps` as well.

PR #5209, opened 2018-04-19
feat(provider/kubernetes): expose optional envvar

ethanfrogers:
exposed the optional field for ConfigMap/Secret environment variables.

fixes spinnaker/spinnaker#2589

### DIFF
--- a/app/scripts/modules/kubernetes/src/container/environmentFrom.component.html
+++ b/app/scripts/modules/kubernetes/src/container/environmentFrom.component.html
@@ -43,7 +43,8 @@
           <div class="col-md-3">
             <input type="checkbox"
                    style="margin-top: 10px; margin-left: 8px;"
-                   ng-model="envFrom.configMapRef.optional"/>
+                   ng-model="envFrom.configMapRef.optional"
+                   ng-init="envFrom.configMapRef.optional = $ctrl.initOptional(envFrom.configMapRef)"/>
           </div>
         </div>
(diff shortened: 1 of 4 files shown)


ethanfrogers:
Disclaimer: I'm not proud of what I had to do to get the desired effect.

app/scripts/modules/kubernetes/src/container/environmentFrom.component.html
```diff
@@ -43,7 +43,8 @@
           <div class="col-md-3">
             <input type="checkbox"
                    style="margin-top: 10px; margin-left: 8px;"
-                   ng-model="envFrom.configMapRef.optional"/>
+                   ng-model="envFrom.configMapRef.optional"
+                   ng-init="envFrom.configMapRef.optional = ((envFrom.configMapRef.optional === null || envFrom.configMapRef.optional === undefined) || envFrom.configMapRef.optional)"/>
```
danielpeach: Might be less surprising for developers if this logic was contained in an angular controller.

danielpeach: Also it's acceptable in JS to use `myVar == null` (rather than triple equals) to check if a variable is either `null` or `undefined`. Though not sure if angular templates fully respect JS syntax.

ethanfrogers:
@danielpeach Updated: PTAL!

PR #5210, opened 2018-04-19
fix(core/serverGroup): Default to decoding User Data as text

erikmunson:
User Data always comes back Base64 encoded. If the encoded data is just text, it's straightforward to decode and show it in the UI rather than making people manually decode every time. If the data is in another format (like gzipped binary), it's far less straightforward and doing the right thing to decode is highly dependent on the individual use case.

Since textual User Data is the most common use case, this adds a checkbox to decode as text that defaults to checked, but allows for viewing the encoded source in case you want to process it yourself.

### DIFF
--- a/app/scripts/modules/core/src/serverGroup/details/ShowUserData.tsx
+++ b/app/scripts/modules/core/src/serverGroup/details/ShowUserData.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import { Modal } from 'react-bootstrap';
 import { BindAll } from 'lodash-decorators';
 
+import { decodeUnicodeBase64 } from 'core/utils/unicodeBase64';
+
 export interface IShowUserDataProps {
   serverGroupName: string;
   title?: string;
@@ -10,6 +12,7 @@ export interface IShowUserDataProps {
 
 export interface IShowUserDataState {
   show: boolean;
+  decodeAsText: boolean;
 }
 
 @BindAll()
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/serverGroup/details/ShowUserData.tsx
```diff
@@ -45,11 +53,20 @@ export class ShowUserData extends React.Component<IShowUserDataProps, IShowUserD
...
-              <textarea readOnly={true} rows={15} className="code">
-                {userData}
-              </textarea>
-            </div>
+            <>
+              <textarea
+                className="code"
+                readOnly={true}
+                rows={15}
+                value={decodeAsText ? decodeUnicodeBase64(userData) : userData}
```
erikmunson: this moved to `value` because you're not actually supposed to use `children` with `textarea` elements in React (it essentially opts you out of React managing the text and means you're responsible for keeping it in sync).

app/scripts/modules/core/src/utils/unicodeBase64.ts
```diff
@@ -0,0 +1,7 @@
+export const decodeUnicodeBase64 = (encodedString: string) =>
+  decodeURIComponent(
+    atob(encodedString)
+      .split('')
+      .map(char => '%' + ('00' + char.charCodeAt(0).toString(16)).slice(-2))
+      .join(''),
+  );
```
erikmunson: turns out `atob` doesn't really know what to do with unicode chars, this is just copy-pasta that properly handles them.

jrsquared: Worth putting a comment in the code about this, or pulling the function out into a nicely named function. None of us will remember what this is doing in a week.

erikmunson: eh, i originally thought about a comment and decided against it because i thought the name was clear enough on its own 😃

Will throw in a quick comment.

erikmunson: @jrsquared 2934804

jrsquared: Noice.

PR #5211, opened 2018-04-19
feat(pipeline_templates) Allow Pipelines to Inherit Pipeline Template Parameters

benjaminws:
Currently, if you specify a pipeline template with parameters and then try to create a template with it in deck, the parameters are not inherited and have to be recreated by hand in the new pipeline. This behavior works as expected with roer / cli / API.
template-inherit

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/templates/configurePipelineTemplateModal.controller.ts
+++ b/app/scripts/modules/core/src/pipeline/config/templates/configurePipelineTemplateModal.controller.ts
@@ -28,6 +28,7 @@ export interface IState {
   loadingError: boolean;
   noVariables: boolean;
   planErrors: IPipelineTemplatePlanError[];
+  inheritTemplateParameters: boolean;
 }
 
 @BindAll()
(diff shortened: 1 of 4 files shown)


wjoel:
Cool! Is there any reason to limit this to just parameters? In my roer pipeline configs I almost always use:
```yaml
configuration:
  inherit:
  - stages
  - expectedArtifacts
  - triggers
```
I've gotten questions from a couple of people at work wondering why their triggers don't show up when they create a templated pipeline from deck. It's also not an uncommon question on Slack.

benjaminws:
Really no reason other than this one feature solved my problem at the time, heh. I'll knock that out!

benjaminws:
My setup typically looks like:

```
configuration:
  inherit:
    - concurrentExecutions
    - triggers
    - parameters
```

wjoel:
Yeah, I think you're right about them being included by default.

benjaminws:
Closing in favor of another PR that merges from a branch rather than master. https://github.com/spinnaker/deck/pull/5214

PR #5214, opened 2018-04-20
feat(pipeline_templates) Allow Pipelines to Inherit Pipeline Template Configuration

benjaminws:
Currently, if you specify a pipeline template with parameters, triggers, or expected artifacts and then try to create a template with it in deck, the configuration is not inherited and has to be recreated by hand in the new pipeline. This behavior works as expected with roer / cli / API.

template-inherit


benjaminws:
Thinking this through a little bit more and wondering if there is a use case for _always_ inheriting these config options, and removing the checkboxes. I am concerned that if someone un-checks one of those options they may not know what they're getting into and end up with a bad pipeline. Seems like a bad experience. Additionally, if you un-check these, because this is a template, you can't really edit the pipeline afterward to add custom configuration.

So maybe these checkboxes aren't needed at all? @wjoel

wjoel:
@benjaminws I always thought the defaults were unreasonable but I didn't have much experience with templates, so I wasn't sure.

Now I have more experience and I'm a bit more sure about them being unreasonable, but maybe there's a reason why they are set the way they are, but I wouldn't be surprised if it's just by accident. I would also like to see them changed to be less surprising. The non-inheritance use cases seem more suited for the roer/API use cases.

app/scripts/modules/core/src/pipeline/config/templates/configurePipelineTemplateModal.controller.ts
```diff
@@ -97,6 +109,14 @@ export class ConfigurePipelineTemplateModalController implements IController {
   }
 
   public buildConfig(): IPipelineTemplateConfig {
+    let inheritConfig = [];
```
danielpeach: `let` is being flagged by the linter because it's never modified (i.e., use `const` instead). `const` means that the reference can't be changed, but the array the variable points to can still be mutated.

danielpeach:
@benjaminws just a comment something you said:

"Additionally, if you un-check these, because this is a template, you can't really edit the pipeline afterward to add custom configuration."

You should be able to change this option (you should be able to go back to this modal), but at some point the UI for this became very badly broken. If that's what you're talking about, that's not the intended behavior, and I'd be grateful for a fix.

app/scripts/modules/core/src/pipeline/config/templates/configurePipelineTemplateModal.controller.ts
benjaminws: 👍 I'll fix this up.

app/scripts/modules/core/src/pipeline/config/templates/configurePipelineTemplateModal.controller.ts
```diff
@@ -97,6 +109,14 @@ export class ConfigurePipelineTemplateModalController implements IController {
...
+    if (this.state.inheritTemplateParameters) {
+      inheritConfig.push('parameters');
+    } else if (this.state.inheritTemplateExpectedArtifacts) {
```
danielpeach: Shouldn't these be separate `if` blocks?

benjaminws: Wasn't a fan of the conditional blocks anyway, cleaned it up

PR #5218, opened 2018-04-20
fix(core/cluster): fix infinite loop toggling listInstances true/false

christopherthielen:
This is a dirty hack.  

I will circle back around later and update the `filterModel` stuff to use router params.  I'll remove the manual params management code.


erikmunson:
👍

PR #5229, opened 2018-04-24
fix(core/pipeline): Only list decorated artifacts

erikmunson:
We've been running into an issue at Netflix where Igor is not configured to decorate artifacts, but Orca is [adding artifact data to manual executions](https://github.com/spinnaker/orca/commit/b1f1686672a5f65bf452779f6d2ee7b49cc51aca). This results in a bunch of artifacts on execution objects without a `type` or `name`, which the current `ArtifactList` just dumps on the screen:

undecorated-artifacts

In general, artifacts that don't have a `type` or `name` are relatively useless from Deck's perspective (hence the decoration that Igor does). This change just omits any undecorated artifacts.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/status/ArtifactList.spec.tsx
+++ b/app/scripts/modules/core/src/pipeline/status/ArtifactList.spec.tsx
@@ -15,7 +15,7 @@ describe('<ArtifactList/>', () => {
   beforeEach(mock.inject(() => {})); // Angular is lazy.
 
   it('renders null when null artifacts are passed in', function() {
-    const artifacts: IArtifact[] = [];
+    const artifacts: IArtifact[] = null;
     component = shallow(<ArtifactList artifacts={artifacts} />);
     expect(component.get(0)).toEqual(null);
   });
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/status/ArtifactList.spec.tsx
```diff
@@ -15,7 +15,7 @@ describe('<ArtifactList/>', () => {
   beforeEach(mock.inject(() => {})); // Angular is lazy.
 
   it('renders null when null artifacts are passed in', function() {
-    const artifacts: IArtifact[] = [];
+    const artifacts: IArtifact[] = null;
```
sbwsg: doh.  I should've written an ArtifactList.spec.tsx.spec.tsx.

jrsquared:
lgtm

app/scripts/modules/core/src/pipeline/status/ArtifactList.spec.tsx
erikmunson: Test Driven Testing

PR #5230, opened 2018-04-24
fix(core) Keep correct stage in summary object for future operations

PerGon:
Keep the `summary` object with the correct stage as the `summary` object is used later on used ( [in here](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/service/executions.transformer.service.ts#L465) ) and without this, the `summary` object changes are discarded since it's changed an object which is not in the array anymore.


jrsquared:
very nice.

PR #5238, opened 2018-04-26
fix(artifacts): limit list of artifacts in execution to those consumed by pipeline

sbwsg:
Before, set of artifacts detailed in execution even if they're not expected by pipeline:
before

After, set of artifacts that weren't expected by pipeline are collapsed:
after

Fixes https://github.com/spinnaker/spinnaker/issues/2567


lwander:
Nice

PR #5254, opened 2018-05-02
fix(ssl/apache2): Fix for ports.conf.gen

laurieodgers:
Change 443 in ports.conf.gen to {%DECK_PORT%} to allow it to be changed by docker entrypoint script. 

Fixes spinnaker/spinnaker#2733


sbwsg:
lgtm but id like to get a +1 from someone who knows the apache config better than i do.

PR #5263, opened 2018-05-03
refactor(google): Refactor config fields to use consistent directive

ezimanyi:
* Make artifact components more re-usable by being more consistent about which tags should be handled by the parent form and which should be embedded in the component.  In particular, push ` <div class="container-fluid form-horizontal">` to the top of each form rather than repeating in each component.
* Replace boiler-plate form layout code with `stage-config-field` directive


ezimanyi:
I suggest ignoring whitespace when looking at this diff.

PR #5270, opened 2018-05-04
feat(artifacts): Allow artifacts to be selected in bake config

ezimanyi:
bake

### DIFF
--- a/app/scripts/modules/google/src/pipeline/stages/bake/bakeStage.html
+++ b/app/scripts/modules/google/src/pipeline/stages/bake/bakeStage.html
@@ -3,6 +3,15 @@
     <input type="text" class="form-control input-sm"
            ng-model="stage.package"/>
   </stage-config-field>
+  <expected-artifact-multi-selector command="stage"
+                                    ids-field="packageArtifactIds"
+                                    artifact-label="Package Artifacts"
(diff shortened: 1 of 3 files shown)


app/scripts/modules/google/src/pipeline/stages/bake/bakeStage.html
```diff
@@ -3,6 +3,15 @@
     <input type="text" class="form-control input-sm"
            ng-model="stage.package"/>
   </stage-config-field>
+  <expected-artifact-multi-selector command="stage"
+                                    ids-field="packageArtifactIds"
```
sbwsg: I think this field might need to be registered as an artifact reference In the `config` function of gceBakeStage.js so that it can be successfully cleaned up if artifacts are deleted.  Something like:

```javascript
artifactReferenceServiceProvider.registerReference('stage', () => [
  ['packageArtifactIds'],
]);
```

See `app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestStage.ts` for a comparable example.

lwander:
Do we want to with "package/deb"? Is there meaning to putting package first? Other artifacts like "github/file", "http/file" put the source first.

PR #5316, opened 2018-05-14
 feat(artifacts): Either string or artifact is sufficient for bake stage

ezimanyi:
Allow packages to be specified as either a string or a list of artifacts on the bake stage.  To implement this, create a new field validator that checks if at least one of a list of fields is non-null.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/validation/anyFieldRequired.validator.ts
+++ b/app/scripts/modules/core/src/pipeline/config/validation/anyFieldRequired.validator.ts
@@ -1,0 +1,47 @@
+import { module } from 'angular';
+import { map, some } from 'lodash';
+
+import { IPipeline, IStage, IStageOrTriggerTypeConfig, ITrigger } from 'core/domain';
+import { BaseRequiredFieldValidator, IRequiredField } from './baseRequiredField.validator';
+import { PIPELINE_CONFIG_VALIDATOR, PipelineConfigValidator } from './pipelineConfig.validator';
+import { IBaseRequiredFieldValidationConfig } from 'core/pipeline/config/validation/baseRequiredField.validator';
+
+export interface IMultiRequiredField extends IBaseRequiredFieldValidationConfig {
+  fields: IRequiredField[];
+}
+
+export type IAnyFieldRequiredValidationConfig = IBaseRequiredFieldValidationConfig & IMultiRequiredField;
+
+export class AnyFieldRequiredValidator extends BaseRequiredFieldValidator {
+  protected passesValidation(
+    pipeline: IPipeline,
+    stage: IStage | ITrigger,
+    validationConfig: IAnyFieldRequiredValidationConfig,
+  ): boolean {
+    return some(validationConfig.fields, (requiredField: IRequiredField) => {
+      return this.fieldIsValid(pipeline, stage, requiredField.fieldName);
+    });
+  }
+
+  protected validationMessage(
+    validationConfig: IAnyFieldRequiredValidationConfig,
+    config: IStageOrTriggerTypeConfig,
+  ): string {
+    let fieldString: string = map(validationConfig.fields, (requiredField: IRequiredField) =>
+      this.printableFieldLabel(requiredField),
+    ).join(', ');
+    return (
+      validationConfig.message ||
+      `At least one of the following fields must be supplied for ${
+        config.label
+      } stages: <strong>${fieldString}</strong>.`
+    );
+  }
+}
+
+export const ANY_FIELD_REQUIRED_VALIDATOR = 'spinnaker.core.pipeline.config.validation.anyFieldRequired';
+module(ANY_FIELD_REQUIRED_VALIDATOR, [PIPELINE_CONFIG_VALIDATOR])
+  .service('anyFieldRequiredValidator', AnyFieldRequiredValidator)
+  .run((pipelineConfigValidator: PipelineConfigValidator, anyFieldRequiredValidator: AnyFieldRequiredValidator) => {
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/pipeline/config/validation/baseRequiredField.validator.ts
```diff
@@ -0,0 +1,56 @@
...
+  ): boolean;
+
+  protected abstract validationMessage(
+    validationConfig: IBaseRequiredFieldValidationConfig,
+    config: IStageOrTriggerTypeConfig,
+  ): string;
+
+  protected printableFieldLabel(field: IRequiredField): string {
+    let fieldLabel: string = field.fieldLabel || field.fieldName;
+    return fieldLabel.charAt(0).toUpperCase() + fieldLabel.substr(1);
```
sbwsg: this is also available as a lodash built-in: https://lodash.com/docs/4.17.10#upperFirst

app/scripts/modules/core/src/pipeline/config/validation/requiredField.validator.ts
```diff
@@ -1,45 +1,32 @@
...
-    const validationMessage =
-      validationConfig.message || `<strong>${fieldLabel}</strong> is a required field for ${config.label} stages.`;
-    const fieldExists = has(stage, validationConfig.fieldName);
-    const field: any = get(stage, validationConfig.fieldName);
-
-    if (!fieldExists || (!field && field !== 0) || (field instanceof Array && field.length === 0)) {
-      return validationMessage;
-    }
-    return null;
+    let fieldLabel: string = this.printableFieldLabel(validationConfig);
```
danielpeach: Linter will be upset here (s/let/const)

app/scripts/modules/core/src/pipeline/config/validation/baseRequiredField.validator.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    const fieldExists = has(stage, fieldName);
+    const field: any = get(stage, fieldName);
+
+    return fieldExists && (field || field === 0) && !(field instanceof Array && field.length === 0);
+  }
+}
+
+export const BASE_REQUIRED_FIELD_VALIDATOR = 'spinnaker.core.pipeline.config.validation.baseRequiredField';
```
danielpeach: The deck convention is that an all caps export indicates an angular module name - would omit this unless needed (I don't think it is).

app/scripts/modules/core/src/pipeline/config/validation/baseRequiredField.validator.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  protected fieldIsValid(pipeline: IPipeline, stage: IStage | ITrigger, fieldName: string): boolean {
+    if (pipeline.strategy === true && ['cluster', 'regions', 'zones', 'credentials'].includes(fieldName)) {
+      return true;
+    }
+
+    const fieldExists = has(stage, fieldName);
+    const field: any = get(stage, fieldName);
+
+    return fieldExists && (field || field === 0) && !(field instanceof Array && field.length === 0);
```
danielpeach: I know this is copied but the preferred check for arrays is `Array.isArray`. I don't totally remember why, something about iframes.

app/scripts/modules/core/src/pipeline/config/validation/baseRequiredField.validator.ts
```diff
@@ -0,0 +1,56 @@
...
+    validationConfig: IBaseRequiredFieldValidationConfig,
+  ): boolean;
+
+  protected abstract validationMessage(
+    validationConfig: IBaseRequiredFieldValidationConfig,
+    config: IStageOrTriggerTypeConfig,
+  ): string;
+
+  protected printableFieldLabel(field: IRequiredField): string {
+    let fieldLabel: string = field.fieldLabel || field.fieldName;
```
danielpeach: Linter will be upset here.

app/scripts/modules/core/src/pipeline/config/validation/anyFieldRequired.validator.ts
```diff
@@ -0,0 +1,47 @@
...
+    return some(validationConfig.fields, (requiredField: IRequiredField) => {
+      return this.fieldIsValid(pipeline, stage, requiredField.fieldName);
+    });
+  }
+
+  protected validationMessage(
+    validationConfig: IAnyFieldRequiredValidationConfig,
+    config: IStageOrTriggerTypeConfig,
+  ): string {
+    let fieldString: string = map(validationConfig.fields, (requiredField: IRequiredField) =>
```
danielpeach: Linter will be upset here. Also should use `Array.prototype.map` here. Lodash gives you null safety here but you should just do `(validationConfig.fields || []).map(...)` if that's what you're worried about.

app/scripts/modules/core/src/pipeline/config/validation/anyFieldRequired.validator.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type IAnyFieldRequiredValidationConfig = IBaseRequiredFieldValidationConfig & IMultiRequiredField;
+
+export class AnyFieldRequiredValidator extends BaseRequiredFieldValidator {
+  protected passesValidation(
+    pipeline: IPipeline,
+    stage: IStage | ITrigger,
+    validationConfig: IAnyFieldRequiredValidationConfig,
+  ): boolean {
+    return some(validationConfig.fields, (requiredField: IRequiredField) => {
```
danielpeach: Please use `Array.prototype.some` here.

PR #5322, opened 2018-05-15
feat(artifacts): List artifacts consumed / produced by executions

sbwsg:
screen shot 2018-05-15 at 10 15 26 am


app/scripts/modules/core/src/artifact/ArtifactIconService.ts
```diff
@@ -0,0 +1,24 @@
...
+    this.typeMap.set(type, iconPath);
+  }
+
+  public static getPath(type: string) {
+    return this.typeMap.get(type) || unknownArtifactPath;
+  }
+}
+
+ArtifactIconService.registerType('docker/image', require('./icons/docker-image-artifact.svg'));
+ArtifactIconService.registerType('kubernetes/manifest', kubernetesLogoPath);
```
lwander: I think this should be able to cover `kubernetes/*`, since technically any kind should be supported

app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/artifactList.ts
```diff
@@ -0,0 +1,28 @@
...
+
+class KubernetesExecutionArtifactListController implements IController {
+  public artifacts: any[];
+
+  public iconPath(type: string): string {
+    return ArtifactIconService.getPath(type);
+  }
+}
+
+class KubernetesExecutionArtifactListComponent implements IComponentOptions {
```
lwander: I think we'll want to use this outside of just the execution tab right?

app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/artifactTab.ts
```diff
@@ -0,0 +1,39 @@
...
+class KubernetesExecutionArtifactTabComponent implements IComponentOptions {
+  public bindings: any = { stage: '<', execution: '<' };
+  public controller: any = KubernetesExecutionArtifactTabController;
+  public controllerAs = 'ctrl';
+  public template = `
+<div class="row execution-artifacts">
+  <div class="col-md-6 artifact-list-container">
+    <h5>Consumed Artifacts</h5>
+    <div ng-if="ctrl.execution && ctrl.execution.trigger && ctrl.execution.trigger.artifacts">
+      <kubernetes-execution-artifact-list artifacts="ctrl.execution.trigger.artifacts">
```
lwander: This isn't necessarily coming from the trigger - I would defer to the artifacts that are explicitly requested in the "req. artifacts to bind" and "manifest artifact" fields.

app/scripts/modules/core/src/artifact/icons/s3-object-artifact.svg
```diff
@@ -0,0 +1,17 @@
+<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" width="100" height="100" viewBox="0 0 100 100">
+  <title>Storage</title>
```
lwander: Are there any restrictions w.r.t. using amazons logos?

app/scripts/modules/core/src/artifact/ArtifactIconService.ts
```diff
@@ -0,0 +1,29 @@
...
+export class ArtifactIconService {
+  private static icons: IArtifactIcon[] = [];
+
+  public static registerType(type: RegExp, path: string) {
+    ArtifactIconService.icons.push({ type, path });
+  }
+
+  public static getPath(type: string) {
+    const icon = ArtifactIconService.icons.find(entry => entry.type.test(type));
+    if (icon == null) {
```
ezimanyi: Minor nit, but `find` returns `undefined` when it doesn't find anything.  This will of course still work given the comparison is `==`.

app/scripts/modules/core/src/artifact/ArtifactIconService.ts
```diff
@@ -0,0 +1,29 @@
...
+    return icon.path;
+  }
+}
+
+ArtifactIconService.registerType(/docker\/image/, require('./icons/docker-image-artifact.svg'));
+ArtifactIconService.registerType(/kubernetes\/.*/, require('./icons/kubernetes-artifact.svg'));
+ArtifactIconService.registerType(/embedded\/base64/, require('./icons/embedded-base64-artifact.svg'));
+ArtifactIconService.registerType(/gcs\/object/, require('./icons/gcs-file-artifact.svg'));
+ArtifactIconService.registerType(/github\/file/, require('./icons/github-file-artifact.svg'));
+ArtifactIconService.registerType(/s3\/object/, require('./icons/s3-object-artifact.svg'));
```
ezimanyi: I'm torn about whether we should use regular expressions here vs. just an exact matching here.  It looks like we're only taking advantage of the regexes for kubernetes---how many different types of `kubernetes/` artifacts are there, and is it a small enough list we know a priori?

I also briefly wondered if it would make sense to just include the icon at the time we register the artifact itself (ex: in `gcs.artifact.ts` include a key for `artifact`) but it looks like not all artifact types are registered in that way, so maybe that is not an option.

sbwsg: I don't know the exact number Spinnaker would have to be concerned with but there are definitely a lot of resource types.  The full list for 1.10 is here: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/.  I also think there's an element of not wanting to have to push code changes if the k8s project adds new kinds.

PR #5325, opened 2018-05-15
feat(ttl): Clearly identify ephemeral server groups

ajordens:
image

### DIFF
--- a/app/scripts/modules/core/src/entityTag/notifications/EntityNotifications.tsx
+++ b/app/scripts/modules/core/src/entityTag/notifications/EntityNotifications.tsx
@@ -5,6 +5,7 @@ import { Placement } from 'core/presentation';
 import { Application } from 'core/application';
 import { noop } from 'core/utils';
 import { NotificationsPopover } from './NotificationsPopover';
+import { EphemeralPopover } from './EphemeralPopover';
 
 export interface IEntityNotificationsProps {
   entity: any;
@@ -89,6 +90,8 @@ export class EntityNotifications extends React.Component<IEntityNotificationsPro
           hOffsetPercent={hOffsetPercent}
           onUpdate={onUpdate}
         />
+
+        <EphemeralPopover entity={entity} />
       </div>
     );
   }
--- a/app/scripts/modules/core/src/entityTag/notifications/EphemeralPopover.tsx
+++ b/app/scripts/modules/core/src/entityTag/notifications/EphemeralPopover.tsx
@@ -1,0 +1,67 @@
+import * as React from 'react';
+
+import { HoverablePopover, IEntityTags } from 'core';
+
+import * as moment from 'moment';
+
+export interface IEphemeralPopoverProps {
+  entity?: any;
+}
+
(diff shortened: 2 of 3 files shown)


ajordens:
No console errors when looking at an application containing server groups w/ and w/o `spinnaker:ttl` entity tags.

@anotherchrisberry How do I go about a colored header in the `HoverablePopover`? The clock icon in the side bar is slightly grayed out ... 

I opt'd for the `HoverablePopover` since at some point we'll add a delete option.

app/scripts/modules/core/src/entityTag/notifications/EphemeralPopover.tsx
```diff
@@ -0,0 +1,59 @@
...
+    this.state = this.getState(props);
+  }
+
+  private getState(props: IEphemeralPopoverProps): IEphemeralPopoverState {
+    const entityTags: IEntityTags = props.entity.entityTags;
+    if (entityTags) {
+      const ephemeralTag = entityTags.tags.filter(x => x.name === 'spinnaker:ttl')[0];
+      if (ephemeralTag) {
+        return {
+          destroyedFromNow: moment(ephemeralTag.value.expiry).fromNow(true),
```
ajordens: pretty trusting that `expiry` is a legitimate value (`"expiry": "2018-05-15T22:34:42.422Z"`)

ajordens:
Actually I think I can control the header if I pass in an actual `Component` to `HoverablePopover`.

ajordens:
image

Revised color scheme -- went one shade darker than what we were using for the account label. No strong opinions on color or title text (`Ephemeral Server Group` ??)

app/scripts/modules/core/src/entityTag/notifications/notifications.less
```diff
@@ -1,6 +1,8 @@
-@import (reference) "~core/presentation/less/imports/commonImports.less";
+@import (reference) '~core/presentation/less/imports/commonImports.less';
```
ajordens: I think the fancy post-commit hook fixed this up?

anotherchrisberry: Yeah, that is prettier doing its thing.

app/scripts/modules/core/src/entityTag/notifications/EntityNotifications.tsx
```diff
@@ -5,6 +5,7 @@ import { Placement } from 'core/presentation';
 import { Application } from 'core/application';
 import { noop } from 'core/utils';
 import { NotificationsPopover } from './NotificationsPopover';
+import { EphemeralPopover } from 'core/entityTag/notifications/EphemeralPopover';
```
anotherchrisberry: can shorten this to:
```javascript
import { EphemeralPopover } from './EphemeralPopover';
```

anotherchrisberry:
How would you feel about https://fontawesome.com/icons/bomb?style=solid as the icon? Too violent?

PR #5326, opened 2018-05-16
feat(bake/manifests): Add namespace support for Helm bakery

posox:
This patch adds support passing namespace for Helm bakery.

link: spinnaker/spinnaker#2763


ezimanyi:
This looks good---but holding off on merging until https://github.com/spinnaker/rosco/pull/261 is finalized, as addressing the comments there may require changes here as well.

PR #5338, opened 2018-05-17
fix(core): Cleanup manual trigger template state when switching pipelines

erikmunson:
Prettier did some extra things so ignore those.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
@@ -86,6 +86,7 @@ module.exports = angular
     let addTriggers = () => {
       let pipeline = this.command.pipeline;
       if (!pipeline || !pipeline.triggers || !pipeline.triggers.length) {
+        this.triggers = null;
         this.command.trigger = null;
         return;
       }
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
```diff
@@ -130,6 +131,8 @@ module.exports = angular
 
       if (command.trigger && pipelineConfig.hasManualExecutionComponentForTriggerType(command.trigger.type)) {
         this.triggerComponent = pipelineConfig.getManualExecutionComponentForTriggerType(command.trigger.type);
+      } else {
+        this.triggerComponent = null;
```
jrsquared: dur.

app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
```diff
@@ -207,7 +210,9 @@ module.exports = angular
 
     this.execute = () => {
       let selectedTrigger = this.command.trigger || {},
-        command = { trigger: selectedTrigger },
+        command = {
+          trigger: selectedTrigger,
```
jrsquared: the important change

erikmunson: there's no way we could've read this without it.

PR #5341, opened 2018-05-17
feat(artifacts): show artifact type as icon in execution summary

sbwsg:
screen shot 2018-05-17 at 12 42 00 pm

screen shot 2018-05-17 at 12 42 17 pm


lwander:
Looking good!

Curious about that first picture, how was that pipeline configured?

sbwsg:
here's the conf for that pipeline:

screen shot 2018-05-17 at 12 48 25 pm

PR #5347, opened 2018-05-18
fix(core): Fix travis triggers

jrsquared:


app/scripts/modules/core/src/pipeline/config/triggers/travis/travisTrigger.module.ts
```diff
@@ -84,27 +84,29 @@ export class TravisTrigger implements IController {
...
-      },
-    ],
-  });
-});
+          preventSave: true,
+        },
+      ],
+    });
+  })
+  .controller('TravisTriggerCtrl', TravisTrigger);
```
jrsquared: This is the only change, the rest is prettier.

PR #5348, opened 2018-05-18
chore(artifacts): de-angularize artifact reference service

sbwsg:
Adds artifact reference removal to the bakeManifest stage for the `expectedArtifacts` and `inputArtifacts` fields.

In the below image an artifact has been removed from the Expected Artifacts section of the config screen & that removal continues to be correctly reflected in the GCE deploy stage:

screen shot 2018-05-18 at 11 47 12 am

And here an artifact has been removed from a config screen and is correctly reflected in the bake manifest stage:

screen shot 2018-05-18 at 11 48 26 am


jrsquared:
Love this.

PR #5351, opened 2018-05-18
fix(artifacts): ensure fieldColumns is not undefined in expected artifact select

sbwsg:
I accidentally broke this with my previous appengine change.

Broken:
screen shot 2018-05-18 at 1 32 20 pm

Fixed:
screen shot 2018-05-18 at 1 32 57 pm


app/scripts/modules/core/src/artifact/expectedArtifactSelector.component.ts
```diff
@@ -16,6 +16,10 @@ class ExpectedArtifactSelectorCtrl implements IController {
   public showIcons: boolean;
   public fieldColumns: number;
 
+  public $onInit() {
+    this.fieldColumns = this.fieldColumns || 8;
```
danielpeach: This could be `public fieldColumns = 8;`

PR #5367, opened 2018-05-22
feat(*/instance): add moniker info + env to instance link templates

erikmunson:
So folks can use things like `{{stack}}` and `{{environment}}` when building links.

To the best of my knowledge I covered all the cloud providers that currently implement custom instance links — there are a few like Azure that didn't seem to use them at all. I kind of went around the existing angular data flow a bit in hopes of making things more amenable to reactification later, but I'm not particularly tied to it.

### DIFF
--- a/app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
+++ b/app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
@@ -32,6 +32,8 @@ module.exports = angular
     instanceReader,
     instance,
     app,
+    moniker,
+    environment,
     $q,
     overrides,
   ) {
@@ -45,6 +47,8 @@ module.exports = angular
     };
 
     $scope.application = app;
+    $scope.moniker = moniker;
+    $scope.environment = environment;
 
     $scope.securityGroupsLabel = FirewallLabels.get('Firewalls');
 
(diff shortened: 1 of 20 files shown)


app/scripts/modules/core/src/cloudProvider/skin.service.ts
```diff
@@ -86,6 +87,27 @@ export class SkinService {
     });
   }
 
+  public static getMonikerForInstance(cloudProvider: string, instanceId: string, app: Application): IPromise<IMoniker> {
```
erikmunson: I'm not entirely sure whether or not we need to look through all these server groups and load balancers to find moniker data like we're currently doing with accounts, so would be cool to get some more context 🤷‍♂️ 

Maybe it's actually the same for this as it is for accounts and we really need this?

lwander: @imosquera might be able to lend some insight

app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
```diff
@@ -32,6 +32,8 @@ module.exports = angular
     instanceReader,
     instance,
     app,
+    moniker,
```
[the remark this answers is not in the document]
erikmunson: from a meatspace convo: the controller registers itself with `overrideRegistry`, and the override code takes any props it gets passed in (in this case from `InstanceDetails.tsx`) and sends them to the controller via DI (see `AngularJSAdapter` for the actual controller invocation).

app/scripts/modules/core/src/cloudProvider/skin.service.ts
erikmunson: would love it if @imosquera or anyone else with moniker knowledge could weigh in, _but_ the worst case here is doing a tiny bit of unnecessary looping/work so I'm going to merge for now and amend this method later if new info arises 🚀

PR #5371, opened 2018-05-22
fix(core): Fix warnings about unused variables

ezimanyi:


ezimanyi:
Saw a couple of new warnings in the logs that looked easy to clean up.

PR #5396, opened 2018-05-29
chore(kubernetes/appengine): bump package versions

danielpeach:
Had to do some work to make it build.


app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/react/DeployStatus.tsx
```diff
@@ -6,10 +6,10 @@ import {
   StageFailureMessage,
   IManifest,
 } from '@spinnaker/core';
-import { KubernetesManifestService } from '../../../../manifest/manifest.service';
```
danielpeach: Also updated `tsconfig.json` so we don't have to have these relative paths everywhere.

danielpeach:
Not a problem, it's very hard to remember.

PR #5402, opened 2018-05-30
fix(provider/kubernetes): ensure key uniqueness in artifact icon list

sbwsg:
Prevent these warnings:

screen shot 2018-05-30 at 10 34 51 am

Occurring because `artifact.reference` isn't guaranteed to be unique.  `artifact.id` isn't present on these resources.


sbwsg:
I'm going to merge this because it's a trivial change and will be equally trivial to revert if need be.

PR #5420, opened 2018-06-04
fix(core/cluster): force stages to lowercase for taskmatcher lookup

christopherthielen:
A task matcher was not found when the stage type was 'resizeServerGroup'
- Removed `good grief`


jrsquared:
fix your tests

PR #5423, opened 2018-06-05
fix(kubernetes): server group manager button alignment

danielpeach:


app/scripts/modules/core/src/cluster/rollups.less
```diff
@@ -1,7 +1,7 @@
-@import (reference) "~core/presentation/less/imports/commonImports.less";
+@import (reference) '~core/presentation/less/imports/commonImports.less';
```
danielpeach: The formatter just did all this, plz don't blame me

PR #5424, opened 2018-06-05
fix(deck): Change RunJob stage to check for logs in HTML

Damonkala:
To fix https://github.com/spinnaker/spinnaker/issues/2869

Checks to see if a RunJob pipeline stage has output logs in the HTML with Angular, rather than in the controller.
The HTML will poll for changes in the logs all the time, and will update without any refreshing, and displays at all times


sbwsg:
Thanks!

PR #5426, opened 2018-06-06
fix(amazon): update help text

emjburns:
Removing the type of load balancer from the help because it is confusing when it appears in ALBs in target groups.

### DIFF
--- a/app/scripts/modules/amazon/src/help/amazon.help.ts
+++ b/app/scripts/modules/amazon/src/help/amazon.help.ts
@@ -101,7 +101,7 @@ const helpContents: { [key: string]: string } = {
   'aws.targetGroup.port':
     'The port on which the targets receive traffic. Cannot be edited after being saved; if you want to use a different port, create a new target group, save the load balancer, move your targets, and then delete this target group.',
   'aws.targetGroup.attributes.deregistrationDelay':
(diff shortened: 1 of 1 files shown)


app/scripts/modules/amazon/src/help/amazon.help.ts
```diff
@@ -101,7 +101,7 @@ const helpContents: { [key: string]: string } = {
   'aws.targetGroup.port':
     'The port on which the targets receive traffic. Cannot be edited after being saved; if you want to use a different port, create a new target group, save the load balancer, move your targets, and then delete this target group.',
   'aws.targetGroup.attributes.deregistrationDelay':
-    'The amount of time for Elastic Load Balancing to wait before changing the state of a deregistering target from draining to unused. The range is 0-3600 seconds. The default value is 300 seconds.',
+    'The amount of time for Load Balancing to wait before changing the state of a deregistering target from draining to unused. The range is 0-3600 seconds. The default value is 300 seconds.',
```
jrsquared: at this point, might as well just change it to `the load balancer`

emjburns: changed

christopherthielen:
neat

PR #5427, opened 2018-06-06
feat(core/presentation): Make CollapsibleSection less style opinionated

christopherthielen:
- Pass in four classNames
- Do not render `<h4>` markup when render prop is provided
- Tweak chevron size (use px)
- Animate chevron rotation


jrsquared:
sure why not.

why is the chevron a prop now? seems to add boiler plate to all the places we are using it?

PR #5435, opened 2018-06-08
fix(artifacts): only show question mark icon for custom artifacts

danielpeach:
i.e., if a configured artifact type doesn't have an icon, don't show a question mark.


danielpeach:
In the dropdown for selecting an artifact type, nothing. In the execution view, the string descriptor of the artifact type e.g., `docker/image`

PR #5439, opened 2018-06-08
fix(amazon/securityGroups): Fix name validator from clearing the name

jrsquared:


christopherthielen:
glad we have an angular expert on the team

PR #5444, opened 2018-06-12
feat(google/iap): Refreshes IAP session after they expire.

stewchen:


app/scripts/modules/core/src/api/network.interceptor.ts
```diff
@@ -72,6 +75,16 @@ export class NetworkInterceptor implements IHttpInterceptor {
         }, (retryCount + 1 + Math.random()) * 1000);
       });
     }
+
+    if (status === 401 && SETTINGS.feature.iapRefresherEnabled) {
+      $http.get('/_gcp_iap/do_session_refresh').then(() => {
+        return this.$q.resolve(this.$injector.get('$http')(config)).then((result: any) => {
```
danielpeach: `this.$injector.get('$http')` is the same thing as `$http` - any reason to import above and inject here?

app/scripts/modules/core/src/api/network.interceptor.ts
```diff
@@ -72,6 +75,16 @@ export class NetworkInterceptor implements IHttpInterceptor {
         }, (retryCount + 1 + Math.random()) * 1000);
       });
     }
+
+    if (status === 401 && SETTINGS.feature.iapRefresherEnabled) {
+      $http.get('/_gcp_iap/do_session_refresh').then(() => {
```
danielpeach: I know this is feature-flagged, but ideally the core Deck library would be unaware of any cloud providers or platforms. 

You could create a separate `IHttpInterceptor` inside either the GCE or App Engine module that does this same check.

app/scripts/modules/core/src/api/network.interceptor.ts
danielpeach: I think the type of `result` is `T`, not `any`

stewchen: changed to $http directly

app/scripts/modules/core/src/api/network.interceptor.ts
stewchen: Done

app/scripts/modules/core/src/api/network.interceptor.ts
stewchen: Done

app/scripts/modules/google/src/interceptors/iap.interceptor.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  public responseError = <T>(response: IHttpPromiseCallbackArg<T>): IPromise<T> => {
+    const { config, status } = response;
+    const retryCount = this.retryQueue[config.url] || 0;
+
+    if (status === 401 && retryCount < this.MAX_RETRIES && SETTINGS.feature.iapRefresherEnabled) {
+      this.retryQueue[config.url] = retryCount + 1;
+
+      $http.get('/_gcp_iap/do_session_refresh').then(() => {
```
danielpeach: I think you need a return here. Otherwise, this method will always return a rejected promise.

app/scripts/modules/google/src/interceptors/iap.interceptor.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * When a Cloud IAP (if enabled) session expires, refreshes session.
+ */
+
+export class IapInterceptor implements IHttpInterceptor {
+  private MAX_RETRIES = 4;
+  private retryQueue: Dictionary<number> = {};
+
+  constructor(private $q: IQService) {
```
danielpeach: Doesn't really matter, but this can be imported from `'ngimport'` as well. If you go that route, remove this constructor.

app/scripts/modules/google/src/interceptors/iap.interceptor.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  private removeFromQueue(config: IRequestConfig): void {
+    delete this.retryQueue[config.url];
+  }
+
+  public responseError = <T>(response: IHttpPromiseCallbackArg<T>): IPromise<T> => {
+    const { config, status } = response;
+    const retryCount = this.retryQueue[config.url] || 0;
+
+    if (status === 401 && retryCount < this.MAX_RETRIES && SETTINGS.feature.iapRefresherEnabled) {
```
danielpeach: What's the retry logic doing here? Is it possible that the IAP flow could fail?

app/scripts/modules/google/src/interceptors/iap.interceptor.ts
stewchen: Done.

app/scripts/modules/google/src/interceptors/iap.interceptor.ts
stewchen: Done

app/scripts/modules/google/src/interceptors/iap.interceptor.ts
stewchen: Removed

app/scripts/modules/google/src/interceptors/iap.interceptor.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+export class IapInterceptor implements IHttpInterceptor {
+  public responseError = <T>(response: IHttpPromiseCallbackArg<T>): IPromise<T> => {
+    const { config, status } = response;
+
+    if (status === 401 && SETTINGS.feature.iapRefresherEnabled) {
+      return $http.get('/_gcp_iap/do_session_refresh').then(
+        () => {
+          return $q.resolve($http.get(config.url, config)).then((result: T) => result, () => $q.reject(response));
```
danielpeach: `$http.get(...)` returns a promise already - so what you have here is equivalent to just `$http.get(config.url, config)`

app/scripts/modules/google/src/interceptors/iap.interceptor.ts
```diff
@@ -0,0 +1,31 @@
...
+        },
+        () => $q.reject(response),
```
danielpeach: This is redundant and can be removed.

PR #5445, opened 2018-06-12
fix(core/pipeline): Don't fail when checking Force Rebake without a trigger

erikmunson:
We were relying on `ng-model`'s deep property setting to add a `trigger` object if it didn't already exist. Since the manual execution controller is already setup to handle us doing it this way I'm not gonna go in and muck with it.


dotdotdotpaul:
This bug is really hurting 1.8.x.  I don't think it can wait to 1.9 to be merged in to that branch.  ref. https://github.com/spinnaker/spinnaker/issues/3079

PR #5449, opened 2018-06-12
fix(core/presentation): add className prop to SubmitButton.tsx

christopherthielen:


christopherthielen:
@archana-s even after the correct 14px size via `primary medium`, the bootstrap button was still misaligned with the spinnaker styleguide button.  This change allows the SubmitButton to use the styleguide styles (but it defaults to bootstrap styles)

PR #5450, opened 2018-06-12
fix(pipeline): Correctly handle saving pipeline templates

robzienert:
If a templated pipeline, save the template configuration, not the rendered pipeline.


jrsquared:
🥇

PR #5458, opened 2018-06-13
feat(provider/gce): Adds mode to autoscaling detail and upsert.

jtk54:
Still needs a bit of work in Orca. Depends on https://github.com/spinnaker/clouddriver/pull/2712.

### DIFF
--- a/app/scripts/modules/google/src/autoscalingPolicy/components/basicSettings/basicSettings.component.html
+++ b/app/scripts/modules/google/src/autoscalingPolicy/components/basicSettings/basicSettings.component.html
@@ -49,4 +49,19 @@
         Must be greater than or equal to 15 seconds.
       </div>
   </div>
+  <div class="row">
+    <div class="col-md-3 sm-label-right">
+      Mode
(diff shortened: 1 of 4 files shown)


jtk54:
Upsert: 
image

Details:
image

app/scripts/modules/google/src/autoscalingPolicy/components/basicSettings/basicSettings.component.html
```diff
@@ -49,4 +49,22 @@
...
+    <div class="col-md-2 content-fields">
+      <select class="form-control input-sm"
+              name="mode"
+              ng-model="$ctrl.policy.mode">
+        <option ng-repeat="mode in $ctrl.modes"
+                value="{{mode}}"
+                ng-selected="$ctrl.policy.mode === mode">{{mode}}</option>
+      </select>
+    </div>
+    <div class="col-md-4 error-message" ng-if="basicSettingsForm.coolDownPeriodSec.$error.min">
```
danielpeach: I think this is tied to the previous form field ("Cool down period seconds") and can be removed.

duftler:
Nice.

PR #5463, opened 2018-06-18
feat(provider/kubernetes): generic details view for unmapped resources

sbwsg:
An "unmapped resource" is one that doesn't map cleanly to Spinnaker's notion of server groups / firewalls / load balancers.  A k8s ConfigMap is an example of this.

Fixes https://github.com/spinnaker/spinnaker/issues/2586
Fixes https://github.com/spinnaker/spinnaker/issues/2327

This PR adds a k8s-specific URL that matches the pattern `/applications/:appId/clusters/manifest/:provider/:accountId/:region/:kind%20:name`.  I haven't leveraged the Override registry here because I don't think any of the other providers have the same need as kubernetes re: unmappable resources.

ConfigMap example:

screen shot 2018-06-18 at 1 17 33 pm

For comparison's sake, here's how a pod renders with the existing k8s instance details view vs. the generic resource details view:

screen shot 2018-06-18 at 1 18 34 pmscreen shot 2018-06-18 at 1 18 19 pm

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/resources/ResourceDetails.tsx
+++ b/app/scripts/modules/kubernetes/src/v2/resources/ResourceDetails.tsx
@@ -1,0 +1,121 @@
+import * as React from 'react';
+import * as moment from 'moment';
+import { get } from 'lodash';
+import { UISref } from '@uirouter/react';
+import { IManifest, Spinner, CloudProviderLogo, CollapsibleSection, AccountTag, timestamp } from '@spinnaker/core';
+import { KubernetesManifestService } from '../manifest/manifest.service';
+import { ManifestEvents } from 'kubernetes/v2/pipelines/stages/deployManifest/react/ManifestEvents';
+import { ManifestLabels } from 'kubernetes/v2/manifest/ManifestLabels';
+import { ManifestAnnotations } from 'kubernetes/v2/manifest/ManifestAnnotations';
+import { ManifestCondition } from 'kubernetes/v2/manifest/status/ManifestCondition';
+
+export interface IKubernetesResourceDetailsProps {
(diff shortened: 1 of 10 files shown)


sbwsg:
Do we want to render `data` fields such as those found in configmap and secrets?  I've avoided this here because I assume that showing a secret's contents would pose a security problem.

app/scripts/modules/kubernetes/src/v2/resources/ResourceDetails.tsx
```diff
@@ -0,0 +1,121 @@
...
+      <div className="details-panel">
+        <div className="header">
+          <div className="close-button">
+            <a className="btn btn-link">
+              <UISref to="^">
+                <span className="glyphicon glyphicon-remove" />
+              </UISref>
+            </a>
+          </div>
+          {(this.state.loading && (
```
danielpeach: A ternary might be a little cleaner here

lwander:
> Do we want to render data fields such as those found in configmap and secrets? I've avoided this here because I assume that showing a secret's contents would pose a security problem.

Yeah let's hold off on `data` until we have a 1st class configmap UI which we can make more configurable to avoid these sorts of problems.

lwander:
In that URL, why not use `manifests` instead of `kubernetesResourceDetails`? There are other manifest based providers (albiet internally) which aren't exactly kubernetes that can rely on this.

lwander:
One more question -- how do users get linked to this? Does it work from the search bar?

sbwsg:
@lwander All manifest resources now show a Details link from executions, matching the behaviour of Deployments and ReplicaSets.

lwander:
Looks great

PR #5473, opened 2018-06-20
feat(pipeline): Support `runAsUser` for implicit pipeline triggers

ajordens:
Previously these pipelines inherited the authentication context of their
parent pipeline.

spinnaker/orca#2295 has altered this behavior.


ajordens:
image

PR #5478, opened 2018-06-21
feat(notification): Add support for Google Chat.

bobocandys:
Add a new notification type Google Chat.

This is part of the issue spinnaker/spinnaker#2936

### DIFF
--- a/app/scripts/modules/core/src/notification/types/googlechat/additionalFields.html
+++ b/app/scripts/modules/core/src/notification/types/googlechat/additionalFields.html
@@ -1,0 +1,12 @@
+<div class="form-group row">
+  <div class="col-sm-3 sm-label-right">Chat Webhook URL</div>
+  <div class="col-sm-9">
+    <input type="text"
+           name="address"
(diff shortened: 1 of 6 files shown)


app/scripts/modules/core/src/notification/types/googlechat/additionalFields.html
```diff
@@ -0,0 +1,12 @@
+<div class="form-group row">
+  <div class="col-sm-3 sm-label-right">Chat Webhook URL</div>
+  <div class="col-sm-9">
+    <input type="text"
+           name="address"
+           class="form-control input-sm "
+           ng-model="vm.notification.address"
+           placeholder="URL starts with https://chat.googleapis.com/v1/spaces/"
+           ng-required="!vm.notification.cc.length">
```
sbwsg: i think you can replace this with just `required`.  The `cc` field is email-specific I believe.

jtk54:
Depends on https://github.com/spinnaker/echo/pull/282.

PR #5480, opened 2018-06-22
Revert "feat(build): add cache bust for builds"

danielpeach:
Reverts spinnaker/deck#5479

See https://github.com/spinnaker/spinnaker/issues/2942

This is more complicated than I thought. The problem is that it adds a hash to `settings.js` that Halyard can't know about beforehand. I can't see a way to prevent `settings.js` from receiving a hash.

Maybe someone with more Webpack knowledge (@christopherthielen ) can think of a solution?

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -30,7 +30,7 @@ function configure(env, webpackOpts) {
     },
     output: {
       path: path.join(__dirname, 'build', 'webpack', process.env.SPINNAKER_ENV || ''),
-      filename: `[name]${IS_PRODUCTION ? '.[chunkhash]' : ''}.js`,
+      filename: '[name].js',
     },
     devtool: IS_PRODUCTION ? 'source-map' : 'eval',
     optimization: {


christopherthielen:
FWIW, we hash our assets in nflx deck builds, but we completely avoid using the hashed settings.js that comes from the deck repo.  We have different settings.js for each environment we deploy to and use a init.d script to move the files to the proper locations.  This information doesn't solve the problem at hand though. 

How does halyard interact with settings.js?

sbwsg:
The HtmlWebpackPlugin has a `hash` config param that will give each injected script tag a cache-busting query param.  Here's some example output in index.html after a `yarn build` using the param:

```html
<script type="text/javascript" src="settings.js?60126dc51605e31ce42c"></script><script type="text/javascript" src="settings-local.js?60126dc51605e31ce42c"></script><script type="text/javascript" src="vendors~app.js?60126dc51605e31ce42c"></script><script type="text/javascript" src="app.js?60126dc51605e31ce42c"></script>
```

That hash changes with every run of webpack regardless of any code changes.  Does this fit the bill or do we want the hash to be dependent on the contents of the files?

danielpeach:
The problem is that Halyard writes its own `settings.js` into `/opt/deck/html` and assumes that `index.html` is pointing to `settings.js` (that exact string, without a hash). If we add a hash to `settings.js`, how could Halyard know that hash without looking inside Deck's container image (or .deb, or whatever)?

PR #5483, opened 2018-06-26
fix(core): Fixed cluster scroll jump

archana-s:
Cluster page scrolls even when the user clicks on a cluster.

We set up scroll to specific server group if it is deep linked. this is set as a state in react. However it is never reset. As a result, even when the user clicks on a different server group, we update the route, but the state still has the previous server group set to scroll. So the page always scrolls to the previous server group. 
The current fix sets the ref as a state. This is because componentDidMount is called before the list component is loaded. So we need the ref and need to check when the component is updating and set it once the very first time to scroll.

### DIFF
--- a/app/scripts/modules/core/src/cluster/AllClustersGroupings.tsx
+++ b/app/scripts/modules/core/src/cluster/AllClustersGroupings.tsx
@@ -6,7 +6,6 @@ import { Subscription } from 'rxjs';
 import { ReactInjector } from 'core/reactShims';
 import { Application } from 'core/application';
 import { ClusterPod } from 'core/cluster/ClusterPod';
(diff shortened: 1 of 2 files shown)


jrsquared:
`This is because componentDidMount is called before the list component is loaded.` Are you sure this is true? `AllClusterGroupings.componentDidMount()` should never get fired until `List` is mounted...

archana-s:
@jrsquared  Resorted to this because of List not being rendered before component mounting. Component mounts when the spinner goes up. 

"This is because componentDidMount is called before the list component is loaded. Are you sure this is true? AllClusterGroupings.componentDidMount() should never get fired until List is mounted..."

archana-s:
give it a go now

jrsquared:
way nicer. love it.

PR #5489, opened 2018-06-27
fix(amazon/loadBalancer): Fix DNS link in target group to match load balancer

jrsquared:
@emjburns


emjburns:
🎈 🎈 woooooow   🎊 LGTM

PR #5494, opened 2018-06-28
feat(core): Adding option to fail pipeline on failed expressions

jeyrschabu:
- adds a UI control to set `failOnFailedExpressions=true`


jeyrschabu:
UI bits for https://github.com/spinnaker/orca/pull/2303

PR #5497, opened 2018-06-29
feat(core): customizing maximum number of pipelines displayed

shraykay:
prefer to have this customizable so we can see further back

### DIFF
--- a/app/scripts/modules/core/src/config/settings.ts
+++ b/app/scripts/modules/core/src/config/settings.ts
@@ -92,6 +92,7 @@ export interface ISpinnakerSettings {
   };
   gateUrl: string;
   gitSources: string[];
+  maxPipelineAge: number;
   notifications: INotificationSettings;
   pagerDuty?: {
     accountName?: string;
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/config/settings.ts
```diff
@@ -92,6 +92,7 @@ export interface ISpinnakerSettings {
   };
   gateUrl: string;
   gitSources: string[];
+  maxPipelineAge: number;
```
sbwsg: Please add the unit of time to the variable name, e.g. `maxPipelineAgeDays`

shraykay:
@sbwsg amended!

halconfig/settings.js
```diff
@@ -98,6 +99,7 @@ window.spinnakerSettings = {
   defaultTimeZone: timezone, // see http://momentjs.com/timezone/docs/#/data-utilities/
   defaultCategory: 'serverGroup',
   defaultInstancePort: 80,
+  maxPipelineAgeDays: 14,
```
sbwsg: ah, i think this would need to be `maxPipelineAgeDays: maxPipielineAgeDays` so that it makes use of the injected variable on line 25.  Is that right, @lwander?

shraykay: ah whoops, thanks! need more coffee!

sbwsg:
Thanks!

PR #5502, opened 2018-07-02
fix(provider/kubernetes): Remove patch manifest stage v1 provider

ezimanyi:
Looks like this was accidentally applied to both v1 and v2, rather than just v2.


sbwsg:
nice!

PR #5504, opened 2018-07-03
feat(provider/kubernetes): add UID to instances

sbwsg:
When instances are returned from Clouddriver via the server groups
or loadbalancer endpoints the serialized InstanceViewModel or
LoadBalancerInstance objects use an instance's `name` as its `id`
in the response.

For Kubernetes instances the `name` is not globally unique.  This
resulted in UI problems such as clicking on one instance in the
infrastructure view leading to two instances being selected.

Clouddriver has been updated to return a `uid` field that is the
same as `id` for all providers except Kubernetes, which sets it
to the resource's UUID.  This new `uid` field needs to be used
in Deck wherever instance uniqueness is communicated or tested,
such as in URLs or during cluster filtering.  The `id` field remains
useful to surface as an instance's human-readable name.

In the below screenshot, two pods have been created in different
clusters of the same application.  Previously, selecting one of them
would have lead to both of them being highlighted and the Details
view displaying incorrect information.  Now, selecting one of them
leaves only one highlighted, and the details displayed are always
correct.

screen shot 2018-07-03 at 9 24 51 am

Also tagging @spinnaker/netflix-reviewers here since this change
modifies core and is not a part of the codebase I'm particularly
familiar with.


app/scripts/modules/core/src/loadBalancer/LoadBalancerClusterContainer.tsx
```diff
@@ -44,7 +44,11 @@ export class LoadBalancerClusterContainer extends React.Component<ILoadBalancerC
         {showServerGroups && ServerGroups}
         {!showServerGroups &&
           showInstances && (
-            <LoadBalancerInstances serverGroups={loadBalancer.serverGroups} instances={loadBalancer.instances} />
+            <LoadBalancerInstances
```
sbwsg: As far as I can see [`LoadBalancerInstance`](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/model/LoadBalancerInstance.java)s from Clouddriver's loadbalancers API do not include a cloudProvider field.  The cloudProvider field is used when the user clicks on an instance fronted by an LB ([here](https://github.com/spinnaker/deck/pull/5504/files#diff-d756080253c17bf1965714e31e06e50eR69)).  I'm not entirely sure how this works for other providers but threading the `cloudProvider` through from here down to `Instances.tsx` fixes the onClick handler.

erikmunson:
While this seems like a reasonable diff for moving to `uid`, I'm not confident adding a `uid` field and keeping `id` non-unique is going to be a long-term positive (though it does stop the pain today).

If we effectively stop considering `id` as unique — and instead as a human-readable name — I think we'll see the degree of confusion when touching instance-related code increase in tandem with the likelihood of introducing subtle bugs like the one this PR fixes. In general it's safe to assume uniqueness across the majority of API data Deck handles when using a field named `id`. Adding another permutation of that contract feels to me like a short-term win that will turn into a longer-term liability. It also means that, realistically, the scope of this change is all code that touches instance ids across all providers (since inconsistent use of `id` vs `uid` will likely make this problem even worse).

Are you open to alternatives? The one that immediately comes to mind is switching `id` to a UUID for k8s, and then adding a `name` field that has the display name (then for other providers `id` and `name` would be identical). I realize that does change the contract of `id` in k8s-land for the short-term, but I think that strategy leaves us in a much more sensical, less human-error-prone state over the long-term. It also has the benefit of effectively scoping this field migration to k8s and core rather than explicitly needing other providers to migrate for consistency (though I'm sure they would over time).

sbwsg:
Good point on the field naming and future pain potential.  I'll spend some time seeing how clouddriver-kubernetes deals with the ID change and take that approach if everything seems sound.

PR #5506, opened 2018-07-03
feat(instances): support name field on IInstance

sbwsg:
DONT MERGE until the parallel PR against clouddriver has been approved + merged (https://github.com/spinnaker/clouddriver/pull/2756)

Enables a provider (in this case, k8s v2) to return both an ID as well as a more human-friendly name in response to a request for instances.  Uses the name (with ID as fallback) wherever an instance is surfaced in the UI.  There's no visible change to the way deck works after this but Stateful Sets should behave as expected now.

Also: fixes an issue where clicking on an Instance in the Load Balancers view would not show details for that instance and would error out because the `provider` field was missing from the state transition parameters.

Fixes https://github.com/spinnaker/spinnaker/issues/2979


app/scripts/modules/core/src/domain/IInstance.ts
```diff
@@ -10,6 +10,7 @@ export interface IInstance {
   id: string;
   launchTime: number;
   loadBalancers?: string[];
+  name?: string;
```
erikmunson: what's your opinion on having `name` be non-nullable and just mirroring the value of `id` if clouddriver doesn't have a distinct human readable name to offer?

sbwsg: Ah, good point.  I've updated the IInstance interface to have a non-null `name`, and I've updated clouddriver to make sure both Load Balancer instances and Server Group instances return ID as name if a human-readable one isn't provided.

PR #5507, opened 2018-07-05
fix(trigger/webhook): add runAsUser to webhook

ethanfrogers:
adds runAsUser to the webhook trigger.

@lwander @danielpeach PTAL. I'll submit a patch for 1.8 once merged. cc @kevinawoo 

fixes spinnaker/spinnaker#2588

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/triggers/webhook/webhook.trigger.ts
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/webhook/webhook.trigger.ts
@@ -2,10 +2,18 @@ import { IController, module } from 'angular';
 
 import { IWebhookTrigger } from 'core/domain';
 import { Registry } from 'core/registry';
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/config/triggers/webhook/webhook.trigger.ts
```diff
@@ -24,7 +32,14 @@ module(WEBHOOK_TRIGGER, [])
       controller: 'WebhookTriggerCtrl',
       controllerAs: 'ctrl',
       templateUrl: require('./webhookTrigger.html'),
-      validators: [],
+      validators: [
+        {
+          type: 'serviceAccountAccess',
+          message: `You do not have access to the service account configured in this pipeline's Travis trigger.
```
danielpeach: s/Travis/webhook

ethanfrogers: smh, nice catch.

PR #5511, opened 2018-07-06
fix(trigger/webhook): fix lint issue - 1.8

ethanfrogers:


ethanfrogers:
@skim1420 FYI
@danielpeach PTAL

PR #5515, opened 2018-07-09
fix(loading): Surface error connecting to gate on applications screen

ezimanyi:
New users frequently end up in a situation where deck is unable to communicate with their gate endpoint. Currently this manifests itself as an infinite "Loading" spinner on the Applications screen (even though we actually give up after three attempts to connect to gate). Let's make this error more obvious to users by surfacing an error message instead of a continuous loading message.


ezimanyi:
Hopefully this will, even slightly, improve a pain point for new users.  Particularly looking for feedback on:
* The text of the error message(s)
* Whether there's a better way of implementing my React change (as this is my first time writing React).

duftler:
A new user may not know what gate is or what the configured endpoint is. Might be helpful to also include this information directly in the error message. Would be ideal if they could be pointed to a location with troubleshooting steps.

ezimanyi:
Made the following changes:
* Re-throw the errors that were caught so they still show up in the console for debugging purposes
* Added a link to our FAQ page, where I'll include some more details on debugging this error

I looked into having the error message include either the gate endpoint or a string representation of the actual error, but there wasn't a really clean way of doing that.  (At the point of catching the error, it could be any type of error, so digging into the `$http`-specific fields from the rendering code wouldn't be great.  Alternatively we could have the `ApiService` replace the error with a string, but then some information would be lost.)  It's definitely possible to improve the error handling here (and maybe even have a structured error class that we throw, and that supports a `.toString()`) but this seemed to be getting a bit out of scope on this change.

I'll have the FAQ direct the user to look in their browser console to see the exact error.

ezimanyi:
New FAQ entry: https://github.com/spinnaker/spinnaker.github.io/pull/900

PR #5517, opened 2018-07-09
refactor(serviceAccounts): Use an interface for service accounts.

dibyom:
The Gate API now returns the entire service account instead of just the roles.
This will eventually allow users to select roles for their pipelines.

Depends on spinnaker/gate/pull/573


dibyom:
May not need this.

PR #5519, opened 2018-07-10
feat(wercker): add wercker trigger and stage

shihchang:
Add WerckerTrigger and WerckerStage

Related PRs
https://github.com/spinnaker/halyard/pull/960
https://github.com/spinnaker/igor/pull/297

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.html
+++ b/app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.html
@@ -1,0 +1,121 @@
+<div class="form-horizontal">
+  <div class="form-group">
+    <label class="col-md-2 col-md-offset-1 sm-label-right">Master</label>
+    <div class="col-md-6">
+      <p class="form-control-static"
+         ng-if="$ctrl.viewState.masterIsParameterized">{{$ctrl.stage.master}}</p>
+      <ui-select class="form-control input-sm"
+                 ng-if="!$ctrl.viewState.masterIsParameterized"
+                 ng-model="$ctrl.stage.master">
+        <ui-select-match placeholder="Select a Wercker Master...">{{$select.selected}}</ui-select-match>
+        <ui-select-choices repeat="master in $ctrl.masters | filter: $select.search">
+          <span ng-bind-html="master | highlight: $select.search"></span>
+        </ui-select-choices>
+      </ui-select>
+    </div>
+    <div class="col-md-1 text-center" ng-if="!$ctrl.viewState.masterIsParameterized">
+      <a href
+         ng-click="$ctrl.refreshMasters()"
+         tooltip-placement="right"
+         uib-tooltip="{{$ctrl.viewState.mastersRefreshing ? 'Masters refreshing.' : 'Refresh masters list' }}"
+        >
+              <span ng-class="{'fa-spin':$ctrl.viewState.mastersRefreshing}"
+                    class="fa fa-sync-alt"></span>
+      </a>
+    </div>
+  </div>
+  
+  <div class="form-group">
+    <label class="col-md-2 col-md-offset-1 sm-label-right">Application</label>
+    <div class="col-md-6">
+      <p class="form-control-static"
+         ng-if="!$ctrl.stage.master">(Select a master)</p>
+      <p class="form-control-static"
+         ng-if="$ctrl.viewState.masterIsParameterized || $ctrl.viewState.appIsParameterized">{{$ctrl.stage.app}}</p>
+      <div ng-if="$ctrl.stage.master && $ctrl.viewState.appsLoaded">
+        <ui-select class="form-control input-sm"
+                   ng-if="!$ctrl.viewState.masterIsParameterized && !$ctrl.viewState.appIsParameterized"
+                   ng-model="$ctrl.stage.app">
+          <ui-select-match placeholder="Select an Application...">{{$select.selected}}</ui-select-match>
+          <ui-select-choices repeat="app in $ctrl.apps | filter: $select.search"><span ng-bind-html="app | highlight: $select.search"></span></ui-select-choices>
+        </ui-select>
+      </div>
+      <div class="horizontal center" ng-if="$ctrl.stage.master && !$ctrl.viewState.appsLoaded">
+        <loading-spinner size="'small'"></loading-spinner>
+      </div>
+    </div>
+    <div class="col-md-1 text-center" ng-if="!$ctrl.viewState.masterIsParameterized && !$ctrl.viewState.appIsParameterized">
+      <a href
+         ng-click="$ctrl.refreshApps()"
+         tooltip-placement="right"
+         uib-tooltip="{{$ctrl.viewState.appsRefreshing ? 'Applications refreshing.' : 'Refresh application list' }}"
+        >
+              <span ng-class="{'fa-spin':$ctrl.viewState.appsRefreshing}"
+                    class="fa fa-sync-alt"></span>
+      </a>
+    </div>
+  </div>
+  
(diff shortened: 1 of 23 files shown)


app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.html
```diff
@@ -0,0 +1,188 @@
...
+         tooltip-placement="right"
+         uib-tooltip="{{viewState.jobsRefreshing ? 'Jobs refreshing.' : 'Refresh job list' }}"
+        >
+              <span ng-class="{'fa-spin':viewState.jobsRefreshing}"
+                    class="fa fa-sync-alt"></span>
+      </a>
+    </div>
+  </div>
+ 
+<!--
```
sbwsg: Ideally dead code should be removed rather than commented out.

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.html
```diff
@@ -0,0 +1,188 @@
...
```
sbwsg: We don't have copyright notices in any other templates in deck so I'm not sure about including these here. I also notice that you didn't include one in `werckerExecutionDetails.html`.  If these aren't necessary then I suggest removing them.

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.js
```diff
@@ -0,0 +1,170 @@
+'use strict';
```
sbwsg: I suggest renaming this file from `.js` to `.ts` and then using the lint output generated by `yarn build`, `yarn tslint` or `yarn start` to guide you through linting your code.

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.js
```diff
@@ -0,0 +1,170 @@
...
+      updateJobsList();
+    };
+
+    function updateJobsList() {
+      if ($scope.stage && $scope.stage.app) {
+        $scope.pipelines = [];
+        $scope.viewState.jobsLoaded = true;
+        $scope.viewState.jobsRefreshing = false;
+        if ($scope.jobs) {
+            var pipelines = {};
```
sbwsg: Prefer `const`/`let` over var declarations.

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.module.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+const angular = require('angular');
+
+import { STAGE_CORE_MODULE } from '../core/stage.core.module';
+import { TIME_FORMATTERS } from 'core/utils/timeFormatters';
+
+module.exports = angular.module('spinnaker.core.pipeline.stage.wercker', [
+  require('./werckerStage.js').name,
```
sbwsg: This branch doesn't currently build for me locally and it looks like TravisCI is having trouble as well.  I believe the reason is that these require statements don't need to include the file extension.  Using output from the `yarn tslint` and `yarn build` commands as a guide can help in tracking down issues.

app/scripts/modules/core/src/pipeline/config/triggers/wercker/werckerTrigger.html
```diff
@@ -0,0 +1,101 @@
...
+         ng-click="werckerTriggerCtrl.refreshJobs()"
+         tooltip-placement="right"
+         uib-tooltip="{{viewState.jobsRefreshing ? 'Jobs refreshing.' : 'Refresh job list' }}"
+        >
+              <span ng-class="{'fa-spin':viewState.jobsRefreshing}"
+                    class="fa fa-sync-alt"></span>
+      </a>
+    </div>
+  </div>
+  <!--
```
sbwsg: Again it's preferable to delete rather than comment out dead code.

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.html
shihchang: Did you mean rename all the the js file to ts? I got "Module not found: Error: Can't resolve './wercker/werckerTrigger.module.js' in '/Users/scchen/wrkspin/deck-ui/app/scripts/modules/core/src/pipeline/config/triggers'
Or which js file should I rename?

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.module.js
shihchang: fixed some broken tests and file name, yarn run test should be good now.

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.html
sbwsg: Yes exactly - rename all the js files to ts.  Then in the places you've used `require()` you don't need to include `.js` in the filename.  This should fix the "Module not found" errors.

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.js
shihchang: It may not be just simply renaming all the js to ts as we are using jeknins trigger and stage as examples to create wercker's, and those module.js files
are using angular.module('spinnaker.core.pipeline.config.trigger.jenkins'...) to config the module. It results in tsModuleNotFoundError if renaming the file from js to ts without changing the codes to, for example, travis ways of module.config.

I did run yarn run build without error in the new codes.

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.js
shihchang: fixed all the var to let

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.js
sbwsg: Yup that makes sense - using the travis trigger as an example would be a good way to go!

shihchang:
@markxnelson

Related PRs:

spinnaker/igor#297
spinnaker/deck#5519
spinnaker/echo#295
spinnaker/gate#574
spinnaker/orca#2313

app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.ts
```diff
@@ -0,0 +1,232 @@
+import { IController, IScope, module } from 'angular';
+import { IModalService } from 'angular-ui-bootstrap';
+import * as moment from 'moment';
+
+import { Registry } from 'core/registry';
+// import { SETTINGS } from 'core/config/settings';
```
sbwsg: remove dead code

app/scripts/modules/core/src/pipeline/config/triggers/wercker/WerckerTriggerTemplate.tsx
```diff
@@ -0,0 +1,158 @@
...
+      buildsLoading: true,
+      loadError: false,
+    });
+
+    if (trigger.buildNumber) {
+      this.updateSelectedBuild(trigger.buildInfo);
+    }
+
+    // do not re-initialize if the trigger has changed to some other type
+    if (trigger.type !== 'wercker') {
```
sbwsg: Check equality against `BuildServiceType.Wercker` enum value here instead of string literal `'wercker'`.  Add `BuildServiceType` to import on line 8.

PR #5524, opened 2018-07-12
feat(kubernetes): Allow deploying containers specified as artifacts

ezimanyi:
Docker artifacts configured in a stage now appear in the 'Containers' and 'Init Containers' field when deploying using the Kubernetes V1 provider.

### DIFF
--- a/app/scripts/modules/kubernetes/src/serverGroup/configure/configuration.service.js
+++ b/app/scripts/modules/kubernetes/src/serverGroup/configure/configuration.service.js
@@ -22,7 +22,7 @@ module.exports = angular
       const containers = command.containers.concat(command.initContainers || []);
       let queries = containers
         .filter(c => {
-          return !c.imageDescription.fromContext;
+          return !c.imageDescription.fromContext && !c.imageDescription.fromArtifact;
         })
         .map(c => {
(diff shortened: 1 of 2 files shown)


ezimanyi:
image

ezimanyi:
This just wires up the basic functionality---two additional pieces to come in a separate PR:
* Updating the configuration when an artifact is deleted
* Showing the artifact as 'Consumed' in the stage details

app/scripts/modules/kubernetes/src/serverGroup/configure/configuration.service.js
```diff
@@ -172,7 +174,11 @@ module.exports = angular
       const validContainers = [];
       const invalidContainers = [];
       containers.forEach(function(container) {
-        if (container.imageDescription.fromContext || container.imageDescription.fromTrigger) {
+        if (
```
sbwsg: Small style nit, this can be reduced to:

```javascript
const { fromContext, fromTrigger, fromArtifact } = container.imageDescription;
if (fromContext || fromTrigger || fromArtifact) {
```

ezimanyi: Good idea; made the change

PR #5525, opened 2018-07-12
docs(artifacts): GitLab example URL typo

kskewes:
1. Add missing slash after `projects` in Gitlab example URL

2. Change example URL from org/repo to Gitlab project :id to match upstream docs.

**Note:**
I'm less certain about 2, as other Spinnaker git sources use a $ORG%2F$REPO style URL path.

- The [Gitlab API v4 docs](https://docs.gitlab.com/ee/api/repository_files.html#get-raw-file-from-repository) say to use a Gitlab project `:id` (int?) value in the URL rather than $ORG%2F$REPO. Should we match this?
- Gitlab project `:id` may remain the same even if the project is renamed or transferred. 
- Both URL paths can be used to retrieve the desired file and I acknowledge it is a lot more intuitive to use $ORG%2F$REPO. 

@ethanfrogers - Did you see $ORG%2F$REPO documented somewhere?

[Gitlab Projects](https://docs.gitlab.com/ee/api/projects.html) - all referencing `:id`
[Initial Deck commit for Gitlab support](https://github.com/spinnaker/deck/pull/5357/commits/3f64bee7fc2087a148dd71e7944551c69f68a5e1#diff-e9a1aae4dcf38984a075ff0aac16cac5R138)


ethanfrogers:
@kskewes yep, it's here: https://docs.gitlab.com/ee/api/#namespaced-path-encoding

even though it's not super user friendly i felt like remembering `myorg%2Fmyrepo` was easier to remember than a number. it supports both so it probably best to call that out instead of changing it one way or the other.

jrsquared:
Is it really `/projects/projects/`?

kskewes:
Thanks @jrsquared !

PR #5535, opened 2018-07-20
feat(wercker): add wercker trigger and stage (#5519)

sbwsg:
This PR reintroduces @shihchang 's work to add Wercker support that I [reverted here](https://github.com/spinnaker/deck/pull/5534).

Merge this PR once these service changes are merged:

https://github.com/spinnaker/igor/pull/297
https://github.com/spinnaker/echo/pull/295
https://github.com/spinnaker/gate/pull/574
https://github.com/spinnaker/orca/pull/2313

@shihchang please confirm with a message here that the code in this PR is as you expect it to be.  Many thanks and apologies again for the merge confusion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@ Thumbs.db
 *.iml
 .idea/
 .vscode/
+.classpath
+.project
+.settings/
 
 # Node Files #
 npm-debug.log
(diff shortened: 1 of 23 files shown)


shihchang:
Confirmed. Or I can create another PR if that is easier for you.

PR #5536, opened 2018-07-20
feat(authz/config): Add a pipeline roles section to pipeline config.

dibyom:
Deck part of https://github.com/spinnaker/spinnaker/issues/3037

Here is what this looks like:
screenshot from 2018-07-20 15-58-46

### DIFF
--- a/app/scripts/modules/core/src/config/settings.ts
+++ b/app/scripts/modules/core/src/config/settings.ts
@@ -47,6 +47,7 @@ export interface IFeatures {
   pipelineTemplates?: boolean;
   versionedProviders?: boolean;
   travis?: boolean;
+  managedServiceAccounts?: boolean;
   [key: string]: any;
 }
 
--- a/app/scripts/modules/core/src/help/help.contents.ts
+++ b/app/scripts/modules/core/src/help/help.contents.ts
@@ -402,6 +402,17 @@ const helpContents: { [key: string]: string } = {
   'pipeline.config.parameter.description': `(Optional): if supplied, will be displayed to users as a tooltip
       when triggering the pipeline manually. You can include HTML in this field.`,
   'pipeline.config.failOnFailedExpressions': `When this option is enabled, the stage will be marked as failed if it contains any failed expressions`,
+  'pipeline.config.roles.help': `
+    <p> When the pipeline is triggered using an automated trigger, these roles will be used to decide if the pipeline has permissions to access a protected application or account.</p>
+    <ul>
+    <li>
+    To read from a protected application or account, the pipeline must have at least one role that has read access to the application or account.
+    </li>
+    <li>
+    To write to a protected application or account, the pipeline must have at least one role that has write access to the application or account.
+    </li>
+    </ul>
+    <p><strong>Note:</strong> To prevent privilege escalation vulnerabilities, a user must be a member of <strong>all</strong> of the groups specified here in order to view, modify, and execute the pipeline.</p>`,
 };
 
 Object.keys(helpContents).forEach(key => HelpContentsRegistry.register(key, helpContents[key]));
--- a/app/scripts/modules/core/src/pipeline/config/triggers/pipelineRoles.component.ts
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/pipelineRoles.component.ts
@@ -1,0 +1,48 @@
+import { IController, module } from 'angular';
+import { AuthenticationService } from 'core/authentication';
+
+class PipelineRolesController implements IController {
+  public pipeline: any;
+  public allowedRoles: string[];
+
+  public $onInit(): void {
+    this.allowedRoles = AuthenticationService.getAuthenticatedUser().roles;
+    this.pipeline.roles = this.pipeline.roles || [];
+  }
+}
+
+class PipelineRolesComponent implements ng.IComponentOptions {
+  public template = `
+    <div class="form-horizontal panel-pipeline-phase">
+      <div class="form-group row">
+        <div class="col-md-10">
+          <div class="row">
+            <label class="col-md-3 sm-label-right">
+              Permissions
+              <help-field key="pipeline.config.roles.help"></help-field>
+            </label>
+            <div class="col-md-9">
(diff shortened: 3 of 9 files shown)


app/scripts/modules/core/src/help/help.contents.ts
```diff
@@ -402,6 +402,9 @@ const helpContents: { [key: string]: string } = {
   'pipeline.config.parameter.description': `(Optional): if supplied, will be displayed to users as a tooltip
       when triggering the pipeline manually. You can include HTML in this field.`,
   'pipeline.config.failOnFailedExpressions': `When this option is enabled, the stage will be marked as failed if it contains any failed expressions`,
+  'pipeline.config.roles.help': `
+  <p>The roles specified will be used for making authorization decisions when the pipeline is triggerd using an automated trigger.
```
danielpeach: Please indent here.

app/scripts/modules/core/src/pipeline/config/triggers/pipelineRoles.component.ts
```diff
@@ -0,0 +1,46 @@
...
+  public pipeline: any;
+  public allowedRoles: string[];
+
+  public $onInit(): void {
+    this.allowedRoles = AuthenticationService.getAuthenticatedUser().roles;
+    this.pipeline.roles = this.pipeline.roles || [];
+  }
+
+  public toggleCheck(role: string) {
+    if (this.pipeline.roles.indexOf(role) === -1) {
```
danielpeach: It's a little more idiomatic to use `if (this.pipeline.roles.includes(role))`, though the boolean is flipped in that case.

app/scripts/modules/core/src/pipeline/config/triggers/pipelineRoles.component.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+class PipelineRolesComponent implements ng.IComponentOptions {
+  public template = `
+  <div class="row">
+    <div class="col-md-11 col-md-offset-1">
+      <div class="row checkbox" ng-repeat="role in permissionCtrl.allowedRoles">
+        <label>
+          <input type="checkbox"
+          ng-checked="permissionCtrl.pipeline.roles.indexOf(role) > -1"
```
danielpeach: Please indent this line and the line below.

app/scripts/modules/core/src/pipeline/config/triggers/runAsUserSelector.component.ts
```diff
@@ -1,25 +1,38 @@
-import { module } from 'angular';
+import { IComponentOptions, IController, module } from 'angular';
+import { SETTINGS } from 'core/config/settings';
 
-class RunAsUserSelectorComponent implements ng.IComponentOptions {
+class RunAsUserSelectorController implements IController {
+  public enabled = true;
+
+  public $onInit(): void {
+    this.enabled = SETTINGS.feature['fiatEnabled'] && !SETTINGS.feature['managedServiceAccounts'];
```
danielpeach: You don't need the bracket notation - can just get the key `SETTINGS.feature.managedServiceAccounts`. It's possible that the TS compiler will complain, but in that case you should add `managedServiceaccounts` to `settings.ts`, which types `settings.js`.

app/scripts/modules/core/src/pipeline/config/triggers/pipelineRoles.component.ts
```diff
@@ -0,0 +1,46 @@
...
+          ng-click="permissionCtrl.toggleCheck(role)">
+          <strong>{{role}}</strong>
```
danielpeach: I think this might be better as a multiselect rather than a checkbox list. If a user has many roles (maybe > 10), this list will become unwieldy. Do a search for `<ui-select>` in this repo - any with the attribute `multiple` should be a good example. The multiselect will also allow users to search for their roles.

app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
```diff
@@ -18,6 +18,17 @@
       </div>
     </div>
   </page-section>
+  <render-if-feature feature="managedServiceAccounts">
+  <page-section key="roles" label="Pipeline Roles" visible="!pipeline.strategy" no-wrapper="true">
```
danielpeach: Please indent this section between the `render-if-feature` tags.

app/scripts/modules/core/src/config/settings.ts
```diff
@@ -47,6 +47,7 @@ export interface IFeatures {
   pipelineTemplates?: boolean;
   versionedProviders?: boolean;
   travis?: boolean;
+  managedServiceAccountsEnabled?: boolean;
```
danielpeach: This doesn't match the `settings.js` value you've configured - I think it should be `managedServiceAccounts`.

app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
danielpeach: I think this would be better positioned under `Automated Triggers` below, rather than as its own page section - what do you think? I'd put it at the top of the trigger configs.

dibyom:
Changed the roles selector to a multi-select and moved it under `Automated Triggers`. Only shows up if a trigger has been added:
screenshot from 2018-07-23 13-20-12

Also reworded the help text to make it clearer.
PTAL.

app/scripts/modules/core/src/pipeline/config/triggers/pipelineRoles.component.ts
```diff
@@ -0,0 +1,48 @@
...
+  public pipeline: any;
+  public allowedRoles: string[];
+
+  public $onInit(): void {
+    this.allowedRoles = AuthenticationService.getAuthenticatedUser().roles;
+    this.pipeline.roles = this.pipeline.roles || [];
+  }
+}
+
+class PipelineRolesComponent implements ng.IComponentOptions {
```
danielpeach: Ideally everything would be pipeline "permissions" or pipeline "roles" - but I am ok with that happening in a different PR.

PR #5538, opened 2018-07-23
feat(kubernetes): Add consumed artifacts to Kubernetes V1 provider

ezimanyi:


app/scripts/modules/kubernetes/src/serverGroup/artifactExtractor.ts
```diff
@@ -0,0 +1,40 @@
...
+  .factory('kubernetesServerGroupArtifactExtractor', function() {
+    function extractArtifacts(cluster: ICluster): string[] {
+      const containers = (cluster.containers || []).concat(cluster.initContainers || []);
+      return containers
+        .filter(c => c.imageDescription && c.imageDescription.fromArtifact)
+        .map(c => c.imageDescription.artifactId);
+    }
+
+    return {
+      extractArtifacts: extractArtifacts,
```
sbwsg: Can shorten to `return { extractArtfiacts };`

ezimanyi: I'm adding a field here in my next PR, so I'll make that change there.

ezimanyi:
https://github.com/spinnaker/spinnaker/issues/2962

PR #5544, opened 2018-07-23
feat(artifacts): Kubernetes V1 provider removes deleted artifacts

ezimanyi:


ezimanyi:
This ended up being a bit more refactoring than it initially seemed this morning---the reason being that for the Kubernetes V1 provider if an artifact is deleted, we can't just delete the field containing the artifact id---we need to delete the whole container to avoid the stage being in an inconsistent state.

I ended up inverting control here; the `ArtifactReferenceService` tells each stage which artifacts were deleted, and it's up to each stage to delete them.  (`ArtifactReferenceService` provides a utility that handles the simple cases, so only more complex stages need to actually implement anything.)  @sbwsg let me know if this refactor makes sense, or if this will make future work you may have planned more difficult.  (Ideally it would make it easier.)

ezimanyi:
Discussed with @sbwsg and just removed the gating to `gce` as the logic itself handles non-existing fields; this means the logic is a no-op now but will just work when artifacts are implemented on other providers.

ezimanyi:
https://github.com/spinnaker/spinnaker/issues/2962

PR #5555, opened 2018-08-01
fix(kubernetes): fixes package build

danielpeach:
@amanya


amanya:
I tested that now builds fine, thanks!

PR #5564, opened 2018-08-03
refactor(core): Convert CancelModal to react

alanmquach:
It doesn't look like anything else (other than Executions) is using the cancel modal, so seems like the old templates/controllers/services can be removed as well.


app/scripts/modules/core/src/cancelModal/CancelModal.tsx
```diff
@@ -0,0 +1,128 @@
...
+    });
+  };
+
+  private submitConfirmation = (values: ICancelModalValues): void => {
+    this.setState({ isSubmitting: true });
+    this.props.submitMethod(values.reason).then(this.close, this.showError);
+  };
+
+  public render() {
+    const { header, body, buttonText, cancelButtonText = 'Cancel' } = this.props;
```
jrsquared: Not used to the pattern of defaults here.

I typically like to see the `public defaultProps: Partial<ICancelModalProps> = { cancelButtonText: 'Cancel' }` pattern, just in case in the future we expect `cancelButtonText` to exist elsewhere than than render function.

app/scripts/modules/core/src/cancelModal/CancelModal.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+import { SubmitButton, ModalClose } from 'core';
+import { ReactModal } from 'core/presentation';
+
+export interface ICancelModalProps {
+  body?: string;
+  buttonText: string;
+  cancelButtonText?: string;
+  header: string;
+  submitMethod: (reason: string, force?: boolean) => ng.IPromise<any>;
```
jrsquared: `import { IPromise } from 'angular';`

app/scripts/modules/core/src/cancelModal/CancelModal.tsx
alanmquach: Ah right, default props. Thanks!

app/scripts/modules/core/src/cancelModal/CancelModal.tsx
```diff
@@ -0,0 +1,133 @@
+import * as React from 'react';
+import { Modal } from 'react-bootstrap';
+import { Field, FieldProps, Form, Formik } from 'formik';
+import { IPromise } from 'angular';
+
+import { SubmitButton, ModalClose } from 'core';
```
christopherthielen: fyi I think or cargo cult says to prefer to import from the submodule of core, i.e.,  `core/modal`

app/scripts/modules/core/src/cancelModal/CancelModal.tsx
```diff
@@ -0,0 +1,133 @@
...
+export class CancelModal extends React.Component<ICancelModalProps, ICancelModalState> {
+  public static defaultProps: Partial<ICancelModalProps> = {
+    cancelButtonText: 'Cancel',
+  };
+
+  public static confirm(props: ICancelModalProps): Promise<void> {
+    return ReactModal.show(CancelModal, props);
+  }
+
+  constructor(props: ICancelModalProps) {
```
christopherthielen: When you don't need to actually *do* anything in the constructor except set initial state, you can omit the constructor and just create a field.  I.e.:

```js
export class CancelModal {
  public state = { isSubmitting: false, error: false };
  ...
```

Note: what you have is fine, and we do it regularly, but I wanted to point out a different option you might not have known about

christopherthielen:
LGTM, nice work

app/scripts/modules/core/src/cancelModal/CancelModal.tsx
alanmquach: I like that, much cleaner. Thanks.

PR #5565, opened 2018-08-03
fix(google): Remove image region validation from find image stage

ezimanyi:
GCE images aren't associated with a region, so there is no field to enter one in the UI. We should remove the validation that warns that a region has not been selected.


ezimanyi:
Fixes https://github.com/spinnaker/spinnaker/issues/2146

PR #5568, opened 2018-08-06
feat(core): Export './AuthenticationInitializer'

ajordens:


ajordens:
discussed offline with @jrsquared

PR #5570, opened 2018-08-07
fix(artifacts): Handle Docker registries with port specifications

wjoel:
Similar to https://github.com/spinnaker/clouddriver/pull/2736, use "lastIndexOf(':')" so "localhost:5000/whatever:0.0.1" gets parsed as name "localhost:5000/whatever" and version "0.0.1" instead of name "localhost" and version "5000/whatever".

Add tests while we're here.


wjoel:
Not sure if I'm doing the tests right, couldn't figure out how to test it as an Angular 2 module since it's not exporting anything, just registering an `artifactKindConfig` (and I don't know Angular).

PR #5577, opened 2018-08-09
feat(core/search): Add "show all" links to each category

christopherthielen:
screenshot 2018-08-09 10 49 48

### DIFF
--- a/app/scripts/modules/core/src/search/global/GlobalSearchResults.tsx
+++ b/app/scripts/modules/core/src/search/global/GlobalSearchResults.tsx
@@ -1,19 +1,18 @@
-import * as React from 'react';
 import { UISref } from '@uirouter/react';
 
 import { SETTINGS } from 'core/config/settings';
 import { ISearchResult } from 'core/search/search.service';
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/search/global/GlobalSearchResults.tsx
```diff
@@ -39,36 +37,42 @@ export const GlobalSearchResults = ({
...
-          to="home.search"
-          // TODO: Setup route redirects to where this is unnecessary
-          params={searchVersion === 2 ? { key: query } : { q: query }}
-        >
-          <a className="expand-results" ref={seeMoreRef} onKeyDown={onItemKeyDown} onClick={onSeeMoreClick}>
-            See more results
-          </a>
-        </UISref>
-      </li>
+      <li key="seeMore" className="result" />
```
erikmunson: am i missing something here? why are these still hanging out?

christopherthielen: nice catch, thanks

PR #5581, opened 2018-08-10
fix(amazon/deploy): Make source capacity work again, other misc. fixes

erikmunson:
Some more fun fixes:

- The radio buttons to select whether to copy source capacity or use new capacity settings were not working for several reasons (both options were checking the same condition, no `value` was specified for either, and there was a missing arrow binding for the `onChange` callback).
- The header/nav label was hard coded to `Security Groups` rather than using `FirewallLabels`.
- When `disableImageSelection` was set, the wizard could never be submitted/completed because `amiName` wasn't present and we were validating for it's existence all the time.
- The `preferSourceCapacity` setting was missing an arrow binding, and also showed the select field as empty/not selected when it was set to the default, rather than showing the actual default.
- The loading spinner + text for the image search select box was an angular HTML string that didn't work anymore.

I also went in and switched some things to `private` just for fun.

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/configure/wizard/capacity/CapacitySelector.tsx
+++ b/app/scripts/modules/amazon/src/serverGroup/configure/wizard/capacity/CapacitySelector.tsx
@@ -12,12 +12,12 @@ export interface ICapacitySelectorProps {
 }
 
 export class CapacitySelector extends React.Component<ICapacitySelectorProps> {
-  public preferSourceCapacityOptions = [
-    { label: 'fail the stage', value: undefined },
+  private preferSourceCapacityOptions = [
+    { label: 'fail the stage', value: false },
     { label: 'use fallback values', value: true },
   ];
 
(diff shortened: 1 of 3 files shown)


app/scripts/modules/amazon/src/serverGroup/configure/wizard/capacity/CapacitySelector.tsx
```diff
@@ -12,12 +12,12 @@ export interface ICapacitySelectorProps {
 }
 
 export class CapacitySelector extends React.Component<ICapacitySelectorProps> {
-  public preferSourceCapacityOptions = [
-    { label: 'fail the stage', value: undefined },
+  private preferSourceCapacityOptions = [
+    { label: 'fail the stage', value: false },
```
erikmunson: apparently using `undefined` makes `react-select` show itself as unselected? 🤷‍♂️ I didn't dig deep enough to tell if I was missing something, since this seemed like a reasonable way to get the default to properly show.

PR #5586, opened 2018-08-13
feat(wercker): feature toggle for wercker stages

gardleopard:
This makes it possible to set feature.wercker to false. We do not want
to expose stages that cannot be used by our users.


app/scripts/modules/core/src/pipeline/config/stages/wercker/werckerStage.ts
```diff
@@ -5,6 +5,8 @@ import * as moment from 'moment';
 import { Registry } from 'core/registry';
 import { IgorService, BuildServiceType } from 'core/ci/igor.service';
 import { IJobConfig, IParameterDefinitionList, IStage } from 'core/domain';
+import { SETTINGS } from 'core/config/settings';
+
```
danielpeach: Extraneous space?

sbwsg:
Can you add an entry to `settings.js` and `halconfig/settings.js` as well to match the travis entries?

gardleopard:
I have tried now PTAL @sbwsg

PR #5598, opened 2018-08-13
feat(provider/kubernetes): hide artifact account selector when only one suitable account is available

sbwsg:
Hides the artifact account selector when the user only has one configured account for a given artifact's type.  Embedded artifacts provide the clearest example of this; the user is unable to configure their own embedded-account and so the account selector is never displayed for embedded/base64 artifacts. More generally if, for example, a user only has a single GCS artifact account configured then Deck will simply choose that account and won't provide a single-entry select box.

An artifact selected that only has one available account associated with its type:

screen shot 2018-08-13 at 5 06 27 pm

An artifact selected that has two available accounts associated with its type:

screen shot 2018-08-13 at 5 05 54 pm

### DIFF
--- a/app/scripts/modules/core/src/account/AccountService.ts
+++ b/app/scripts/modules/core/src/account/AccountService.ts
@@ -27,6 +27,11 @@ export interface IAccount {
   skin?: string;
 }
 
+export interface IArtifactAccount {
+  name: string;
+  types: string[];
+}
+
 export interface IAccountDetails extends IAccount {
   accountType: string;
   authorized: boolean;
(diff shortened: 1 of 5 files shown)


sbwsg:
This is intended as a step towards https://github.com/spinnaker/spinnaker/issues/3168

lwander:
We can push the manual association out of the client -- but the client will still need to know not to show the account in these cases.

danielpeach:
Yep, we'll still need to have most of this PR.

MPT users would appreciate not having to know about this artifact account.

lwander:
Behavior SGTM

PR #5599, opened 2018-08-13
fix(kubernetes): pick correct account when opening edit manifest wizard

danielpeach:
https://github.com/spinnaker/spinnaker/issues/3159


lwander:
Worth cherry picking?

PR #5604, opened 2018-08-14
refactor(instances): Convert instance-load-balancer-health to react

alanmquach:


app/scripts/modules/core/src/instance/loadBalancer/InstanceLoadBalancerHealth.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import { OverlayTrigger, Tooltip } from 'react-bootstrap';
+
+export interface InstanceLoadBalancer {
+  healthState?: string;
```
jrsquared: are all of these really optional?

alanmquach: The way the old code relied on it, I couldn't tell if anything was strictly required because there were fallbacks.

When the primary candidate is available, the fallbacks seem optional? Is there a way in TS to say "at least one of these"?

jrsquared: Yes, but it's ugly. Fine as is!

alanmquach:
Added some comments for context

PR #5612, opened 2018-08-14
feat(artifacts): filter set of available artifact kinds by configured artifact accounts

sbwsg:
Before, all artifact kinds are available regardless of which artifact accounts have been configured:

After, only the artifact kinds matching the set of configured accounts are available. In the below screenshot I only have a GCS credential configured. Base64 and Custom artifacts are credential-less and Docker artifacts are publicly accessible from docker hub:


sbwsg:
Now that I'm thinking about this I'm wondering if it's too restrictive.  Could a docker artifact be from a public repo that doesn't require an artifact account?

sbwsg:
@lwander @danielpeach this PR is ready for review now

lwander:
Nice

PR #5613, opened 2018-08-14
fix(loadBalancer): Use correct provider for titus server groups/instances

erikmunson:
The fact that there can be Titus server groups/instances underneath AWS load balancers in the hierarchy breaks some of the assumptions we originally made when building the LB view for AWS. This led to an end state where we tried to represent Titus entities using our Amazon components, which made the instance details pane make no sense (even though it rendered without issue).

I chose to patch this up by hydrating the provider onto instances when we get the provider for each server group, and switching things that were pulling providers off of target groups to use the server group instead (in places where the logic is scoped to a given server group). A couple of things that came to mind:

- It feels like it'd be nice to tack the provider onto the server group when it comes back from `/loadBalancers`, but hydrating on the client via the account is not horrible if done right.

- I started hydrating instances because we expect a provider to get tacked onto instances throughout the rest of the instance-related components (in fact the clusters view has never given `<Instances />` a separate `cloudProvider` prop), and it feels like relying on a parent association to get a provider is asking for trouble.

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/details/targetGroupDetails.html
+++ b/app/scripts/modules/amazon/src/loadBalancer/details/targetGroupDetails.html
@@ -91,7 +91,7 @@
               <a ui-sref="^.serverGroup({region: serverGroup.region,
                                           accountId: serverGroup.account,
                                           serverGroup: serverGroup.name,
-                                          provider: 'aws'})">
+                                          provider: serverGroup.cloudProvider})">
               {{serverGroup.name}}
               </a>
             </li>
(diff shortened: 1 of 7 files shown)


app/scripts/modules/amazon/src/loadBalancer/details/targetGroupDetails.html
```diff
@@ -91,7 +91,7 @@ <h3 class="horizontal middle space-between flex-1" select-on-dbl-click>
               <a ui-sref="^.serverGroup({region: serverGroup.region,
                                           accountId: serverGroup.account,
                                           serverGroup: serverGroup.name,
-                                          provider: 'aws'})">
+                                          provider: serverGroup.cloudProvider})">
```
erikmunson: this link was also broken, so i fixed it 🤷‍♂️

jrsquared: 🥇

PR #5617, opened 2018-08-15
feat(kubernetes): new deployment representation in cluster view

danielpeach:
new_deployments

Clicking the top of the grouping opens the deployment details on the left side.

When selected, looks like this:
selected_deployment


danielpeach:
@ethanfrogers @sbwsg @benjaminws @lwander @demobox What do you think? I'll fix it up, add tests, etc, etc, if this is an improvement.

danielpeach:
Addresses https://github.com/spinnaker/spinnaker/issues/3160

demobox:
> I wonder if there would be any value in including the kind somewhere in each box.

Agreed that, if we're going to surface kinds, it would be more consistent to do that in multiple places, e.g. also on Server Groups (and perhaps as a toggleable option in the UI?)

For now, I think it would be interesting to consider what, if anything, would make sense top-left in the bounding box for the server group manager, next to the icon. Is there any useful name or other identifying information we could put there?

PR #5623, opened 2018-08-16
fix(canary): Make deploy dialog command handling and cluster edits work

erikmunson:
Noticed this while I was fixing related stuff. Looks like the react deploy dialog wasn't hooked up to editing and the resultant command from adding a cluster pair wasn't getting handled.i

I don't even know if anybody uses this, but hey, it's fixed!

### DIFF
--- a/app/scripts/modules/canary/canary/canaryStage.js
+++ b/app/scripts/modules/canary/canary/canaryStage.js
@@ -420,8 +420,8 @@ module.exports = angular
         const config = CloudProviderRegistry.getValue(getCloudProvider(), 'serverGroup');
 
         const handleResult = function(command) {
-          var baselineCluster = awsServerGroupTransformer.convertServerGroupCommandToDeployConfiguration(command),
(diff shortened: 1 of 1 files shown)


app/scripts/modules/canary/canary/canaryStage.js
```diff
@@ -420,8 +420,8 @@ module.exports = angular
         const config = CloudProviderRegistry.getValue(getCloudProvider(), 'serverGroup');
 
         const handleResult = function(command) {
-          var baselineCluster = awsServerGroupTransformer.convertServerGroupCommandToDeployConfiguration(command),
-            canaryCluster = _.cloneDeep(baselineCluster);
+          const baselineCluster = awsServerGroupTransformer.convertServerGroupCommandToDeployConfiguration(command);
+          const canaryCluster = _.cloneDeep(baselineCluster);
```
jrsquared: best part of this pr

erikmunson: `const`-ing things brings joy to my life.

PR #5634, opened 2018-08-17
fix(core/mpt): Fix configure template checkbox initial state

jrsquared:


app/scripts/modules/core/src/pipeline/config/templates/configurePipelineTemplateModal.controller.ts
```diff
@@ -66,8 +66,13 @@ export class ConfigurePipelineTemplateModalController implements IController {
   }
 
   public initialize(): void {
-    this.pipelineName = this.pipelineTemplateConfig.config.pipeline.name;
-    this.source = this.pipelineTemplateConfig.config.pipeline.template.source;
+    const { config } = this.pipelineTemplateConfig;
+    const inherit: string[] = (config.configuration || { inherit: [] }).inherit;
```
christopherthielen:
```suggestion
    const inherit: string[] = (config.configuration || { inherit: ['expectedArtifacts', 'parameters', 'triggers'] }).inherit;
```

PR #5635, opened 2018-08-17
refactor(trigger): pull link and or text from ITrigger

emjburns:
Trigger object can contain (optionally) a `link` and/or `linkText`. If that exists, use that for display purposes

### DIFF
--- a/app/scripts/modules/core/src/pipeline/executionBuild/ExecutionBuildLink.tsx
+++ b/app/scripts/modules/core/src/pipeline/executionBuild/ExecutionBuildLink.tsx
@@ -40,18 +40,52 @@ export class ExecutionBuildLink extends React.Component<IExecutionBuildLinkProps
               {trigger.parentExecution.name}
             </a>
           )}
-        {this.props.execution.buildInfo &&
-          this.props.execution.buildInfo.number && (
-            <a
-              className="execution-build-number clickable"
-              onClick={this.handleBuildInfoClick}
-              href={this.props.execution.buildInfo.url}
-              target="_blank"
-            >
-              <span className="build-label">Build</span> #{this.props.execution.buildInfo.number}
-            </a>
-          )}
+        {this.getBuildLink()}
       </span>
     );
   }
+
+  private getBuildText(execution: IExecution) {
+    if (execution.trigger.linkText) {
+      return <span className="build-label">{execution.trigger.linkText}</span>;
+    } else {
+      return (
+        <>
+          <span className="build-label">Build</span> #{execution.buildInfo.number}
+        </>
+      );
+    }
+  }
+
+  private getBuildLink = () => {
+    const { execution } = this.props;
+
+    if (
+      !(execution.trigger.linkText && execution.trigger.link) &&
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/executionBuild/ExecutionBuildLink.tsx
```diff
@@ -40,18 +40,52 @@ export class ExecutionBuildLink extends React.Component<IExecutionBuildLinkProps
...
+    const { execution } = this.props;
+
+    if (
+      !(execution.trigger.linkText && execution.trigger.link) &&
+      !(execution.buildInfo && execution.buildInfo.number)
+    ) {
+      return null;
+    }
+
+    if (execution.trigger.linkText && !(execution.trigger.link && execution.buildInfo && execution.buildInfo.number)) {
```
erikmunson: This is saying that if `linkText` is present, you must have `link` __and__ `buildInfo.number` to show anything, but I believe your intent was that if `linkText` is present you need __one of the two__. Is that right?

I know you had mentioned that `buildInfo` will be present in the uses we have for `link`/`linkText` right now so it probably doesn't matter in that context, but is it conceivable there may be future cases where `buildInfo` is not present?

app/scripts/modules/core/src/pipeline/executionBuild/ExecutionBuildLink.tsx
```diff
@@ -40,18 +40,52 @@ export class ExecutionBuildLink extends React.Component<IExecutionBuildLinkProps
...
+      // Can't have just link text, but can parse link text from buildInfo
+      return null;
+    }
+
+    if (execution.trigger.link && !(execution.trigger.linkText && execution.buildInfo && execution.buildInfo.number)) {
```
erikmunson: Same question here as above.

app/scripts/modules/core/src/pipeline/executionBuild/ExecutionBuildLink.tsx
erikmunson: oh one other thought — now that we're splitting these conditionals up, does it make sense to check `buildInfo.url` in this check rather than `buildInfo.number` since that's what we actually use for the href?

app/scripts/modules/core/src/pipeline/executionBuild/ExecutionBuildLink.tsx
```diff
@@ -40,18 +40,52 @@ export class ExecutionBuildLink extends React.Component<IExecutionBuildLinkProps
...
+      // Can't have just link, but can parse link from buildInfo
```
erikmunson: `but can parse link text from buildInfo`?

app/scripts/modules/core/src/pipeline/executionBuild/ExecutionBuildLink.tsx
emjburns: Ok I think I made it better?

erikmunson:
🚀 🚢 🇮🇹

PR #5640, opened 2018-08-20
fix(core): Long app names hide and make refresh unusable

archana-s:
Long app names along with app icon and refresh do not shrink well for smaller laptops and devices. 

Long app names push the refresh button in to the navigation menu and as a result the refresh button is not accessible. The temp solution now is to wrap the icon and bring it down if there isn't enough room.


erikmunson:
👍

PR #5647, opened 2018-08-21
fix(amazon): Security Group cloning did not refresh fields with chang…

archana-s:
Bug fixes for security group clone modal.

There were three problems that are fixed in this PR:
_Problem 1_: Security Group cloning today does not retain the correct VPC selection when regions are updated even though the same VPC is available for this newly selected region also. This ended up being the case because we compared vpc ids to exist across regions. I updated this to use vpc name comparison so we can choose the one they already had. Why is this important? When changing regions, there is no feedback about change to the vpc.  People assume everyhting is correct until they get to the ingress section. 

_Problem 2_: Security groups are not completely loaded before we attempt to show them in the ingress section dropdown. Fixed it to wait for the security groups to load first. 

_Problem 3_: Existing selected security groups were not cleared with changes to account or region. This happened because there was an error in how we read the rules and compared them. That is fixed here too.


app/scripts/modules/amazon/src/securityGroup/configure/configSecurityGroup.mixin.controller.js
```diff
@@ -169,7 +169,7 @@ module.exports = angular
         }
       }
 
-      const match = (available || []).find(vpc => vpc.ids.includes($scope.securityGroup.vpcId));
+      const match = (available || []).find(vpc => vpc.label === $scope.securityGroup.vpcName);
```
archana-s: Using the vpn name here to ensure that it is selected if its available for the region and account.

app/scripts/modules/amazon/src/securityGroup/configure/configSecurityGroup.mixin.controller.js
```diff
@@ -227,7 +228,11 @@ module.exports = angular
       var removed = $scope.state.removedRules,
         securityGroup = $scope.securityGroup;
       $scope.securityGroup.securityGroupIngress = securityGroup.securityGroupIngress.filter(rule => {
-        if (rule.accountName !== securityGroup.accountName || (rule.vpcId && rule.vpcId !== securityGroup.vpcId)) {
+        if (
```
archana-s: in some of these cases, the rule does not have an accountname or vpcid, as a result, it was never refreshing the security groups for the region/account/vpc combination.

jrsquared: it was already checking for `rule.vpcId` before comparing `vpcId`s, was that not sufficient? Either just add `rule.accountName &&` to the `accountName` check, or remove the `(rule.vpcId &&`...

archana-s: good catch

PR #5648, opened 2018-08-22
feat(artifacts): move artifact creation into central service

sbwsg:
First step towards https://github.com/spinnaker/spinnaker/issues/2921 is to get expected artifact creation in to a central place for components to access.

### DIFF
--- a/app/scripts/modules/core/src/artifact/expectedArtifact.service.ts
+++ b/app/scripts/modules/core/src/artifact/expectedArtifact.service.ts
@@ -1,5 +1,6 @@
 import { PipelineConfigService } from 'core/pipeline/config/services/PipelineConfigService';
 import { IPipeline, IStage, IExpectedArtifact, IExecutionContext } from 'core/domain';
+import { UUIDGenerator } from 'core/utils/uuid.service';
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/artifact/expectedArtifact.service.ts
```diff
@@ -31,4 +32,29 @@ export class ExpectedArtifactService {
...
         .reduce((array, value) => array.concat(value), []);
   }
+
+  public static createEmptyArtifact(kind: string): IExpectedArtifact {
+    return {
+      id: UUIDGenerator.generateUuid(),
+      usePriorArtifact: false,
+      useDefaultArtifact: false,
+      matchArtifact: {
+        id: UUIDGenerator.generateUuid(),
```
sbwsg: I am unclear if the ID for an artifact is intended to equal the id of its ExpectedArtifact so I've made them all different for now.

ezimanyi: I think the intent is for them to be different, so this looks good

PR #5652, opened 2018-08-23
feat(artifacts): add react implementations of artifact editors

sbwsg:
Part 2 of many towards https://github.com/spinnaker/spinnaker/issues/2921

This adds react components for editing each artifact type. Improved type safety over html template strings and they're a bit more fungible for use in different places.

These aren't used yet. I'm breaking up a larger set of diffs into individual PRs. The next in this series adds a manifest artifact editor that makes use of these.

### DIFF
--- a/app/scripts/modules/core/src/domain/IArtifactEditorProps.ts
+++ b/app/scripts/modules/core/src/domain/IArtifactEditorProps.ts
@@ -1,0 +1,8 @@
+import { IArtifact } from 'core';
+
+export interface IArtifactEditorProps {
+  artifact: IArtifact;
+  onChange: (a: IArtifact) => void;
+  labelColumns: number;
+  fieldColumns: number;
+}
--- a/app/scripts/modules/core/src/domain/IArtifactKindConfig.ts
+++ b/app/scripts/modules/core/src/domain/IArtifactKindConfig.ts
@@ -1,3 +1,6 @@
+import { Component, SFC } from 'react';
+import { IArtifactEditorProps } from 'core/domain';
+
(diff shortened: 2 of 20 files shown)


app/scripts/modules/core/src/pipeline/config/triggers/artifacts/SingleFieldArtifactEditor.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import { isFinite } from 'lodash';
+import { HelpField } from 'core';
+import { IArtifactEditorProps, IArtifact } from 'core/domain';
+
+export const SingleFieldArtifactEditor = (
+  fieldKey: keyof IArtifact,
+  label: string,
+  placeholder: string,
+  helpText: string,
```
danielpeach: Maybe `helpTextKey`? (`helpText` sounds like the content of the key itself)

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/SingleFieldArtifactEditor.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import { isFinite } from 'lodash';
+import { HelpField } from 'core';
+import { IArtifactEditorProps, IArtifact } from 'core/domain';
+
+export const SingleFieldArtifactEditor = (
```
danielpeach: I'd make the first letter lower case, since it's not a component itself.

PR #5660, opened 2018-08-24
feat(kubernetes): add ace yaml editor

danielpeach:
yaml_editor

Yaml parsing error:
yaml_editor_error

In stage config:
in_stage_config

Probably has some bugs!


app/scripts/modules/kubernetes/src/v2/manifest/edit/editManifestWizard.controller.ts
```diff
@@ -27,11 +27,13 @@ class KubernetesEditManifestCtrl implements IController {
     private application: Application,
   ) {
     'ngInject';
-    if (sourceManifest &&
-        sourceManifest.metadata &&
-        sourceManifest.metadata.annotations &&
-        sourceManifest.metadata.annotations['kubectl.kubernetes.io/last-applied-configuration']) {
-      sourceManifest = load(sourceManifest.metadata.annotations['kubectl.kubernetes.io/last-applied-configuration'])
+    if (
```
danielpeach: Not sure why the formatter picked this up now.

app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestConfig.controller.spec.ts
```diff
@@ -1,102 +0,0 @@
-import { KubernetesV2DeployManifestConfigCtrl as Controller } from './deployManifestConfig.controller';
```
danielpeach: The `YamlEditor` is now responsible for translating JS objects to YAML, rather than the parent controllers. My understanding of this test is that it tested that behavior for the deploy manifest controller.

danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/3156

PR #5663, opened 2018-08-24
feat(kubernetes): better undo rollout ux

danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/2769

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/serverGroupManager/details/details.controller.ts
+++ b/app/scripts/modules/kubernetes/src/v2/serverGroupManager/details/details.controller.ts
@@ -1,9 +1,16 @@
 import { IController, IScope, module } from 'angular';
 import { IModalService } from 'angular-ui-bootstrap';
+import { orderBy } from 'lodash';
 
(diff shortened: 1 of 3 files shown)


app/scripts/modules/kubernetes/src/v2/serverGroupManager/details/details.controller.ts
```diff
@@ -86,13 +93,13 @@ class KubernetesServerGroupManagerDetailsController implements IController {
...
           account: this.serverGroupManager.account,
         },
-        revisions: () =>
-          this.serverGroupManager.serverGroups.map(sg => {
-            return {
-              name: sg.name,
-              revision: sg.moniker.sequence,
-            };
-          }),
+        revisions: () => {
```
lwander: Should we remove the current active revision?

danielpeach: The `const [, ...rest]` bit below pulls out the first revision (after sorting by the sequence).

PR #5666, opened 2018-08-25
feat(provider/ecs): Add option to associate public IP address

clareliguori:
Configuration for public IP address option ("Associate Public IP Address") when using VPC mode in ECS server groups

Clouddriver-side change is here: https://github.com/spinnaker/clouddriver/pull/2908

public-ip-addr

### DIFF
--- a/app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
@@ -57,6 +57,26 @@
     </div>
   </div>
 
+  <div ng-if="$ctrl.command.networkMode == 'awsvpc'">
+    <div class="form-group">
+      <div class="col-md-4 sm-label-right">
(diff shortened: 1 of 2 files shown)


app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
```diff
@@ -57,6 +57,26 @@
...
   </div>
 
+  <div ng-if="$ctrl.command.networkMode == 'awsvpc'">
+    <div class="form-group">
+      <div class="col-md-4 sm-label-right">
+        <b>Associate Public IP Address</b> <help-field key="ecs.publicip"></help-field>
+      </div>
+      <div class="col-md-1 radio">
+        <label>
+          <input type="radio" ng-model="$ctrl.command.associatePublicIpAddress" ng-value="true" id="associatePublicIpAddressTrue"/>
```
danielpeach: A checkbox is more common UX for binary choices, especially for yes/no, true/false choices. It's up to you though.

PR #5667, opened 2018-08-25
feat(provider/ecs): Add launch type option to support Fargate

clareliguori:
Configuration for ECS Launch Type option

Clouddriver-side change is here: spinnaker/clouddriver#2909

launch-type


danielpeach:
Can you please fix this compiler error? Then we can merge this. 

screen shot 2018-08-27 at 7 56 48 am

PR #5669, opened 2018-08-27
fix(provider/ecs): Now satisfying the expectation of a securityGroupTransformer for ECS

BrunoCarrier:


BrunoCarrier:
Without this, server groups can no longer be displayed in the infrastructure tab

PR #5672, opened 2018-08-27
fix(core): Smarter aws instance search when i- prefix is missing

alanmquach:
AWS instances take the format `i-xxxxxxxxxxxxxxxxx`, but when copying and pasting, selection logic typically delimits on the `-` causing only the 17 character hex identifier to be copied without the `i-` prefix.

For convenience, we can detect this condition (only for AWS) and search for `i-{searchterm}` instead. And because instance searching is an [exact match](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-core/src/main/groovy/com/netflix/spinnaker/clouddriver/cache/CatsSearchProvider.groovy#L287) a 17 character hex search term would never get any results since it would be missing the `i-` prefix.


app/scripts/modules/core/src/instance/instanceSearchResultType.tsx
```diff
@@ -76,6 +79,18 @@ class InstancesSearchResultType extends SearchResultType<IInstanceSearchResult>
...
   };
 
+  public search(
+    params: ISearchParams,
+    _otherResults: Observable<ISearchResultSet>,
+  ): Observable<ISearchResults<IInstanceSearchResult>> {
+    const { cloudProvider, key, ...otherParams } = params;
+    // Because the instance search is an exact match, we can detect when an aws search is missing 'i-' and prepend it
+    if (cloudProvider === 'aws' && key.match(new RegExp('^[A-Fa-f0-9]{17}$'))) {
+      return super.search({ ...otherParams, key: `i-${key}` }, _otherResults);
```
erikmunson: does using `otherParams` here mean that `cloudProvider` will be excluded (vs. the call below that uses `params`)? Is that a concern?

erikmunson:
👍It does feel like we could theoretically make use of some kind of provider-specific search intercept pattern so things like this could live inside `amazon`, but until another use case for it pops up it's likely not worth the effort to come up with a framework/registry for forking out.

app/scripts/modules/core/src/instance/instanceSearchResultType.tsx
alanmquach: Ah, you're right. That's unintentional. Thanks for catching that.

christopherthielen:
This feels like it belongs in clouddriver to me

alanmquach:
Closing.

This will be addressed in clouddriver instead: https://github.com/spinnaker/clouddriver/pull/2923

PR #5673, opened 2018-08-27
fix(pipeline): fix invisible parameter when default is not in options

alanmquach:
When a pipeline is configured with a parameter that can be selected from an options list, yet the default value is not part of that options list, it looks as if nothing is selected by default.

In actuality, the default option is still selected (just not shown in the UI) and running the pipeline will result in running with the default value, even though that was not present to the user.

Additionally, if the user selects from one of the options, there is no way to select the pipeline's configured default.

This PR forces the default value to be part of the options list if it is absent.


erikmunson:
v cool — think we could benefit from any sort of annotation on the default option to indicate that it's the intended default for the pipeline rather than just the first selected option when the form loads?

gcomstock:
Hey! Sorry for the delay; I didn't realize my github notifications were still being sent to my personal email. I think appending " (default)" to the first dropdown value wouldn't hurt, but it sounds like a user probably already has a good understanding of what that is already, so I defer to you. Thanks for asking!

PR #5674, opened 2018-08-27
fix(ux): deploy stage table layout

gcomstock:
minor change to the "deploy stage" table to allow word breaking, so the action icons don't get pushed out of view when the table's cell content is too long.


jrsquared:
v.nice

PR #5675, opened 2018-08-27
fix(kubernetes): links from server groups + instances in load balance…

danielpeach:
…r view


danielpeach:
Addresses https://github.com/spinnaker/spinnaker/issues/3252

PR #5679, opened 2018-08-29
fix(core/bootstrap): Remove vis=true/vis=false from URL after toggling visualizer

christopherthielen:
Adding ?vis=true to the end of the URL shows the visualizer.
However, many links in spinnaker will strip the vis=true from the URL, so the visualizer tends to disappear.
This change will enable/disable the visualizer when changing between `vis=true`/`vis=false`, but will then strip the parameter from the URL.
The visualizer will remain in its current state (enabled/disabled) until the parameter changes again.

This change exposed a bug in `@uirouter/core` which has now been fixed, but until a new version of ui-router is released and included in deck, the "remove parameter from the URL" behavior won't work reliably.


app/scripts/modules/core/src/navigation/state.provider.ts
```diff
@@ -31,12 +31,8 @@ export class StateConfigProvider implements IServiceProvider {
...
     name: 'home',
     abstract: true,
-    url: '?{debug:boolean}&{vis:boolean}&{trace:query}',
-    params: {
-      debug: { dynamic: true },
-      vis: { dynamic: true, value: false, squash: true },
-      trace: { dynamic: true, value: null, squash: true },
-    },
+    url: '?{debug:boolean}&{vis:query}&{trace:query}',
+    dynamic: true,
```
christopherthielen: latest ui-router supports `dynamic` on the state def instead of individually on each parameter.

@erikmunson @archana-s @jrsquared @danielpeach @sbwsg FYI

PR #5681, opened 2018-08-29
feat(halconfig): Have AWS settings configured by halyard to be picked up in settings.js

mizzy:
Current halconfig/settings.js has fixed values for aws settings.
It should pick up settings configured by halyard.

This works with https://github.com/spinnaker/halyard/pull/1024 .

Ref: https://github.com/spinnaker/spinnaker/issues/3271

### DIFF
--- a/halconfig/settings.js
+++ b/halconfig/settings.js
@@ -77,6 +77,20 @@ var dcos = {
     account: '{%dcos.default.account%}',
   },
 };
+var aws = {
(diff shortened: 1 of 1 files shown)


lwander:
Keep in mind this will take affect in Spinnaker 1.10 and the nightly builds.

mizzy:
Thanks!

PR #5682, opened 2018-08-29
fix(core/projects): remove duplicative home.projects.project.** substates

christopherthielen:
The projects states are under `home.project.**`, but were also duplicated underneath `home.projects.project.**`.
The second set of states are not used, AFAICT

This is a baby step.  More steps to come, I hope.


christopherthielen:
screenshot 2018-08-29 09 56 52

PR #5688, opened 2018-09-04
fix(core): Server group details could not be opened

archana-s:
Fixed passing the cloud provider correctly to server groups so their details can render from the load balancer page.

https://github.com/spinnaker/deck/commit/060ad9cc0eed5cf389d2923436f02911a2a25b6a#diff-0051f567bb321bdd6ec4d5e2652f82f1 This introduced a bug where the details of the server groups could not be accessed because they are missing the provider value during state transition. 
This is fixed by passing the cloud provider to the server groups. This will still keep the instance provider intact as it is hydrated with the provider information.


erikmunson:
🎉

PR #5690, opened 2018-09-04
feat(artifacts): add a select box for choosing an artifact source

sbwsg:
artifact_source_selector

Part of the editor functionality for https://github.com/spinnaker/spinnaker/issues/2921

### DIFF
--- a/app/scripts/modules/core/src/artifact/react/ExpectedArtifactSourceSelector.tsx
+++ b/app/scripts/modules/core/src/artifact/react/ExpectedArtifactSourceSelector.tsx
@@ -1,0 +1,72 @@
+import { module } from 'angular';
+import * as React from 'react';
+import { react2angular } from 'react2angular';
+import { UUIDGenerator, TetheredSelect, IExpectedArtifact } from 'core';
+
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/artifact/react/ExpectedArtifactSourceSelector.tsx
```diff
@@ -0,0 +1,72 @@
...
+    name: string;
+  };
+  value?: string;
+  label: string;
+}
+
+export interface IExpectedArtifactSourceSelectorProps {
+  sources: IExpectedArtifactSourceOption[];
+  selected: IExpectedArtifactSourceOption;
+  onChange: (_: IExpectedArtifactSourceOption) => void;
```
danielpeach: `_` implies to me that the parameter is unused

sbwsg: ah yeah fair point. mostly just didn't want to have to think of a name for a type annotation where's it's otherwise unused.

PR #5692, opened 2018-09-04
feat(artifacts): add inline artifact editor

sbwsg:
This is the inline editor component for https://github.com/spinnaker/spinnaker/issues/2921

artifact_editor

A GIF of using the editor in the deploy manifest stage:
inline_artifact_editor

### DIFF
--- a/app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
+++ b/app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
@@ -1,0 +1,146 @@
+import { module } from 'angular';
+import * as React from 'react';
+import { react2angular } from 'react2angular';
+import { cloneDeep } from 'lodash';
+import { IExpectedArtifact, IArtifactAccount, IArtifactKindConfig, IArtifact } from 'core';
+import { StageConfigField } from 'core/pipeline/config/stages/core/stageConfigField/StageConfigField';
+import { ExpectedArtifactService } from '../expectedArtifact.service';
+import {
+  ExpectedArtifactKindSelector,
+  EXPECTED_ARTIFACT_KIND_SELECTOR_COMPONENT_REACT,
+} from './ExpectedArtifactKindSelector';
+import {
+  ExpectedArtifactSourceSelector,
+  IExpectedArtifactSourceOption,
+  EXPECTED_ARTIFACT_SOURCE_SELECTOR_COMPONENT_REACT,
+} from './ExpectedArtifactSourceSelector';
+import { ArtifactAccountSelector } from './ArtifactAccountSelector';
+
+export interface IExpectedArtifactEditorProps {
+  default?: IExpectedArtifact;
+  kinds: IArtifactKindConfig[];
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
```diff
@@ -0,0 +1,144 @@
...
+        </StageConfigField>
+        <StageConfigField label="Artifact Type" fieldColumns={8}>
+          <ExpectedArtifactKindSelector kinds={kinds} selected={kind} onChange={this.onKindChange} />
+        </StageConfigField>
+        <StageConfigField label="Artifact Account" fieldColumns={8}>
+          <ArtifactAccountSelector accounts={accounts} selected={account} onChange={this.onAccountChange} />
+        </StageConfigField>
+        {EditCmp && <EditCmp artifact={artifact} onChange={this.onArtifactEdit} labelColumns={3} fieldColumns={3} />}
+        <StageConfigField label="" fieldColumns={8}>
+          <button onClick={this.onSave}>Create New Artifact</button>
```
danielpeach: Would add standard spinnaker styles to this button. Probably `add-new btn btn-block btn-sm`.

app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
```diff
@@ -0,0 +1,144 @@
...
+  public render() {
+    const { sources } = this.props;
+    const { expectedArtifact, source, account } = this.state;
+    const accounts = this.accountsForExpectedArtifact(expectedArtifact);
+    const artifact = ExpectedArtifactService.artifactFromExpected(expectedArtifact);
+    const kinds = this.availableKinds().sort((a, b) => a.label.localeCompare(b.label));
+    const kind = artifact ? kinds.find(k => k.key === artifact.kind) : null;
+    const EditCmp = kind && kind.editCmp;
+    return (
+      <div>
```
danielpeach: This can be a `<> ... </>` pair

app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
danielpeach: I'm wondering if the language is a little confusing - since inside your dropdown you have "Create new", and then here you have "Create New Artifact". What about "Save" or "Confirm"?

sbwsg: Agreed. I thought about this while I was writing it and figured "Save" would clash in meaning with the "Save Changes" button for the pipeline.  I'll use "Confirm".

sbwsg: I ended up using `btn-primary` instead of `add-new` because the dotted outline from `add-new` gives the sense that it's going to add something / fill a gap in the form rather than commit the existing changes:

screen shot 2018-09-05 at 11 04 33 am

PR #5700, opened 2018-09-05
fix(aws): Fix clone stage.

jtk54:
A series of fixes related to AWS Clone stage.

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/AvailabilityZoneSelector.tsx
+++ b/app/scripts/modules/amazon/src/serverGroup/AvailabilityZoneSelector.tsx
@@ -39,8 +39,8 @@ export class AvailabilityZoneSelector extends React.Component<
   private setDefaultZones(props: IAvailabilityZoneSelectorProps) {
     const { credentials, onChange, region } = props;
 
-    AccountService.getAvailabilityZonesForAccountAndRegion('aws', credentials, region).then(preferredZones =>
-      onChange(preferredZones.slice()),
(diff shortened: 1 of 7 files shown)


app/scripts/modules/amazon/src/serverGroup/AvailabilityZoneSelector.tsx
```diff
@@ -39,8 +39,8 @@ export class AvailabilityZoneSelector extends React.Component<
   private setDefaultZones(props: IAvailabilityZoneSelectorProps) {
     const { credentials, onChange, region } = props;
 
-    AccountService.getAvailabilityZonesForAccountAndRegion('aws', credentials, region).then(preferredZones =>
-      onChange(preferredZones.slice()),
+    AccountService.getAvailabilityZonesForAccountAndRegion('aws', credentials, region).then(
+      preferredZones => preferredZones && onChange(preferredZones.slice()),
```
danielpeach: Would wrap the function body here as `{ preferred Zones && onChange(preferredZones.slice()) }`. As is, the value of the promise (even though it's not being used) is either whatever falsey value `preferredZones` is or the return type of `onChange`.

jtk54: Should I add this as a subsequent patch on top of the cherry-picked commits?

jtk54:
We deliberately included the 4 commits in the PR since they were adding fixes users were seeing in 1.9.2. I can make the suggested on top as another commit, but I think we need each.

PR #5701, opened 2018-09-05
feat(artifacts): add expected artifact selector with option to create new artifacts

sbwsg:
This PR continues building towards https://github.com/spinnaker/spinnaker/issues/2921

Rather than adding to the existing expected artifact selector I've opted to create a new one. The reasons for doing so are to move away from a large unlinted template string and to avoid some tricky behaviour in the original related to angular two-way data binding. I've also added some UTs since the component is progressively growing larger.

screen shot 2018-09-05 at 3 40 03 pm


app/scripts/modules/core/src/artifact/react/ExpectedArtifactSelector.spec.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { ExpectedArtifactService, IExpectedArtifact } from 'core';
+import { ExpectedArtifactSelector } from './ExpectedArtifactSelector';
+
+const artifact = (kind: string, type: string): IExpectedArtifact => {
+  const ea = ExpectedArtifactService.createEmptyArtifact(kind);
+  ea.matchArtifact.type = type;
+  return ea;
+};
+
+fdescribe('<ExpectedArtifactSelector/>', () => {
```
danielpeach: `fdescribe` I think causes only this test to run (and all others to be ignored)

sbwsg:
I've used the artifact name instead of the summarizeArtifact helper for this component.  The reason being that we're now creating artifacts with UUIDs and those IDs are so long that they quickly gum up the select box and prevent the user from seeing useful information about the artifacts.

app/scripts/modules/core/src/artifact/react/ExpectedArtifactSelector.spec.tsx
sbwsg: Doh. _Artifact_ of development.

PR #5705, opened 2018-09-06
fix(kubernetes): Fix container selection with multiple triggers

ezimanyi:
The logic for matching containers to upstream images matches only on the container description, and doesn't take into account whether the source of the image (artifact, trigger, etc.).  The matching logic should be updated to account for image source.


ezimanyi:
Fixes https://github.com/spinnaker/spinnaker/issues/3291

PR #5707, opened 2018-09-06
feat(kubernetes): use inline artifact editor for deploy manifest artifact

sbwsg:
This PR enables a user to define an expected artifact while in-situ on the Deploy Manifest stage configuration screen. This pulls together the previous merges for https://github.com/spinnaker/spinnaker/issues/2921

I'm not particularly happy with the amount of code that's needed for a stage to utilize the inline editor. At the moment adding support to a new stage will involve writing a delegate class that meets the `IExpectedArtifactSelectorViewControllerDelegate` interface. In this PR it's an extra 100 lines of code that's specific to the deploy manifest stage. I'm still looking for ways to reduce this burden but there's no reason to hold back the functionality while I think about that. Suggestions welcome here.


sbwsg:
deploy_manifest_inline_artifact_editor

PR #5711, opened 2018-09-06
fix(kubernetes): Replace all periods in image name

ezimanyi:
When generating an image name from the registry, we were only replacing the first period, which breaks for registries that have more than one period.


ezimanyi:
Fixes https://github.com/spinnaker/spinnaker/issues/3290

PR #5717, opened 2018-09-07
feat(kubernetes): add inline artifact editor to patch manifest stage

sbwsg:
screen shot 2018-09-07 at 4 22 55 pm

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestConfig.controller.ts
+++ b/app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestConfig.controller.ts
@@ -1,5 +1,5 @@
 import { IController, IScope } from 'angular';
-import { get } from 'lodash';
+import { get, defaults } from 'lodash';
 
 import {
   IKubernetesManifestCommandMetadata,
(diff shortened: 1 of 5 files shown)


app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestConfig.controller.ts
```diff
@@ -34,21 +30,18 @@ export class KubernetesV2DeployManifestConfigCtrl implements IController {
       this.$scope.stage.moniker,
     ).then((builtCommand: IKubernetesManifestCommandData) => {
       if (this.$scope.stage.isNew) {
-        Object.assign(this.$scope.stage, builtCommand.command);
-        this.$scope.stage.source = this.textSource;
+        defaults(this.$scope.stage, builtCommand.command, {
```
sbwsg: `defaults` fixes a small issue when adding a new stage: Currently if a user selects an artifact source in a new patch or deploy stage, then leaves to another stage, and then comes back, the artifact source is overwritten to be a text source. Using defaults provides continuity from any previous edits to this new stage upon revisiting it.

PR #5719, opened 2018-09-07
refactor(core/entityTag): Remove dataUpdated() call from DataSource onLoad

christopherthielen:
I verified that no current afterLoad() implementation mutates the current data source's `data[]`, therefore calling dataUpdated() should be unnecessary.


app/scripts/modules/core/src/task/task.dataSource.spec.ts
```diff
@@ -72,7 +72,8 @@ describe('Task Data Source', function() {
       expect(application.getDataSource('tasks').loading).toBe(false);
       expect(application.getDataSource('tasks').loadFailure).toBe(false);
 
-      expect(nextCalls).toBe(1);
+      // TODO: Uncomment in next PR
```
christopherthielen: this test will be addressed in an upcoming PR

PR #5723, opened 2018-09-10
test(travis): test to make sure modules are buildable and fix typings

kevinawoo:


jrsquared:
It would actually be great if you could add a `package.json` script to test this too so we can run something easy locally to make sure things will build before they hit travis.

kevinawoo:
sounds like a GREAT idea!

I'll add in the ability to do:
```
yarn run module   # which builds all modules
```

```
yarn run modules core  # to build a single module
```

christopherthielen:
FYI: although it won't work with Deck's current project structure, I built a "test downstream projects" system for my OSS projects which we may want to consider in the future.

https://github.com/ui-router/publish-scripts/blob/master/test_downstream_projects.js

It works like this:

1) `yarn add -D @uirouter/publish-scripts`
2) Create a `downstream_projects.json` which points to the downstream projects, either local paths or GIT URLs:

```json
{
  "typescript2.6": "./test/typescript2.6",
  "react": "https://github.com/ui-router/react.git"
}
```

3) Add a script to package.json

```json
{ 
  "scripts": {
    "test:downstream": "test_downstream_projects"
  }
}
```

4) invoke via `npm run test:downstream`

kevinawoo:
I made a separate script to build the modules now and you can now run 
```bash
yarn run modules  # builds all modules except the modules listed in SKIPPED_MODULES

yarn run modules amazon core docker   # builds modules amazon, core, and docker only.
```

I'll need to fix some typings in some modules before this passes travis.

kevinawoo:
A future improvement might be to use travis build stages to parallelize the builds for speed.

christopherthielen:
nice

kevinawoo:
https://github.com/spinnaker/deck/pull/5742 and https://github.com/spinnaker/deck/pull/5743 needs to be merged for these tests to pass.

christopherthielen:
here's how I handle parallel CI in my OSS projects: 

https://github.com/ui-router/react/blob/41004901caed40ad3ab1c6689046cb84e8422cf7/.travis.yml#L7-L16

I haven't seen the build stages Travis feature before, but it looks very cool.  ... whatever makes the most sense

PR #5725, opened 2018-09-11
feat(api): reject invalid content API responses with message

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/3277


sbwsg:
screen shot 2018-09-11 at 12 36 18 pm

app/scripts/modules/core/src/api/ApiService.ts
```diff
@@ -25,6 +33,8 @@ export class API {
     },
   };
 
+  public static invalidContentMessage = 'API response was neither JSON nor zero-length html or text';
```
danielpeach: could add a `readonly` modifier

PR #5726, opened 2018-09-11
refactor(core): All region-select-field to use RegionSelectField.tsx

alanmquach:
Removing the angular template and wiring up the directive to a proxy component that delegates to the react version.


app/scripts/modules/core/src/region/regionSelectField.directive.js
```diff
@@ -1,13 +1,72 @@
...
+      'field',
+      'account',
+      'provider',
+      'onChange',
+      'labelColumns',
+      'fieldColumns',
+      'readOnly',
+    ]),
+  )
+  .component('regionSelectFieldProxy', {
```
christopherthielen: seems like this intermediate component might not be necessary?  or is it?

Have a look at this similar example in https://github.com/spinnaker/deck/blob/0e984db58d5d41ca67f7b8ca98d6972977a021d2/app/scripts/modules/core/src/entityTag/notifications/entityNotifications.component.ts

alanmquach: You're right, it's not necessary. 👍

PR #5730, opened 2018-09-12
fix(core/serverGroup): Guard for existence of callbacks

christopherthielen:


christopherthielen:
@jrsquared @sbwsg

sbwsg:
nice

PR #5732, opened 2018-09-12
refactor(core/application): Add strong typing to ApplicationModelBuilder.createApplication()

christopherthielen:


app/scripts/modules/core/src/application/applicationModel.builder.ts
```diff
@@ -15,10 +15,7 @@ export class ApplicationModelBuilder {
   /**
    * This is mostly used in tests
    */
-  public createApplication(name: string, ...dataSources: any[]): Application {
-    if (Array.isArray(dataSources[0])) {
-      dataSources = dataSources[0];
-    }
+  public createApplication(name: string, ...dataSources: IDataSourceConfig[]): Application {
```
christopherthielen: `IDataSourceConfig[]` not `any[]`

PR #5733, opened 2018-09-13
fix(amazon): Fix empty security group dropdown due to nulled vpcId

alanmquach:
Since we filter security groups by vpcId, we end up filtering **all security groups** from the dropdown because the `vpcId` gets nulled when `vpcName` is missing from a `$scope.securityGroup`.


alanmquach:
Closing since this revert would break retaining VPC choice when cloning across accounts (since that match can only be done by `vpcName` as `vpcId` is unique across accounts.

PR #5735, opened 2018-09-13
feat(artifacts): add inline artifact editor for appengine artifacts

sbwsg:


app/scripts/modules/core/src/artifact/ExpectedArtifactSelectorViewControllerAngularDelegate.ts
```diff
@@ -0,0 +1,18 @@
+import { IScope } from 'angular';
+import { IArtifactAccount, IArtifactKindConfig } from 'core';
+
+export class ExpectedArtifactSelectorViewControllerAngularDelegate<ArtifactSource> {
```
danielpeach: would make this an abstract class

PR #5738, opened 2018-09-14
 fix(ux): hover flicker on cluster instances

gcomstock:
this flicker was annoying me


christopherthielen:
IDK what this does, but fine with me

PR #5740, opened 2018-09-14
feat(kubernetes): bake manifest artifacts inline editor

sbwsg:


sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/2921

PR #5742, opened 2018-09-14
fix(cloudfoundry): module builds require the use of alias instead of a path directly in core

kevinawoo:


jrsquared:
`s/cloudfoundary/cloudfoundry/`

kevinawoo:
good catch, pushed

kevinawoo:
😞 another good catch. English is hard.

PR #5743, opened 2018-09-14
fix(ecs): angular is declared as global, so let's not redefine it

kevinawoo:


christopherthielen:
FWIW, I'd rather we align all typescript code to use `import` and `export` style, i.e.: https://github.com/spinnaker/deck/blob/532ab7784ca93569308c8f2ab80a18d313b910f9/app/scripts/modules/amazon/src/serverGroup/details/scalingPolicy/createScalingPolicyButton.component.ts

kevinawoo:
I thought I'd give it a try, but https://github.com/spinnaker/deck/blob/master/app/scripts/modules/ecs/src/ecs.module.ts#L39 I believe prevents me from doing that.

christopherthielen:
Where it's being used, switch to `import { CREATE_SCALING_POLICY_BUTTON } from './path/to/file'`, like so: 

https://github.com/spinnaker/deck/blob/master/app/scripts/modules/ecs/src/ecs.module.ts#L8

kevinawoo:
OH man, I derped hard on that, Thanks! It's been a while since I've been in actually js/ts app code -_-"

PR #5744, opened 2018-09-16
fix(amazon/instance): Fix standalone instance view

jrsquared:
@erikmunson FYI


erikmunson:
...so we have a standalone instance view. 😮🙈 

thx for catching

christopherthielen:
we have a standalone instance view?

PR #5746, opened 2018-09-17
feat(artifacts): default artifact type auto selected from match artifact

sbwsg:
Closes https://github.com/spinnaker/spinnaker/issues/2711


sbwsg:
# Before

default_artifact_before


# After

default_artifact_copy_type

PR #5750, opened 2018-09-18
chore(travis): parallelize module build tests

kevinawoo:
This will build all the modules in parallel like:

<kbd></kbd>

which will help reduce the time that travis takes to test.
- prior to #5750 being merged it took ~8 mins
- after #5750  was merged in, it took: ~11 mins
- with this PR, it now takes: ~6 mins. 😱 🎉 

> note: travis is usually limited to building 5 parallel jobs at a time, so we'll use that to our advantage and stack in the module builds.

> note: the build of "core" module is duplicated because each cloud provider's module requires the core module to be built.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,19 +17,8 @@ jobs:
     - stage: tests
       name: run tests for deck
       script: gradle/buildViaTravis.sh
-
-    # these modules have a specific build order because of dependancies
-    - name: build the core, amazon, ecs, and titus module
-      script: gradle/buildModules.sh core amazon ecs titus
-
-    # these modules do not depend on anything besides "core".
-    # Let's group together just to take advantage of travis's parallel job limit (5) to speed up builds
-    - name: build the core, appengine, and cloudfoundry module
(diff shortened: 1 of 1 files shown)


kevinawoo:
cc @christopherthielen

christopherthielen:
Travis seems really flakey.  Adding so many parallel builds seems to exacerbate the flakiness.  I've been clicking "Restart Build" and one or more of the jobs have been failing more often than not.  

Can we try splitting into only two jobs?  perhaps 1) tests and 2) libs?

kevinawoo:
👍 hopefully it's a bit more stable

christopherthielen:
Thanks Kevin. We can iterate in this in the future if we see room for improvement.

christopherthielen:
Another unfortunate travis thing, sometimes jobs get queued up (Behind other spinnaker builds? who knows?) . For example:

screen shot 2018-09-19 at 12 03 13 pm

kevinawoo:
Hmm, that is unfortunate :disappointed: using env matrixes would run into the same issue too.

PR #5751, opened 2018-09-19
feat(kubernetes): allow a copy from running infrastructure into deplo…

danielpeach:
…y manifest stage

Link to open modal:
copy_button

Modal:
copy_options

### DIFF
--- a/app/scripts/modules/core/src/manifest/ManifestReader.ts
+++ b/app/scripts/modules/core/src/manifest/ManifestReader.ts
@@ -1,8 +1,9 @@
+import { IPromise } from 'angular';
 import { API } from 'core/api/ApiService';
 import { IManifest } from 'core/domain';
 
 export class ManifestReader {
-  public static getManifest(account: string, location: string, name: string): Promise<IManifest> {
+  public static getManifest(account: string, location: string, name: string): IPromise<IManifest> {
     return API.all('manifests')
       .all(account)
       .all(location)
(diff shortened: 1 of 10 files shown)


danielpeach:
- Mostly captures the functionality of the `DeployInitializer`, but with fewer opinions (just copies a running manifest). 
- Like the `DeployInitializer`, it only offers the most recent server group in a cluster
- Offers other Kubernetes kinds (deployments, services, etc).
- Doesn't offer a replicaSet if it's managed by a deployment

app/scripts/modules/kubernetes/src/v2/pipelines/stages/ManifestCopier.tsx
```diff
@@ -0,0 +1,222 @@
...
+export interface IManifestOption {
+  account: string;
+  location: string;
+  name: string;
+  kind: string;
+  data: any;
+  key: string;
+}
+
+const LAST_APPLIED_CONFIGURATION = 'kubectl.kubernetes.io/last-applied-configuration';
```
lwander: :+1:

app/scripts/modules/kubernetes/src/v2/pipelines/stages/ManifestCopier.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+/*
+ * A modal that allows a user to copy a running Kubernetes resource.
+ **/
+export class ManifestCopier extends React.Component<IManifestCopierProps, IManifestCopierState> {
+  public static getDerivedStateFromProps = (props: IManifestCopierProps): IManifestCopierState => {
+    let manifests: IManifestOption[] = [];
+
+    const serverGroups: IServerGroup[] = props.application.getDataSource('serverGroups').data.filter(
```
lwander: Why only server groups?

danielpeach: What do you mean? They're just handled a little differently, but the other kinds are included...

lwander: Sorry got sidetracked halfway through. I second (from your todo) using the account map to derive kinds to load

PR #5753, opened 2018-09-19
refactor(core/task): Migrated task progress bar to react

alanmquach:


app/scripts/modules/core/src/task/TaskProgressBar.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { OverlayTrigger, Tooltip } from 'react-bootstrap';
+
+import { ITask } from 'core/domain';
+import { robotToHuman } from 'core/presentation/robotToHumanFilter/robotToHuman.filter';
+
+export interface ITaskProgressBarProps {
+  task: ITask;
+}
+
+export function TaskProgressBar(props: ITaskProgressBarProps) {
```
christopherthielen: I think we're using  either `class TaskProgressBar extends Component...` or `const TaskProgressBar = (props: ITaskProgressBarProps) => ` everywhere else (as opposed to `function`)

app/scripts/modules/core/src/task/TaskProgressBar.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const {
+    task: { steps, id },
+    task,
+  } = props;
```
christopherthielen: Personal opinion: I find this destructure somewhat odd to read.  

How do you feel about this instead?
```js
const { task } = props;
const { steps, id } = task;
```

christopherthielen:
sorry my prettier commit must have caused conflicts :(

app/scripts/modules/core/src/task/TaskProgressBar.tsx
```diff
@@ -0,0 +1,79 @@
...
+        </Tooltip>
+      );
+    } else {
+      tooltip = <Tooltip id={id}>Task failed; sorry, no reason provided</Tooltip>;
+    }
+  }
+
+  const stepsComplete = steps.filter(step => step.isCompleted);
+  const progressBarClass =
+    task.isRunning || task.hasNotStarted
```
erikmunson: are these states all mutually exclusive? If so, might be nicer to just use `classnames` and give it an object with all of these.

app/scripts/modules/core/src/task/TaskProgressBar.tsx
christopherthielen: please elaborate on 👎 @erikmunson

erikmunson: 😄 no strong opinion from my end (hence the reaction vs. commenting). I personally prefer the single assignment form in terms of readability, but it's deeply subjective. Also worth noting that there even _another_ option which is destructuring some or all of this in the function parameter itself rather than as a `const` assignment.

alanmquach: I started with that, then tried making it a single line:
```js
const { task: { steps, id }, task } = props;
```
which I liked, but after `prettier` reformatted it, it does seem more confusing. Happy to go back to
```js
const { task } = props;
const { steps, id } = task;
```

app/scripts/modules/core/src/task/TaskProgressBar.tsx
alanmquach: Love it, thanks.

app/scripts/modules/core/src/task/TaskProgressBar.tsx
alanmquach: I actually just followed what we were doing in other places. For example [ExecutionStepDetails](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/stages/core/ExecutionStepDetails.tsx#L13).

But we should probably pick one pattern to use moving forward.

erikmunson: yeah IME we're very inconsistent. My personal tendency has been to just go with whatever is in the file already, or use an arrow function if brand new.

archana-s:
lgtm

PR #5754, opened 2018-09-19
chore(prettier): Just Update Prettier™

christopherthielen:
Update to latest `prettier`

Run latest `prettier` on all `js`/`ts` sources


jrsquared:
I don't think this bumped up your lines of code as much as you hoped it would.

PR #5756, opened 2018-09-19
refactor(formik): Use TSX generics to render Formik components

christopherthielen:
Typescript 2.9 added support for rendering components with generic type parameters.
By switching to this style, the `values` and `errors` can be destructured in the render function, and the types are inferred.

```ts
<Component<ConcreteType> />
```


jrsquared:
Ooooo 😍

PR #5759, opened 2018-09-20
fix(core/application): Use relative import of ICluster

christopherthielen:
For some reason importing ICluster fro core/domain broke our downstream deck project


erikmunson:
ytho

PR #5761, opened 2018-09-20
fix(bake/oracle): always show rebake option in bake configuration.

guoyongzhang:
In "Bake Configuration" for Oracle Bakery, we always want to show "Rebake" option.


guoyongzhang:
Thanks for the approval.  Could you merge it once it passes all the checks?

sbwsg:
will do

PR #5767, opened 2018-09-21
feat(kubernetes): show images in server group details

danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/3263

screen shot 2018-09-21 at 2 34 11 pm

Tooltip shows which container or init container the image is running on:
screen shot 2018-09-21 at 2 34 05 pm

@Rigdon is this what you're looking for?


danielpeach:
How about this - the labels only show up if you have both containers & initContainers

init_containers

Rigdon:
> How about this - the labels only show up if you have both containers & initContainers
> 
> init_containers

I think that would be a very helpful distinction when looking at the image list. I also noticed that the busybox example doesn't have a tag. Can we include the tag on all images, even if it's `latest`?

lwander:
Looks great :+1: also +1 to @Rigdon's comment

PR #5773, opened 2018-09-24
feat(provider/kubernetes): show manifest condition times as relative

sbwsg:
### Before:

screen shot 2018-09-24 at 6 46 46 pm

### After:

screen shot 2018-09-24 at 6 44 16 pm

Closes https://github.com/spinnaker/spinnaker/issues/3235


danielpeach:
Nice

PR #5778, opened 2018-09-25
feat(core): collapse execution group account tags if there are more than 2

sbwsg:
Having too many accounts utilized in a pipeline results in a messy execution group header:

screen shot 2018-09-25 at 4 07 59 pm

This PR hides account tags if there are more than 2 and provides an on-hover handler to show those that are hidden:

multiple_account_tags_unroll

It's not the prettiest thing in the world. Any objections to merging? Open to suggestions on the animation or layout.

Closes https://github.com/spinnaker/spinnaker/issues/3200

### DIFF
--- a/app/scripts/modules/core/src/account/AccountTag.tsx
+++ b/app/scripts/modules/core/src/account/AccountTag.tsx
@@ -4,6 +4,7 @@ import { AccountService } from 'core/account/AccountService';
 
 export interface IAccountTagProps {
   account: string;
+  className?: string;
 }
 
 export interface IAccountTagState {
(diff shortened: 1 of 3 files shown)


sbwsg:
@erikmunson @archana-s @christopherthielen this PR changes some layout in core so I'm pinging you  for review here as well.

christopherthielen:
Very nice.  Do we use _lists_ of accounts elsewhere?  If so this "account list with overflow" functionality could be extracted to a separate component.

sbwsg:
Good thinking; I'll keep an eye out for account lists and refactor if i spot some.

PR #5780, opened 2018-09-25
fix(amazon/loadBalancer): Stop adding default security groups when editing

jrsquared:
@ajordens FYI


app/scripts/modules/amazon/src/loadBalancer/configure/application/CreateApplicationLoadBalancer.tsx
```diff
@@ -266,7 +266,7 @@ export class CreateApplicationLoadBalancer extends React.Component<
     }
 
     return (
-      <WizardModal<IAmazonApplicationLoadBalancerUpsertCommand>
+      <WizardModal
```
jrsquared: had to do this to make prettier not cry.

PR #5783, opened 2018-09-26
fix(amazon/loadBalancer): Do not ask for an SSL cert unless it's an SSL protocol

christopherthielen:


app/scripts/modules/amazon/src/loadBalancer/configure/classic/Listeners.tsx
```diff
@@ -163,7 +191,9 @@ class ListenersImpl extends React.Component<IListenersProps, IListenersState> {
                           this.listenerExternalProtocolChanged(listener, event.target.value as ClassicListenerProtocol)
                         }
                       >
-                        {this.protocols.map(p => <option key={p}>{p}</option>)}
+                        {this.protocols.map(p => (
```
christopherthielen: weird, I think your prettier and my prettier are fighting

erikmunson: we need a prettier for keeping our prettiers in sync

PR #5785, opened 2018-09-26
 refactor(docker/pipeline): Convert docker trigger config to react

jrsquared:


app/scripts/modules/docker/src/pipeline/trigger/DockerTrigger.ts
```diff
@@ -0,0 +1,30 @@
...
+import { DockerTriggerTemplate } from './DockerTriggerTemplate';
+import { DockerTriggerConfig } from './DockerTriggerConfig';
+
+Registry.pipeline.registerTrigger({
+  label: 'Docker Registry',
+  description: 'Executes the pipeline on an image update',
+  key: 'docker',
+  component: DockerTriggerConfig,
+  manualExecutionComponent: DockerTriggerTemplate,
+  validators: [
```
christopherthielen: where do these get used?

jrsquared: https://github.com/spinnaker/deck/blob/0e984db58d5d41ca67f7b8ca98d6972977a021d2/app/scripts/modules/core/src/pipeline/config/validation/PipelineConfigValidator.ts#L72

PR #5793, opened 2018-09-26
fix(core): Fix error when changing execution grouping

ezimanyi:


ezimanyi:
@sbwsg : I only noticed for the first time today that you can change the grouping of executions at the top of the executions screen.  When I tried it I got an NPE introduced by https://github.com/spinnaker/deck/pull/5778

PR #5794, opened 2018-09-26
feat(core): Call new manual execution endpoint behind a flag

ezimanyi:
A new endpoint that triggers manual executions via Echo has been added to Gate. Update Deck to send manual execution requests to this endpoint behind a flag that will default to off (until we are confident there
are no regressions).

### DIFF
--- a/app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
+++ b/app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
@@ -12,7 +12,10 @@ import { IExecution, IExecutionGroup, IExecutionTrigger, IPipeline, IPipelineCom
 import { NextRunTag } from 'core/pipeline/triggers/NextRunTag';
 import { Popover } from 'core/presentation/Popover';
 import { ExecutionState } from 'core/state';
-import { PipelineConfigService } from 'core/pipeline/config/services/PipelineConfigService';
+import {
+  IEchoTriggerPipelineResponse,
+  PipelineConfigService,
+} from 'core/pipeline/config/services/PipelineConfigService';
 
 import { TriggersTag } from 'core/pipeline/triggers/TriggersTag';
 import { AccountTag } from 'core/account';
@@ -20,6 +23,7 @@ import { ModalInjector, ReactInjector } from 'core/reactShims';
 import { Spinner } from 'core/widgets/spinners/Spinner';
 
 import './executionGroup.less';
+import { SETTINGS } from 'core';
 
 const ACCOUNT_TAG_OVERFLOW_LIMIT = 2;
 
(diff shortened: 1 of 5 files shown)


ezimanyi:
@sbwsg : Do you have any idea what code path calls the `startPipeline` function in `Executions.tsx` (https://github.com/spinnaker/deck/blob/3adbf368d6e15e063280c6a684623a43857f9f1b/app/scripts/modules/core/src/pipeline/executions/Executions.tsx#L142)?  I've only been able to hit the one in `ExecutionGroups.tsx` that I changed here.  If the one in `Executions.tsx` isn't dead code, I'd like to change it as well, but I haven't been able to figure out how to test my change.

app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -12,14 +12,18 @@ import { IExecution, IExecutionGroup, IExecutionTrigger, IPipeline, IPipelineCom
...
+  PipelineConfigService,
+} from 'core/pipeline/config/services/PipelineConfigService';
 
 import { TriggersTag } from 'core/pipeline/triggers/TriggersTag';
 import { AccountTag } from 'core/account';
 import { ModalInjector, ReactInjector } from 'core/reactShims';
 import { Spinner } from 'core/widgets/spinners/Spinner';
 
 import './executionGroup.less';
+import { SETTINGS } from 'core';
```
danielpeach: Please sort this above with the other TS imports.

app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -114,13 +118,21 @@ export class ExecutionGroup extends React.Component<IExecutionGroupProps, IExecu
...
     this.setState({ triggeringExecution: true });
-    return PipelineConfigService.triggerPipeline(
-      this.props.application.name,
-      command.pipelineName,
-      command.trigger,
-    ).then(
-      newPipelineId => {
-        const monitor = executionService.waitUntilNewTriggeredPipelineAppears(this.props.application, newPipelineId);
+
+    let triggerFunction: (a: string, b: string, c: any) => IPromise<string>;
```
danielpeach: The parameter names could be `app: string, pipeline: string, trigger: any`.

app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -114,13 +118,21 @@ export class ExecutionGroup extends React.Component<IExecutionGroupProps, IExecu
...
+    let monitorFunction: (a: string) => IPromise<any>;
+    if (SETTINGS.feature.triggerViaEcho) {
+      triggerFunction = PipelineConfigService.triggerPipelineViaEcho;
```
danielpeach: Please bind `PipelineConfigService` here and below, since someone could unwittingly use `this` inside `triggerViaEcho` or `triggerPipeline`.

sbwsg:
The code path in Executions.tsx can be triggered by clicking the play button at the top of the pipelines list:

fgmh8j7e0qv

PR #5795, opened 2018-09-27
fix(kubernetes): copy into pipeline stage

danielpeach:

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/wizard/ManifestEntry.tsx
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/wizard/ManifestEntry.tsx
@@ -1,35 +1,55 @@
 import * as React from 'react';
 import { FormikErrors } from 'formik';
+import { dump } from 'js-yaml';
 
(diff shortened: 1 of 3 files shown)


app/scripts/modules/kubernetes/src/v2/manifest/wizard/ManifestEntry.tsx
```diff
@@ -1,36 +1,57 @@
...
+      this.state = { manifestString: null };
+    }
+  }
+
   public validate = (_values: IKubernetesManifestCommandData) => {
     return {} as FormikErrors<IKubernetesManifestCommandData>;
   };
 
-  private handleChange = (manifests: any) => {
+  private handleChange = (manifestString: string) => {
```
sbwsg: One API suggestion but not a blocker at all: make the onChange signature 
```typescript
(raw: string, manifest: any, errors: []E)
```
so that the YamlEditor remains the only one responsible for `load`ing.

danielpeach: yes, that's better.

PR #5798, opened 2018-09-27
fix(core): Fix parameter in webpack dev configuration

ezimanyi:
The ca certificate should be passed as the parameter 'ca', not 'cacert'.


ezimanyi:
Docs: https://webpack.js.org/configuration/dev-server/#devserver-https

PR #5799, opened 2018-09-27
feat(ecs): support private registry credentials

clareliguori:
Addresses spinnaker/spinnaker#3339

Related to https://github.com/spinnaker/clouddriver/pull/2998


clareliguori:
screenshot from 2018-09-27 15-53-48

clareliguori:
Rerun build

jrsquared:
something odd trying to download yarn during tests it seems...
Also, please remove package-lock.json. We use `yarn` for deck.

app/scripts/modules/ecs/src/serverGroup/configure/#serverGroupConfiguration.service.ts#
```diff
@@ -0,0 +1,497 @@
+import { module, IPromise, IQService } from 'angular';
```
christopherthielen: What is this file?  It looks to be a dupe of `app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts` but with pound signs in the filename?

iiro: Looks like an Emacs buffer auto-save... :)

clareliguori: Whoops you are right, removed

christopherthielen: I didn't know.  I use vim, like a normal person.  ;)

PR #5806, opened 2018-09-28
chore(package): prepare -> prepublishOnly for everything

erikmunson:
As far as I can tell we don't rely on the behavior of `prepare` (which also gets triggered on `install` in addition to `publish`). This just updates us to a more use-case appropriate `prepublishOnly` script across all packages.

An added side-benefit of this is a more convenient experience for those of us that use [`yalc`](https://github.com/whitecolor/yalc).


jrsquared:
😴

erikmunson:
actual photo of trying to use `npm link` unless you're on @jrsquared's laptop:

fall

PR #5808, opened 2018-09-29
feat(core/presentation): Support validation when fields are touched

christopherthielen:
(…and some other things)

WizardModal: support formClassName prop
TextField: Only show validation errors when field is touched
BasicLayout: Support input field with no label
formikField/TextField: support touched prop and set touched from onBlur

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/fields/TextField.tsx
+++ b/app/scripts/modules/core/src/presentation/forms/fields/TextField.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as classNames from 'classnames';
 
 import { BasicLayout } from '../layouts';
 import { formikField, IFormikField } from './formikField';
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/presentation/forms/fields/TextField.tsx
```diff
@@ -10,14 +11,26 @@ export class TextField extends React.Component<ITextFieldProps> {
   public static defaultProps: Partial<IFieldProps> = { FieldLayout: BasicLayout };
 
   public render() {
-    const { label, help, error, warning, preview, actions, ...rest } = this.props;
-    const { FieldLayout, value, onChange, required, ...inputProps } = rest;
+    const { label, help, touched, error, warning, preview, actions, ...rest } = this.props;
+    const { FieldLayout, value, onChange, onBlur, required, ...inputProps } = rest;
+
+    const className = classNames({
```
christopherthielen: using these for now until consolidated forms design is complete and the style guide is updated

PR #5813, opened 2018-10-01
fix(google): backend service selection in server group wizard

danielpeach:


app/scripts/modules/core/src/forms/uiSelect.decorator.js
```diff
@@ -2,25 +2,23 @@
...
-            $select.selected.push(item);
-            scope.$selectMultiple.updateModel();
-          }
-        });
-      };
+    const [uiSelect] = $delegate,
+      originalLink = uiSelect.link,
+      SELECT_EVENT_KEY = 'uis:select';
+
+    uiSelect.link = function(scope, element, attrs, ctrls) {
```
sbwsg: this is a part of angular I don't know much about. What's the distinction between a `.compile` and a `.link` function? It looke like the function is doing the same thing in both cases.

PR #5816, opened 2018-10-02
 fix(titus/runJob): Set defaults earlier in lifecycle

jrsquared:


app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
```diff
@@ -34,6 +34,48 @@ export class TitusRunJobStageConfig extends React.Component<IStageConfigProps, I
...
+    }
+
+    const clusterDefaults = {
+      application: application.name,
+      env: {},
+      resources: {
+        cpu: 1,
+        disk: 10000,
+        gpu: 0,
+        memory: 512,
```
christopherthielen: `menu`

PR #5830, opened 2018-10-05
Create react version of account select and use it in titus run job

jrsquared:


app/scripts/modules/core/src/account/AccountSelectField.tsx
```diff
@@ -1,13 +1,142 @@
...
+import { $q } from 'ngimport';
+import { flatten, has, intersection, isEqual, map, uniq, xor } from 'lodash';
+
 import { IAccount } from 'core/account';
+import { AccountService } from 'core/account/AccountService';
 
 export interface IAccountSelectFieldProps {
   accounts: IAccount[] | string[];
-  component: Object;
+  component: { [key: string]: any };
```
christopherthielen: I think you could use `object` with lower case `o` if you prefer

app/scripts/modules/core/src/account/AccountSelectField.tsx
```diff
@@ -1,13 +1,142 @@
...
+    if (!accounts || !accounts.length) {
+      return;
+    }
+    if (has(component, field) && isExpression(component[field])) {
+      this.setState({ accountContainsExpression: true });
+      return;
+    }
+
+    const accountsAreObjects = Boolean((accounts[0] as IAccount).name);
+    let getAccountDetails = provider ? AccountService.getAllAccountDetailsForProvider(provider) : $q.when([]);
```
christopherthielen: looks like an extra api call to `getAllAccountDetailsForProvider` is made then thrown away when `accountsAreObjects === true`

christopherthielen: looks like an extra api call to `getAllAccountDetailsForProvider` is made then thrown away when `accountsAreObjects === true`

christopherthielen: Also, looks like normally `getAccountDetails` is a promise for an array of objects.  But when `provider` is truthy and `accountsAreObjects` are falsey, it's a promise for an object not an array of objects?

app/scripts/modules/core/src/account/AccountSelectField.tsx
```diff
@@ -1,13 +1,142 @@
...
+      if (accountNames) {
+        primaryAccounts = accountNames.sort();
+      }
+      if (accountNames && accountNames.length && details.length) {
+        primaryAccounts = accountNames
+          .filter(account => {
+            return details.some(detail => detail.name === account && detail.primaryAccount);
+          })
+          .sort();
+        secondaryAccounts = xor(accountNames, primaryAccounts).sort();
```
christopherthielen: fancy `xor` look at you go wheeeee

app/scripts/modules/core/src/account/AccountSelectField.tsx
```diff
@@ -1,13 +1,142 @@
...
+        mergedAccounts = flatten([primaryAccounts, secondaryAccounts]);
+      }
+
+      if (component) {
+        mergedAccounts = mergedAccounts || [];
+        const comp = flatten([component[field]]);
```
christopherthielen: this seems off since `flatten` is a shallow operation and you're passing in an array of one element.  I think this will always be equivalent to `const comp = component[field]`

app/scripts/modules/core/src/account/AccountSelectField.tsx
jrsquared: Oh nice, didn't know about that one.

jrsquared: Ah, then can't do `object[key]` because it complains of no index signature. sad face.

app/scripts/modules/core/src/account/AccountSelectField.tsx
jrsquared: Yar, good catch. Fixed.

app/scripts/modules/core/src/account/AccountSelectField.tsx
jrsquared: Good catch. Leftover cruft from not copying over the multi select property from the angular directive since we aren't using it anywhere.

PR #5835, opened 2018-10-05
feat(core): skip stage button

danielpeach:
Corresponds to https://github.com/spinnaker/orca/pull/2445

Deploy stages can't be skipped - just using deploy as an example.

Button:
skip_deploy
Modal:
skip_deploy_modal
After skipping a stage:
skip_deploy_tooltip


christopherthielen:
@gcomstock here's an example of a tooltip over a date

PR #5850, opened 2018-10-12
feat(webhook): show status code and response body

emjburns:


app/scripts/modules/core/src/pipeline/config/stages/webhook/webhookExecutionDetails.html
```diff
@@ -32,6 +32,10 @@ <h4>Results</h4>
         <dd class="ng-binding">{{ctrl.stage.status}}</dd>
         <dt>Info</dt>
         <dd class="ng-binding">{{ctrl.stage.context.progressMessage}}</dd>
+        <dt>Code</dt>
+        <dd class="ng-binding">{{ctrl.stage.context.webhook.statusCodeValue}}</dd>
+        <dt>Response</dt>
+        <dd class="ng-binding" style="max-height:400px;overflow-y:auto;">{{ctrl.stage.context.webhook.body}}</dd>
```
lwander: Just wondering, if the webhook's body has malicious javascript, is it run inside of deck?

christopherthielen: no, it is sanitized by angularjs

PR #5851, opened 2018-10-12
feat(rollback): Support wait before disable during an orchestrated rollback

ajordens:
We support this for a normal red/black and have been asked to also allow it during a rollback.

related to spinnaker/orca#2456


ajordens:
image

ajordens:
image

ajordens:
Made the change for aws and titus since those are easy enough for us to validate ... 

Other providers should be able to make a very similar if not identical change -- @danielpeach @bruno-carrier-lookout @owainlewis @Jammy-Louie

ajordens:
Just realized I could update the text from `before disabling old server group` to `before disabling spintestajordens-v026`

PR #5852, opened 2018-10-14
feat(deck): Support Github Status notification type

amanya:
This is to support the github status notification type introduced in https://github.com/spinnaker/echo/pull/344

image


app/scripts/modules/core/src/notification/types/githubstatus/githubstatus.notification.type.module.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const angular = require('angular');
+
+module.exports = angular
+  .module('spinnaker.core.notification.types.githubstatus', [])
+  .config(function(notificationTypeConfigProvider) {
+    notificationTypeConfigProvider.registerNotificationType({
+      label: 'Github Status',
+      key: 'githubStatus',
```
danielpeach: This is the only camel cased `githubStatus` - is this one wrong?

app/scripts/modules/core/src/notification/types/githubstatus/githubstatus.notification.type.module.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
```
danielpeach: The convention for new files in Deck is to use TypeScript rather than JavaScript.

app/scripts/modules/core/src/notification/types/githubstatus/githubstatus.notification.type.module.js
amanya: No, it's the notification type in echo: https://github.com/spinnaker/echo/pull/344/files#diff-59ef91a5c716e1ef92630c9252135c89R131

sbwsg:
Hi @amanya do you intend to finish up the remaining change in this PR (JavaScript -> TypeScript)?  I can make the change if you don't have time.

amanya:
@sbwsg Hi! If you could do it would be great :) Thanks!

sbwsg:
Updated to TypeScript in https://github.com/spinnaker/deck/pull/6084.  Closing this one.

PR #5854, opened 2018-10-16
feat(pipeline_templates): Display "Force Rebake" checkbox for MPT

jervi:
When manually executing templated pipelines, the "Force Rebake" checkbox is missing, even though the pipeline contains a bake stage. Templated pipelines do not contain any stages before runtime, so the bake stage is never asked to inject its custom logic into the manual execution modal.

This is a simple 90/10-solution that just ~~iterates through previous executions, finds the previous execution that actually ran, and naïvely copies the~~ plans a new execution and adds the stages to the pipeline, so that the bake stage magic is executed.
It might result in false positives or negatives if you are changing the pipeline template, but as the main structure of a pipeline rarely changes and the fact that there are no bad consequences of displaying a "Force Rebake" checkbox when there is no bake stage, I think this is good enough.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
@@ -132,6 +132,17 @@ module.exports = angular
       const pipeline = this.command.pipeline,
         executions = application.executions.data || [];
 
+      if (
+        pipeline.type === 'templatedPipeline' &&
+        executions.length > 0 &&
+        (pipeline.stages === undefined || pipeline.stages.length === 0)
+      ) {
+        const previousSuccessfulExecution = executions.find(e => e.stages.length > 0);
+        if (previousSuccessfulExecution) {
+          pipeline.stages = previousSuccessfulExecution.stages;
+        }
+      }
+
       pipelineNotifications = pipeline.notifications || [];
       synchronizeNotifications();
 
--- a/app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
@@ -7,6 +7,7 @@ import { AuthenticationService } from 'core/authentication';
 import { Registry } from 'core/registry';
 import { SETTINGS } from 'core/config/settings';
 import { AppNotificationsService } from 'core/notification/AppNotificationsService';
+import { PipelineTemplateReader } from 'core/pipeline/config/templates/PipelineTemplateReader';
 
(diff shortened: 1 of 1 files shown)


jervi:
@spinnaker/reviewers PTAL

christopherthielen:
I'm not sure I'm comfortable with this, but I don't know MPT well enough to offer a better suggestion

robzienert:
It's awkward for sure, but is the pattern we've gone with on the backend as well for outlier behaviors in MPT.

app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
```diff
@@ -132,6 +132,17 @@ module.exports = angular
...
         executions = application.executions.data || [];
 
+      if (
+        pipeline.type === 'templatedPipeline' &&
+        executions.length > 0 &&
+        (pipeline.stages === undefined || pipeline.stages.length === 0)
+      ) {
+        const previousSuccessfulExecution = executions.find(e => e.stages.length > 0);
+        if (previousSuccessfulExecution) {
+          pipeline.stages = previousSuccessfulExecution.stages;
```
danielpeach: Could you plan the pipeline here instead? Just a thought.

jtk54: :+1:

app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.controller.js
```diff
@@ -109,6 +110,21 @@ module.exports = angular
       this.command.trigger = suppliedTriggerCanBeInvoked ? trigger : _.head(this.triggers);
     };
 
+    const updatePipelinePlan = pipeline => {
+      if (pipeline.type === 'templatedPipeline' && (pipeline.stages === undefined || pipeline.stages.length === 0)) {
+        PipelineTemplateReader.getPipelinePlan(pipeline)
```
jervi: `getPipelinePlan` takes an optional second parameter `executionId`. Ideally I should've used the build selected in the build drop down if available, but that turned out to be a rabbit hole. The builds in that list only have the CI build number, not the `executionId`. The `executions` variable only have a few executions for each pipeline (depending on the number of pipelines being displayed). And since the dropdown is a react component, I have to use a $watch expression to detect when it change.
So the current implementation will only consider the newest version of the pipeline when deciding wether to show the force rebake box or not. But I think this is Good Enough™, for the same reasons mentioned earlier. It's certainly better than never showing it.

jervi:
Take 2! It's not perfect, but it's better. I have implemented the plan suggestion from @danielpeach, which is of course the way it should have been done. The implementation is still a bit naïve, but I don't think it's worth it to put a lot more effort into this.

PR #5858, opened 2018-10-16
fix(amazon): Fixed cloning security group across accounts

alanmquach:
Cloning a security group gets the entire source security group object, then updates it with any modifications made in dialog. Unfortunately, "account" exists as multiple attributes:

- `account`
- `accountId`
- `accountName`
- `credentials`

and the `<account-select-field>` can only update a single attribute. To address this, we can copy the updated account to the rest of the fields.


alanmquach:
me too

PR #5860, opened 2018-10-16
fix(core/account): Fixed AccountSelectField missing state initialization

alanmquach:
Dropdowns were empty since `componentWillReceiveProps` is not called on mount and `props.account` never change after that so `groupAccounts` never has a chance to map them into `state`.


jrsquared:
Lame. The cases I tested this with must have had some other props set async to cause the side effect, thanks for fixing this!

PR #5861, opened 2018-10-17
fix(kubernetes): add support for json patch #3418

maggieneterval:
Closes #3418
- Adds JSON editor component
- Patch body editor toggles between YAML and JSON depending on selected merge strategy, and converts existing input between YAML/JSON appropriately
- See corresponding [CloudDriver change](https://github.com/spinnaker/clouddriver/pull/3047).

json patch


app/scripts/modules/kubernetes/src/v2/manifest/json/JsonEditor.tsx
```diff
@@ -0,0 +1,56 @@
...
+import AceEditor from 'react-ace';
+import { $log } from 'ngimport';
+
+import 'brace/theme/textmate';
+import 'brace/mode/json';
+import './jsonEditor.less';
+
+export interface IJsonEditorProps {
+  value: string;
+  onChange(raw: string, manifest: any): void;
```
danielpeach: Using the word `manifest` here for the yaml editor was a mistake on my part - maybe `obj` or something? (Doesn't seem like this component necessarily has to be used for editing manifests)

app/scripts/modules/kubernetes/src/v2/manifest/yaml/yamlEditor.less
```diff
@@ -1,6 +1,6 @@
 #yaml-editor {
```
danielpeach: maybe this file could be renamed and the selectors could apply to both editors?

app/scripts/modules/kubernetes/src/v2/manifest/json/jsonEditor.component.ts
```diff
@@ -0,0 +1,7 @@
+import { module } from 'angular';
+import { react2angular } from 'react2angular';
+
+import { JsonEditor } from './JsonEditor';
+
+export const JSON_EDITOR_COMPONENT = 'spinnaker.core.jsonEditor.component';
```
danielpeach: s/`spinnaker.core.jsonEditor.component`/`spinnaker.kubernetes.jsonEditor.component`

app/scripts/modules/kubernetes/src/v2/pipelines/stages/patchManifest/patchManifestConfig.controller.ts
```diff
@@ -84,4 +97,42 @@ export class KubernetesV2PatchManifestConfigCtrl implements IController {
   public handleManifestSelectorChange = (): void => {
     this.$scope.$applyAsync();
   };
+
+  public getMergeStrategy = (): MergeStrategy => {
```
danielpeach: if this method is only called within this controller, the access modifier can be `private`

app/scripts/modules/kubernetes/src/v2/pipelines/stages/patchManifest/patchManifestConfig.controller.ts
```diff
@@ -84,4 +97,42 @@ export class KubernetesV2PatchManifestConfigCtrl implements IController {
...
+    return this.$scope.stage.options.mergeStrategy;
+  };
+
+  public getEditorMode = (): EditorMode => {
+    return get(mergeStrategyToEditorMode, this.getMergeStrategy());
+  };
+
+  public setRawPatchBody = (mergeStrategy: MergeStrategy, patchBody: any): void => {
```
danielpeach: same here re: access modifier

app/scripts/modules/kubernetes/src/v2/pipelines/stages/patchManifest/patchManifestConfig.controller.ts
```diff
@@ -84,4 +97,42 @@ export class KubernetesV2PatchManifestConfigCtrl implements IController {
   public handleManifestSelectorChange = (): void => {
     this.$scope.$applyAsync();
   };
+
+  public getMergeStrategy = (): MergeStrategy => {
+    return this.$scope.stage.options.mergeStrategy;
+  };
+
+  public getEditorMode = (): EditorMode => {
+    return get(mergeStrategyToEditorMode, this.getMergeStrategy());
```
danielpeach: I think you could get away with not using `get` here

app/scripts/modules/kubernetes/src/v2/manifest/json/JsonEditor.tsx
maggieneterval: Good call, replaced with `obj` here and in the `YamlEditor`.

app/scripts/modules/kubernetes/src/v2/manifest/yaml/yamlEditor.less
maggieneterval: Done!

PR #5871, opened 2018-10-19
feat(stages/evaluatevariables): Add UI for Evaluate Variables stage

alanmquach:
Just stage config and execution details UI to support the new stage @emjburns built!


app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/evaluateVariablesStage.ts
```diff
@@ -0,0 +1,16 @@
+import { Registry } from 'core/registry';
+
+import { ExecutionDetailsTasks } from '../core';
+import { EvaluateVariablesExecutionDetails } from './EvaluateVariablesExecutionDetails';
+import { EvaluateVariablesStageConfig } from './EvaluateVariablesStageConfig';
+
+Registry.pipeline.registerStage({
+  label: 'Evaluate Variables',
+  description: 'evalutes variables',
```
emjburns: Can we change to "Evaluates variables for use in SpEL expressions in downstream stages. Variables can be accessed by their key."?

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
+
+import { StageExecutionLogs, StageFailureMessage } from 'core/pipeline/details';
```
christopherthielen: Let’s export all this stuff from index files then import from `core/pipeline`

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
```diff
@@ -0,0 +1,44 @@
...
+export function EvaluateVariablesExecutionDetails(props: IExecutionDetailsSectionProps) {
+  const {
+    stage: { context = {}, outputs = {} },
+  } = props;
+
+  const evaluatedVariables = context.variables ? (
+    <div>
+      <dl>
+        {context.variables.map((evalPair: IEvaluatedVariables) => (
+          <React.Fragment key={evalPair.key}>
```
christopherthielen: Is this necessary syntax because of the `key`?

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
```diff
@@ -0,0 +1,44 @@
...
+  ) : (
+    <div>
+      <b>No variables were evaluated.</b>
+    </div>
+  );
+
+  return (
+    <ExecutionDetailsSection name={props.name} current={props.current}>
+      {evaluatedVariables}
+      <StageFailureMessage stage={props.stage} message={props.stage.failureMessage} />
```
christopherthielen: Would prefer to destruct these from `props`

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <StageExecutionLogs stage={props.stage} />
+    </ExecutionDetailsSection>
+  );
+}
+
+export namespace EvaluateVariablesExecutionDetails {
```
christopherthielen: Not understanding the namespace?

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -0,0 +1,75 @@
+import { IStageConfigProps } from 'core/pipeline/config/stages/core/IStageConfigProps';
```
christopherthielen: Imports order

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export interface IEvaluateVariablesStageConfigState {
+  variables: any;
+}
+
+export class EvaluateVariablesStageConfig extends React.Component<
+  IStageConfigProps,
+  IEvaluateVariablesStageConfigState
+> {
+  public static getDerivedStateFromProps(props: IStageConfigProps): IEvaluateVariablesStageConfigState {
```
christopherthielen: I assume you have read https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html ?
I there a better approach?

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const {
+      stage: { variables = [] },
+    } = props;
+    return {
+      variables: EvaluateVariablesStageConfig.compress(variables),
+    };
+  }
+
+  private static compress(variables: any) {
```
christopherthielen: Lodash has `indexBy`

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
alanmquach: Yes. `< key={x.key}>` is not valid syntax.

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
alanmquach: I guess we don't need state at all. It's just passing everything through to `MapEditor` anyways.

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
emjburns: For context, this will allow for other fields like 'type' in the future.

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
alanmquach: `keyBy` unfortunately uses the entire original object as the value so it would require another transform iteration to pluck.

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
alanmquach: Looks like it's already exported. I'll use `core/pipeline` and fix the other stages in a separate PR

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
alanmquach: I think it just allows the title to be accessible under `EvaluateVariablesExecutionDetails.title`

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
alanmquach: Nice catch, thanks. Copied the wrong order from `WaitStageConfig` without noticing!

I'll fix the order there as well in the aforementioned `other PR`.

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
christopherthielen: Hmmmmmmmm... I'm not happy with this pattern.   Seems like this should be a static property of the component class.  however, that should happen in a different PR.

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
christopherthielen: I learned by watching @jrsquared

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+
+import { StageExecutionLogs, StageFailureMessage } from 'core/pipeline';
+import { ExecutionDetailsSection, IExecutionDetailsSectionProps } from 'core/pipeline';
```
jrsquared: combine these two imports?

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import { map, set } from 'lodash';
+
+import { IStageConfigProps, StageConfigField, MapEditor } from '@spinnaker/core';
```
jrsquared: When in `core`, don't import from `@spinnaker/core`

alanmquach: Thanks! I was just looking at that.

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
alanmquach: 👍

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
```diff
@@ -0,0 +1,51 @@
...
+  StageExecutionLogs,
+  StageFailureMessage,
+} from 'core/pipeline';
+
+export interface IEvaluatedVariables {
+  key: string;
+  value: string;
+}
+
+export function EvaluateVariablesExecutionDetails(props: IExecutionDetailsSectionProps) {
```
christopherthielen:
```suggestion
export const EvaluateVariablesExecutionDetails = (props: IExecutionDetailsSectionProps) => {
```

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import { map, set } from 'lodash';
+
+import { IStageConfigProps, StageConfigField, MapEditor } from 'core';
```
christopherthielen:
```suggestion
import { IStageConfigProps, StageConfigField } from 'core/pipeline';
import { MapEditor } from 'core/forms';
```

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
alanmquach: I can apply this convention to the other stages that I'm cleaning up too

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -0,0 +1,47 @@
...
+export class EvaluateVariablesStageConfig extends React.Component<
+  IStageConfigProps,
+  IEvaluateVariablesStageConfigState
+> {
+  private expand(variables: any) {
+    return map(variables, (value, key) => ({ key, value }));
+  }
+
+  private stageFieldChanged = (fieldIndex: string, value: any) => {
+    set(this.props.stage, fieldIndex, this.expand(value));
```
christopherthielen: is this mutation necessary because of angularjs?

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import { map, set } from 'lodash';
+
+import { IStageConfigProps, StageConfigField, MapEditor } from 'core';
+
+export interface IEvaluateVariablesStageConfigState {
```
christopherthielen: I don't see this state being used anywhere?

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesExecutionDetails.tsx
alanmquach: Hmmm... this breaks the `namespace` thing.
```
[ts] Cannot redeclare block-scoped variable 'EvaluateVariablesExecutionDetails'.
```

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
alanmquach: I don't have the context on why it's done this way and I'm not sure it's entirely necessary, but it feels very awkward. I was going to look at whether there's a better alternative next.

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
alanmquach: doh! forgot to remove the types after I chucked out state completely.

nice catch. thanks!

PR #5873, opened 2018-10-20
fix(dependencies): Use fontawesome-free instead of deprecated fonrtawesome-free-webfonts

jaypz:
@fortawesome/fontawesome-free-webfonts is deprecated and no longer properly resolves from public npm, makes trying to run this locally a nuisance


christopherthielen:
thanks!  make sure you're using `yarn` and not `npm` to install packages.

christopherthielen:
I spot checked the UI and it looks fine with the updated package dependency/

PR #5876, opened 2018-10-22
fix(core/application): Add workaround for applications with { accounts: ['array','of','strings'] }

christopherthielen:


christopherthielen:
Not sure if this is necessary or not.  Maybe we should just update the data in front50.  @jrsquared ?

christopherthielen:
Updated the bad data in front50

PR #5885, opened 2018-10-24
feat(provider/aws): allow editing of load balancer idle timeout and deletion protection

tomaslin:


app/scripts/modules/amazon/src/loadBalancer/configure/classic/AdvancedSettings.tsx
```diff
@@ -79,11 +79,22 @@ class AdvancedSettingsImpl extends React.Component<IAdvancedSettingsProps> {
               <Field className="form-control input-sm" type="number" min="0" name="unhealthyThreshold" />
             </div>
           </div>
+
```
tomaslin: Classic load balancers only support idleTimeout

PR #5895, opened 2018-10-24
feat(stages/evaluatevariables): Default failOnFailedExpressions to true

alanmquach:
As this stage purely evaluates variables, it makes sense to enable `failOnFailedExpressions` by default so that it is opt-out instead of opt-in.


app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
```diff
@@ -178,6 +178,9 @@ module.exports = angular
           $scope.description = config.description;
           $scope.extendedDescription = config.extendedDescription;
           $scope.label = config.label;
+          if (config.configuration && config.configuration.failOnFailedExpressions) {
```
christopherthielen: yo dawg, maybe we can call this `config.defaults` or something else?

alanmquach: I actually started with `config.defaults`. :)

But `config.configuration` is already a thing, and it felt ok to chuck it in there instead of creating something new.

emjburns: wow online @christopherthielen uses much more slang than IRL @christopherthielen

alanmquach: Went back to `config.defaults` since that can also be used for other things.

PR #5902, opened 2018-10-26
fix(docker): fix binding of `handleRefreshImages` method #3514

maggieneterval:
Per [this issue](https://github.com/spinnaker/spinnaker/issues/3514), refreshing Docker images was failing with the following error due to unbound method:

console output


MrBlaise:
I think this should have been included in the latest 1.10.2 version. This is a small fix and without it, the feature is kind of broken.

PR #5907, opened 2018-10-29
feat(core): Support a date constraint on trigger parameters

jrsquared:
Support a date picker for trigger parameters.

If the parameter has `"constraint": "date"` in the parameter config in the pipeline JSON, then when manually triggering the pipeline, the input for the parameter will only accept a valid date string and will have a date picker.

Currently no UI to actually pick the constraint type, JSON only feature right now.

screen shot 2018-10-29 at 11 24 28 am

@asher FYI

### DIFF
--- a/app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.html
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.html
@@ -80,7 +80,25 @@
             <span ng-if="parameter.required">*</span>
             <help-field content="{{parameter.description}}" ng-if="parameter.description"></help-field>
           </div>
-          <div class="col-md-6" ng-if="!parameter.hasOptions">
+          <div class="col-md-6" ng-if="!parameter.hasOptions && parameter.constraint === 'date'">
+            <p class="input-group">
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/pipeline/manualExecution/manualPipelineExecution.html
```diff
@@ -80,7 +80,25 @@ <h3 ng-if="!vm.command.pipeline">Select Pipeline</h3>
             <span ng-if="parameter.required">*</span>
             <help-field content="{{parameter.description}}" ng-if="parameter.description"></help-field>
           </div>
-          <div class="col-md-6" ng-if="!parameter.hasOptions">
+          <div class="col-md-6" ng-if="!parameter.hasOptions && parameter.constraint === 'date'">
+            <p class="input-group">
+              <input type="text"
+                class="form-control input-sm"
+                uib-datepicker-popup
```
alanmquach: Cool, so this already existed and we just never used it?

jrsquared: Yeah, a part of the angular-ui-bootstrap

PR #5913, opened 2018-10-30
feat(core/presentation): Introduce Validation class for reusable validation fns.

christopherthielen:
- Support multiple validators and `required` flag in `FormField` and `FormikFormField`

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/Validation.ts
+++ b/app/scripts/modules/core/src/presentation/forms/Validation.ts
@@ -1,0 +1,11 @@
+export type ValidationFunction = (value: any) => string | Function | Promise<any>;
+
+export class Validation {
+  public static compose = (...validationFns: ValidationFunction[]): ValidationFunction => {
+    return (value: any) => validationFns.reduce((error, validationFn) => error || validationFn(value), null);
+  };
+
+  public static isRequired: ValidationFunction = (val: any) => {
+    return (val === undefined || val === null || val === '') && 'This field is required';
+  };
+}
--- a/app/scripts/modules/core/src/presentation/forms/Validation.ts
+++ b/app/scripts/modules/core/src/presentation/forms/Validation.ts
@@ -1,3 +1,4 @@
+export type ValidationFunctionFactory = (message?: string) => ValidationFunction;
 export type ValidationFunction = (value: any) => string | Function | Promise<any>;
 
 export class Validation {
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/presentation/forms/Validation.ts
```diff
@@ -0,0 +1,11 @@
+export type ValidationFunction = (value: any) => string | Function | Promise<any>;
+
+export class Validation {
+  public static compose = (...validationFns: ValidationFunction[]): ValidationFunction => {
+    return (value: any) => validationFns.reduce((error, validationFn) => error || validationFn(value), null);
+  };
+
+  public static isRequired: ValidationFunction = (val: any) => {
```
erikmunson: should there be some mechanism for providing a custom message that's not `This field is required`, or are you intending to use the `message` prop on both field components to solve those use cases?

christopherthielen: great idea 👍

christopherthielen: Maybe we should just use `yup`? https://github.com/jquense/yup#string

app/scripts/modules/core/src/presentation/forms/Validation.ts
```diff
@@ -0,0 +1,11 @@
+export type ValidationFunction = (value: any) => string | Function | Promise<any>;
+
+export class Validation {
```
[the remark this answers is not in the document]
christopherthielen: I originally had individual exports, but switched to static properties because:

- it seemed more explicit to write `Validation.isRequired` than `isRequired`.
- it would force us to be more consistent (not switching between `isRequired` and `Validation.isRequired` arbitrarily).  
- it would enforce the variable name for the latter style (no change of typing `import * as Validators from 'core/presentation'`)
- it allows you to discover validations by typing `Validation.<ctrl-space>`

At any rate, feel free to argue back if you prefer it.

app/scripts/modules/core/src/presentation/forms/Validation.ts
christopherthielen: I added another commit which supports a custom validation message.

erikmunson: I've been a bit ambivalent on `yup`. It feels like a pretty significant step into magic DSL territory and it's unclear to me how the method chaining style would feel with inline validations. On the flip side, as we're starting to build out validation constructs that will likely start getting lots of `yup`-esque assertions added over time I'm increasingly feeling more positive about it. Perhaps worth a little experimentation 🤷‍♂️

app/scripts/modules/core/src/presentation/forms/Validation.ts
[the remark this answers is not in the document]
erikmunson: I definitely agree with the point on enforcing import naming. I was surprised to see the thing about autocomplete though — I use ctrl + space autocomplete for that import style frequently in VS Code 🤭

PR #5917, opened 2018-10-31
 chore(core/presentation): Update formik from 0.11.11 to 1.3.1

christopherthielen:
Removes `formik-effect` which seems to have been abandoned.
Adds `create-react-context` which is a dependency of `formik` 1.x, and for some reason is not being hoisted by `yarn`.
Temporarily types some errors as `any` until we find a better way to manage `FormikError` typing (I think I want to move more towards inline validation, which doesn't depend on the type).


erikmunson:
🤘

PR #5918, opened 2018-10-31
feat(artifacts): Add support for helm/chart artifacts

Tizaniy:
issue: spinnaker/spinnaker#3284


app/scripts/modules/core/src/artifact/ExpectedArtifactSelectorViewController.ts
```diff
@@ -27,7 +27,10 @@ export class ExpectedArtifactSelectorViewController {
     } else {
       const artifact = ExpectedArtifactService.artifactFromExpected(expectedArtifact);
       const allAccounts = this.delegate.getExpectedArtifactAccounts();
-      this.accountsForArtifact = allAccounts.filter(a => a.types.includes(artifact.type));
+      this.accountsForArtifact =
+        artifact.type === 'helm/chart'
+          ? allAccounts.filter(a => a.types.includes(artifact.type) && a.name === artifact.reference)
```
sbwsg: I am a little confused as to why the set of accounts must be limited to those matching the artifact's reference.

Artifacts already have an artifactAccount field (https://github.com/spinnaker/kork/blob/master/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/model/Artifact.java#L46)

Does that field better suit this purpose?

Tizaniy: Thanks for your reply @sbwsg. I have rewritten code according to IArtifact interface.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/helm/HelmArtifactEditor.tsx
```diff
@@ -0,0 +1,94 @@
...
+  private onNameChange = () => {
+    ArtifactService.getArtifactVersions('helm', this.state.artifact.artifactAccount, this.state.artifact.name).then(
+      versions => {
+        this.setState({ versions });
+      },
+    );
+  };
+}
+
+export const ARTIFACT_ACCOUNT_SELECTOR_COMPONENT_REACT = 'spinnaker.core.artifacts.account.selector.react';
```
sbwsg: It looks like this is copy/paste from the ArtifactAccountSelector.tsx but not used anywhere else in this PR.  I think you can safely delete lines 83 - 94.

If you do need this module declaration for some reason then change these lines to something like 
```js
export const HELM_ARTIFACT_SELECTOR_COMPONENT_REACT = 'spinnaker.core.artifacts.helm.selector.react';
module(HELM_ARTIFACT_SELECTOR_COMPONENT_REACT, []).comp// ... and so on
```

Tizaniy: It's my fault. I deleted these lines

PR #5940, opened 2018-11-01
feat(kubernetes): enable / disable stage configs

danielpeach:
enable_stage


app/scripts/modules/core/src/widgets/ScopeClusterSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+  freeFormClusterField: boolean;
+}
+
+export class ScopeClusterSelector extends React.Component<IScopeClusterSelectorProps, IScopeClusterSelectorState> {
+  constructor(props: IScopeClusterSelectorProps) {
+    super(props);
+    const { clusters, model } = this.props;
+    const selectedNotInClusterList = !(
+      Array.isArray(clusters) &&
+      clusters.length &&
```
maggieneterval: Not sure you need this line since I think `some` will just return false if called on an empty array

danielpeach: Ah yep good catch

PR #5947, opened 2018-11-02
fix(artifacts): bake manifest artifact scope used before defined

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/3549


app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/bakeManifestConfig.controller.ts
```diff
@@ -116,9 +116,9 @@ class InputArtifact implements IArtifactAccountPair {
 
   constructor($scope: IScope, artifact = { id: '', account: '' }) {
     const unserializable = { configurable: false, enumerable: false, writable: false };
+    Object.defineProperty(this, '$scope', { ...unserializable, value: $scope });
```
sbwsg: This $scope is needed by the delegate in the line below

PR #5949, opened 2018-11-02
chore(dependencies): bump deck-kayenta to 0.0.63

sbwsg:


package.json
```diff
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-free-webfonts": "^1.0.9",
-    "@spinnaker/kayenta": "0.0.62",
+    "@spinnaker/kayenta": "^0.0.63",
```
maggieneterval: Just confirming that we want to add the caret rather than pin an exact version -- I think `^0.0.63` will update to most recent `0.x.y`

sbwsg: doh, good spot.  I did not intend a range.  Will update.

PR #5952, opened 2018-11-02
feat(kubernetes): include server group manager name in clickable clusters box

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/3516

Before:

screen shot 2018-11-02 at 5 10 49 pm

After:

screen shot 2018-11-02 at 5 04 22 pm

Open to suggestions of a better way to indicate clickability. The duplication definitely isn't ideal.


demobox:
> Open to suggestions of a better way to indicate clickability. The duplication definitely isn't ideal.

Agreed. In the screenshots, what is `deployment nginx-deployment` actually? Is that the name of the cluster (and deployment)? Or is the name `nginx-deployment` with an automatically added prefix `deployment`?

I ask because, if `deployment` is a _prefix_, it could probably be omitted from the server group manager name?

PR #5953, opened 2018-11-02
feat(core): Reuse project config for teams page

archana-s:
Reuse of existing project configuration component for Netflix-specific teams page. 

Projects and teams share the same configuration. Changes in this diff will make it easier to abstract out information from projects that are not (or not yet) relevant to teams. 
Ideal state for this would be to rename these files to remove Project from them and have "projects/teams" option. I will tackle it in a subsequent PR.


app/scripts/modules/core/src/projects/configure/ConfigureProjectModal.tsx
```diff
@@ -24,6 +24,7 @@ import './ConfigureProjectModal.css';
 export interface IConfigureProjectModalProps extends IModalComponentProps {
   title: string;
   projectConfiguration: IProject;
+  isTeams?: boolean;
```
christopherthielen: Maybe instead of an `isTeams` boolean we can have a discriminator like `type?: 'team' | 'project'`

christopherthielen:
I don't think this belongs in OSS deck.  Let's chat about moving it.

app/scripts/modules/core/src/projects/configure/ConfigureProjectModal.tsx
archana-s: +1. will fix.

archana-s:
Do you recommend creating a copy of this in deck-nflx specific to teams? That would be fine too.

PR #5966, opened 2018-11-05
fix(build): re-enable minification

danielpeach:
Closes https://github.com/spinnaker/spinnaker/issues/3524


christopherthielen:
CI has been timing out after 10 minutes.  I suspect it's because minification is taking too long.

danielpeach:
Something strange is happening. `yarn build` works fine, `./gradlew build` hangs.

PR #5976, opened 2018-11-06
fix(kubernetes): defaults can be passed in as null (#5975)

sbwsg:
cherry pick of https://github.com/spinnaker/deck/pull/5975


sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/3559

PR #5978, opened 2018-11-06
fix(kubernetes): add k8s kind in artifact icon list

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/3567

K8s kind now appears in list alongside artifact name:

screen shot 2018-11-06 at 3 45 37 pm


duftler:
nice

PR #5986, opened 2018-11-07
Create collapsible container list in ServerGroupHeader if more than 1 container per pod

jensvdh:
implements https://github.com/spinnaker/spinnaker/issues/3555

Collapses the list of containers per pod if there is more than 1 container. In cases with many sidecars, this can really clutter the Cluster View UI


app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
```diff
@@ -51,6 +51,65 @@ export class CloudProviderIcon extends React.Component<IServerGroupHeaderProps>
...
+      collapsed: true,
+    };
+    this.toggle = this.toggle.bind(this);
+  }
+
+  public render() {
+    const { images } = this.props;
+    const { collapsed } = this.state;
+
+    // if we only have 1 image, render as simple string.
```
sbwsg: i think the code here does a decent job of self-describing.  this comment doesn't add much beyond it.

app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
```diff
@@ -51,6 +51,65 @@ export class CloudProviderIcon extends React.Component<IServerGroupHeaderProps>
...
+    if (images.length === 1) {
+      return <span>{images[0]}</span>;
+    }
+
+    // otherwise, collapsed by default, with the option to toggle all.
+    return (
+      <>
+        {collapsed && (
+          <>
+            <span>{images.length} containers per pod. </span>
```
sbwsg: "container" and "pod" are both container-specific, whereas this header is shared amongst all providers where, for example, "VM Images" or "Instances" might be more apt descriptions.  I suggest slightly tweaking the approach here to render a single image in the collapsed view with a "(+ X more)" link that, when clicked, shows the complete list.  This way we can avoid introducing provider-specific language completely.

jensvdh: I see. Yeah, that sounds good!

jensvdh:
@sbwsg I implemented your feedback. PTAL when you have some time

app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
```diff
@@ -51,6 +51,62 @@ export class CloudProviderIcon extends React.Component<IServerGroupHeaderProps>
...
+    const { collapsed } = this.state;
+
+    return (
+      <>
+        {collapsed && (
+          <>
+            <span>{images[0]}</span>
+            &nbsp;
+            {images.length > 1 && (
+              <a href="javascript:void()" onClick={this.toggle}>
```
sbwsg: I think it might be semantically more appropriate to use a button here since the anchor isn't referencing anything.  Deck provides for a link-style button with the following:

```ts
<button className="link" onClick={this.toggle}>
  (+ {images.length - 1} more)
</button>
```

This may also require a little further styling to get padding & font size squared away.

app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
```diff
@@ -51,6 +51,62 @@ export class CloudProviderIcon extends React.Component<IServerGroupHeaderProps>
...
+          <>
+            {images.map((image, index) => (
+              <span key={image}>
+                {index > 0 && <br />}
+                {image}
+                {index < images.length - 1 ? ',' : ''}
+              </span>
+            ))}
+            <br />
+            <a href="javascript:;" onClick={this.toggle}>
```
sbwsg: Same here; replace with button.

app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
jensvdh: @sbwsg makes sense! Were you thinking of any specifics in terms of styling ? 
(color, padding, .. ?)

sbwsg: Only for it to match the neighbouring text as closely as possible. Same font-size, font, line-spacing etc as the image name text next to it.  It may be that the `link` classname does all of that in the context of this header.

jensvdh: Good catch! I think the only thing that had to be adjusted was the fontSize. Let me know if you want to change the color to match as well. But I'm assuming we want to keep the colors associated with the `link` style.

sbwsg:
Thanks @jensvdh !

PR #5992, opened 2018-11-08
chore(*): Update uirouter/react-hybrid+react+angularjs to latest versions

christopherthielen:
Related: https://github.com/ui-router/react-hybrid/pull/138

This PR adds support for rendering UIView content using React Portals. This allows the child views to use the react context from the parent uiview. This should make global context APIs  like Redux easier to use.


christopherthielen:
@danielpeach @sbwsg FYI

PR #5995, opened 2018-11-09
feat(core/presentation): Support internal Validator(s) in Inputs

christopherthielen:
- Add internal Validators in NumberInput for min/max
- Use internal validators for NumberInputs

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/classic/AdvancedSettings.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/classic/AdvancedSettings.tsx
@@ -23,44 +23,43 @@ class AdvancedSettingsImpl extends React.Component<IAdvancedSettingsProps> {
         <FormikFormField
           name="healthTimeout"
           label="Timeout"
+          required={true}
+          fastField={false} /* This field depends on  healthInterval */
           help={<HelpField id="loadBalancer.advancedSettings.healthTimeout" />}
-          input={NumberInput}
-          validate={[
-            Validation.minValue(0),
-            Validation.maxValue(values.healthInterval, 'Timeout must be less than the health Interval.'),
-          ]}
+          input={props => <NumberInput {...props} min={0} max={values.healthInterval} />}
+          validate={Validation.maxValue(values.healthInterval, 'Timeout must be less than the health Interval.')}
         />
 
         <FormikFormField
           name="healthInterval"
           label="Interval"
+          required={true}
           help={<HelpField id="loadBalancer.advancedSettings.healthInterval" />}
-          input={NumberInput}
-          validate={Validation.minValue(0)}
+          input={props => <NumberInput {...props} min={0} />}
         />
 
         <FormikFormField
           name="healthyThreshold"
           label="Healthy Threshold"
+          required={true}
           help={<HelpField id="loadBalancer.advancedSettings.healthyThreshold" />}
-          input={NumberInput}
-          validate={Validation.minValue(0)}
+          input={props => <NumberInput {...props} min={0} />}
         />
 
         <FormikFormField
           name="unhealthyThreshold"
           label="Unhealthy Threshold"
+          required={true}
           help={<HelpField id="loadBalancer.advancedSettings.unhealthyThreshold" />}
-          input={NumberInput}
-          validate={Validation.minValue(0)}
+          input={props => <NumberInput {...props} min={0} />}
         />
 
         <FormikFormField
           name="idleTimeout"
           label="Idle Timeout"
+          required={true}
           help={<HelpField id="loadBalancer.advancedSettings.idleTimeout" />}
-          input={NumberInput}
-          validate={Validation.minValue(0)}
+          input={props => <NumberInput {...props} min={0} />}
         />
 
         <div className="col-md-12">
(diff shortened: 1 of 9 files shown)


app/scripts/modules/amazon/src/loadBalancer/configure/classic/AdvancedSettings.tsx
```diff
@@ -23,44 +23,43 @@ class AdvancedSettingsImpl extends React.Component<IAdvancedSettingsProps> {
         <FormikFormField
           name="healthTimeout"
           label="Timeout"
+          required={true}
+          fastField={false} /* This field depends on  healthInterval */
```
alanmquach: If "depends on other field" is the test, would it make sense to rename this prop to reflect that? Would just make it more obvious.

app/scripts/modules/core/src/presentation/forms/inputs/NumberInput.tsx
```diff
@@ -9,10 +11,37 @@ interface INumberInputProps extends IFormInputProps, React.InputHTMLAttributes<a
...
+export class NumberInput extends React.Component<INumberInputProps> {
+  private getMinValidator = memoizeOne((min: any) => (isNumber(min) ? Validation.minValue(min) : undefined));
+  private getMaxValidator = memoizeOne((max: any) => (isNumber(max) ? Validation.maxValue(max) : undefined));
+
+  private validator: Validator = (val: any, label?: string) => {
+    const { min, max } = this.props;
+    const validators = [this.getMinValidator(min), this.getMaxValidator(max)].filter(x => !!x);
+
+    if (validators.length === 1) {
+      return validators[0](val, label);
```
alanmquach: Can `composeValidators` just handle this case?

christopherthielen: Yes! great idea, thanks.  I wanted to optimize this since this gets called a lot, but the optimization definitely belongs in `composeValidators`

christopherthielen:
"everything", heh

app/scripts/modules/amazon/src/loadBalancer/configure/classic/AdvancedSettings.tsx
```diff
@@ -23,44 +23,43 @@ class AdvancedSettingsImpl extends React.Component<IAdvancedSettingsProps> {
...
           help={<HelpField id="loadBalancer.advancedSettings.healthTimeout" />}
-          input={NumberInput}
-          validate={[
-            Validation.minValue(0),
-            Validation.maxValue(values.healthInterval, 'Timeout must be less than the health Interval.'),
-          ]}
+          input={props => <NumberInput {...props} min={0} max={values.healthInterval} />}
+          validate={Validation.maxValue(values.healthInterval, 'Timeout must be less than the health Interval.')}
```
jrsquared:
```suggestion
          validate={Validation.maxValue(values.healthInterval, 'Timeout must be less than the health interval.')}
```

I know you just copy/pasted my bad capitalization, but might as well fix it here ;)

app/scripts/modules/amazon/src/loadBalancer/configure/classic/AdvancedSettings.tsx
erikmunson: `derived` is accurate but a little opaque. `dependsOnOtherField` seems ok.

It's unfortunate that we have to push this cognitive overhead up to form code, but a boolean is a lot better than needing to explicitly define field dependencies so I'll take it.

alanmquach:
```suggestion
          fastField={false} /* This field depends on healthInterval */
```

erikmunson: re: `dependsOnField="healthInterval` in my experience adding dummy config rarely buys as much future value as one anticipates, requirements change and the dummy config eventually has to keep up or die. It may also get confusing when we run into the first case of a field depending on multiple other fields (or some kind of relationship that goes beyond the concept of specific field dependencies). Do we add some additional no-op array syntax to the mix then?

There's something to be said for not abstracting over Formik nomenclature — I wouldn't be upset about it staying as `fastField`.

erikmunson:
I have mixed feelings about this. I'm all for the cleaner boilerplate around validations for a standard input, very nice to read. It's also really positive that we totally sidestep concerns about folks setting things like `min`/`max` on the input but then forgetting to do validation that enforces it!

Where I get a little uneasy:
- Stateful things are now happening behind the scenes between underlying input components and their parent fields. It feels like a level of coupling that isn't obvious and complicates the flow of state.

- We originally went back and forth on centralized vs. field level validation for formik in part because fragmented validation can get harder to reason about. This goes even further and gets _inputs_ involved, so one now has to consider multiple layers and work out what validators an input will use based on props, then combine that with any overrides for the same validation at the field level.

I realize I'm the minority here, and that's fine 😄 — if we restrict usage of this pattern to very simple validation cases like `min`/`max` maybe the downsides will be minimal 🤷‍♂️

app/scripts/modules/amazon/src/loadBalancer/configure/classic/AdvancedSettings.tsx
christopherthielen: It's funny because I basically had this whole conversation in my head and settled on `fastField` before I submitted this PR.   Keeping as is then.

christopherthielen:
> I have mixed feelings about this.

Same.  The hidden complexity also makes me a bit uneasy, but like formik itself (which is very stateful and magic-y) I hope it improves DX overall, especially for non-ui engineers who just want to write a form.

> Stateful things are now happening behind the scenes

I don't care much that this approach is stateful, but would love to see any other suggestions!

> We originally went back and forth on centralized vs. field level validation

I don't have super strong opinions on this, but I am now leaning towards preferring inline validation for basic validation, and external validation for more complicated field interactions.

> if we restrict usage of this pattern

I think the use case is clearly limited to validations that are inherent to the input.  Expressions come to mind.

app/scripts/modules/core/src/presentation/forms/inputs/NumberInput.tsx
christopherthielen: Moved into `composeValidators` and this feels much better now, thanks.

PR #6001, opened 2018-11-09
fix(kubernetes): Allow resource name to be added as expression

maggieneterval:
…e can be entered as an expression

Resolves: https://github.com/spinnaker/spinnaker/issues/3513


christopherthielen:
@gcomstock PTAL

maggieneterval:
Played around with it a little more and pushed a change to clarify existing Creatable select functionality rather than adding the text/select toggle:

creatable

@gcomstock I defer to your design sensibilities :)

christopherthielen:
@maggieneterval I didn’t mean that you should not merge as is. I wanted Greg to use this as a real world example. We’ve been discussing how to handle expressions all across spinnaker and this is a good data point.

maggieneterval:
@christopherthielen I didn't realize how the Creatable component worked so I think the second fix is cleaner for now. But curious what we land on regarding handling expressions across Spinnaker as a whole!

gcomstock:
Thanks for tagging me on this! Sorry I didn't catch it Friday. I did some early work on how we might validate/render expressions in spinnaker here:

https://gallery.io/projects/MCHbtQVoQ2HCZcM1NWkzJjWA/files/MCGgtPMLb5Pry9LxMg_LJ1em

In the case of a dropdown + ability to use a custom expression, let me add this use case to the form field design deck tomorrow. I'll share it in the OSS spinnaker channel as soon as I put it in. I don't think there's anything wrong with merging this now though. Thanks for asking me!

maggieneterval:
@gcomstock Looks great, thanks for putting that together!

maggieneterval:
@danielpeach Would it be possible to include this with 1.10.4? This issue is a blocker for the user who posted the issue

gcomstock:
Here’s a few different options for how a key can take either a predefined values or a custom expression. Please share your thoughts on the slide at your convenience: https://gallery.io/projects/MCHbtQVoQ2HCZcM1NWkzJjWA/files/MCHtA7U1iMGr6xG3sIbzweygP8eVUeK21tE

maggieneterval:
Thanks @gcomstock, this is great. I like options A and B, especially the use of the `${*}` icon to identify all expression-enabled inputs.

PR #6008, opened 2018-11-10
feat(core/pipeline): Indicate quiet period enabled on pipeline trigger status

jrsquared:
Tooltip could use less wrapping, but functional for now.

screen shot 2018-11-10 at 11 54 28 am
screen shot 2018-11-10 at 11 53 50 am

@gcomstock @skandragon PTAL


app/scripts/modules/core/src/pipeline/triggers/TriggersTag.tsx
```diff
@@ -25,27 +29,41 @@ export class TriggersTag extends React.Component<ITriggersTagProps, ITriggersTag
       activeTriggerCount = filter(pipeline.triggers, { enabled: true }).length;
     }
 
+    const hasQuietPeriod = SETTINGS.feature.quietPeriod && SETTINGS.quietPeriod && SETTINGS.quietPeriod.length === 2;
+
     this.state = {
       triggerCount,
       activeTriggerCount,
+      quietPeriodStart: hasQuietPeriod ? new Date(SETTINGS.quietPeriod[0]) : undefined,
```
christopherthielen: why is this in state?

jrsquared: 🤷‍♂️ It doesn't have to be

PR #6015, opened 2018-11-12
fix(build): Fix yarn lockfile

ezimanyi:
It looks like the yarn lockfile wasn't properly updated after upgrading uirouter to 5.0.21; commit the update that is generated by running yarn on the current package.json.


christopherthielen:
My bad.  I actually need to bump `@uirouter/react-hybrid` again today. Won't hurt to merge this first though.

PR #6022, opened 2018-11-12
fix(core): Fixed DeploymentStrategySelector not updating strategy

alanmquach:
If `onStrategyChange` is not used, the new strategy doesn't go anywhere.

This used to be done by `<ui-select ng-model="$ctrl.command.strategy">`.


jrsquared:
oh angular...

PR #6024, opened 2018-11-12
feat(kubernetes): add execution details for scale + delete stages

danielpeach:


app/scripts/modules/kubernetes/src/v2/pipelines/stages/deleteManifest/deleteManifestStage.ts
```diff
@@ -1,24 +1,27 @@
...
       Registry.pipeline.registerStage({
         label: 'Delete (Manifest)',
         description: 'Destroy a Kubernetes object created from a manifest.',
-        key: 'deleteManifest',
+        key: STAGE_KEY,
         cloudProvider: 'kubernetes',
         templateUrl: require('./deleteManifestConfig.html'),
         controller: 'KubernetesV2DeleteManifestConfigCtrl',
         controllerAs: 'ctrl',
+        executionDetailsSections: [manifestExecutionDetails(STAGE_KEY), ExecutionDetailsTasks],
```
lwander: Delete can delete multiple manifests when label selectors are used -- how is that rendered?

danielpeach: updated to render the label selectors and the deleted manifest names

lwander: SGTM

PR #6028, opened 2018-11-13
fix(kubernetes): prevent namespace as expression from resetting to null in ManifestSelector

maggieneterval:


app/scripts/modules/kubernetes/src/v2/manifest/selector/ManifestSelector.tsx
```diff
@@ -179,7 +179,8 @@ export class ManifestSelector extends React.Component<IManifestSelectorProps, IM
       .map(([kind]) => kind)
       .sort();
 
-    if (namespaces.every(ns => ns !== this.state.selector.location)) {
+    const isLocationExpression = this.state.selector.location.includes('${');
```
danielpeach: Could use `this.isExpression(...)` instead

maggieneterval: Oh duh didn't notice that method, thanks Dan!

PR #6044, opened 2018-11-15
fix(provider/kubernetes): Change FileSystem to Filesystem

tpayne1:
This addresses issue: #2993. The problem is that deck pushes `FileSystem` and cloud driver expects `Filesystem`. This causes the pod yamls to not be updated and thus the deployed state to not reflect state as displayed in deck.


ethanfrogers:
spinnaker/spinnaker#2993

lwander:
@spinnakerbot cherry-pick 1.10

lwander:
@spinnakerbot cherry-pick 1.10

sorry for the noise :)

ethanfrogers:
`sudo cherry-pick 1.10` 🤣

PR #6052, opened 2018-11-16
 fix(amazon/serverGroup): Show the list of changed values (min/max/desired)

christopherthielen:
In a previous PR we change deck to only submit the changed min/max/desired values to the back end.
This addresses feedback from @dreynaud about surfacing the details about which fields have actually changed.  The user may assume they're submitting all three values when they're only changing one value.

This also fixes duplicate validation messages appearing.

aws resize 2


christopherthielen:
@tomaslin would you like this treatment for titus as well?

PR #6053, opened 2018-11-16
fix(webhook): Move from buildInfo to webhook field. buildInfo is deprecated.

PerGon:


PerGon:
@spinnaker/reviewers ptal, thanks

PR #6061, opened 2018-11-20
fix(kubernetes): yaml editor support for multi docs

maggieneterval:
Fixes support in YAML editor for multiple manifests entered as either a list of maps or a YAML multi-doc
- Replace js-yaml `load` with `loadAll` to handle YAML multi-docs
- Update ys-yaml to 3.9.0 so iterator is not required for `loadAll` method (will return list of YAML documents by default)

@danielpeach I removed a bunch of logic that it doesn't look like we need anymore but let me know if I'm breaking something I don't understand...

zrgtttbccmk


maggieneterval:

PR #6062, opened 2018-11-20
refactor(provider/cf): use DeploymentStrategySelector react component

Jammy-Louie:
refactor the ui to use the React component `DeploymentStrategySelector`

spinnaker/spinnaker#3622


app/scripts/modules/cloudfoundry/src/serverGroup/configure/wizard/sections/basicSettings.cf.tsx
```diff
@@ -142,28 +124,15 @@ class BasicSettingsImpl extends React.Component<
             Start on creation <HelpField id="cf.serverGroup.startApplication" />
           </div>
           <div className="checkbox checkbox-inline">
-            <input
-              type="checkbox"
-              checked={this.props.formik.values.startApplication}
-              onChange={this.startApplicationUpdated}
-            />
+            <input type="checkbox" checked={values.startApplication} onChange={this.startApplicationUpdated} />
```
jkschneider: Can we take advantage of what @jmelchio now knows about using the formik `Field` element?

Jammy-Louie: 👍

PR #6063, opened 2018-11-21
feat(provider/cf):  Allow health check config on deploy

claymccoy:


app/scripts/modules/cloudfoundry/src/serverGroup/configure/wizard/sections/configurationSettings.cf.tsx
```diff
@@ -111,6 +117,55 @@ class ConfigurationSettingsImpl extends React.Component<ICloudFoundryServerGroup
...
+                      value="port"
+                      checked={manifest.healthCheckType === 'port'}
+                      onChange={() => this.healthCheckTypeUpdated('port')}
+                    />{' '}
+                    port
+                  </label>
+                </div>
+                <div className="radio radio-inline">
+                  <label>
+                    <input
```
jkschneider: Can we use Formik's `Field` here?

claymccoy: I don't know. I followed the example of how the other radio buttons were done here and I'm inclined to keep it consistent for now.

christopherthielen: I'm working on a consolidated forms system for Spinnaker in React.  I will submit a PR next week converting some cloudfoundry forms to use `<FormikFormField />` and/or `<FormField />`.  See https://github.com/spinnaker/deck/tree/master/app/scripts/modules/core/src/presentation/forms

christopherthielen: Here are some examples:

### TextInput

https://github.com/spinnaker/deck/blob/bd5d3d25a045490e6dadb63bb2a2be6f3bf3b77f/app/scripts/modules/amazon/src/loadBalancer/configure/application/ConfigureOidcConfigModal.tsx#L84-L89

### NumberInput

(has built in min/max validation)

https://github.com/spinnaker/deck/blob/d08b202d637333b9b505b85593b48207b69db478/app/scripts/modules/amazon/src/loadBalancer/configure/classic/HealthCheck.tsx#L51-L55

### SelectInput

https://github.com/spinnaker/deck/blob/d08b202d637333b9b505b85593b48207b69db478/app/scripts/modules/amazon/src/loadBalancer/configure/classic/HealthCheck.tsx#L44-L48

### ReactSelectInput

https://github.com/spinnaker/deck/blob/19b83ce4f5039ccad7a4cb7f40c08da7a57a425f/app/scripts/modules/core/src/projects/configure/Clusters.tsx#L111-L117

### CheckboxInput

https://github.com/spinnaker/deck/blob/d08b202d637333b9b505b85593b48207b69db478/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx#L27-L32

### RadioButtonInput

This example uses `<FormField/>` and `onChange/value` instead of `<FormikFormField/>`.  This is because the value applied to the model is an object, not a simple `string`.

https://github.com/spinnaker/deck/blob/7a4b0cd77d993d801f4b9b96178a8f0b4a69fd8b/app/scripts/modules/core/src/entityTag/EntityTagEditor.tsx#L184-L194

christopherthielen: Your radio button would be something like:

```ts
const options: Array<Option<string>> = [
  { label: 'port', value: 'port' },
  { label: 'HTTP', value: 'http' },
  { label: 'process', value: 'process' },
]
```


```js
 <FormikFormField 
   name="manifest.healthCheckType" 
   label="Health Check Type" 
   required={true} 
   input={props => <RadioButtonInput {...props} options={options} />} 
 /> 
```

PR #6067, opened 2018-11-22
feat(provider/cf): option to define timeouts for service creation/destroy

Jammy-Louie:
add an option to override the time out setting when deploying or destroying services in pipeline stages.

spinnaker/spinnaker#3637

### DIFF
--- a/app/scripts/modules/cloudfoundry/src/help/cloudfoundry.help.ts
+++ b/app/scripts/modules/cloudfoundry/src/help/cloudfoundry.help.ts
@@ -1,12 +1,16 @@
 import { HelpContentsRegistry } from '@spinnaker/core';
 
 const helpContents: { [key: string]: string } = {
+  'cf.service.deploy.timeout':
+    '(Optional) <b>Override Deploy Timeout</b> is the maximum amount of time allowed for service deployment before Spinnaker reports a timeout error (default: 450 seconds). If a timeout error is reported by Spinnaker, the request may still be processed and a service may still be created.',
+  'cf.service.destroy.timeout':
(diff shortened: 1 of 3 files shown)


app/scripts/modules/cloudfoundry/src/help/cloudfoundry.help.ts
```diff
@@ -1,6 +1,10 @@
 import { HelpContentsRegistry } from '@spinnaker/core';
 
 const helpContents: { [key: string]: string } = {
+  'cf.service.deploy.timeout':
+    '(Optional) <b>Override Deploy Timeout</b> is to override the default timeout value of 450 seconds',
```
claymccoy: seems like the help content should explain the impact that value has without reusing the terms being described

Jammy-Louie: I've changed the text but I'm open for suggestions on better help text.

claymccoy: I meant answering what a deploy or destroy timeout mean. I actually don't know exactly. Does it expect the deployment to happen and become healthy within a certain amount of time? If it doesn't then what happens? Does it try again, does the server group get deleted? What if it times out on a destroy? I guess it is still taking traffic? What info could help me understand what these values should be for my application?

Jammy-Louie: 👍

jkschneider: @claymccoy If I understand you correctly, you'd still expect to see some text that says something to the effect that "even if the operation times out, a service may yet get created and be bindable" and the opposite for destroy?

Jammy-Louie: 👍 
I've added additional clarification to the help text

PR #6072, opened 2018-11-23
fix(provider/cf): populate red/black strategy additional fields

Jammy-Louie:
populate the red/black additional fields with the previously saved values from the pipeline stages.


Jammy-Louie:
removed the population of property scale down

PR #6086, opened 2018-11-28
fix(core/pipeline): Fix manual execution dropdown when execution has no buildInfo

christopherthielen:
This should fix this problem: 

whops


erikmunson:
noice 🥇

PR #6087, opened 2018-11-28
fix(core/executions): Fix NPE in ExecutionGroup->Notifications when there is a Strategy visible

christopherthielen:
Fixes this error:

screen shot 2018-11-27 at 5 22 24 pm

### DIFF
--- a/app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
+++ b/app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
@@ -41,14 +41,14 @@ export interface IExecutionGroupState {
 }
 
 export class ExecutionGroup extends React.Component<IExecutionGroupProps, IExecutionGroupState> {
-  private strategyConfig: IPipeline;
+  public state: IExecutionGroupState;
   private expandUpdatedSubscription: Subscription;
   private stateChangeSuccessSubscription: Subscription;
 
   constructor(props: IExecutionGroupProps) {
     super(props);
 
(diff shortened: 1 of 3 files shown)


erikmunson:
🤷‍♂️ 👍

app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -276,14 +276,16 @@ export class ExecutionGroup extends React.Component<IExecutionGroupProps, IExecu
...
                 </h4>
-                <EntityNotifications
-                  entity={pipelineConfig}
-                  application={this.props.application}
-                  entity-type="pipeline"
-                  hOffsetPercent="20%"
-                  placement="top"
-                  onUpdate={() => this.props.application.refresh()}
-                />
+                {pipelineConfig && (
```
christopherthielen: I'm not sure if a `strategyConfig` can have entity tags.  Seems like it could, in which case this code should pass `pipelineConfig || strategyConfig` to the `EntityNotifications` component.   

@nvndnl do you know?

christopherthielen: I've found the code that adds the tags to pipelines and it does not add tags to strategies.

https://github.com/spinnaker/deck/blob/7a4b0cd77d993d801f4b9b96178a8f0b4a69fd8b/app/scripts/modules/core/src/entityTag/entityTags.dataSource.ts#L24-L49

ezimanyi:

PR #6094, opened 2018-11-28
feat(provider/cf): Deploy Service via manifest

jmelchio:
- spinnaker/spinnaker#3592


jmelchio:
Should merge together with: https://github.com/spinnaker/clouddriver/pull/3188

PR #6097, opened 2018-11-28
fix(core): Actually default to using new manual trigger endpoint

ezimanyi:
The last change to default to using the new manual trigger endpoint
only affected users running a localgit setup; to default to enabled
for halyard users, we need to change this file.

To turn off the settings, users can add the following to
settings-local.js:
window.spinnakerSettings.feature.triggerViaEcho = false;


ezimanyi:

PR #6105, opened 2018-11-29
feat(provider/cf): shorten the appsManagerUri

Jammy-Louie:
update the display text of apps manager uri to something human readable

spinnaker/spinnaker#3664


jkschneider:
Could we shorten further by removing SG name? SG name appears at the top of the details heading anyway.

Jammy-Louie:
I've removed the SG name from the link text

PR #6106, opened 2018-11-29
fix(provider/cf): Rectify scaling ASGs

jmelchio:
Now CF uses the Capacity struct in clouddriver to handle the case of
scaling-down ASGs to 0 instances, so CF's implementation in deck
must also use it.

spinnaker/spinnaker#3670


jmelchio:
This pull request must be merged at the same time as: 
https://github.com/spinnaker/clouddriver/pull/3193

PR #6108, opened 2018-11-29
fix(amazon/loadBalancer) Fixed load balancer with multiple accounts

ncknt:
* Fix "component is changing an uncontrolled" on LB name
* Fix subnet without "purpose" showing blank in dropdown
* Fix cannot go back to EC2 classic
Should fix https://github.com/spinnaker/spinnaker/issues/3242

### DIFF
--- a/app/scripts/modules/amazon/src/subnet/subnetSelectField.component.ts
+++ b/app/scripts/modules/amazon/src/subnet/subnetSelectField.component.ts
@@ -53,7 +53,7 @@ class SubnetSelectFieldController implements IController {
     this.activeSubnets = subnets.filter(s => !s.deprecated);
     this.deprecatedSubnets = subnets.filter(s => s.deprecated);
     if (subnets.length) {
-      if (!this.component[this.field] && !this.readOnly) {
+      if (this.component[this.field] === null && !this.readOnly) {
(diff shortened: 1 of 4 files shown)


app/scripts/modules/amazon/src/subnet/subnetSelectField.component.ts
```diff
@@ -53,7 +53,7 @@ class SubnetSelectFieldController implements IController {
     this.activeSubnets = subnets.filter(s => !s.deprecated);
     this.deprecatedSubnets = subnets.filter(s => s.deprecated);
     if (subnets.length) {
-      if (!this.component[this.field] && !this.readOnly) {
+      if (this.component[this.field] === null && !this.readOnly) {
```
jrsquared: What was this change fixing?

ncknt: The subnetType is initially null and is changed to "" (empty string) for EC2 classic. Without this, `this.change()` is called which in turns recomputes the subnets (loop).

I should say that I don't think this directly fixes 3242 as I wrote in the summary. It's already more or less fixed but I'm preparing fixes for release branches as well where that issue still exist.

PR #6113, opened 2018-11-30
fix(core/pipeline): Fix grouped stages rendering in execution graph

jrsquared:
I remember why i used `<g>` now... 

FYI @christopherthielen @erikmunson


erikmunson:
terrific

PR #6118, opened 2018-11-30
refactor(amazon/image): Convert amazon image reader to a TS class

christopherthielen:


app/scripts/modules/amazon/src/image/image.reader.ts
```diff
@@ -0,0 +1,49 @@
...
+  imageName: string;
+  tags: {
+    [tag: string]: string;
+  };
+  tagsByImageId: {
+    [imageId: string]: IAmazonImage['tags'];
+  };
+}
+
+export class AwsImageReader {
```
jrsquared: Looks like this doesn't need angular at all... how about just make it a static non-angular class while you're here?

christopherthielen: fine

christopherthielen: nope. this is still needed by aws module 😢 

https://github.com/spinnaker/deck/blob/e1a33930a6b522f9349c4ccbe1abc4aa93bb6e1f/app/scripts/modules/amazon/src/aws.module.ts#L82-L92

christopherthielen: De-angularized this in a new commit.

PR #6120, opened 2018-11-30
fix(kubernetes): clicking execution details link filters clusters screen

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/3443

Before, clicking a Details link in an execution showed the clusters screen with all clusters shown.

screen shot 2018-11-30 at 3 56 09 pm

After, clicking a Details link in an execution shows the clusters screen filtered to the namespace that the resource you clicked on belongs.

screen shot 2018-11-30 at 3 55 14 pm


lwander:
Can we tag this as a fix? It's not really a feature and can go into 1.11 this way

sbwsg:

PR #6121, opened 2018-11-30
fix(provider/cf): rectify font sizes on dropdowns

jmelchio:
- spinnaker/spinnaker#3673

This change merely affects the size of the components; it does not make
all of the select boxes in Cloud Foundry's forms use the same React widget.


jmelchio:
Based on discussions with core team we withdraw the PR for the time being.

PR #6127, opened 2018-12-03
chore: add tests for yaml editor utils

maggieneterval:
- Convert yaml string-to-doc and doc-to-string logic to testable utils
- Add unit tests


lwander:
Uh oh I think spinbot is broken

maggieneterval:
Uh oh, well good catch larsbot

PR #6135, opened 2018-12-04
fix(provider/ecs): Add memory limit to server group details

clareliguori:
Suppress container port, memory reservation, and hard memory limit if not set in the task definition

Depends on https://github.com/spinnaker/clouddriver/pull/3203

screenshot


clareliguori:
spinnaker/clouddriver#3203 is now merged in

PR #6148, opened 2018-12-06
fix(imports): Avoid "import { thing } from 'core'"

christopherthielen:


sbwsg:
lgtm. Is this purely a style thing or does it a solve a problem?

christopherthielen:
This solves a problem with module resolution across modules

erikmunson:
excellent 👍

PR #6150, opened 2018-12-06
 refactor(core/account): Refactor AccountSelectInput to use 'value' prop

christopherthielen:


christopherthielen:
@Jammy-Louie I've refactored the `AccountSelectField` into an `Input`.  I also extracted an `Input` from the `RegionSelectField`.

Both these `Input` components should be usable from a `<FormikFormField input={} />`

PR #6151, opened 2018-12-06
feat(core/pipelines): Add an apply entity tags stage

jrsquared:
The stage already exists, just adding a simple UI for it. Could definitely use some love, but it's functional for now.

screen shot 2018-12-06 at 1 17 02 pm

@gcomstock FYI

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsExecutionDetails.tsx
@@ -1,0 +1,72 @@
+import * as React from 'react';
+
+import {
+  IExecutionDetailsSectionProps,
+  ExecutionDetailsSection,
+  StageExecutionLogs,
+  StageFailureMessage,
+} from 'core/pipeline';
+export function ApplyEntityTagsExecutionDetails(props: IExecutionDetailsSectionProps) {
+  const { stage, current } = props;
+
+  const entityRef = stage.context.entityRef || {};
+  const tags = stage.context.tags || {};
+
+  const entityRefSection = (
+    <div className="row">
+      <div className="col-md-12">
+        Entity Reference
+        <dl className="dl-narrow dl-horizontal">
+          {entityRef.cloudProvider && <dt>Provider</dt>}
+          {entityRef.cloudProvider && <dd>{entityRef.cloudProvider}</dd>}
+          {entityRef.entityType && <dt>Entity Type</dt>}
+          {entityRef.entityType && <dd>{entityRef.entityType}</dd>}
+          {entityRef.entityId && <dt>Entity ID</dt>}
+          {entityRef.entityId && <dd>{entityRef.entityId}</dd>}
+          {entityRef.account && <dt>Account</dt>}
+          {entityRef.account && <dd>{entityRef.account}</dd>}
+          {entityRef.region && <dt>Region</dt>}
+          {entityRef.region && <dd>{entityRef.region}</dd>}
+          {entityRef.vpcId && <dt>VPC ID</dt>}
+          {entityRef.vpcId && <dd>{entityRef.vpcId}</dd>}
+        </dl>
+      </div>
+    </div>
+  );
+
+  const tagsSection = (
+    <div className="row">
+      <div className="col-md-12">
+        Tags
+        <dl className="dl-narrow dl-horizontal">
+          {tags.map(({ name, value }: any) => {
+            if (typeof value === 'object') {
+              try {
+                value = JSON.stringify(value);
+              } catch (e) {}
+            }
+            return (
+              <React.Fragment key={name}>
+                <dt title={name}>{name}</dt>
+                <dd title={value}>{value}</dd>
+              </React.Fragment>
+            );
+          })}
+        </dl>
+      </div>
+    </div>
+  );
+
+  return (
+    <ExecutionDetailsSection name={props.name} current={current}>
+      {entityRefSection}
+      {tagsSection}
+      <StageFailureMessage stage={stage} message={stage.failureMessage} />
+      <StageExecutionLogs stage={stage} />
+    </ExecutionDetailsSection>
+  );
+}
+
+export namespace ApplyEntityTagsExecutionDetails {
+  export const title = 'applyEntityTagsConfig';
+}
--- a/app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsExecutionDetails.tsx
@@ -7,7 +7,7 @@ import {
   StageFailureMessage,
 } from 'core/pipeline';
 export function ApplyEntityTagsExecutionDetails(props: IExecutionDetailsSectionProps) {
-  const { stage, current } = props;
+  const { current, name, stage } = props;
 
   const entityRef = stage.context.entityRef || {};
   const tags = stage.context.tags || {};
@@ -39,15 +39,16 @@ export function ApplyEntityTagsExecutionDetails(props: IExecutionDetailsSectionP
       <div className="col-md-12">
         Tags
         <dl className="dl-narrow dl-horizontal">
-          {tags.map(({ name, value }: any) => {
+          {tags.map(({ name: tagName, value }: any) => {
             if (typeof value === 'object') {
               try {
                 value = JSON.stringify(value);
               } catch (e) {}
             }
+
             return (
-              <React.Fragment key={name}>
-                <dt title={name}>{name}</dt>
+              <React.Fragment key={tagName}>
+                <dt title={tagName}>{tagName}</dt>
                 <dd title={value}>{value}</dd>
               </React.Fragment>
             );
@@ -58,7 +59,7 @@ export function ApplyEntityTagsExecutionDetails(props: IExecutionDetailsSectionP
   );
 
   return (
-    <ExecutionDetailsSection name={props.name} current={current}>
+    <ExecutionDetailsSection name={name} current={current}>
       {entityRefSection}
       {tagsSection}
       <StageFailureMessage stage={stage} message={stage.failureMessage} />
--- a/app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
@@ -1,0 +1,135 @@
+import * as React from 'react';
+import Select, { Option } from 'react-select';
+
+import { IEntityRef, IEntityTag, IStage } from 'core/domain';
+import { IStageConfigProps, StageConfigField } from 'core/pipeline';
+import { SETTINGS } from 'core/config/settings';
+
+import { TagEditor } from './TagEditor';
+import { FormField, ReactSelectInput, StringsAsOptions } from 'core/presentation';
+
+export interface IApplyEntityTagsStage extends IStage {
+  entityRef: IEntityRef;
+  tags: IEntityTag[];
+}
+
+export interface IApplyEntityTagsStageConfigProps extends IStageConfigProps {
+  stage: IApplyEntityTagsStage;
+}
+
+export class ApplyEntityTagsStageConfig extends React.Component<IApplyEntityTagsStageConfigProps> {
+  private entityOptions = [
+    { label: 'Server Group', value: 'servergroup' },
+    { label: 'Load Balancer', value: 'loadbalancer' },
+    { label: 'Application', value: 'application' },
+    { label: 'Cluster', value: 'cluster' },
+    { label: 'Security Group', value: 'securitygroup' },
+  ];
+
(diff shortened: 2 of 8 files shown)


app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsExecutionDetails.tsx
```diff
@@ -0,0 +1,72 @@
...
+              </React.Fragment>
+            );
+          })}
+        </dl>
+      </div>
+    </div>
+  );
+
+  return (
+    <ExecutionDetailsSection name={props.name} current={current}>
```
christopherthielen: might as well destructure `name` too

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
```diff
@@ -0,0 +1,135 @@
+import * as React from 'react';
+import Select, { Option } from 'react-select';
+
+import { IEntityRef, IEntityTag, IStage } from 'core/domain';
+import { IStageConfigProps, StageConfigField } from 'core/pipeline';
+import { SETTINGS } from 'core/config/settings';
```
christopherthielen: can this be `core/config`?

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+import { TagEditor } from './TagEditor';
```
christopherthielen: move down 👇

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
```diff
@@ -0,0 +1,135 @@
...
+    { label: 'Security Group', value: 'securitygroup' },
+  ];
+
+  private entityRefTypeChanged = (entityType: string) => {
+    const entityRef = { entityType };
+    this.props.updateStageField({ entityRef });
+  };
+
+  private entityRefFieldChanged = (key: string, value: string) => {
+    const entityRef = { ...this.props.stage.entityRef, ...{ [key]: value } };
```
christopherthielen: can this?
```suggestion
    const entityRef = { ...this.props.stage.entityRef, [key]: value };
```

christopherthielen: why does this callback spread the current entityRef attributes, but `entityRefTypeChanged` does not?

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  private tagChanged = (tag: IEntityTag, index: number) => {
+    const tags = this.props.stage.tags.slice();
+    tags[index] = tag;
+    this.props.updateStageField({ tags });
+  };
+
+  private addTag = () => {
+    this.props.stage.tags.push({ name: '', value: '' });
+    this.props.updateStageField({ tags: this.props.stage.tags });
```
christopherthielen: To avoid  array mutation?
```suggestion
    this.props.updateStageField({ tags: this.props.stage.tags.concat({ name: '', value: '' }) });
```

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
```diff
@@ -0,0 +1,135 @@
...
+  };
+
+  public render() {
+    const {
+      stage: { entityRef = {} as IEntityRef, tags = [] as IEntityTag[] },
+      application: {
+        attributes: { cloudProviders = SETTINGS.defaultProviders },
+      },
+    } = this.props;
```
christopherthielen: still not sure how i feel about these fancy destructs 🤔

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
```diff
@@ -0,0 +1,135 @@
...
+                value={entityRef.entityId || ''}
+                onChange={e => this.entityRefFieldChanged('entityId', e.target.value)}
+              />
+            </StageConfigField>
+            {entityRef.entityType !== 'application' && (
+              <>
+                <StageConfigField label="Cloud Provider">
+                  <FormField
+                    input={fieldProps => (
+                      <StringsAsOptions strings={cloudProviders}>
```
christopherthielen: ReactSelectInput can take `stringOptions` array

https://github.com/spinnaker/deck/blob/49ec7716c0130264648e3a5c6e4664d79e1853c1/app/scripts/modules/core/src/presentation/forms/inputs/ReactSelectInput.tsx#L10-L13

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
```diff
@@ -0,0 +1,135 @@
...
+              <input
+                className="form-control input-sm"
+                value={entityRef.entityId || ''}
+                onChange={e => this.entityRefFieldChanged('entityId', e.target.value)}
+              />
+            </StageConfigField>
+            {entityRef.entityType !== 'application' && (
+              <>
+                <StageConfigField label="Cloud Provider">
+                  <FormField
```
christopherthielen: curious why you used `FormField` here but nowhere else?

app/scripts/modules/core/src/pipeline/config/stages/entityTags/TagEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    const valueIsObject = typeof props.tag.value === 'object';
+
+    this.state = {
+      type,
+      valueIsObject,
+    };
+  }
+
+  public componentWillReceiveProps(nextProps: ITagEditorProps) {
```
christopherthielen: suggest combining this with the code in the constructor and moving it all into `componentDidUpdate`

app/scripts/modules/core/src/pipeline/config/stages/entityTags/TagEditor.tsx
```diff
@@ -0,0 +1,169 @@
...
+              className="form-control input-sm"
+              value={tag.name || ''}
+              onChange={event => this.tagValueChanged('name', event.target.value)}
+              type="text"
+            />
+          </div>
+        </div>
+      ) : null;
+
+    const valueInput =
```
christopherthielen: I find these ternaries a bit hard to read.  Maybe `renderFoo()` methods would read better?

gcomstock:
This design looks like the right level of effort until we get the new form fields in. Thanks for running it by me. Let's just make sure when a user selects to add another tag, that there's enough white space between each group to show its for a new tag. Merrrrci

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
jrsquared: Nooooooo....

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
jrsquared: I only used it to use ReactSelectInput. I guess it makes sense to use it for "all the things"?

app/scripts/modules/core/src/pipeline/config/stages/entityTags/TagEditor.tsx
jrsquared: 👍

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsExecutionDetails.tsx
```diff
@@ -0,0 +1,72 @@
...
+          {tags.map(({ name, value }: any) => {
+            if (typeof value === 'object') {
+              try {
+                value = JSON.stringify(value);
+              } catch (e) {}
+            }
+            return (
+              <React.Fragment key={name}>
+                <dt title={name}>{name}</dt>
+                <dd title={value}>{value}</dd>
```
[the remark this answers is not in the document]
jrsquared: Probably blows up.

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsExecutionDetails.tsx
jrsquared: true

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
jrsquared: Apparently.

app/scripts/modules/core/src/pipeline/config/stages/entityTags/TagEditor.tsx
jrsquared: Happy to change it, but not sure what you find hard to read or what exactly you mean by a `renderFoo()` method...

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
jrsquared: when the type changes, we clear the values since switching from `custom` to the others have very different values.

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
jrsquared: 👍

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
jrsquared: Trying it out, blame @alanmquach if you hate them.

erikmunson: ❤️been doing this for a while and I don't regret it

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsExecutionDetails.tsx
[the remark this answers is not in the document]
christopherthielen: I suppose if there's an object reference loop?

```
            if (typeof value === 'object') {
              try {
                value = JSON.stringify(value);
              } catch (e) {}
            }
```

app/scripts/modules/core/src/pipeline/config/stages/entityTags/ApplyEntityTagsStageConfig.tsx
christopherthielen: ok then I'll get over it

app/scripts/modules/core/src/pipeline/config/stages/entityTags/TagEditor.tsx
christopherthielen: I guess I meant something like this:

```
private renderValueInput(value: any)
      if (type === 'custom') {
        return (
          <div className="row">
          <label className="col-md-3 sm-label-right">
            Value <HelpField id="pipeline.config.entitytags.value" />
          </label>
          <div className="col-md-8">
            <FormField
              input={TextInput}
              value={value || ''}
              onChange={event => this.tagValueChanged('value', event.target.value)}
            />
          </div>
        </div>
      ) 

      return (
        <div className="row">
          <label className="col-md-3 sm-label-right">Message</label>
          <div className="col-md-8">
            <FormField
              input={TextInput}
              value={tag.value.message || ''}
              onChange={event => this.tagValueChanged('value.message', event.target.value)}
            />
          </div>
        </div>
      );
}
```

but tbh I'm not sure how much better it is.  I think the ternaries are slightly harder to scan

erikmunson: I've found that most ternaries where the JSX blocks are > ~4 lines get tough to scan if you don't already know the structure. Also makes it easier to throw more logic in later without needing to contort yourself into a bunch of extra ternaries. That being said it's a minor annoyance, don't twist yourself in knots.

jrsquared: since no strong opinions, leaving as is for now.

PR #6152, opened 2018-12-07
feat(runJob): Adding support for preconfigured run job stages

alanmquach:
Mostly driven by server data. Titus execution logs will need to be placed at `stage.context.execution.logs` for the Execution Details to properly link out to it though.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/core/IStageConfigProps.ts
+++ b/app/scripts/modules/core/src/pipeline/config/stages/core/IStageConfigProps.ts
@@ -4,7 +4,7 @@ import { IStage } from 'core/domain';
 export interface IStageConfigProps {
   application: Application;
   stage: IStage;
-  configuration: any;
+  configuration?: any;
   stageFieldUpdated: () => void;
   updateStageField: (changes: { [key: string]: any }) => void;
 }
--- a/app/scripts/modules/core/src/pipeline/config/stages/core/IStageConfigProps.ts
+++ b/app/scripts/modules/core/src/pipeline/config/stages/core/IStageConfigProps.ts
@@ -4,6 +4,7 @@ import { IStage } from 'core/domain';
 export interface IStageConfigProps {
   application: Application;
   stage: IStage;
+  configuration: any;
   stageFieldUpdated: () => void;
   updateStageField: (changes: { [key: string]: any }) => void;
 }
--- a/app/scripts/modules/core/src/pipeline/config/stages/preconfiguredJob/preconfiguredJobStage.module.ts
+++ b/app/scripts/modules/core/src/pipeline/config/stages/preconfiguredJob/preconfiguredJobStage.module.ts
@@ -1,8 +1,8 @@
 import { module } from 'angular';
 
-import { PRECONFIGUREDJOB_STAGE } from './preconfiguredJobStage';
-import { STAGE_CORE_MODULE } from '../core/stage.core.module';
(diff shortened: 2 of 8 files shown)


app/scripts/modules/core/src/pipeline/config/stages/preconfiguredJob/preconfiguredJobStage.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export const PRECONFIGUREDJOB_STAGE = 'spinnaker.core.pipeline.stage.preconfiguredJobStage';
+
+module(PRECONFIGUREDJOB_STAGE, []).run(() => {
+  API.one('jobs')
+    .all('preconfigured')
+    .getList()
+    .then((preconfiguredJobs: IPreconfiguredJob[]) => {
+      preconfiguredJobs.forEach((preconfiguredJob: IPreconfiguredJob) => {
```
jrsquared: pretty sure TS already knows this is an `IPreconfiguredJob`

app/scripts/modules/core/src/pipeline/config/stages/core/IStageConfigProps.ts
```diff
@@ -4,6 +4,7 @@ import { IStage } from 'core/domain';
 export interface IStageConfigProps {
   application: Application;
   stage: IStage;
+  configuration: any;
```
jrsquared: I don't think this needs to be here... just in `PreconfiguredJobStageConfig.tsx` something like `export interface IPreconfiguredJobStageConfigProps extends IStageConfigProps { configuration: waitForCompletion: ?, parameters: ? }`?

app/scripts/modules/core/src/pipeline/config/stages/preconfiguredJob/preconfiguredJobStage.ts
alanmquach: You're right! I got a little type happy.

app/scripts/modules/core/src/pipeline/config/stages/core/IStageConfigProps.ts
alanmquach: Since it's passed in from [stage.module.js](https://github.com/spinnaker/deck/pull/6152/files#diff-f0f41e28f10844cbe8077af8dfc669edR203) and a few other stages have it, it belongs further up?

I think it should be optional though, so I'll change that.

jrsquared: ah, didn't realize a few other stages have it since it was just added here. carry on.

app/scripts/modules/core/src/pipeline/config/stages/preconfiguredJob/preconfiguredJobStage.module.ts
```diff
@@ -0,0 +1,13 @@
+import { module } from 'angular';
+
+import { PRECONFIGUREDJOB_STAGE } from './preconfiguredJobStage';
+import { STAGE_CORE_MODULE } from '../core/stage.core.module';
+import { TIME_FORMATTERS } from 'core/utils/timeFormatters';
```
christopherthielen: nitpick move `core` import up ☝️

alanmquach: Thanks @jrsquared

PR #6171, opened 2018-12-11
fix(formik): Fixed validation, AccountSelectInput and RegionSelectInput props

alanmquach:
1. Gnarly corner case where validation doesn't render for falsy values (i.e. `0`) because it was thought to be not touched

2. The `<select>` for `AccountSelectInput` and `RegionSelectInput` needs `name` so just passing all the props.


christopherthielen:
closing/reopening to trigger CI

PR #6193, opened 2018-12-13
fix(core): add missing parenthesis

vinhlh:
Looks like we missed a `)` in https://github.com/spinnaker/deck/pull/6177/commits


erikmunson:
lgtm, thanks for catching 🙏

PR #6204, opened 2018-12-14
feat(tests): simple static server for functional tests

sbwsg:
Introduce a very simple express server to allow automation to exercise the production build of Deck without also installing and configuring an apache / nginx server.


christopherthielen:
I think we can use `webpack-dev-server --mode=production` and avoid pulling in all of express @sbwsg

PR #6206, opened 2018-12-14
fix(quiet): Add caveat that quietPeriod does not affect pipeline triggers

alanmquach:


app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
```diff
@@ -33,7 +33,7 @@
                     <div class="checkbox">
                       <label>
                         <input type="checkbox" ng-model="pipeline.respectQuietPeriod">
-                        Disable automatic triggers during quiet period. <help-field key="pipeline.config.triggers.respectQuietPeriod"></help-field>
+                        Disable automated triggers during quiet period (<strong>does not affect Pipeline triggers</strong>). <help-field key="pipeline.config.triggers.respectQuietPeriod"></help-field>
```
erikmunson: worth noting that the implementation of quiet periods involves whitelisting a custom set of triggers to disable, so this may be somewhat misleading depending on how the backend is configured: https://github.com/spinnaker/echo/pull/377

PR #6207, opened 2018-12-14
fix(tests): remove express from static server test runner class

sbwsg:
Per https://github.com/spinnaker/deck/pull/6204#issuecomment-447405054


test/functional/tools/StaticServer.ts
```diff
@@ -1,33 +1,40 @@
...
+        },
+        (err, _stdout, _stderr) => {
+          if (err != null) {
+            reject(err);
+          }
+        },
+      );
+      this.proc.stdout.on('data', (data: Buffer) => {
+        const str = String(data);
+        if (str.match(/compiled/i)) {
```
sbwsg: Not convinced this is sound but I couldn't think of another way of detecting when the dev server was ready to serve up the build.

christopherthielen: I've used `wait-on` npm package to wait for 200 responses.  

https://github.com/jeffbski/wait-on

see https://github.com/ui-router/cypress-runner/blob/05116d8b5782025a2ddf1f6c4b6423235a20e9af/cypress-runner.js#L73-L84

PR #6210, opened 2018-12-14
feat(core/presentation): remove hideSections prop from wizardModal api

christopherthielen:
@jrsquared This is simpler.  I don't know if there will be any unexpected side effects due to the hidden components not being registered in the wizard, but it doesn't look like it.


christopherthielen:
>  I think the security groups section in the load balancer wizard?
Glad you remembered this because thus change does indeed break security groups.

christopherthielen:
This is in #6348

PR #6225, opened 2018-12-18
feat(core): add ability to search clusters by labels

maggieneterval:
Adding label search functionality as a separate PR from label filter functionality as the latter is a bit messier (#3238)

a0bd8221-3a10-4d85-a1b1-49c2da51989d

### DIFF
--- a/app/scripts/modules/core/src/cluster/filter/ClusterFilterService.ts
+++ b/app/scripts/modules/core/src/cluster/filter/ClusterFilterService.ts
@@ -1,4 +1,4 @@
-import { each, forOwn, groupBy, sortBy } from 'lodash';
+import { each, every, forOwn, groupBy, some, sortBy } from 'lodash';
 import { Debounce } from 'lodash-decorators';
 import { $log } from 'ngimport';
 import { Subject } from 'rxjs';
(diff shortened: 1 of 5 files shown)


lwander:
The demo is curious, why does applying a more restrictive filter (more labels) result in more results? Are the labels being OR'd instead of AND'd?

maggieneterval:
@lwander Yeah I tried to keep it consistent with how `clusters` search works, which is an OR, but if you think AND makes more sense for labels I can switch it and just clarify that in the help box. (https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/cluster/filter/ClusterFilterService.spec.ts#L135)

lwander:
Ah that's difficult... I'd prefer to remain consistent with what people expect coming from kubernetes, especially since this label query will help them build expressions to use either in pipelines or manifests when selecting groups of resources in k8s, via spinnaker.

I wonder if we can guide users towards this feature by expanding it from being a custom search query to something with a little more treatment in a future PR

maggieneterval:
Makes sense, updated search to match on AND.
Incoming PR for a hopefully more intuitive UI like this:

e1rapwuvaot

However, sounds like we don't need to have multiple acceptable values for a given label key so I'll change that multi-select to a single select.

lwander:
That looks awesome

app/scripts/modules/core/src/cluster/filter/ClusterFilterService.ts
```diff
@@ -334,6 +334,25 @@ export class ClusterFilterService {
...
+        let labelKey: string;
+        let labelValue: string;
+        if (label.includes('=')) {
+          [labelKey, labelValue] = label.split('=');
+        }
+        if (!labelKey || !labelValue) {
+          return false;
+        }
+        return some(serverGroup.labels || {}, (val: string, key: string) => {
+          return labelKey.toLowerCase() === key.toLowerCase() && val.toLowerCase().includes(labelValue.toLowerCase());
```
lwander: I think these are case-sensitive... do the k8s docs say otherwise?

maggieneterval: Ah, another case of inconsistency with existing search filters vs. inconsistency with k8s patterns -- I'll remove case insensitivity + partial value match acceptance

PR #6234, opened 2018-12-19
feat(notifications): Adds pubsub notification module. Alphabetize the…

ttomsu:
… notification list. Adds all notification types to `settings.js` file(s).

For the notifications in `settings.js` - I marked them all as enabled because the [notificationTypeConfig.provider.js](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/notification/notificationTypeConfig.provider.js#L13) file enables them by default if no config is present. I think this is wrong, and would happily change them all to disabled because the user shouldn't be allowed to choose a notification method that is not configured.

WDYT @anotherchrisberry @maggieneterval @sbwsg @lwander ?


halconfig/settings.js
```diff
@@ -170,17 +170,32 @@ window.spinnakerSettings = {
...
+      enabled: true,
+    },
     hipchat: {
       botName: 'Skynet T-800',
       enabled: true,
     },
+    pubsub: {
+      enabled: true,
+    },
+    slack: {
```
sbwsg: It seems like we're removing configurability here, is that intentional?

The `slack` object that used to be assigned to this field includes template literals that I assume halyard populates?

If it's intentional then let's also remove the slack object defined on line 30.

ttomsu: Good catch - this is a copy/paste error.

ttomsu: Fixed.

PR #6241, opened 2018-12-19
feat(artifacts): add ivy and maven expected artifact editors

jkschneider:
image

In many cases artifact repository UI's surface coordinates as part of text helping you inject them into builds. Splitting the coordinate into 3 independent input text boxes would be annoying to the user:

image


jkschneider:
> Looks like the .svg files are missing.

Yep, got them added. Must have had a gitignore on them somewhere.

sbwsg:
Code lgtm. Where are the logos from & what terms are they released under?

jkschneider:
> Code lgtm. Where are the logos from & what terms are they released under?

Both Ivy and Maven are Apache projects, and logos are taken from their project sites.

PR #6249, opened 2018-12-20
feat(core): add label filter UI to clusters view

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/3238:

6a78d305-cc3e-4e77-ab7c-1103d4d6459e


anotherchrisberry:
Which cloud providers support labels?

If it's only Kubernetes, I'd suggest not showing this unless an app is configured with Kubernetes as a cloud provider - or maybe hide it if there are no labels in any of the server groups, if that's easier.

maggieneterval:
@anotherchrisberry Good idea, just updated to hide the Labels section if none of the server groups have labels. [This change](https://github.com/spinnaker/clouddriver/pull/3070/files) allows all infrastructure objects regardless of provider to have labels but I think only Kubernetes currently uses them.

duftler:
GCE server groups have labels:
image

maggieneterval:
Oh nice, thanks @duftler!

PR #6256, opened 2018-12-20
fix(bake): Allow null extended attributes in bake stages

ezimanyi:
Many optional paramters are defaulted by various config files during bake stages; it is useful to be able to explicitly clear these values using extended attributes. In order to allow this, allow the value of an extended attribute to be empty.


ezimanyi:
Addresses the suggestion in [this comment](https://github.com/spinnaker/spinnaker/issues/3764#issuecomment-449127497).

I don't plan on cherry-picking this into 1.11.

duftler:
Defer to @maggieneterval or @sbwsg on merging.

PR #6257, opened 2018-12-20
feat(core): Extend app name to be modified in custom builds of deck

archana-s:
Ability to extend Application Name displayed in the navigation

Allowing Application Name component to be extended will help with adding a custom component at Netflix. The custom component can display all apps that belong to the team, link to teams page; almost similar to applications displayed within projects page. The benefit of having this directly on the application name itself is to ensure it is visible, discoverable and easy to hop between apps one owns.


archana-s:
Will re-open this after design discussions with Netflix spinnaker UI team.

PR #6306, opened 2019-01-04
fix(artifacts): Fix null artifact kind

ezimanyi:
Some workflows can lead to the expected artifact widget having a null artifact kind.  Examples are pipelines that were created before 'kind' was persisted to the config JSON, and pipeline templates that don't explicitly set 'kind'.  Explicitly set the artifact kind when loading the widget to fix these issues.


ezimanyi:
Addresses one of the issues in https://github.com/spinnaker/spinnaker/issues/2654

ezimanyi:
@lwander : Yes, spinnaker/spinnaker#3740 is exactly the issue I'm trying to solve here.  I hadn't seen that issue, I just came across the bug myself...but that issue is exactly what I observed.

@sbwsg : It does cause the "Save" button to be highlighted...good catch.

Given that it probably makes sense to reduce reliance on the 'kind' field, and that this causes the save button to highlight, let me take a look to see if there's an easy way to make the component render properly in the absence of a kind instead of having it populate the kind.

ezimanyi:
Replaced with:
https://github.com/spinnaker/deck/pull/6351
https://github.com/spinnaker/deck/pull/6357
https://github.com/spinnaker/deck/pull/6359

PR #6307, opened 2019-01-05
fix(dcos): Catch null argument in new servergroup command

jasonkshar:
When upgrading to 1.10, noticed that when clicking the "Create Server Group" button on the CLUSTERS page for an application and choosing DCOS as the provider, nothing would happen. The web console showed: `TypeError: "defaults is null"`

After some digging, I think it's coming down to the [allClusters.controller.js](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/cluster/allClusters.controller.js#L108), which goes to the [serverGroupCommandBuilder.service.ts](https://github.com/spinnaker/deck/blob/655f7311deb5e00d648a63ca2036f78eba01e612/app/scripts/modules/core/src/serverGroup/configure/common/serverGroupCommandBuilder.service.ts#L145), which goes to the [CommandBuilder.js](https://github.com/spinnaker/deck/blob/655f7311deb5e00d648a63ca2036f78eba01e612/app/scripts/modules/dcos/serverGroup/configure/CommandBuilder.js#L126) for DCOS, and then you get a nice defaults is null there.

Admittedly, this change does work for me, but I have to admit I'm not terribly familiar with Deck code. I noticed that a code search through the repo shows that there are serverGroupCommandBuilder classes for the various providers (kube, google, etc) but not one for DCOS which is probably why it falls into that core one, I think. Point being, should there be an implementation of that class for DCOS as well?


jasonkshar:
Hmm, well actually, I should've left that default parameter.

app/scripts/modules/dcos/serverGroup/configure/CommandBuilder.js
```diff
@@ -124,6 +124,8 @@ module.exports = angular
     }
 
     function buildNewServerGroupCommand(application, defaults = {}) {
+      defaults = defaults || {};
+
```
jrsquared: this shouldn't actually be doing anything, since the function signature already sets it to `{}` if it's null?

jasonkshar: hmmm idk, somehow defaults was getting to be null without it. I admit, I know very little about javascript.

christopherthielen: If you explicitly pass a falsey value, then the default parameter isn't applied.  i.e.:
```
buildNewServerGroupCommand(application, null);
```
I suggest removing the default parameter on line 126 and keep the reassignment on line 127.

app/scripts/modules/dcos/serverGroup/configure/CommandBuilder.js
```diff
@@ -124,6 +124,8 @@ module.exports = angular
     }
 
     function buildNewServerGroupCommand(application, defaults = {}) {
```
christopherthielen:
```suggestion
    function buildNewServerGroupCommand(application, defaults) {
```

app/scripts/modules/dcos/serverGroup/configure/CommandBuilder.js
jasonkshar: but by removing the parameter entirely, wouldn't that be a non-passive change? LIke, then you wouldn't have the option of not specifying the second parameter.

christopherthielen: I mean just removing the default value in the parameter declaration and leaving it as `buildNewServerGroupCommand(application, defaults)`.   In javascript you can omit arguments when calling a function. In TypeScript, the argument would have to be explicitly declared as optional, but this file is `.js`

PR #6316, opened 2019-01-07
refactor(amazon/loadBalancer): Simplify the "security group not found" code

christopherthielen:
screen shot 2019-01-07 at 1 48 29 pm

- Re-adds "missing security groups" when switching *back* to an account in which they do exist
- Use Formik validate() instead of dirtyCallback()
- Remove sorter fn (no longer needed with react-select multi-select)


christopherthielen:
This was in 
https://github.com/spinnaker/deck/pull/6318/commits already, sorry

PR #6317, opened 2019-01-07
feat(provider/gce): Moniker support for GCE server groups

jervi:
Makes Deck use moniker for GCE server groups. Depends on https://github.com/spinnaker/clouddriver/pull/3258

### DIFF
--- a/app/scripts/modules/core/src/cluster/cluster.service.ts
+++ b/app/scripts/modules/core/src/cluster/cluster.service.ts
@@ -273,6 +273,13 @@ export class ClusterService {
     const nameParts = NameUtils.parseServerGroupName(serverGroup.name);
     if (serverGroup.moniker) {
       Object.assign(serverGroup, serverGroup.moniker);
+      if (serverGroup.moniker.cluster == null && (serverGroup.moniker.stack || serverGroup.moniker.detail)) {
+        serverGroup.cluster = NameUtils.getClusterName(
+          serverGroup.moniker.app || nameParts.application,
+          serverGroup.stack,
+          serverGroup.detail,
+        );
+      }
     } else {
       serverGroup.app = nameParts.application;
       serverGroup.stack = nameParts.stack;
--- a/app/scripts/modules/core/src/cluster/cluster.service.ts
+++ b/app/scripts/modules/core/src/cluster/cluster.service.ts
@@ -273,13 +273,6 @@ export class ClusterService {
     const nameParts = NameUtils.parseServerGroupName(serverGroup.name);
     if (serverGroup.moniker) {
       Object.assign(serverGroup, serverGroup.moniker);
-      if (serverGroup.moniker.cluster == null && (serverGroup.moniker.stack || serverGroup.moniker.detail)) {
-        serverGroup.cluster = NameUtils.getClusterName(
-          serverGroup.moniker.app || nameParts.application,
-          serverGroup.stack,
-          serverGroup.detail,
-        );
-      }
     } else {
       serverGroup.app = nameParts.application;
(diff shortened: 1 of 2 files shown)


jervi:
@spinnaker/reviewers PTAL (clouddriver PR is merged)

app/scripts/modules/core/src/cluster/cluster.service.ts
```diff
@@ -273,6 +273,13 @@ export class ClusterService {
     const nameParts = NameUtils.parseServerGroupName(serverGroup.name);
     if (serverGroup.moniker) {
       Object.assign(serverGroup, serverGroup.moniker);
+      if (serverGroup.moniker.cluster == null && (serverGroup.moniker.stack || serverGroup.moniker.detail)) {
```
christopherthielen: What scenario is this testing for? A code comment might be useful here.

cloudbuild-tagged.yaml
```diff
@@ -1,14 +1,20 @@
...
-  entrypoint: "bash"
-  args: [ "-c", "./gradlew build -PskipTests"]
-- name: 'gcr.io/cloud-builders/docker'
-  args: ["build",
-         "-t", "gcr.io/$PROJECT_ID/$REPO_NAME:$COMMIT_SHA",
-         "-t", "gcr.io/$PROJECT_ID/$REPO_NAME:$TAG_NAME",
-         "-f", "Dockerfile.slim",
-         "."]
+  - name: 'openjdk:8-stretch'
+    env: ['GRADLE_USER_HOME=cache']
```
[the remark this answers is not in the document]
jervi: Yes, they happened automatically. I'll remove them from the PR 👍

app/scripts/modules/core/src/cluster/cluster.service.ts
jervi: I ended up putting the same code in the corresponding backend changes in Clouddriver, so it can be removed from here.

jervi: The rationale is explained here: https://github.com/spinnaker/clouddriver/pull/3258#discussion_r252848244

app/scripts/modules/google/src/serverGroup/configure/serverGroupCommandBuilder.service.js
```diff
@@ -379,14 +383,14 @@ module.exports = angular
 
       function buildServerGroupCommandFromExisting(application, serverGroup, mode) {
         mode = mode || 'clone';
-        const serverGroupName = NameUtils.parseServerGroupName(serverGroup.name);
+        const moniker = serverGroup.moniker;
```
jtk54: What is the fallback behavior here if the previous server group was deployed before moniker support? Is this all handled in the Clouddriver changes?

app/scripts/modules/google/src/serverGroup/configure/serverGroupCommandBuilder.service.js
```diff
@@ -274,6 +274,10 @@ module.exports = angular
           if (command.labels['spinnaker-server-group']) {
             delete command.labels['spinnaker-server-group'];
           }
+
+          if (command.labels['spinnaker-moniker-sequence']) {
```
jtk54: Can you drop a comment here explaining the expected behavior here? It's not too clear why it's important to clear this key.

app/scripts/modules/google/src/serverGroup/configure/serverGroupCommandBuilder.service.js
jervi: Yes, all server groups will have moniker set by Clouddriver.

PR #6344, opened 2019-01-10
feat(artifacts): Human readable expected artifact display names

jkschneider:
* Whenever the display name changes in the expected artifact configuration, it immediately changes everywhere else as well.
* Expected artifacts are still referred to in pipeline JSON by ID and not display name.
* Expected artifact summarization has been replaced by the display name.
* I toyed with the idea of generating display names like `maven1`...`mavenN` based on the artifact kind, but didn't see a good way to change the generated display name when the kind changed (differentiating a still-generated display name from one the user has chosen after the fact is not nice). The `human-readable-ids` package provides reasonably easy-to-remember names without users having to do anything unless they feel like changing the name.
* The backend needn't be aware of the display name for any specific purpose. It is stored because it is serialized from the front-end without any changes.

image

The expected artifact editor has been updated to include a display name as well which is also generated by default:

image


app/scripts/modules/core/src/pipeline/config/pipelineConfig.controller.js
```diff
@@ -18,6 +19,15 @@ module.exports = angular
 
     this.initialize = () => {
       this.pipelineConfig = _.find(app.pipelineConfigs.data, { id: $stateParams.pipelineId });
+
+      if (this.pipelineConfig) {
+        for (const artifact of this.pipelineConfig.expectedArtifacts) {
+          if (!artifact.displayName || artifact.displayName.length === 0) {
+            artifact.displayName = HumanReadableIds.random();
```
jkschneider: This is how we retrofit existing pipelines with display names.

jkschneider:
I think in a future iteration it would be nice to include an artifact account selector under the "Assign to a matched artifact" heading so that later stages don't need to select the account. At least this is my justification for placing `Display Name` under its own subsection right now. Thoughts welcome on this.

ezimanyi:
Looks good, though I have the same concern as @sbwsg .  I'd probably prefer if we left the current behavior in place, where display name is optional (and not auto-generated), and where the artifact summary falls back to the JSON when there's no display name.

jkschneider:
@sbwsg @ezimanyi Perhaps a compromise is to generate a name with higher information density in `pipelineConfig.controller.js` which is where we're retrofitting existing pipelines with artifact display names.

jkschneider:
I do strongly want to kill the JSON. I really think it looks bad. It's subjective though, I know.

sbwsg:
Agreed; generating a name based on the artifact data in that specific case seems like a good approach to me.

jkschneider:
@sbwsg Also will probably be nice going forward as the retrofitting isn't strictly to deal with "legacy" pipelines. I could see pipeline templating mechanisms generating expected artifact associations without display names too, but this situation is similar in that much of the artifact structure is more or less set by the time we go to generate a display name.

ezimanyi:
@lwander in case you have thoughts about this

lwander:
I think we can get pretty far generating a random name that includes at least a few details (like suggested by @sbwsg and @ezimanyi) about the artifact, and only do this when the human-readable name is empty and a pipeline is being saved. For example, if I have a docker image `gcr.io/project/image:tag`, the human readable name `docker-image-project/image` makes a lot of sense. Similar for a github file to path `/src/manifests/deployment.yml` -> `github-file-deployment.yml`.

app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
```diff
@@ -127,6 +127,9 @@ export class ExpectedArtifactEditor extends React.Component<
     const EditCmp = kind && kind.editCmp;
     return (
       <>
+        <StageConfigField label="Display Name" fieldColumns={fieldColumns} groupClassName={fieldGroupClassName}>
+          <input className="form-control input-sm" value={expectedArtifact.displayName} />
```
sbwsg: Is the intention for this field to be read-only?  I wouldn't expect that the user's edits to this input box would have any effect on the saved artifact.  React doesn't implement data binding the way that Angular might here.  I think you'd need to include an `onChange` handler to save modifications to the displayName in `this.state.expectedArtifact`.  Something like:

```jsx
// method on the class
public onDisplayNameEdit = (e: React.FormEvent<HTMLInputElement>) => {
  const newName = e.value;
  const expectedArtifact = { ...this.state.expectedArtifact };
  expectedArtifact.displayName = newName;
  this.setState({ expectedArtifact });
};

// in the render() method...
  <input className="..." value={expectedArtifact.displayName} onChange={this.onDisplayNameEdit} />
```

jkschneider: Good point. Added with 5d313f424.

jkschneider:
@lwander Updated to retrofit existing pipelines with display names that have higher information density based closely on your recommended scheme.

image

PR #6357, opened 2019-01-11
 refactor(artifacts): Add customKind flag to artifact config

ezimanyi:


app/scripts/modules/core/src/artifact/expectedArtifact.service.spec.ts
```diff
@@ -35,33 +35,34 @@ describe('ExpectedArtifactService', () => {
         isDefault: true,
       },
     ].map(k => ({ ...baseKindConfig, ...k }));
-    const defaultKindConfig = {
+    const customKindConfig = {
       ...baseKindConfig,
       key: 'custom',
+      isCustom: true,
```
sbwsg: Is there a mismatch here or is this property name intentional? `isCustom` here vs `customKind` elsewhere.

app/scripts/modules/core/src/artifact/expectedArtifact.service.spec.ts
```diff
@@ -87,15 +88,15 @@ describe('ExpectedArtifactService', () => {
...
     it('returns the default kind if neither kind nor type are stored on artifact when isDefault is true', () => {
       const artifact: IArtifact = {
         id: 'artifact-id',
       };
       const kindConfig = ExpectedArtifactService.getKindConfig(artifact, true);
-      expect(kindConfig).toEqual(defaultKindConfig);
+      expect(kindConfig).toEqual(customKindConfig);
     });
   });
 });
```
sbwsg: Would it be worth adding a test ensuring that an artifact with no `kind` field but a `customKind` flag w/ a known type correctly results in `getKindConfig` returning the `customKindConfig`? E.g. something like:

```js
it('returns the default kind if an artifact has a customKind flag, regardless of its type', () => {
  const artifact: IArtifact = {
    id: 'artifact-id',
    customKind: true,
    type: 'foo-type'
  };
  const kindConfig = ExpectedArtifactService.getKindConfig(artifact, false);
  expect(kindConfig).toEqual(customKindConfig);
});
```

app/scripts/modules/core/src/artifact/expectedArtifact.service.spec.ts
ezimanyi: Good catch, I renamed it and missed this test

app/scripts/modules/core/src/artifact/expectedArtifact.service.spec.ts
ezimanyi: Yes, good point; I'll add such a test

PR #6359, opened 2019-01-11
refactor(artifacts): Deprecate kind field

ezimanyi:
Replace all places where we write the kind field of artifacts with writing the customKind field. Delete artifact.kind any time we're updating customKind, as we want customKind alone to control whether to show the custom editing form once the artifact has this field.


ezimanyi:
Closes https://github.com/spinnaker/spinnaker/issues/3820

ezimanyi:
Closes https://github.com/spinnaker/spinnaker/issues/2654

PR #6360, opened 2019-01-11
fix(validation): Validator should pass checkParentTriggers through

alanmquach:
Created a new validator to check upstream for stages/triggers that provide version information for bakes (https://github.com/spinnaker/deck/pull/6328), but forgot to pass through the flag that also checks stages/triggers of parent pipeline triggers. Whoops!


emjburns:
🍾

PR #6361, opened 2019-01-12
fix(core/presentation): make detail dropdowns not visually crop

gcomstock:
detail view will be reconsidered soon as part of larger infrastructure view redesigns. For now, this just lets the dropdowns not get cropped. 

screen shot 2019-01-11 at 4 04 04 pm


erikmunson:
quality PR, would review again

PR #6363, opened 2019-01-14
fix(google): fix autohealing clone logic

maggieneterval:
When testing this fix, my stack was also named 'clone' so I somehow confused this with the correct check on viewState mode. :see_no_evil: 
Will patch into 1.10 and 1.11 as well.


maggieneterval:
@spinnakerbot cherry-pick 1.10

maggieneterval:

PR #6382, opened 2019-01-16
fix(google): differentiate among autohealing health check kinds

maggieneterval:
Previously, we specified autohealing health checks by name, which is not guaranteed to be a unique key because it is possible to have up to three health checks with the same name (one of each kind: `healthCheck`, `httpHealthCheck`, and `httpsHealthCheck`). This can lead to [this Angular error](https://code.angularjs.org/1.6.4/docs/error/ngRepeat/dupes?p0=healthCheck%20in%20$select.items&p1=string:routing-test-hc&p2=routing-test-hc), which prevents the autohealing health check dropdown from working.

This changes the display logic to key on the health check URL, which can be parsed to infer its name and kind. Additionally, health checks will now be displayed alphabetically, and duplicated names will now be displayed along with their kinds. Corresponding Clouddriver PR: https://github.com/spinnaker/clouddriver/pull/3282

h7qw2e5vtup

### DIFF
--- a/app/scripts/modules/google/src/healthCheck/healthCheckUtils.ts
+++ b/app/scripts/modules/google/src/healthCheck/healthCheckUtils.ts
@@ -1,0 +1,47 @@
+import { IGceHealthCheck, IGceHealthCheckKind } from 'google/domain';
+
+export interface IGceHealthCheckOption {
+  displayName: string;
+  kind: IGceHealthCheckKind;
+  name: string;
+  selfLink: string;
+}
+
+export function parseHealthCheckUrl(url: string): { healthCheckName: string; healthCheckKind: IGceHealthCheckKind } {
+  const healthCheckPathParts = url.split('/');
+  if (healthCheckPathParts.length < 2) {
+    throw new Error(`Health check url ${url} missing expected segments.`);
+  }
+  const healthCheckName = healthCheckPathParts[healthCheckPathParts.length - 1];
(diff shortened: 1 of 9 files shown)


app/scripts/modules/google/src/healthCheck/healthCheckUtils.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export interface IGceHealthCheckOption {
+  displayName: string;
+  kind: IGceHealthCheckKind;
+  name: string;
+  selfLink: string;
+}
+
+export function parseHealthCheckUrl(url: string): { healthCheckName: string; healthCheckKind: IGceHealthCheckKind } {
+  const healthCheckPathParts = url.split('/');
```
jtk54: May want to validate `healthCheckPathParts` here.

jtk54: Might want to validate the length of `healthCheckPathParts`.

jtk54:
Heh reviewed it twice, thought first submit didn't go through :+1:

app/scripts/modules/google/src/healthCheck/healthCheckUtils.ts
maggieneterval: :+1: Updated!

maggieneterval:

PR #6401, opened 2019-01-18
config(google): Update GCE's base URL to the new domain (spinnaker#3863)

CodingFanSteve:
GCE is rolling out a new domain "compute.googleapis.com/compute".
Update GCE's base URL in Spinnaker code to point to this new domain.
The old domain "www.googleapis.com/compute" will still be valid, so the old release will still be working fine.


ezimanyi:
Since this is primarily touching the new functional tests that @sbwsg wrote, makes sense for him to review this.

PR #6402, opened 2019-01-19
fix(amazon): Fixing load balancers isInternal flag

alanmquach:
Fixing `isInternal` derived by the subnet purpose.

Also it looks like the checkbox input wasn't wired up to the `Formik`. I wonder how many people actually checked the unchecked checkbox!

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/common/LoadBalancerLocation.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/common/LoadBalancerLocation.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import * as classNames from 'classnames';
 import { IPromise } from 'angular';
 import { chain, isNil, uniq, groupBy } from 'lodash';
-import { Field, FormikErrors } from 'formik';
+import { Field, FormikErrors, FieldProps } from 'formik';
 import { Observable, Subject } from 'rxjs';
 
 import {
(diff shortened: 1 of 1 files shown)


app/scripts/modules/amazon/src/loadBalancer/configure/common/LoadBalancerLocation.tsx
```diff
@@ -186,6 +186,9 @@ class LoadBalancerLocationImpl extends React.Component<ILoadBalancerLocationProp
     subnet$.takeUntil(this.destroy$).subscribe(subnet => {
       this.props.formik.setFieldValue('vpcId', subnet && subnet.vpcIds[0]);
       this.props.formik.setFieldValue('subnetType', subnet && subnet.purpose);
+      if (!this.state.hideInternalFlag && !this.state.internalFlagToggled && subnet && subnet.purpose) {
+        this.props.formik.setFieldValue('isInternal', subnet.purpose.includes('internal'));
```
christopherthielen: This logic seems wrong.  I think we should only perform this when `inferInternalFlagFromSubnet === true`

Also, nflx configuration is currently set to `false`:
```
     loadBalancers: {
        certificateTypes: ['iam'],
        inferInternalFlagFromSubnet: false,
        disableManualOidcDialog: true,
      }
```

https://github.com/spinnaker/deck/blob/286d4ba44ed9fd1afcae447995ba5194e0c62b6a/settings.js#L126-L128


Perhaps we should first find out why we haven't toggled this setting on.

alanmquach: This is the same as the previous behavior:
https://github.com/spinnaker/deck/blob/b6de8f8268b20865ee64751e999f9166c6312d8d/app/scripts/modules/amazon/src/loadBalancer/configure/common/LoadBalancerLocation.tsx#L190-L191

Though it's not clear to me whether it is fully `deck`'s responsibility to infer, or if it's done somewhere else in the backend. Though the latter would make sense with the way things are implemented:

1. `inferInternalFlagFromSubnet` is `true`, deck doesn't surface or send `isInternal`
2. `inferInternalFlagFromSubnet` is `false`, deck surfaces the option and still computes what it thinks may be what you want, but you have the option to toggle.

christopherthielen: Oh ok, thanks for clarifying

christopherthielen: Oh ok, thanks for clarifying

alanmquach: Added some clarifying comments as well.

PR #6405, opened 2019-01-22
feat(tests): add pipelines_list functional test for GCE

sbwsg:


test/functional/tests/google/pipelines_list.spec.ts.mountebank_fixture.json
```diff
@@ -0,0 +1,2683 @@
...
+        },
+        {
+          "caseSensitive": true,
+          "deepEquals": {
+            "query": {}
+          },
+          "keyCaseSensitive": true
+        }
+      ],
+      "responses": [
```
jtk54: Is this the recorded traffic from Gate? Is the longer term plan to just keep these in Deck?

sbwsg: Yeah that's the recorded traffic from Gate.  I think it makes sense to keep them alongside the tests.  The fixture data is going to be required for anyone who wants to reproduce the tests and who doesn't have the exact same cloud setup that a test's author does.

jtk54: Makes sense, I'm just curious how large these will be as we add tests across providers. I think we can cross that bridge when we get to it.

PR #6406, opened 2019-01-22
fix(amazon): Fixed similar image finder so concat happens

alanmquach:
`find` always succeeded at the first element so we never `concat`ed


app/scripts/modules/amazon/src/serverGroup/configure/AmazonImageSelectInput.tsx
```diff
@@ -98,7 +98,7 @@ export class AmazonImageSelectInput extends React.Component<IAmazonImageSelector
       }
 
       return this.searchForImages(this.buildQueryForSimilarImages(image.imageName)).then(similarImages => {
-        if (!similarImages.find(img => img.imageName !== image.imageName)) {
+        if (!similarImages.find(img => img.imageName === image.imageName)) {
```
christopherthielen: oh, good lord.

PR #6407, opened 2019-01-23
fix(webhooks): Various webhook stage improvements

marchello2000:
* Allow user to specify HTTP status codes that will fail fast
  (see corresponding [orca](https://github.com/spinnaker/orca/pull/2602) change
* Ensure that body and error are always shown


app/scripts/modules/core/src/pipeline/config/stages/webhook/webhookStage.ts
```diff
@@ -101,6 +109,12 @@ export class WebhookStage implements IController {
     this.stage.statusUrlResolution = this.viewState.statusUrlResolution;
   }
 
+  public failFastCodesChanged(): void {
+    const failFastCodes = this.viewState.failFastStatusCodes.split(/(?:,| |;)+/);
```
anotherchrisberry: What is that regex matching? If it's just a comma separated list we're after, it might read easier as:
```javascript
const failFastCodes = this.viewState.failFastStatusCodes.split(',').map(c => c.trim());
```
If we're trying to match on space or semicolon separated choices, we should update the help text above to indicate the flexibility in separators - but it's probably better to be consistent with the rest of the app and only allow comma-separated values here.

app/scripts/modules/core/src/help/help.contents.ts
```diff
@@ -420,6 +420,8 @@ const helpContents: { [key: string]: string } = {
   'pipeline.config.webhook.terminalStatuses':
     'Comma-separated list of strings that will be considered as TERMINAL status.',
   'pipeline.config.webhook.customHeaders': 'Key-value pairs to be sent as additional headers to the service.',
+  'pipeline.config.webhook.failFastCodes':
+    'Comma separated HTTP status codes (4xx or 5xx) that will cause this webhook stage to fail without retying.',
```
anotherchrisberry: * should probably be `Comma-separated` just to be consistent with the other help text on the stage right above this entry - we are not consistent throughout Deck on "comma separated" vs. "comma-separated".
* s/retying/retrying

app/scripts/modules/core/src/pipeline/config/stages/webhook/webhookStage.ts
marchello2000: I was trying to be more permissive, but you're right, better to be consistent and have the behavior match the docs. Will update to just accept commas

christopherthielen: I'd also be happy with `.split(/[,\s]+/)`

app/scripts/modules/core/src/help/help.contents.ts
marchello2000: 👍

app/scripts/modules/core/src/pipeline/config/stages/webhook/webhookStage.ts
```diff
@@ -54,6 +56,9 @@ export class WebhookStage implements IController {
     this.viewState = {
       waitForCompletion: this.stage.waitForCompletion || false,
       statusUrlResolution: this.stage.statusUrlResolution || 'getMethod',
+      failFastStatusCodes: isNullOrUndefined(this.stage.failFastStatusCodes)
```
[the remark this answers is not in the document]
marchello2000: for some reason that was giving me trouble, let me try again

app/scripts/modules/core/src/pipeline/config/stages/webhook/webhookStage.ts
anotherchrisberry: I like the initiative!

app/scripts/modules/core/src/pipeline/config/stages/webhook/webhookStage.ts
```diff
@@ -4,10 +4,12 @@ import { IModalService } from 'angular-ui-bootstrap';
 import { API } from 'core/api/ApiService';
 import { JsonUtils } from 'core/utils';
 import { Registry } from 'core/registry';
+import { isNullOrUndefined } from 'util';
```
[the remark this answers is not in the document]
marchello2000: got it for future, for now, removed

app/scripts/modules/core/src/pipeline/config/stages/webhook/webhookStage.ts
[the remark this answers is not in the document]
marchello2000: 👍

app/scripts/modules/core/src/pipeline/details/StageFailureMessage.tsx
```diff
@@ -72,9 +72,9 @@ export class StageFailureMessage extends React.Component<IStageFailureMessagePro
       const exceptionTitle = isFailed ? (messages.length ? 'Exceptions' : 'Exception') : 'Warning';
       const displayMessages =
         message || !messages.length ? (
-          <Markdown message={message || 'No reason provided.'} className="break-word" />
+          <Markdown message={message || 'No reason provided.'} />
```
marchello2000: @christopherthielen @anotherchrisberry : what do you think about this. Since `break-word` is actually defined as `break-all` in our CSS error messages look a little hard to read. e.g.
image

Looks like there was a bit of back and forth (https://github.com/spinnaker/deck/pull/4344) on this. Happy to undo if you think this isn't the right thing to do.

ps. here is with the word break defaulted to breaking on words (after this change)
image

christopherthielen: nice 👍

anotherchrisberry: It's unfortunate that we're in this situation - would be nice to use `break-word` instead of `break-all`. There's probably a better CSS solution - maybe hiding `overflow-x` is a better approach.

Let me try to find a stacktrace in an error message and see if we can do better...

anotherchrisberry: It's been long enough since we ran into whatever issues we were seeing with the `break-all` thing that I don't remember what it was. I dare you to change the CSS class for `break-word` to this:
```css
.break-word {
  overflow-wrap: break-word
}
```
as part of the PR (or a separate one) and we can see where it goes wrong...

marchello2000: Challenge accepted...?
My head hurts reading the explanation of the diff on mdn :)
Let me make a separate PR

PR #6416, opened 2019-01-23
feat(kubernetes/v2): Adds ClipboardText component to ease getting information from labels in UI

DarrenN:
features:

- Adds `ClipboardText` component to be used in text-labels to make it easy to copy/paste information directly from the UI. (Is a React version of existing Angular component).
- When you click on a label that uses the component, it will be auto-selected and the text will be copied to your clipboard automatically.
- Example: https://cl.ly/493c95d5c56f

copywidgetthing

fixes:

Issue in which it is difficult to double-click to select text labels (this is probably due to the use of dictionary lists in the rendered HTML).

### DIFF
--- a/app/scripts/modules/core/src/utils/ClipboardText.less
+++ b/app/scripts/modules/core/src/utils/ClipboardText.less
@@ -1,0 +1,14 @@
+.clipboard-btn {
+  border-width: 0;
+  background-color: transparent;
+  color: var(--color-dovegray);
+  display: inline-block;
+}
+.clipboard-btn:hover {
+  background-color: transparent;
+}
+&.copy-to-clipboard-sm {
+  .clipboard-btn {
+    margin-bottom: 3px;
+  }
+}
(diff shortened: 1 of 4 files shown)


sbwsg:
Deck has a CopyToClipboard component that I think provides this behaviour already. Did you explore using that as a first step?  You can find example usage in `app/scripts/modules/core/src/pipeline/executions/execution/Execution.tsx`

DarrenN:
@sbwsg my understanding was that `CopyToClipboard` was an Angular component, re: `app/scripts/modules/core/src/utils/clipboard/copyToClipboard.component.ts`

As noted in the description above this is a straight React port.

sbwsg:
First, I really like the feature. I think there's value in having the copy to clipboard button in this new location.

Notice that there's an adapter from the angular copy-to-clipboard component to React (the example I gave is `Executions.tsx`, a React component using CopyToClipboard in JSX).

So I don't think there's a ton of need for three components (1 angular, 1 adapter, 1 "pure" react) that all do the exact same thing living in the codebase all at the same time.

Again, really like the feature. If you want to completely replace the angular component with a React one + react2angular adapter then I'd definitely like to see that in a separate PR.

DarrenN:
Allrighty.

DarrenN:
@christopherthielen makes sense, will take a swing at it - thanks for the tips.

PR #6431, opened 2019-01-24
Dont import using file extension

christopherthielen:
Doing some housekeeping and modernization. 
I would love to get to ES6 modules and eliminate use of commonJS `require()` and `exports`


christopherthielen:
testing this @spinnaker/netflix-deck-reviewers tag

PR #6433, opened 2019-01-24
chore(package): remove unused (hopefully) spin.js package

christopherthielen:
I don't think this package `spin.js` is used.  @spinnaker/netflix-deck-reviewers anyone have any insight?
I searched for `.spin(` and got 0 results.


christopherthielen:
it's being used by `angular-spinner` but is also a direct dependency of `angular-spinner`

christopherthielen:
Turns out this is a direct dependency of `angular-spinner` and putting `Spinner` on `window` is not even necessary.  Merging this PR because this is confusing and doesn't help anyone.

PR #6438, opened 2019-01-28
fix(provider/azure): Fix the account list can't show when "Create Server Group" in cluster page 

michaeljqzq:
Root cause: In **serverGroupCommandBuilder.service**, the `defaultCredentials` and `defaultRegion` are not set to correct value. So that the "credential" property passed to **accountSelectField.directive.js** is a Map instead of a string. And calling `includes` function on the Map cause the error which prevent the account list show


kevinheins:

PR #6348, opened 2019-01-10
Rejigger wizard modal and wizard wizard page to use render props

christopherthielen:
The goal of this refactor is to 

a) minimize wizard modal magic 
b) allow more flexibility with how Wizard Page components can be rendered

This PR is split into multiple commits:

- refactor(core/modal): Rewrite wizard modal to use render props
  - Move label from wrapped component to WizardPage prop
  - Expose 'formik' as a separate prop in the render prop callback
  - Access wrapped component validate() function using 'innerRef'
  - Remove wizardPage() HOC
  - Remove hideSections prop
- refactor(core/projects): Migrate to wizardmodal render props
- refactor(amazon/modal): Refactor amazon modals to use new WizardPage component
- refactor(cloudfoundry/modal): Refactor cloudfoundry modals to use WizardPage component
- refactor(kubernetes/modal): Refactor kubernetes modals to use WizardPage component
- refactor(titus/modal): Refactor titus modals to use WizardPage component

Pros: Render props make the data flow more obvious
Cons: This is waaaaaay more verbose than the HOC

PR #6449, opened 2019-01-30
feat(cf/serverGroup): Use discriminator 'type' field to improve type checking

christopherthielen:
Discriminated union types:  https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions

I ran into this while refactoring the modals in #6348.  You can improve type checking by adding a discriminator to each interface in the union type.  Then, TypeScript can tell which one of the union types is present based on the value of the discriminator.

@Jammy-Louie PTAL...  if you like it I'll merge when you say so.


app/scripts/modules/cloudfoundry/src/serverGroup/configure/serverGroupCommandBuilder.service.cf.ts
```diff
@@ -93,9 +93,8 @@ export class CloudFoundryServerGroupCommandBuilder {
         routes: serverGroup.loadBalancers,
         environment: CloudFoundryServerGroupCommandBuilder.envVarsFromObject(serverGroup.env),
         services: (serverGroup.serviceInstances || []).map(serviceInstance => serviceInstance.name),
-        reference: '',
```
christopherthielen: After adding the discriminator, these fields `reference`, `account`, and `pattern` were flagged as a type error.  It also told me that `healthCheckType` and `healthCheckHttpEndpoint` were required when `type === 'direct'`.  Can you validate that this is correct?

PR #6453, opened 2019-01-30
fix(core/ga): Re-enable google analytics

christopherthielen:
Fixes https://github.com/spinnaker/spinnaker/issues/3914


ezimanyi:
Just discussed this on slack, but noting here for posterity---this doesn't actually enable Google analytics by default.  Users can if desired set Deck to report to their own Google Analytics account; this commit just fixes that integration, which was temporarily broken.

PR #6457, opened 2019-01-31
fix(provider/azure): update stack and detail field validation for azure

neil-yechenwei:
Currently spinnaker allows dot and underscore special character in the stack and detail field while creating server group for Azure. 
Actually, it doesn't support them for Azure after testing.
So update the code to not allow dot and underscore in the stack and detail field for Azure.


mybayern1974:
@sbwsg, can this PR be merged since you've approved it?

duftler:
I believe the format of the commit message needs to be updated.

duftler:
Also please rebase to master and force push the branch instead of merging master using the github ui.

sbwsg:
I updated the commit title during the squash and merge process to match the title of this PR. Also just double checked and it looks like the merge has ended up at HEAD of the master branch, as if it was rebased on top.  So I think the formatting errors that spinnakerbot notes here can be fixed at merge time. 🤞

PR #6462, opened 2019-01-31
fix(projects): config state was stale after update

alanmquach:
Issue was:
1. Update a `Project Configuration`
2. Submit
3. Updates are persisted to server
4. Click `Project Configuration`
5. Updates are gone!

Spoiler alert: we actually did update it, the in-memory props were just stale.


alanmquach:
So this fixes it but I'm also curious if the actual bug is we are short-circuiting before this code gets to execute:
https://github.com/spinnaker/deck/blob/1246086496255ea7be933380d5b7a7cd857598e2/app/scripts/modules/core/src/projects/ProjectHeader.tsx#L51-L58

alanmquach:
I tried it the other way (https://github.com/spinnaker/deck/pull/6464) and it required adding the post-creation transition to the newly created project https://github.com/spinnaker/deck/blob/44b0d19b8352f71898d1a749bf5d3b2cc9edca0a/app/scripts/modules/core/src/insight/InsightMenu.tsx#L42-L47

so I think I like this way better as the state transitions can remain local to projects. Thoughts?

alanmquach:
That does feel more natural, let's go with that.

PR #6473, opened 2019-02-02
fix(provider/azure): Add Application Name Validator for Azure

neil-yechenwei:
Currently spinnaker doesn't validate application name for azure.
So add an application name validator for azure.


mybayern1974:
@sbwsg , @lwander , @duftler , a kindly ping to review this PR. The purpose of this PR is, without this fix, the user is likely to create an invalid application name that will be used in the later Azure deploy stage, in which things will break w/o friendly error notifications. So the PR is to prevent that at the very beginning.

duftler:
I think SpinnakerBot is pointing out that the commit message needs its format updated, not that the PR description needs its format updated.

maggieneterval:
@neil-yechenwei this looks good to go, would you mind resolving the conflicts when you get a chance so we can merge? Thanks!

PR #6483, opened 2019-02-05
fix(core): read displayTimestampsInUserLocalTime off SETTINGS.feature

maggieneterval:
[This](https://github.com/spinnaker/deck/pull/6362) PR added the `displayTimestampsInUserLocalTime` feature flag to `SETTINGS.feature` in `settings.js` and `halconfig/settings.js` but tried to read it off of `SETTINGS` directly. This change makes sure that it is read off of `SETTINGS.feature` everywhere.


maggieneterval:

PR #6504, opened 2019-02-07
fix(cf): fix SG pipeline modal display issues

Jammy-Louie:


christopherthielen:
is this a workaround or do you think this is the proper way to fix the issue with blank modal?

Jammy-Louie:
This is a work around for now. Not sure yet what a proper fix would be yet.
I think the amazon cloud driver would run into a similar issue but they also incidentally also call 2 set states which trigger 2 renders to make it work.

PR #6505, opened 2019-02-07
fix(bootstrap): make sure jquery loads before angular

christopherthielen:
why did this fix it i dont know

screen shot 2019-02-07 at 2 25 29 pm


maggieneterval:
Ahhh thanks so much for the fix, this has been driving me crazy all day!

PR #6508, opened 2019-02-07
fix(kubernetes): do not override `location` and `replicas` in new Sca…

maggieneterval:
…le Manifest stage

Closes https://github.com/spinnaker/spinnaker/issues/3949

We do not remove the `isNew` field from a stage [until the pipeline is saved](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/services/PipelineConfigService.ts#L66), so previously we were overwriting fields on the Scale Manifest stage each time the form re-rendered when clicking back and forth among other stages and the new Scale Manifest stage. This switches `Object.assign` --> `_.defaults` so non-undefined values are not overwritten.


maggieneterval:

PR #6509, opened 2019-02-08
fix(amazon): do not call setState inside validate

alanmquach:
Causes looping, also isn't actually needed.


app/scripts/modules/amazon/src/loadBalancer/configure/network/CreateNetworkLoadBalancer.tsx
```diff
@@ -221,7 +221,6 @@ export class CreateNetworkLoadBalancer extends React.Component<
   };
 
   private validate = (): FormikErrors<IAmazonNetworkLoadBalancerUpsertCommand> => {
-    this.setState({});
```
christopherthielen: hah

PR #6516, opened 2019-02-08
fix(core): introduce state when the modal has been initialized

Jammy-Louie:


Jammy-Louie:
@christopherthielen just a proposal to fix the problem we ran into yesterday.

app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
```diff
@@ -62,19 +63,25 @@ export class WizardModal<T = {}> extends React.Component<IWizardModalProps<T>, I
...
+  public componentDidMount(): void {
+    this.setState({ initialized: true });
+  }
+
   public onWizardPageAdded = (wizardPage: WizardPage<T>): void => {
     this.setState(prevState => {
       const pages = prevState.pages.concat(wizardPage);
       const currentPage = prevState.currentPage || pages[0];
-      return { pages, currentPage };
+      const initialized = prevState.initialized;
```
christopherthielen: this shouldn't be necessary (`setState()` with the same value of `initialized`)

app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
```diff
@@ -62,19 +63,25 @@ export class WizardModal<T = {}> extends React.Component<IWizardModalProps<T>, I
...
+      return { pages, currentPage, initialized };
     }, this.revalidate);
   };
 
   public onWizardPageRemoved = (wizardPage: WizardPage<T>): void => {
     this.setState(prevState => {
       const pages = without(prevState.pages, wizardPage);
       const currentPage = prevState.currentPage || pages[0];
-      return { pages, currentPage };
+      const initialized = prevState.initialized;
```
christopherthielen: same

app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
```diff
@@ -171,7 +178,7 @@ export class WizardModal<T = {}> extends React.Component<IWizardModalProps<T>, I
               <Modal.Header>{heading && <h3>{heading}</h3>}</Modal.Header>
 
               <Modal.Body>
-                {loading ? (
+                {loading && initialized ? (
```
christopherthielen: should this be `loading || !initialized` ?  maybe it doesn't make any difference.

app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
Jammy-Louie: 👍

app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
Jammy-Louie: 👍

app/scripts/modules/core/src/modal/wizard/WizardModal.tsx
Jammy-Louie: 👍

Jammy-Louie:
changes made

PR #6520, opened 2019-02-08
feat(core): add feature flag for managed pipeline templates v2

louisjimenez:
https://github.com/spinnaker/spinnaker/issues/3894

Adds a feature flag for Managed Pipeline Templates v2. Halyard changes to follow in separate PR.


settings.js
```diff
@@ -62,6 +62,7 @@ window.spinnakerSettings = {
     // whether stages affecting infrastructure (like "Create Load Balancer") should be enabled or not
     infrastructureStages: infrastructureEnabled,
     jobs: false,
+    managedPipelineTemplatesV2: false,
```
sbwsg: It might be worth adding an environment variable that can be used to enable the feature independently of halyard, similarly to the `templatesEnabled` var above.

louisjimenez: Good call, added!

PR #6523, opened 2019-02-11
chore(cf): rename serviceName to serviceInstanceName

Jammy-Louie:
renaming `serviceName` to `serviceInstanceName` so it is clear that `serviceInstanceName` represents the new service instance and not the name of the Service


Jammy-Louie:
requires PR from Clouddriver: https://github.com/spinnaker/clouddriver/pull/3358

PR #6539, opened 2019-02-13
fix(provider/azure): subnet is null after select same loadbalancer

neil-yechenwei:
Currently it sets subnet as null as default value in drop down list after select same loadbalancer.
So update the logic to not change subnet after select same loadbalancer.


sbwsg:
I reviewed this prior to merging but forgot to click the Approve button.  LGTM!

PR #6544, opened 2019-02-14
fix(appengine): older apps that dont have cloudProviders field dont select correct serverGroup modal

sbwsg:
If an application does not include a `cloudProviders` string in its JSON then certain UI features will fall back to the set of `defaultProviders` defined in the settings object when determining whether that application includes specific providers. If `"appengine"` is missing from the list of default providers then older appengine applications that don't include the `cloudProviders` field can end up showing AWS modals by mistake.

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -49,7 +49,18 @@ window.spinnakerSettings = {
   debugEnabled: debugEnabled,
   defaultCategory: 'serverGroup',
   defaultInstancePort: 80,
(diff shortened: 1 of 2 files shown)


settings.js
```diff
@@ -49,7 +49,18 @@ window.spinnakerSettings = {
...
+    'aws',
+    'gce',
+    'azure',
+    'cloudfoundry',
+    'kubernetes',
+    'dcos',
+    'openstack',
+    'oracle',
+    'ecs',
+    'appengine',
```
ttomsu: Nit: Alphabetize here and the other file.

sbwsg:

christopherthielen:
we're trying to get ya'll to use AWS

PR #6546, opened 2019-02-14
fix(core): Match order of pipeline config nav items to page sections

louisjimenez:
https://github.com/spinnaker/spinnaker/issues/3970

The feature flag check now occurs in the `triggers.directive` giving each `<page-section>` a chance to init and register its page in the correct order with the `PageNavigationState`. `PageNavigationState` can then be used to render navigation items as long as they are `visible`.

This includes `prettier` changes that I can split into a separate commit if helpful.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
@@ -4,13 +4,13 @@
       <div class="col-md-11 col-md-offset-1">
         <div class="checkbox">
           <label>
-            <input type="checkbox" ng-model="pipeline.limitConcurrent">
+            <input type="checkbox" ng-model="pipeline.limitConcurrent" />
             <strong>Disable concurrent pipeline executions (only run one at a time). </strong>
           </label>
         </div>
         <div ng-if="pipeline.limitConcurrent" class="checkbox">
           <label>
-            <input type="checkbox" ng-model="pipeline.keepWaitingPipelines">
+            <input type="checkbox" ng-model="pipeline.keepWaitingPipelines" />
             <strong>Do not automatically cancel pipelines waiting in queue.</strong>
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
```diff
@@ -4,50 +4,55 @@
...
-            <input type="checkbox" ng-model="pipeline.keepWaitingPipelines">
+            <input type="checkbox" ng-model="pipeline.keepWaitingPipelines" />
             <strong>Do not automatically cancel pipelines waiting in queue.</strong>
             <help-field key="pipeline.config.parallel.cancel.queue"></help-field>
           </label>
         </div>
       </div>
     </div>
   </page-section>
-  <render-if-feature feature="artifacts">
```
louisjimenez: Most of this file is prettier changes, but this page section is taken out of the `<render-if-feature>` and instead relies on `checkFeatureFlag`

app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
```diff
@@ -4,50 +4,55 @@
...
-          <button class="btn btn-block btn-add-trigger add-new" ng-click="triggersCtrl.addArtifact()">
-            <span class="glyphicon glyphicon-plus-sign"></span> Add Artifact
-          </button>
-        </div>
+  <page-section
+    key="artifacts"
+    label="Expected Artifacts"
+    badge="pipeline.expectedArtifacts.length"
+    no-wrapper="true"
+    visible="triggersCtrl.checkFeatureFlag('artifacts')"
```
louisjimenez: Calls `checkFeatureFlag` to determine visibility

louisjimenez:
@sbwsg @maggieneterval fyi

sbwsg:

PR #6552, opened 2019-02-14
fix(titus): Help users with iam profile setting

archana-s:
IAM instance profile for existing run job stages will get a recommended value.

Recently, we made IAM instance profile mandatory for run job titus stages. There are some pipelines which have run job stages without IAM profile set. This led to some user errors. One of the users updated other settings, and our application set the default application specific IAM profile. This change was unintentional and led to errors since they relied on titus's default IAM profile. In order to avoid these errors, we show default iam profile as a placeholder for existing run job stages without iam profiles. For newer stages, we automatically set iam profile so users do not have to worry about it. 

Treatment for existing run job stages: 
image

### DIFF
--- a/app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
+++ b/app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
@@ -26,8 +26,27 @@ export interface ITitusRunJobStageConfigState {
   loaded: boolean;
 }
 
+interface IClusterDefaults {
+  application: string;
+  containerAttributes: object;
+  env: object;
+  labels: object;
+  resources: {
+    cpu: number;
+    disk: number;
+    gpu: number;
+    memory: number;
+    networkMbps: number;
+  };
+  retries: number;
+  runtimeLimitSecs: number;
+  securityGroups: string[];
+  iamProfile?: string;
+}
+
 export class TitusRunJobStageConfig extends React.Component<IStageConfigProps, ITitusRunJobStageConfigState> {
   private credentialsKeyedByAccount: IAggregatedAccounts = {};
+  private defaultIamProfile = '';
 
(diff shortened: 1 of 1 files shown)


emjburns:
I think that "Use default" would be better phrasing for that checkbox, because it's more clear what the option does with that wording.

app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
```diff
@@ -64,7 +64,7 @@ export class TitusRunJobStageConfig extends React.Component<IStageConfigProps, I
       application: application.name,
       containerAttributes: {},
       env: {},
-      iamProfile: defaultIamProfile,
+      defaultIamProfile,
```
jrsquared: this value will be sent down in the deploy config, and it's not a valid field, right?

archana-s: This is appended to the existing stage object. Let's say a stage has the iamProfile, we don't need to set it here; we use it directly. Let's say the value is not set, then we set the value from defaultIamProfile or set this as a placeholder. Is that what you are asking?

jrsquared: Yeah, I don't think we should set `defaultIamProfile` on the cluster object itself, since that object is what gets sent in the api call to gate.

archana-s:
> I think that "Use default" would be better phrasing for that checkbox, because it's more clear what the option does with that wording.

Makes sense. Fixed.

app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
archana-s: Thanks for catching it! Will send out a fix shortly.

PR #6554, opened 2019-02-15
fix(core/css): be explicit on which file we're importing

kevinawoo:
in the module `source-sans-pro@2.0.10` `packageJson.main` is specified to be `source-sans-pro.css`, however this is dropped in following versions. Here's a screenshot of available versions:

To address this, let's directly reference the file which should help prevent build issues when upgrading and also with importing `@spinnaker/core` as a module in deck extensions since the current package.json specifies `"source-sans-pro": "^2.0.10",` which allows using the `2.45.0 (latest)` version.


kevinawoo:
@spinnakerbot 	cherry-pick 1.11

kevinawoo:
@spinnakerbot cherry-pick 1.11.x

kevinawoo:

kevinawoo:

PR #6569, opened 2019-02-18
fix(securityGroups): User `securityGroupName` for upsertSecurityGroupTask

marchello2000:
Add `securityGroupName` to the `upsertSecurityGroupTask` as per `orca` change
https://github.com/spinnaker/orca/pull/2668

### DIFF
--- a/app/scripts/modules/azure/securityGroup/securityGroup.write.service.js
+++ b/app/scripts/modules/azure/securityGroup/securityGroup.write.service.js
@@ -9,6 +9,8 @@ module.exports = angular
   .module('spinnaker.azure.securityGroup.write.service', [require('@uirouter/angularjs').default])
   .factory('azureSecurityGroupWriter', function() {
     function upsertSecurityGroup(securityGroup, application, descriptor, params = {}) {
+      params.securityGroupName = securityGroup.name;
+
       // We want to extend params with all attributes from securityGroup, but only if they don't already exist.
       _.assignWith(params, securityGroup, function(value, other) {
         return _.isUndefined(value) ? other : value;
--- a/app/scripts/modules/azure/securityGroup/clone/cloneSecurityGroup.controller.js
+++ b/app/scripts/modules/azure/securityGroup/clone/cloneSecurityGroup.controller.js
@@ -147,7 +147,6 @@ module.exports = angular
         let params = {
           cloudProvider: 'azure',
           appName: application.name,
-          securityGroupName: $scope.securityGroup.name,
           region: $scope.securityGroup.region,
           subnet: 'none',
           vpcId: 'null',
--- a/app/scripts/modules/azure/securityGroup/configure/CreateSecurityGroupCtrl.js
+++ b/app/scripts/modules/azure/securityGroup/configure/CreateSecurityGroupCtrl.js
@@ -105,7 +105,6 @@ module.exports = angular
         let params = {
           cloudProvider: 'azure',
           appName: application.name,
-          securityGroupName: $scope.securityGroup.name,
           region: $scope.securityGroup.region,
           subnet: 'none',
           vpcId: 'null',
--- a/app/scripts/modules/azure/securityGroup/configure/EditSecurityGroupCtrl.js
+++ b/app/scripts/modules/azure/securityGroup/configure/EditSecurityGroupCtrl.js
@@ -153,7 +153,6 @@ module.exports = angular
         let params = {
           cloudProvider: 'azure',
           appName: application.name,
-          securityGroupName: $scope.securityGroup.name,
           region: $scope.securityGroup.region,
           subnet: 'none',
           vpcId: 'null',
--- a/app/scripts/modules/core/src/securityGroup/SecurityGroupWriter.spec.ts
+++ b/app/scripts/modules/core/src/securityGroup/SecurityGroupWriter.spec.ts
@@ -1,0 +1,13 @@
+import { SecurityGroupWriter } from './SecurityGroupWriter';
+import { TaskExecutor } from 'core/task';
+
+describe('SecurityGroupWriter', () => {
+  it('copies the security group name on upsert', () => {
+    const spy = spyOn(TaskExecutor, 'executeTask');
+    SecurityGroupWriter.upsertSecurityGroup({ name: 'mySecurityGroupName' }, null, 'myDescription', {});
+
+    expect(spy).toHaveBeenCalled();
+    const job = spy.calls.mostRecent().args[0].job[0];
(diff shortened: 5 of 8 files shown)


app/scripts/modules/amazon/src/securityGroup/configure/EditSecurityGroupCtrl.js
```diff
@@ -90,6 +90,7 @@ module.exports = angular
       let command = {
         credentials: group.accountName,
         name: group.name,
```
christopherthielen: can't we remove `name` from the payload?

app/scripts/modules/amazon/src/securityGroup/clone/cloneSecurityGroup.controller.js
```diff
@@ -85,6 +85,7 @@ module.exports = angular
         account: credentials,
         accountName: credentials,
         accountId: credentials,
+        securityGroupName: $scope.securityGroup.name,
```
christopherthielen: instead of `$scope.securityGroup.name` destructure `name` on line 83

```
      const { credentials, name } = $scope.securityGroup;
```
then do 
```suggestion
        securityGroupName: name,
```

app/scripts/modules/amazon/src/securityGroup/configure/EditSecurityGroupCtrl.js
```diff
@@ -90,6 +90,7 @@ module.exports = angular
...
+        securityGroupName: group.name,
```
marchello2000: @christopherthielen : not sure if there are other places I need to change

app/scripts/modules/amazon/src/securityGroup/configure/EditSecurityGroupCtrl.js
marchello2000: unfortunately no because orca needs it

christopherthielen: I'm not sure I understand, I thought you submitted a PR against orca to switch from `name` to `securityGroupName`?

app/scripts/modules/amazon/src/securityGroup/clone/cloneSecurityGroup.controller.js
marchello2000: thank you!

app/scripts/modules/amazon/src/securityGroup/configure/EditSecurityGroupCtrl.js
marchello2000: I take it back. We could remove it... But that would make logs hard to read.
Basically, what happens, `orca` assumes the `name` field is for it and removes it for context but uses it for all human readable references to this task.
But because it removes it the task doesn't get it and hence can't execute. This is, as we discovered only an issue in execution.

But you're right since the UI only calls this on the `/tasks` endpoint we should be able to safely remove it. Let me try

app/scripts/modules/amazon/src/securityGroup/configure/EditSecurityGroupCtrl.js
christopherthielen: unfortunately, yes.   look for usages of  `SecurityGroupWriter.upsertSecurityGroup`

christopherthielen: interestingly, some call sites already use `securityGroupName`, i.e.: 

https://github.com/spinnaker/deck/blob/master/app/scripts/modules/openstack/src/securityGroup/configure/wizard/upsert.controller.js#L161-L164

https://github.com/spinnaker/deck/blob/master/app/scripts/modules/kubernetes/src/securityGroup/configure/wizard/upsert.controller.js#L166-L169

marchello2000: 🤕 the deleteTask expects this format but these are both for upsert..

app/scripts/modules/amazon/src/securityGroup/configure/EditSecurityGroupCtrl.js
marchello2000: after some more investigation, this is ok to remove, thanks!

marchello2000:
don't merge till https://github.com/spinnaker/orca/pull/2668 is in main

app/scripts/modules/amazon/src/securityGroup/clone/cloneSecurityGroup.controller.js
```diff
@@ -80,11 +80,13 @@ module.exports = angular
...
       // <account-select-field> only updates securityGroup.credentials, but Orca looks at account* before looking at credentials
       // Updating the rest of the attributes to send the correct (expected) account for all attributes
-      const { credentials } = $scope.securityGroup;
+      const { credentials, name } = $scope.securityGroup;
+
       Object.assign($scope.securityGroup, {
         account: credentials,
         accountName: credentials,
         accountId: credentials,
+        securityGroupName: name,
```
marchello2000: forgot to remove this

app/scripts/modules/azure/securityGroup/securityGroup.write.service.js
```diff
@@ -9,6 +9,8 @@ module.exports = angular
   .module('spinnaker.azure.securityGroup.write.service', [require('@uirouter/angularjs').default])
   .factory('azureSecurityGroupWriter', function() {
     function upsertSecurityGroup(securityGroup, application, descriptor, params = {}) {
+      params.securityGroupName = securityGroup.name;
```
marchello2000: @christopherthielen wanted to ask you about this, not sure why azure has a different writer, just wanted to make sure I understood it correctly

christopherthielen: Not sure why.  This looks fine.

app/scripts/modules/amazon/src/securityGroup/clone/cloneSecurityGroup.controller.js
```diff
@@ -80,7 +80,8 @@ module.exports = angular
     vm.upsert = function() {
       // <account-select-field> only updates securityGroup.credentials, but Orca looks at account* before looking at credentials
       // Updating the rest of the attributes to send the correct (expected) account for all attributes
-      const { credentials } = $scope.securityGroup;
+      const { credentials, name } = $scope.securityGroup;
```
christopherthielen: `name` is unused and is causing a linter error

marchello2000: thanks!

PR #6578, opened 2019-02-20
fix(kubernetes): properly detect if autoscaler is attached to server groups

emptywee:
`serverGroup.autoscaler` comes as `undefined` in angular scope when there's no autoscaler attached to a server group, thus wrong modals were shown to users.

@christopherthielen @maggieneterval - feel free to test it with Kubernetes V2. I've tested it with Kubernetes V1 and now everyone is happy that the Resize feature can be used again.

### DIFF
--- a/app/scripts/modules/kubernetes/src/serverGroup/details/details.html
+++ b/app/scripts/modules/kubernetes/src/serverGroup/details/details.html
@@ -111,13 +111,13 @@
     </collapsible-section>
     <collapsible-section heading="Replicas" expanded="true">
       <dl class="dl-horizontal dl-flex">
-        <div ng-show="serverGroup.autoscalerStatus === null">
+        <div ng-show="!serverGroup.autoscalerStatus">
           <dt>Current</dt>
           <dd>{{serverGroup.instances.length}}</dd>
           <dt>Desired</dt>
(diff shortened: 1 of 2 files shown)


app/scripts/modules/kubernetes/src/serverGroup/details/details.html
```diff
@@ -191,7 +191,7 @@ <h4 class="text-center" ng-if="serverGroup.isDisabled">[SERVER GROUP IS DISABLED
     </collapsible-section>
     <collapsible-section heading="Events">
       <div ng-if="serverGroup.events && serverGroup.events.length > 0">
-        <kubernetes-event ng-repeat="event in serverGroup.events" event="event"></kubernetes-event>
+        <kubernetes-event ng-repeat="event in serverGroup.events" event="event"/>
```
christopherthielen: this needs to be reverted because self closing custom tags are not valid html

```suggestion
        <kubernetes-event ng-repeat="event in serverGroup.events" event="event"></kubernetes-event>
```

emptywee: Ooops, that was my IDE'd doing :\

emptywee:
@christopherthielen reverted the invalid html due to self-closing tag.

PR #6582, opened 2019-02-20
fix(provider/azure): respect vm scale set schema for custom extensions

mybayern1974:
**Background:**
Azure Resource Template is used to provision VM Scale Set (VMSS). Within the template, users could add the section of extensionProfile if they want to execute any post actions after VMSS is provisioned.  

**What to fix**
The extensionProfile supports multiple custom scripts to be downloaded but the current behavior only allows to config one script. So fix it to respect the VMSS feature design.

This PR covers another trivial fix that is to remove some hard coded credential texts which indeed are not used anywhere.


app/scripts/modules/azure/serverGroup/serverGroup.transformer.js
```diff
@@ -79,10 +77,22 @@ module.exports = angular
 
       if (typeof command.customScriptsSettings !== 'undefined') {
         configuration.customScriptsSettings.commandToExecute = command.customScriptsSettings.commandToExecute;
-        if (Array.isArray(command.customScriptsSettings.fileUris)) {
-          configuration.customScriptsSettings.fileUris = command.customScriptsSettings.fileUris;
-        } else {
-          configuration.customScriptsSettings.fileUris = [command.customScriptsSettings.fileUris];
+        if (
+          typeof command.customScriptsSettings.fileUris !== 'undefined' &&
```
sbwsg: You can use lodash's `isEmpty` method here to simplify a bit:

```javascript
if (!_.isEmpty(command.customScriptsSettings.fileUris)) {
  var fileUrisTemp = command.customScriptsSettings.fileUris;
  // ...
}
```

You'll need to add lodash as an import to the top of the file: `const _ = require('lodash');`

app/scripts/modules/azure/serverGroup/serverGroup.transformer.js
```diff
@@ -79,10 +77,22 @@ module.exports = angular
...
+          if (fileUrisTemp.includes(',')) {
+            configuration.customScriptsSettings.fileUris = fileUrisTemp.split(',');
+          } else if (fileUrisTemp.includes(';')) {
+            configuration.customScriptsSettings.fileUris = fileUrisTemp.split(';');
+          } else {
+            configuration.customScriptsSettings.fileUris = [fileUrisTemp];
+          }
+
+          configuration.customScriptsSettings.fileUris.forEach(function(v, index) {
+            configuration.customScriptsSettings.fileUris[index] = v.trim();
```
sbwsg: It is possible the array will end up with empty strings here.  Is that a problem?

For example, if `command.customScriptsSettings.fileUris` is `" "` then `configuration.customScriptsSettings.fileUris` would equal `['']`

sbwsg:
I'm going to merge this since it's been open so long. The minor issues I raised can be resolved in a follow-up commit.

PR #6586, opened 2019-02-20
fix(core): Child pipeline should route back correctly

archana-s:
Child pipeline in different application should route back to the initiating pipeline/application.

With growing number of teams that create an application only with pipelines, the number of pipeline stages that point to pipelines in other applications have increased. When someone navigates from a parent pipeline to a child pipeline execution through the pipelineStage, the back arrow takes the user back to the current application executions. However this can be annoying for users who landed here from other applications. This diff fixes where the back arrow takes the user.


app/scripts/modules/core/src/pipeline/details/SingleExecutionDetails.tsx
```diff
@@ -128,6 +128,9 @@ export class SingleExecutionDetails extends React.Component<
   public render() {
     const { app } = this.props;
     const { execution, sortFilter, stateNotFound } = this.state;
+    let { senderApplication: application, senderExecutionId: executionId } = ReactInjector.$state.params;
```
christopherthielen: If you pass an object in the params such as `executionParams`, then you could do:
```
const defaultExecutionParams = { application: app.name, executionId: execution && execution.id };
const executionParams = ReactInjector.$state.params.executionParams || defaultExecutionParams;
```
and then use `executionParams` in the sref:
```
<UISref to="^.executions.execution" params={executionParams}>
```

app/scripts/modules/core/src/pipeline/pipeline.states.ts
```diff
@@ -55,6 +55,14 @@ module(PIPELINE_STATES, [APPLICATION_STATE_PROVIDER]).config(
       views: {
         pipelines: { component: SingleExecutionDetails, $type: 'react' },
       },
+      params: {
+        senderApplication: {
```
christopherthielen: the name `sender` seems ambigious to me, how about something like `executionParams`

app/scripts/modules/core/src/pipeline/pipeline.states.ts
```diff
@@ -55,6 +55,14 @@ module(PIPELINE_STATES, [APPLICATION_STATE_PROVIDER]).config(
       views: {
         pipelines: { component: SingleExecutionDetails, $type: 'react' },
       },
+      params: {
```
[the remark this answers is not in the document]
archana-s: I like that. I will update it.

PR #6603, opened 2019-02-22
fix(core/executions): Disable text selection when re-ordering pipelines

erikmunson:
When re-ordering pipelines in the filter sidebar, it can get kinda wonky if you accidentally make the browser think you're selecting text. If you manage to click + drag just right, sometimes it even thinks you're trying to drag a multi-line text selection for copy-pasting.

This switches off text selection when entering re-ordering mode for strategies and normal pipelines. Outside re-ordering mode you can still select the pipeline names to copy-paste, if that's something you're into.


erikmunson:
yeah it's great for making user-hostile interfaces without the fuss:

```
* {
  user-select: none;
}
```

PR #6610, opened 2019-02-25
chore(*): Fix inconsistent @uirouter/react-hybrid versions

ezimanyi:
We're pinning react-hybrid to 0.3.8 in resolutions, but explicitly requesting 0.3.9 at the top level (as well as in deck-kayenta). This leads to an extra unused entry in yarn.lock for the 0.3.8 and also confuses yarn's caching as running yarn is downloading 0.3.9 every time which slows local development.


ezimanyi:
This makes `yarn` complete quickly when there's no work to do (as due to some strange caching behavior in `yarn` the inconsistentcy above was making it re-download ui-router every time, even when nothing changed).

More broadly, do we still need this `resolutions` block at all?  It looks like it was intended to be temporary, and it looks like most of the entries there don't actually have an effect.

christopherthielen:
I'm trying to remember why I added this block instead of simply updating deck-kayenta dependencies and cannot recall.    Feel free to try removing the resolutions block altogether.  I'll likely be working on all these dependencies in Q2 (they're a giant mess right now).

PR #6611, opened 2019-02-25
fix(build): Pass --frozen-lockfile to travis builds

ezimanyi:
By default, yarn will update yarn.lock if it's out of date with package.json. We don't want this to happen during CI builds, as the CI system is not committing yarn.lock and so is testing an unreproducible state. Yarn recommends passing --frozen-lockfile on CI builds for this reason. This will cause Travis builds to
fail if someone bumps dependencies in package.json, but forgets to commit yarn.lock.


ezimanyi:
This will block merging if `package.json` and `yarn.lock` are not in sync.

The Travis build on this PR will fail until I merge in #6610 but I wanted to create the PR now to double check that it's failing as expected (and that it passes once I merge in that change).

christopherthielen:
love it, thanks!

PR #6613, opened 2019-02-25
refactor(artifacts): Combine expected artifacts and trigger artifact constraints

jkschneider:
The last image uses a new common React component called `StageArtifactSelector`. K8S and AppEngine providers would have to be updated to use this as well.

Also, selecting Artifact Account up front in the artifact definition certainly has an impact on existing pipelines which might have defined expected artifacts and either later defined the account (e.g. in a K8S Deploy Manifest stage) or not. Perhaps for backwards compat, we could look forward in the pipeline to determine what account was selected and drag that forward to the pipeline definition of the expected artifact. In the case where no such account was selected later, we could introduce a composite artifact account for each kind and select that by default. The implementation would simply iterate over all available accounts of that type when attempting a download. But I think it's easier to use: folks don't need to understand the distinction between kind and account anymore.

image

Once an artifact is defined and added to a trigger constraint, it may be used by name in other triggers as well. If you delete a named artifact in every trigger, it is dropped from the pipeline definition.

image

image

image


christopherthielen:
I hope you were able to rebase over all my sweeping changes 🤞

jkschneider:
@christopherthielen Yep, got it done ;)

jkschneider:
I think the whole `customKind` thing can just go away. I'll wait to see what you think about this in general before dealing with those tests.

jkschneider:
Some sources that would be removed I think after this:

ExpectedArtifactEditor
ExpectedArtifactKindSelector
ExpectedArtifactSourceSelector
ExpectedArtifactSelector
IArtifactAccountPair (accounts are expected to be defined on expected artifacts up front)

Expect substantial change to:

NgAppEngineDeployArtifactDelegate
NgAppengineConfigArtifactDelegate
NgBakeManifestArtifactDelegate
NgGenericArtifactDelegate
NgGCEImageArtifactDelegate

PR #6625, opened 2019-02-27
fix(google/auth): Only show authorized accounts during Server group creation

dibyom:
Addresses https://github.com/spinnaker/spinnaker/issues/4005


duftler:

PR #6628, opened 2019-02-28
feat(kubernetes): add label mode to manifest selector component to en…

maggieneterval:
…able dynamic target selection in delete manifest stage

Partially resolves https://github.com/spinnaker/spinnaker/issues/3829 (will address Patch Stage in separate PR)

Changes:
- Removes Angular MultiManifestSelector component
- Adds Label mode to existing React ManifestSelector component
- Converts Angular LabelSelector to React
- Converts Angular DeleteManifestOptions to React
- Consolidates stage validation for Enable, Disable and Delete stages

ka5gzvyhghd


app/scripts/modules/kubernetes/src/v2/manifest/selector/labelEditor/LabelEditor.tsx
```diff
@@ -0,0 +1,126 @@
...
+                    this.handleChange(idx, 'values', LabelEditor.convertValueStringToArray(e.target.value))
+                  }
+                  placeholder="Comma-separated values"
+                  value={selector.values.join(', ')}
+                  type="text"
+                />
+              </td>
+              <td>
+                <div className="form-control-static">
+                  <a className="label-editor-remove" onClick={() => this.removeField(idx)}>
```
archana-s: you can use <button className="link"> if you prefer. That way you don't have to use click event on <a>

maggieneterval: ah sorry I am not understanding, use what if I prefer?

app/scripts/modules/kubernetes/src/v2/manifest/selector/ManifestSelector.tsx
```diff
@@ -248,15 +272,25 @@ export class ManifestSelector extends React.Component<IManifestSelectorProps, IM
     this.setStateAndUpdateStage({ selector: this.state.selector });
   };
 
+  public handleKindsChange = (kinds: string[]): void => {
+    this.state.selector.kinds = kinds;
```
archana-s: This does not mutate the state object since its editing existing properties which is good. but I think it might be more idiomatic and easy to read if the edits were made to a copy of the state object and passed along to `setState()` rather than editing it directly.

app/scripts/modules/kubernetes/src/v2/manifest/selector/labelEditor/LabelEditor.tsx
archana-s: Forgot to quote it :D

maggieneterval: Oh good call, I kinda blindly copied the html from the angular template. Updated!

app/scripts/modules/kubernetes/src/v2/manifest/selector/ManifestSelector.tsx
maggieneterval: Yeah, the way this component handles state is a bit wonky and probably deserves a thorough refactor. Going to leave as-is for now, only because this is the pattern across all of the pre-existing methods (see `handleClusterChange`, `handleCriteriaChange`, `handleNamespaceChange`, etc. above) but I will try to revisit this next time we have to edit this component.

PR #6630, opened 2019-03-01
fix(kubernetes): allow text input for replicas in `Scale (Manifest)` stage

perek:
Currently the deck UI forces numeric input for replicas in the `Scale (Manifest)` stage. This prohibits configuring the stage to accept SpEL. Changed the input type to text, built and tested in our production cluster.


maggieneterval:
cc @gcomstock because I know you were working on some designs for inputs that toggle between a validated input type and SpEL expression, here's another data point in favor of that being useful :)

PR #6627, opened 2019-02-28
fix(bake/oracle): Added extendedAttributes

guoyongzhang:
Adding extendedAttributes so that additional packer variables can be passed in to Oracle bakery.

PR #6631, opened 2019-03-01
fix(bake/oracle): Added extendedAttributes

guoyongzhang:
back port the fix (https://github.com/spinnaker/deck/pull/6627) into Spinnaker 1.12 release so that additional packer variables can be passed in to Oracle bakery.


guoyongzhang:
@anotherchrisberry Thank you for merging https://github.com/spinnaker/deck/pull/6627.  Could you review and merge this PR (the same fix, but for release-1.12.x)?

PR #6634, opened 2019-03-02
refactor(artifacts): Combine expected artifacts and trigger artifact constraints

jkschneider:
See https://github.com/spinnaker/deck/pull/6613 for details about why and further screenshots.

Depends on https://github.com/spinnaker/clouddriver/pull/3424


jkschneider:
This is the "Deploy Manifest" stage after incorporating the new selector:

image

It adds `requiredArtifacts` and `matchArtifact` to the existing `matchArtifactId` and `requiredArtifactIds`. These fields are of type `IArtifact` and are present when one of the bindable artifacts or the match artifact are defined inline in the stage.

image

jkschneider:
Here is a screenshot of the new Kubernetes artifact editor:

image

app/scripts/modules/core/src/pipeline/config/stages/stage.html
```diff
@@ -100,7 +100,7 @@ <h4 ng-bind="stage.name || '[new stage]'"></h4>
     </page-section>
     <render-if-feature feature="artifacts">
       <page-section key="producesArtifacts" label="Produces Artifacts" visible="stageProducesArtifacts()">
-        <produces-artifacts stage="stage" pipeline="pipeline"></produces-artifacts>
+        <produces-artifacts-react stage="stage" pipeline="pipeline" on-produces-changed="producesArtifactsChanged" />
```
sbwsg: Suggest wrapping this in a feature flag so that unflagged users get the existing experience.

app/scripts/modules/core/src/domain/IArtifact.ts
```diff
@@ -1,13 +1,14 @@
...
   name?: string;
   version?: string;
   location?: string;
   reference?: string;
   metadata?: any;
   artifactAccount?: string;
   provenance?: string;
+  // Legacy artifacts properties
+  kind?: string; // TODO delete
+  customKind?: boolean; // TODO delete
```
sbwsg: Is the intention in this PR to remove custom kinds or not? There are lots of places where it looks like the functionality has been pulled out but others where it looks like we're trying to maintain it. I would like to see them remain until 1.13's release so that there is ample time to give clear notice to anyone who is currently using them.

sbwsg:
Looks like the unflagged experience is still a bit broken. In this screenshot the GCS editable input where I would enter the bucket's `gs://` URL is missing from the inline editor:

screen shot 2019-03-04 at 11 34 57 am

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
```diff
@@ -70,9 +70,8 @@ class ArtifactCtrl implements IController {
     if (this.artifactAccounts) {
       options = options.filter(o => {
         const isCustomArtifact = o.customKind;
-        const isPublic = !!o.isPubliclyAccessible;
         const hasCredential = this.artifactAccounts.find(a => a.types.includes(o.type));
-        return isCustomArtifact || isPublic || hasCredential;
```
sbwsg: I'm a bit confused about the scope of changes.  If we're flagging the new artifact experience why do removals like this remain?

sbwsg:
Looks like the unflagged experience is still a bit broken. In this screenshot the GCS editable input where I would enter the bucket's `gs://` URL is missing from the inline editor:

screen shot 2019-03-04 at 11 34 57 am

Additionally looks as though adding an HTTP default artifact causes console errors with no input field showing up:

screen shot 2019-03-04 at 11 46 06 am

screen shot 2019-03-04 at 11 47 01 am

app/scripts/modules/core/src/pipeline/config/PipelineRegistry.ts
```diff
@@ -109,10 +118,6 @@ export class PipelineRegistry {
     return cloneDeep(this.artifactKinds.filter(k => k.isDefault));
   }
 
-  public getCustomArtifactKind(): IArtifactKindConfig {
```
sbwsg: Not 100% clear why this needs to be removed when the new experience is intended to be flagged.

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
```diff
@@ -96,20 +98,23 @@ module.exports = angular
...
+        $scope.options.dependencies = availableDependencyStages.map(function (stage) {
           return {
             name: stage.name,
             refId: stage.refId,
             available: true,
           };
         });
 
-        $scope.pipeline.stages.forEach(function(stage) {
+        $scope.pipeline.stages.forEach(function (stage) {
```
sbwsg: There are a lot of style changes throughout this PR that add noise to the diff, many of which are not consistent with the rest of the Deck codebase.  Please try to avoid mixing large feature modifications alongside lints.

sbwsg:
I think I see the issue with the http default artifact - it looks like it shouldn't be presented there at all; it doesn't appear on master.

app/scripts/modules/core/src/domain/IArtifact.ts
jkschneider: just indicating that customKind will be removed with 1.14. not aware of anywhere that previously referenced customKind that doesn't now...

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts
jkschneider: because with https://github.com/spinnaker/clouddriver/pull/3424, isPublic is not needed.

app/scripts/modules/core/src/pipeline/config/PipelineRegistry.ts
jkschneider: it was completely unused except in tests

jkschneider: effectively always returned `undefined` in production code

jkschneider:
> I think I see the issue with the http default artifact - it looks like it shouldn't be presented there at all; it doesn't appear on master.
yes, i suppose we are going to have to add an angular default http editor :/

app/scripts/modules/core/src/artifact/expectedArtifact.service.ts
```diff
@@ -104,13 +102,13 @@ export class ExpectedArtifactService {
 
   public static getKindConfig(artifact: IArtifact, isDefault: boolean): IArtifactKindConfig {
     if (artifact == null || artifact.customKind || artifact.kind === 'custom') {
-      return Registry.pipeline.getCustomArtifactKind();
```
sbwsg: Given I have an artifact that is already defined with a `customKind` field or a `kind` set to `'custom'`, this change results in my artifact being dropped from the expected-artifacts section in the config UI. This happens because `app/scripts/modules/core/src/pipeline/config/triggers/artifacts/artifact.component.ts` calls out to `getKindConfig()` to determine which template to render for a given artifact.

sbwsg: Before:

screen shot 2019-03-04 at 12 43 29 pm

After:

screen shot 2019-03-04 at 12 42 10 pm

app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
```diff
@@ -85,24 +86,24 @@ export class ExpectedArtifactEditor extends React.Component<
...
-    expectedArtifact.displayName = newName;
-    this.setState({ expectedArtifact });
-  };
-
   private onKindChange = (kind: IArtifactKindConfig) => {
     const expectedArtifact = cloneDeep(this.state.expectedArtifact);
     expectedArtifact.matchArtifact.type = kind.type;
     // kind is deprecated; remove it from artifacts as they are updated
     delete expectedArtifact.matchArtifact.kind;
-    expectedArtifact.matchArtifact.customKind = kind.customKind;
```
sbwsg: note: `customKind` in use here.

app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
```diff
@@ -113,21 +114,38 @@ export class ExpectedArtifactEditor extends React.Component<
...
+        type: artifact.type,
+      });
     this.setState({ expectedArtifact, account });
   };
 
   private availableKinds = () => {
     const kinds = this.props.kinds || [];
     const accounts = this.props.accounts || [];
     if (this.props.showAccounts) {
-      return kinds.filter(k => k.customKind || accounts.find(a => a.types.includes(k.type)));
```
sbwsg: `customKind` still in use here.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/custom/CustomArtifactEditor.tsx
```diff
@@ -1,82 +1,107 @@
...
-        <label className={labelClassName + ' sm-label-right'}>Reference</label>
-        <div className={fieldClassName}>{input(artifact, 'reference', onChange)}</div>
-      </div>
+export const CustomMatch: IArtifactKindConfig = {
+  label: 'Custom',
+  typePattern: /UNMATCHABLE/,
+  description: 'A custom-defined artifact.',
+  key: 'custom',
+  isDefault: true,
+  isMatch: true,
```
sbwsg: Custom Artifact is not currently selectable in the expected-artifact selector component in the config UI.

This artifact kind config needs a `customKind: true` field to workaround this issue.  It appears to have been lost in the rewrite from angular to react.

app/scripts/modules/core/src/pipeline/config/stages/stage.html
jkschneider: Done in 4dff56206.

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
jkschneider: As far as I can tell, it only happened in this file. Reverted style changes in 4dff56206.

app/scripts/modules/core/src/pipeline/config/PipelineRegistry.ts
jkschneider: Restored in 5ab4885a4.

app/scripts/modules/core/src/artifact/expectedArtifact.service.ts
jkschneider: Restored in 5ab4885a4. There was more to do afterward. Turns out that the old angular custom editor was calling a special function `Registry.pipeline.registerCustomKind`. That was essential to the custom editor showing up here. Since I'm suggesting in 1.14 that we do away with the custom editor entirely, I've just removed the new Custom react component and reinstated the angular component.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/custom/CustomArtifactEditor.tsx
jkschneider: Restored in 5ab4885a4.

app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
jkschneider: Restored in 1f97ac2a4.

app/scripts/modules/core/src/artifact/react/ExpectedArtifactEditor.tsx
jkschneider: Restored in 1f97ac2a4.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/docker/docker.artifact.ts
```diff
@@ -1,26 +1,25 @@
...
 module(DOCKER_ARTIFACT, []).config(() => {
-  Registry.pipeline.registerArtifactKind({
+  Registry.pipeline.mergeArtifactKind({
     label: 'Docker',
+    typePattern: ArtifactTypePatterns.DOCKER_IMAGE,
     type: 'docker/image',
     isDefault: false,
     isMatch: true,
     // docker hub image artifacts can be bound to manifests without an associated artifact-account
-    isPubliclyAccessible: true,
```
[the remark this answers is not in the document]
jkschneider: Removed in 50dac9ec6.

jkschneider:
@sbwsg I added a default HTTP editor so that works now:

image

PR #6637, opened 2019-03-04
feat(cf): prefix apps manager and metrics urls with https://

Jammy-Louie:


app/scripts/modules/cloudfoundry/src/serverGroup/details/sections/ApplicationManagerSection.tsx
```diff
@@ -10,12 +10,16 @@ export class ApplicationManagerSection extends React.Component<ICloudFoundryServ
...
     const { serverGroup } = this.props;
+    const { appsManagerUri } = serverGroup;
     return (
       <>
-        {serverGroup.appsManagerUri && (
+        {appsManagerUri && (
           <CollapsibleSection heading="Apps Manager" defaultExpanded={true}>
             <div>
-              <a href={serverGroup.appsManagerUri} target="_blank">
+              <a
```
archana-s: Quick suggestion here: 
You could replace `https://` with just `//` which will fallback to the protocol being used. 

You could do something like this too: 
```
const pattern = /^http(s)*.*/;
pattern.test(appsManagerUri) ? appsManagerUri : `//${appsManagerUri}`;
```

jmelchio: I agree with your suggestion that this could be made more generic but we don't really _*want*_ people to use plain `http` connections to the CF foundation they are connecting to.

PR #6644, opened 2019-03-05
fix(core): Remove configure button and setup redirect for mptv2 pipeline

louisjimenez:
https://github.com/spinnaker/spinnaker/issues/4056

Removes configure as an option for pipelines from mptv2 templates:
image

Removes mptv2 pipelines from the Configure menu in the "Pipelines" view:
image

Removes the configure button from the single execution view if a execution has a mptv2 pipeline config:
image

Redirects to "Pipelines" view if user navigates to mptv2 pipeline config directly.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/details/singleExecutionDetails.less
+++ b/app/scripts/modules/core/src/pipeline/details/singleExecutionDetails.less
@@ -1,5 +1,3 @@
-.single-execution-details {
-  .checkbox {
-    padding: 0 1em;
-  }
+.single-execution-details__configure {
+  margin: 0 5px 0 1em;
 }
--- a/app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
+++ b/app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
@@ -55,6 +55,8 @@ export class ExecutionGroup extends React.Component<IExecutionGroupProps, IExecu
     const pipelineConfig = find(this.props.application.pipelineConfigs.data, {
       name: this.props.group.heading,
     }) as IPipeline;
+
+    const hasNonMPTV2PipelineConfig = pipelineConfig && pipelineConfig.schema !== 'v2';
     const sectionCacheKey = this.getSectionCacheKey();
 
     this.state = {
@@ -66,7 +68,7 @@ export class ExecutionGroup extends React.Component<IExecutionGroupProps, IExecu
         CollapsibleSectionStateCache.isExpanded(sectionCacheKey),
       poll: null,
       canTriggerPipelineManually: !!pipelineConfig,
-      canConfigure: !!(pipelineConfig || strategyConfig),
+      canConfigure: !!(hasNonMPTV2PipelineConfig || strategyConfig),
       showAccounts: ExecutionState.filterModel.asFilterModel.sortFilter.groupBy === 'name',
       pipelineConfig,
       showOverflowAccountTags: false,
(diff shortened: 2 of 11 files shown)


app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -66,7 +68,7 @@ export class ExecutionGroup extends React.Component<IExecutionGroupProps, IExecu
         CollapsibleSectionStateCache.isExpanded(sectionCacheKey),
       poll: null,
       canTriggerPipelineManually: !!pipelineConfig,
-      canConfigure: !!(pipelineConfig || strategyConfig),
+      canConfigure: !!(hasNonMPTV2PipelineConfig || strategyConfig),
```
louisjimenez: Will also lead to "Start Manual Execution" not being rendered for mptv2 pipelines. The UI does not currently support kicking off mptv2 pipeline executions so this works out. Going to follow up this PR with some work that addresses other parts of the application that incorrectly allow starting manual executions for mptv2 pipelines

app/scripts/modules/core/src/pipeline/details/singleExecutionDetails.less
```diff
@@ -1,5 +1,3 @@
-.single-execution-details {
-  .checkbox {
-    padding: 0 1em;
-  }
+.single-execution-details__configure {
+  margin: 0 5px 0 1em;
```
sbwsg: This ends up having the same effect as the replaced `.single-execution-details .checkbox` style rule?

app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -55,6 +55,8 @@ export class ExecutionGroup extends React.Component<IExecutionGroupProps, IExecu
     const pipelineConfig = find(this.props.application.pipelineConfigs.data, {
       name: this.props.group.heading,
     }) as IPipeline;
+
+    const hasNonMPTV2PipelineConfig = pipelineConfig && pipelineConfig.schema !== 'v2';
```
maggieneterval: since there are three different places we're checking `config.schema` against "v2", maybe abstract into some sort of helper function or at least throw v2 into a constant?

maggieneterval: if it's easier to keep the way it is i don't really have a strong opinion about this

app/scripts/modules/core/src/pipeline/details/singleExecutionDetails.less
louisjimenez: The aim is to have the right padding of 1em come off of the checkbox and instead have the configure button use it as 1em left margin. When the checkbox is alone it can sit flush on the right side and when the configure button is present it can maintain 1em spacing. What are you thoughts on it ? 😄 
image
image

sbwsg: 👍

app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
louisjimenez: Great point, i'll dry this up

louisjimenez: updated, https://github.com/spinnaker/deck/pull/6644/commits/292cb5b53f1e600fadb819e704058b896e63c0b3

louisjimenez:

PR #6645, opened 2019-03-06
fix({core,amazon}/serverGroup): filter out empty tags, change 'tags' field type

erikmunson:
When messing with AWS deploy configs people sometimes click to add a tag but never populate it, and then end up saving it into the config. Example:

screen shot 2019-03-05 at 9 46 24 pm

Thing is, the task blows up because you can't actually set an empty tag, and at that point the app owner has no memory of adding an empty field in a form somewhere (maybe days/weeks/months prior). We could handle this better on the API side too (it gets all the way to Amazon and errors out there), but that empty tag is more like ephemeral UI state in this context and really ought to disappear when you leave the modal.

I also discovered that the `tags` field on `IServerGroupCommand` is defined as `IEntityTags`, not an object of strings. Based on some searching it doesn't seem like any code actually treats the `tags` field that way 🤷‍♂️.


app/scripts/modules/amazon/src/serverGroup/configure/wizard/AmazonCloneServerGroupModal.tsx
```diff
@@ -149,6 +160,7 @@ export class AmazonCloneServerGroupModal extends React.Component<
   }
 
   private submit = (command: IAmazonServerGroupCommand): void => {
+    this.normalizeCommand(command);
```
erikmunson: I went back and forth for a minute on putting this in `convertServerGroupCommandToDeployConfiguration`, but IMO empty keys are ephemeral component/view state that should be scrubbed before they hit any transformer/writer code. Happy to be convinced otherwise.

archana-s:
I feel like we should not store the values of this field and pass it if its empty.  Handling it in the form field file might feel cleaner: `ServerGroupAdvancedSettingsCommon.tsx`

app/scripts/modules/amazon/src/serverGroup/configure/wizard/AmazonCloneServerGroupModal.tsx
```diff
@@ -141,6 +141,17 @@ export class AmazonCloneServerGroupModal extends React.Component<
...
 
+  private normalizeCommand = ({ tags }: IAmazonServerGroupCommand) => {
+    if (!tags) {
+      return;
+    }
+    Object.keys(tags).forEach(key => {
+      if (!key.length && !tags[key].length) {
+        delete tags[key];
+      }
+    });
```
erikmunson: @jrsquared do you know whether this field needs reference equality for angular shenanigans once it hits the submit flow? I wrote it this way under that anxiety but if possible would love to generate a new `tags` object.

jrsquared: I do not _believe_ so, but I think I left it this way under the same anxiety.

erikmunson: glad to hear we're both entirely driven by FUD, will leave it as-is 😉

erikmunson:
@archana-s I agree with the philosophy of pushing ephemeral state as close to the field it comes from as possible, and if the stage config code this dialog interacts with was all React I would likely do that. In this case, however, the server group dialog is architected to think of the `command` that gets passed to Formik as a big, mutable source of truth for all the things happening in the form that Angular code then relies on to do some validation/other stage config stuff. As a result we typically do sanitization and transformation when the form is done using the command. Since we're already working in that architecture/mindset I'm skeptical of adding some kind of separate state inside the child component that tries to mutably keep a single `tags` object in sync with some other component local state.

Will hopefully not need all these awkward patterns once most/all of stage configuration is using React.

PR #6649, opened 2019-03-06
fix(core): titus run jobs override all other providers (#6647)

sbwsg:
Cherry-pick of https://github.com/spinnaker/deck/pull/6647

Since the Titus Run Job stage is written in React its provider config
includes a component field. stage.module.js responds to a component
field by rendering it into the DOM.

When multiple providers are available for a given stage (for example
with the Run Job stage) the baseProviderStage, an angular component, is
meant to be shown to the user for them to pick one.

Unfortunately the rendering of a React component from a provider config
short circuits the rendering of the baseProviderStage.

Resolving this first problem requires nulling out the `component` field
when baseProviderStage needs to be shown.

The follow-on issue is that baseProviderStage did not previously know
how to render react components. This meant that the user could pick a
Run Job stage, then choose the Titus provider, but they wouldn't see any
stage rendered.

The solution I've implemented is the fewest LOC change I could make - it
duplicates the rendering logic from `stage.module.js` into
baseProviderStage and threads the constructed props object from
stage.module.js down into the baseProviderStage.

The alternatives proposed were: 1. To try and split up baseProviderStage
or 2. Try to merge baseProviderStage into stage.module.js. I tried both
of these but didn't get either solution into a working state and ended
up with a huge mess of changed lines on my hands.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/baseProviderStage/baseProviderStage.html
+++ b/app/scripts/modules/core/src/pipeline/config/stages/baseProviderStage/baseProviderStage.html
@@ -1,8 +1,12 @@
 <div class="form-horizontal" ng-if="!viewState.loading">
   <stage-config-field label="Provider" ng-if="providers.length > 1">
-      <provider-selector providers="providers"
-                         component="stage" field="cloudProviderType"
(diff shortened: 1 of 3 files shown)


louisjimenez:
LGTM

PR #6654, opened 2019-03-08
feat(core): save pipelines stage was added

claymccoy:
Adds a new stage that can extract pipelines from an artifact and save them.
This exposes spinnaker/orca#2712 in the UI.


app/scripts/modules/core/src/pipeline/config/stages/savePipelines/savePipelinesConfig.controller.ts
```diff
@@ -0,0 +1,38 @@
+import { IController, IScope } from 'angular';
+import { defaults } from 'lodash';
+
+export class SavePipelinesConfigCtrl implements IController {
```
jkschneider: I think we should avoid net new angular code. Take a look at `EvaluateVariablesStageConfig` for an example of a React-based stage that doesn't require an angular bridge.

claymccoy: I started with one like that as an example, but abandoned it because stage-artifact-selector-react wasn't working correctly in the react HTML, and the only example of using stage-artifact-selector-react is like this.

claymccoy: yes, I'll update it to look like that

app/scripts/modules/core/src/pipeline/pipeline.module.ts
```diff
@@ -20,6 +20,7 @@ import './config/stages/gremlin/gremlinStage';
 import { GROUP_STAGE_MODULE } from './config/stages/group/groupStage.module';
 import { MANUAL_JUDGMENT_STAGE_MODULE } from './config/stages/manualJudgment/manualJudgmentStage.module';
 import { RESIZE_ASG_STAGE } from './config/stages/resizeAsg/resizeAsgStage';
+import { SAVE_PIPELINES_STAGE } from 'core/pipeline/config/stages/savePipelines/savePipelinesStage';
```
jkschneider: Just `import from 'core/pipeline/config/stages/savePipelines/savePipelinesStage';`

app/scripts/modules/core/src/pipeline/pipeline.module.ts
```diff
@@ -91,6 +92,7 @@ module(PIPELINE_MODULE, [
   PRODUCES_ARTIFACTS,
   RESIZE_ASG_STAGE,
   require('./config/stages/runJob/runJobStage.module').name,
+  SAVE_PIPELINES_STAGE,
```
jkschneider: Omit this.

app/scripts/modules/core/src/pipeline/config/stages/savePipelines/savePipelinesStage.ts
```diff
@@ -0,0 +1,26 @@
...
+module(SAVE_PIPELINES_STAGE, []).config(() => {
+  Registry.pipeline.registerStage({
+    label: 'Save Pipelines',
+    description: 'Saves pipelines defined in an artifact.',
+    key: 'savePipelinesFromArtifact',
+    component: SavePipelinesStageConfig,
+    executionDetailsSections: [ExecutionDetailsTasks, ExecutionArtifactTab],
+    producesArtifacts: false,
+    defaultTimeoutMs: 30 * 60 * 1000, // 30 minutes
+    validators: [],
```
jkschneider: This is the default

app/scripts/modules/core/src/pipeline/config/stages/savePipelines/savePipelinesStage.ts
```diff
@@ -0,0 +1,26 @@
...
+export const SAVE_PIPELINES_STAGE = 'spinnaker.core.pipeline.stage.savePipelinesStage';
+
+module(SAVE_PIPELINES_STAGE, []).config(() => {
+  Registry.pipeline.registerStage({
+    label: 'Save Pipelines',
+    description: 'Saves pipelines defined in an artifact.',
+    key: 'savePipelinesFromArtifact',
+    component: SavePipelinesStageConfig,
+    executionDetailsSections: [ExecutionDetailsTasks, ExecutionArtifactTab],
+    producesArtifacts: false,
```
jkschneider: This is the default

app/scripts/modules/core/src/pipeline/config/stages/savePipelines/SavePipelinesStageConfig.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+
+import { IStageConfigProps, StageConfigField } from 'core/pipeline';
+import { IArtifact, IExpectedArtifact, StageArtifactSelector } from '@spinnaker/core';
+
+export class SavePipelinesStageConfig extends React.Component<IStageConfigProps> {
```
jkschneider: Make `React.SFC`

app/scripts/modules/core/src/pipeline/config/stages/savePipelines/savePipelinesStage.ts
```diff
@@ -0,0 +1,26 @@
...
+  ExecutionDetailsTasks,
+  ExpectedArtifactService,
+  Registry,
+} from '@spinnaker/core';
+import { SavePipelinesStageConfig } from 'core/pipeline/config/stages/savePipelines/SavePipelinesStageConfig';
+import { module } from 'angular';
+
+export const SAVE_PIPELINES_STAGE = 'spinnaker.core.pipeline.stage.savePipelinesStage';
+
+module(SAVE_PIPELINES_STAGE, []).config(() => {
```
jkschneider: You don't need to wrap this in an angular module at all.

app/scripts/modules/core/src/pipeline/config/stages/savePipelines/savePipelinesStage.ts
```diff
@@ -0,0 +1,23 @@
+import {
+  ArtifactReferenceService,
+  ExecutionArtifactTab,
+  ExecutionDetailsTasks,
+  ExpectedArtifactService,
+  Registry,
+  SETTINGS,
+} from '@spinnaker/core';
```
jkschneider: core code should not refer to `@spinnaker/core`. use a relative path instead.

PR #6657, opened 2019-03-08
chore(provider/azure): Enable the firewall work and add source IP/CIDR filtering

mybayern1974:
Currently the Azure Network Security Group (NSG for short), which plays the role of Spinnaker firewall, does not work besides being provisioned. This PR includes:

1. Users can bind the NSG to a subnet where the load balancer is located, thus the NSG could take real effects;
2. Add filters against IP/CIDR with format validation
3. Fix some tiny UI bugs

image


mybayern1974:
This PR has a feature link PR: https://github.com/spinnaker/clouddriver/pull/3439

PR #6661, opened 2019-03-08
fix(artifacts): Correct render-if-feature for new artifacts on stage …

jkschneider:
Now the artifact definition component shows up:

image


claymccoy:
Lgtm

PR #6663, opened 2019-03-11
feat(artifacts): Re-use artifacts when re-running a pipeline

ezimanyi:
Closes spinnaker/spinnaker#3748

* refactor(artifacts): Rename `ArtifactList` to `ResolvedArtifactList`
 This component is specific to a list of resolved artifacts, and requires the pipeline's expected artifacts as input. Rename it to free up ArtifactList for a more generic list of artifacts.

* refactor(artifacts): Create a react component for an artifact list
    
  The logic for rendering a single artifact is currently embedded in `ResolveArtifactList`; pull this into its own component so it can be reused and create a new `ArtifactList` component that generically displays a list of artifacts, without expected artifact logic.

* feat(artifacts): Re-use artifacts when re-running a pipeline
    
  The manual execution widget does not currently support injecting artifacts into a pipeline. This means that pipelines that depend on artifacts cannot be manually started; support this workflow for the re-run pipeline case by re-sending the same artifacts as the initial run. These will show up as a read-only list of artifacts in the Confirm Execution dialog.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/status/Artifact.spec.tsx
+++ b/app/scripts/modules/core/src/pipeline/status/Artifact.spec.tsx
@@ -2,15 +2,17 @@ import * as React from 'react';
 import { ShallowWrapper, shallow } from 'enzyme';
 import { mock } from 'angular';
 import { REACT_MODULE } from 'core/reactShims';
+
 import { IArtifact } from 'core/domain';
-import { Artifact, IArtifactProps, IArtifactState } from './Artifact';
+
+import { Artifact, IArtifactProps } from './Artifact';
 
 const ARTIFACT_TYPE = 'docker/image';
 const ARTIFACT_NAME = 'example.com/container';
 const ARTIFACT_REFERENCE = 'docker.io/example.com/container:latest';
 
 describe('<Artifact/>', () => {
-  let component: ShallowWrapper<IArtifactProps, IArtifactState>;
+  let component: ShallowWrapper<IArtifactProps>;
 
   beforeEach(mock.module(REACT_MODULE));
   beforeEach(mock.inject(() => {})); // Angular is lazy.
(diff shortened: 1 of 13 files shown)


app/scripts/modules/core/src/pipeline/status/Artifact.spec.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import { ShallowWrapper, shallow } from 'enzyme';
+import { mock } from 'angular';
+import { REACT_MODULE } from 'core/reactShims';
```
maggieneterval: Typically third-party imports, core imports, and local-module imports are newline-separated per this hard-to-find conventions doc: https://github.com/spinnaker/deck/wiki/Conventions#imports

app/scripts/modules/core/src/pipeline/status/Artifact.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import './artifact.less';
+
+export interface IArtifactProps {
+  artifact: IArtifact;
+  isDefault?: boolean;
+  sequence?: number;
+}
+
+export interface IArtifactState {}
```
maggieneterval: Can probably just omit this empty interface, and the constructor

app/scripts/modules/core/src/pipeline/status/Artifact.spec.tsx
ezimanyi: Thanks!  I had no idea about those conventions---I fixed all the imports I touch to follow them

app/scripts/modules/core/src/pipeline/status/Artifact.tsx
ezimanyi: Great point, I didn't realize you could omit state entirely, fixed

app/scripts/modules/core/src/pipeline/status/Artifact.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export class Artifact extends React.Component<IArtifactProps, IArtifactState> {
```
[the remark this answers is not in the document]
ezimanyi: Done, much less verbose now

ezimanyi:
image

PR #6667, opened 2019-03-11
fix(appengine): allow sequence to be suppressed in servergroup names

sbwsg:
Twinned with Clouddriver PR https://github.com/spinnaker/clouddriver/pull/3453 

Exposes a checkbox to let the user suppress the sequence number normally added to a server group when deployed:

Screen Shot 2019-03-11 at 5 45 45 PM

And has a tooltip that explains how very not-normal it is to use this checkbox:

Screen Shot 2019-03-11 at 5 45 49 PM


sbwsg:

jianglai:
Tested and working. However I'd like to know how much work is there to support a user-supplied version string instead of hardcoding it to the application name?

PR #6668, opened 2019-03-11
fix(core): Surface invalid params for a pipeline stage

archana-s:
Parameters not accepted by a pipeline are highlighted in a pipeline stage. 

The stage config can end up having parameters that are no longer needed/accepted by the pipeline stage. Having them around might lead to issues during executions and silently too. This difff adds a message with all params that the stage is going to pass to the pipeline that are not accepted by the pipeline execution. It also removes them all if the user chooses to. Image attached below.

image


app/scripts/modules/core/src/pipeline/config/triggers/artifacts/triggerArtifactConstraintSelector.less
```diff
@@ -12,7 +12,7 @@
 }
 
 .artifact-select .Select-arrow-zone {
-  border-left: 1px solid rgb(204, 204, 204);
+  border-left: 1px solid var(--color-alto);
```
archana-s: @sbwsg This gets rid of the lint warning. Hope this ok. (They are the same color)

archana-s:
> very nice. can we do the same for the jenkins stage?

yup, upcoming.

PR #6670, opened 2019-03-11
fix(artifacts): Simplify display name generation

ezimanyi:
Fixes spinnaker/spinnaker#4097

* fix(artifacts): Use artifact type to determine readable name

  The logic for getting a human-readable name from an artifact currently uses the kind field, which is deprecated and not populated for new artifacts. Switch to using 'type' which is equivalent.

* fix(artifacts): Simplify display name generation

  The current logic to auto-generate a display name for artifacts without one assumes that the user will enter a fully-qualified path for each artifact type. If this is not true, an unhelpful (and often duplicate) name such as GCS-no-bucket/no-file is generated.

  While we could make the logic more complex, it seems easier to maintain and just as (if not more) useful to just default the display name to the name.


ezimanyi:
Interested particularly in thoughts about my change to make things simpler and just default the display name to the name; as I dug into it, there was a lot of logic to just transform `gs://abc/def` to `GCS-abc/def` (and likewise for other artifacts) and rather than make it even more complex to deal with edge cases, just dropping the name in as the default seemed just as good and much easier to maintain.

jkschneider:
Great, sounds good.

PR #6673, opened 2019-03-12
fix(core): In baseProvider, fixed autoselection of React stage

alanmquach:
This doesn't feel great to me but it fixes things.

The problem was that when

1. There is only a single cloud provider configured for a stage that supports multiple
2. That stage is a React stage

We get an `Invariant Violation: Target container is not a DOM element.` because we try to immediately auto-select the only cloud provider, but the `.react-stage-details` div has not been mounted yet.

This defers the React render until after the digest cycle completes. Not great, but I think this can probably hold us over until we port the `baseProvider` code over to React (soon).


alanmquach:
@sbwsg Anytime. Does this need to be cherry-picked to the other releases?

sbwsg:
Ah yes good call.

sbwsg:

PR #6684, opened 2019-03-13
feat(core/deploy): UI widget for `Delay Before Scale Down`

ajordens:
image

### DIFF
--- a/app/scripts/modules/core/src/config/settings.ts
+++ b/app/scripts/modules/core/src/config/settings.ts
@@ -55,7 +55,6 @@ export interface IFeatures {
   roscoMode?: boolean;
   snapshots?: boolean;
   travis?: boolean;
-  triggerViaEcho?: boolean;
   versionedProviders?: boolean;
   wercker?: boolean;
 }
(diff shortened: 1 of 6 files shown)


ajordens:
Basically a c&p of what normal red/black has and rolling red/black was missing.

I did add a bit of inline margining which could come out if we'd like consistent squishiness across the board.

image

app/scripts/modules/core/src/deploymentStrategy/strategies/rollingredblack/AdditionalFields.tsx
```diff
@@ -44,6 +46,11 @@ export class AdditionalFields extends React.Component<IRollingRedBlackStrategyAd
     this.forceUpdate();
   };
 
+  private handleChange = (key: string, value: any) => {
```
erikmunson: did typing `value` as a string blow up on you?

ajordens: 🤷‍♀️ This was totally c&p from the redblack AdditionalFields.

I'll give it a whirl.

ajordens: Gave it the `value: string` treatment

PR #6687, opened 2019-03-14
feat(ecs): docker image selection

clareliguori:
Enables deploying a specific Docker image from a registered Docker registry, or a trigger or baked image

Fixes https://github.com/spinnaker/spinnaker/issues/3455

Screenshot from 2019-03-13 11-39-54


clareliguori:
Dependent orca PR: https://github.com/spinnaker/orca/pull/2751

christopherthielen:
I don't have the context around this addition, but I'd like to point out that the UI we built for Titus looks like this: 


"Select from list" shows indexed images:
Screen Shot 2019-03-13 at 5 37 07 PM

Screen Shot 2019-03-13 at 5 38 37 PM

"Manually" is free form and assumes the user knows what they are doing
Screen Shot 2019-03-13 at 5 37 17 PM



I'm curious, how does "Images from triggers" work?

clareliguori:
@christopherthielen the UI in this PR is largely consistent with the Docker image selectors for kubernetes and dcos providers, and eventually I'll add in Docker artifact support like the kubernetes one has.

The UI you showed looks close to the Docker Triggers selector.  The actual tag of the image that triggered the pipeline (on a docker push) is the one that will get deployed when that's selected in the server group.

Screenshot from 2019-03-13 17-48-53

anotherchrisberry:
Are we good with this? Happy to merge it and the dependent Orca PR...

clareliguori:
@anotherchrisberry I think we're good?

PR #6688, opened 2019-03-14
feat(provider/azure): Add availability zones support in server group …

michaeljqzq:
Add availability zones support in server group configuration


michaeljqzq:
Linked with PR in `clouddriver` https://github.com/spinnaker/clouddriver/pull/3467

app/scripts/modules/azure/serverGroup/configure/serverGroupConfiguration.service.spec.js
```diff
@@ -164,5 +164,33 @@ describe('Service: azureServerGroupConfiguration', function() {
...
+        regionsSupportZones: [],
+        availabilityZones: ['1', '2', '3'],
+      };
+
+      service.configureZones(this.command);
+
+      expect(this.command.backingData.filtered.zones).toEqual([]);
+    });
+
+    it('returns actual zone list if region not supported', function() {
```
sbwsg: I think this description string is intended to end with "if region **is** supported".

In this test, `'eastasia'` is in the set of supported zones, right?

app/scripts/modules/azure/serverGroup/configure/serverGroupConfiguration.service.js
```diff
@@ -143,6 +143,22 @@ module.exports = angular
         return result;
       }
 
+      function configureZones(command) {
+        const result = { dirty: {} };
+        const filteredData = command.backingData.filtered;
+        if (command.region === null) {
```
sbwsg: Should this also return early if `command.region` is `undefined`?  If so I suggest changing this to:

```
if (!command.region) {
  return result;
}
```

app/scripts/modules/azure/serverGroup/configure/serverGroupConfiguration.service.spec.js
```diff
@@ -164,5 +164,33 @@ describe('Service: azureServerGroupConfiguration', function() {
...
+        regionsSupportZones: ['eastasia'],
+        availabilityZones: ['1', '2', '3'],
+      };
+
+      service.configureZones(this.command);
+
+      expect(this.command.backingData.filtered.zones).toEqual(
+        this.command.backingData.credentialsKeyedByAccount[this.command.credentials].availabilityZones,
+      );
+    });
```
sbwsg: I suggest adding a test for behaviour expected when `this.command.region` is null or undefined.

app/scripts/modules/azure/serverGroup/configure/serverGroupConfiguration.service.spec.js
michaeljqzq: you're right, made mistake here

michaeljqzq:
> I have added a few suggestions. Once those are addressed or rejected I will merge.

Thanks @sbwsg , resolved these suggestions

PR #6691, opened 2019-03-14
fix(kubernetes): Use apiGroup when looking up deploy status for CRDs

willgorman:
Part of the fix for spinnaker/spinnaker#3387.
Relates to spinnaker/clouddriver#3003.  Without this change, the
status pills (stable, available, etc) would not be found for CRD
manifests after applying that patch to clouddriver (and vice-versa
so these should be released together).


app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/react/DeployStatus.tsx
```diff
@@ -9,6 +9,12 @@ import {
 import { KubernetesManifestService } from 'kubernetes/v2/manifest/manifest.service';
 import { ManifestStatus } from './ManifestStatus';
 
+// from https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/
+const BUILT_IN_GROUPS = ['', 'core', 'batch', 'apps', 'extensions', 'storage.k8s.io',
+  'apiextensions.k8s.io', 'apiregistration.k8s.io', 'policy', 'scheduling.k8s.io',
+  'settings.k8s.io', 'authorization.k8s.io', 'authentication.k8s.io', 'rbac.authorization.k8s.io',
+  'certifcates.k8s.io', 'networking.k8s.io']
+
```
maggieneterval: @willgorman hopefully this list is not expected to change frequently / get out of sync with the list in Clouddriver?

PR #6696, opened 2019-03-14
feat(kubernetes): add expression evaluation options to bake and deplo…

maggieneterval:
…y manifest stages

Corresponding Orca PR: https://github.com/spinnaker/orca/pull/2761

The primary purpose of this change is to prevent Deploy Manifest stage failure for users attempting to deploy third-party Helm manifests that include `${}`-wrapped expressions, while allowing SpEL overrides to be explicitly evaluated just prior to baking. These options should probably be deprecated when expression evaluation is refactored in Spinnaker (see this [proposal](https://github.com/spinnaker/spinnaker/issues/3500)).

- Adds `evaluateOverrideExpressions` option to Bake Manifest stage. This defaults to false. When it is set to true, SpEL expressions evaluation of manifest overrides will occur during the Bake stage.
- Adds `skipExpressionEvaluation` option to Deploy Manifest stage for manifest artifacts. This defaults to false. When it is set to true, SpEL expression of the manifest artifact will not occur during the Deploy stage.
- Important caveat: Orca attempts to evaluate expressions in the entire pipeline context multiple times in logic that does not belong to these stages (for example, [here](https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L167)). Therefore, there will still be "failed to evaluate" warnings for all `${}`-wrapped expressions not intended for Spinnaker.

Related issues:
- https://github.com/spinnaker/spinnaker/issues/2809
- https://github.com/spinnaker/spinnaker/issues/3028
- https://github.com/spinnaker/spinnaker/issues/3939
- https://github.com/spinnaker/spinnaker/issues/4109
- https://github.com/spinnaker/spinnaker/issues/3644
- https://github.com/spinnaker/spinnaker/issues/3335

Before:
2kX9K7zFvuJ

After:
2N55McgwoEf

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestConfig.html
+++ b/app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestConfig.html
@@ -41,6 +41,18 @@
         >
         </stage-artifact-selector-react>
       </stage-config-field>
+      <stage-config-field
+        help-key="kubernetes.manifest.skipExpressionEvaluation"
+        label="Expression Evaluation"
+        ng-if="ctrl.$scope.stage.source === ctrl.artifactSource"
+      >
+        <div class="checkbox">
+          <label>
+            <input type="checkbox" ng-model="ctrl.$scope.stage.skipExpressionEvaluation" />
+            Skip SpEL expression evaluation
+          </label>
+        </div>
+      </stage-config-field>
       <hr />
       <stage-config-field
         label="Req. Artifacts to Bind"
@@ -89,6 +101,18 @@
         on-save="ctrl.manifestArtifactController.onArtifactCreated"
       >
       </expected-artifact-editor-react>
+      <stage-config-field
+        help-key="kubernetes.manifest.skipExpressionEvaluation"
+        label="Expression Evaluation"
+        ng-if="ctrl.$scope.stage.source === ctrl.artifactSource"
+      >
+        <div class="checkbox">
+          <label>
+            <input type="checkbox" ng-model="ctrl.$scope.stage.skipExpressionEvaluation" />
+            Skip SpEL expression evaluation
+          </label>
+        </div>
+      </stage-config-field>
       <hr />
       <expected-artifact-multi-selector
         command="ctrl.$scope.stage"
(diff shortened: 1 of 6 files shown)


app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestConfig.html
```diff
@@ -41,6 +41,14 @@ <h4>Manifest Configuration</h4>
         >
         </stage-artifact-selector-react>
       </stage-config-field>
+      <stage-config-field label="Expression Evaluation" ng-if="ctrl.$scope.stage.source === ctrl.artifactSource">
```
ezimanyi: Do we only want to show the textbox if the manifest comes from an artifact?  I guess that's probably the most common use case, as if it's a text box, you can probably just not type SpEL in the first place...but just wondering if the checkbox should be there for both.

maggieneterval: That was my initial thought as well, but it seems misleading since as far as I can tell we are [only explicitly evaluating (and failing the stage on the failure to evaluate) SpEL expressions for manifest artifacts](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestTask.java#L85) in the Deploy Stage.

ezimanyi: Oh, good catch---I had no idea SpEL evaluation only happened for artifacts.  Makes sense to only show the checkbox in that case then.

maggieneterval:
@sbwsg Great idea on tooltip text, how does this sound?

Bake stage:
`
Explicitly evaluate SpEL expressions in overrides just prior to manifest baking. Can be paired with the "Skip SpEL evaluation" option in the Deploy Manifest stage when baking a third-party manifest artifact with expressions not meant for Spinnaker to evaluate as SpEL.
`

Deploy stage:
`
Skip SpEL expression evaluation of the manifest artifact in this stage. Can be paired with the "Evaluate SpEL expressions in overrides at bake time" option in the Bake Manifest stage when baking a third-party manifest artifact with expressions not meant for Spinnaker to evaluate as SpEL.
`

app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/deployManifestConfig.html
maggieneterval: Confusingly, I think it happens for text manifests as well, during one of the possibly many non-failing passes at expression evaluation that occur when the pipeline begins. We're _explicitly_ evaluating SpEL in artifact manifests just prior to deploy since the full manifest text may only be available _after_ those initial passes.

maggieneterval:
Screen grabs including new tooltips:

G95u6K0UQTE

Kw7gVNcRya7

sbwsg:
Nice, lgtm!

PR #6698, opened 2019-03-14
feat(core/deploy): UI widget for 'Delay Before Scale Down'

ajordens:


ajordens:
https://github.com/spinnaker/deck/pull/6684 was unhappy for one reason or another.

PR #6699, opened 2019-03-15
fix({core,cloudfoundry}/deploy): better red/black, rolling red/black help text

erikmunson:
Turns out that the language we use for help text about the various red/black strategy options gives people the impression that we'll _only_ disable/scale down the single Server Group we're 'replacing' by doing the deploy. This makes sense because, well, the wording uses the singular 'Server Group'. In reality though we'll disable/scale down _all_ the old server groups in the cluster.

I wanted to make this super explicit as it's counterintuitive and often confuses people so I also added 'in this cluster'.


app/scripts/modules/cloudfoundry/src/deploymentStrategy/strategies/redblack/AdditionalFields.tsx
```diff
@@ -47,7 +47,7 @@ export class AdditionalFields extends React.Component<IRedBlackStrategyAdditiona
         <div className="col-md-12 form-inline">
           <label>
             Wait Before Disable &nbsp;
-            <HelpField content="Time to wait before disabling old server group" />
+            <HelpField content="Time to wait before disabling all old server groups in this cluster" />
```
erikmunson: @jkschneider you cool with this?

app/scripts/modules/core/src/deploymentStrategy/strategies/rollingredblack/AdditionalFields.tsx
```diff
@@ -103,7 +103,7 @@ export class AdditionalFields extends React.Component<IRollingRedBlackStrategyAd
         <div className="col-md-12 form-inline">
           <label>
             Wait
-            <HelpField content="Time to wait before disabling instances in old server group" />
+            <HelpField content="Time to wait before disabling instances in all old server groups" />
```
erikmunson: is the behavior of these options meaningfully different enough that it's valuable to say 'instances in'? the wording is a little awkward and adding 'in this cluster' felt like it would make the sentence even more awkward, so i left that part out for RRB.

app/scripts/modules/cloudfoundry/src/deploymentStrategy/strategies/redblack/AdditionalFields.tsx
jkschneider: Yes, I think that is more clear. Thanks!

app/scripts/modules/core/src/deploymentStrategy/strategies/rollingredblack/AdditionalFields.tsx
jmelchio: Would it make sense to put these strings in `core/src/help/help.contents.ts` while this is being changed?

mybayern1974:
Hi all here, especially the initiator @erikmunson, by reading the words "Disable all old server groups" in Red/Black, will there be down time in the story? Per my reading the code of RB in orca, the sequence of RB is:
Step 1: Deploy new SG;
Step 2: Shrink Cluster;
Step 3: Disable old SGs
After step 3, will there be service down time? The only point i can think of to avoid downtime is Step 1 covers the behavior of Enabling(or to say activating) the newly created SG thus the new SG serves the service all the time during those post shrink/disable things happened. Is my guess correct?

dreynaud:
That is correct, in step 1 the new server group is created enabled by
default, hence no downtime

mybayern1974:
> That is correct, in step 1 the new server group is created enabled by
> default, hence no downtime

Thank you @dreynaud for the confirmation! May I further ask one question: is it a recommendation (or even a hard requirement) to deallocate (or to say deactivate) all instances/nodes within a server group in the DISABLE SERVER GROUP stage? I'm asking this because I saw in code some cloud providers made this while some seem not.

dreynaud:
> Thank you @dreynaud for the confirmation! May I further ask one question: is it a recommendation (or even a hard requirement) to deallocate (or to say deactivate) all instances/nodes within a server group in the DISABLE SERVER GROUP stage? I'm asking this because I saw in code some cloud providers made this while some seem not.

I am not necessarily very familiar with other cloud providers. No hard requirements that I can think of, other that we do want to make sure that there are no instances taking traffic in disabled server groups.

mybayern1974:
> Thank you @dreynaud for the confirmation! May I further ask one question: is it a recommendation (or even a hard requirement) to deallocate (or to say deactivate) all instances/nodes within a server group in the DISABLE SERVER GROUP stage? I'm asking this because I saw in code some cloud providers made this while some seem not.
> 
> I am not necessarily very familiar with other cloud providers. No hard requirements that I can think of, other that we do want to make sure that there are no instances taking traffic in disabled server groups.

Thank you @dreynaud for your clarification.

PR #6700, opened 2019-03-15
fix(cf): map single route at a time

jmelchio:
- spinnaker/spinnaker#4122

In the infrastructure view only allow for mapping of a single route
at a time for server groups.

### DIFF
--- a/app/scripts/modules/cloudfoundry/src/serverGroup/details/mapLoadBalancers/CloudFoundryMapLoadBalancersModal.tsx
+++ b/app/scripts/modules/cloudfoundry/src/serverGroup/details/mapLoadBalancers/CloudFoundryMapLoadBalancersModal.tsx
@@ -57,7 +57,7 @@ export class CloudFoundryMapLoadBalancersModal extends React.Component<
 
     this.state = {
       initialValues: {
-        routes: [],
+        routes: [''],
       },
       taskMonitor: new TaskMonitor({
         application: props.application,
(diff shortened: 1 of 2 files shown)


app/scripts/modules/cloudfoundry/src/serverGroup/details/mapLoadBalancers/CloudFoundryMapLoadBalancersModal.tsx
```diff
@@ -57,7 +57,7 @@ export class CloudFoundryMapLoadBalancersModal extends React.Component<
 
     this.state = {
       initialValues: {
-        routes: [],
+        routes: [''],
```
claymccoy: Seems like it would be cleaner for the UI to handle an empty array rather than a single blank value.

jmelchio: If we don't want to allow the users to add more routes in the dialog we have to provide them with the first value to edit. There may be other ways to do that but upon review I still think this is the most straightforward way to go about it.

jkschneider: if they can only edit one route, why is it an array still?

jmelchio: We have a single component to handle the routes. We're re-using it for either single or multiple routes depending on where we use it.

PR #6701, opened 2019-03-15
fix(cf): fix the alignment issue for artifacts in deploy SG

Jammy-Louie:
spinnaker/spinnaker#4124


Jammy-Louie:
proposed change will result in the following ui:

Screen Shot 2019-03-15 at 10 51 55 AM

jkschneider:
Looks cleaner, thanks @Jammy-Louie !

PR #6703, opened 2019-03-15
fix(triggers): Add pipeline name to search request

ezimanyi:
When searching for a manually started pipeline by eventId, we should also pass in the pipeline name to reduce the amount of data returned from the persistence store.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/service/execution.service.ts
+++ b/app/scripts/modules/core/src/pipeline/service/execution.service.ts
@@ -120,9 +120,9 @@ export class ExecutionService {
       });
   }
 
-  public getExecutionByEventId(application: string, eventId: string): IPromise<IExecution> {
+  public getExecutionByEventId(application: string, pipelineName: string, eventId: string): IPromise<IExecution> {
     return API.all('applications', application, 'executions', 'search')
-      .get({ eventId })
+      .get({ pipelineName, eventId })
       .then((data: IExecution[]) => {
         if (data.length > 0) {
           const execution = data[0];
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/service/execution.service.ts
```diff
@@ -240,13 +240,17 @@ export class ExecutionService {
   public startAndMonitorPipeline(app: Application, pipeline: string, trigger: any): IPromise<IRetryablePromise<void>> {
     const { executionService } = ReactInjector;
     return PipelineConfigService.triggerPipeline(app.name, pipeline, trigger).then(triggerResult =>
-      executionService.waitUntilTriggeredPipelineAppears(app, triggerResult),
+      executionService.waitUntilTriggeredPipelineAppears(app, pipeline, triggerResult),
     );
   }
 
-  public waitUntilTriggeredPipelineAppears(application: Application, eventId: string): IRetryablePromise<any> {
+  public waitUntilTriggeredPipelineAppears(
```
jtk54: Can we also limit the retry period here to something reasonable?

ezimanyi: Yeah, I'm working on another PR that will either limit the retry period, or move the retry to look in the execution data source already in the front end rather than hitting the back end again.

jtk54: 👍

ezimanyi: #6707

PR #6704, opened 2019-03-15
fix(cf): provide meaningful labels in resize dialog

jmelchio:
- spinnaker/spinnaker#4063

### DIFF
--- a/app/scripts/modules/cloudfoundry/src/serverGroup/details/resize/CloudFoundryResizeServerGroupModal.tsx
+++ b/app/scripts/modules/cloudfoundry/src/serverGroup/details/resize/CloudFoundryResizeServerGroupModal.tsx
@@ -147,22 +147,25 @@ export class CloudFoundryResizeServerGroupModal extends React.Component<
   private renderQuota(
     formik: FormikProps<ICloudFoundryResizeServerGroupValues>,
     field: string,
+    fieldLabel: string,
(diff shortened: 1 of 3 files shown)


app/scripts/modules/cloudfoundry/src/serverGroup/details/resize/CloudFoundryResizeServerGroupModal.tsx
```diff
@@ -204,8 +207,8 @@ export class CloudFoundryResizeServerGroupModal extends React.Component<
                 <Modal.Body>
                   <Form className="form-horizontal">
                     {this.renderDesired(formik)}
-                    {this.renderQuota(formik, 'diskQuota', diskQuota)}
-                    {this.renderQuota(formik, 'memory', memory)}
+                    {this.renderQuota(formik, 'diskQuota', 'Disk Mb', diskQuota)}
```
jkschneider: Picky, but since this is a labeling concern ;) http://www.dewassoc.com/performance/memory/MB_vs_Mb.htm. Can we make this `Disk (MB)`?

jmelchio: Yes I can but then we have to change it in our pipeline stage as well. Do you want that included in this change then?

PR #6706, opened 2019-03-15
feat(cf): Add Updatable Flag To Create Service

stuart-pollock:
- also repaired validators for required fields
- also converted CreateUserProvidedInput's handling of data to be consistent with CreateServiceInstanceDirectInput

spinnaker/spinnaker#4084


stuart-pollock:
direct
user-provided

PR #6709, opened 2019-03-18
fix(artifacts): Exclude unmatchable expected artifact types

jkschneider:
* Exclude front50 artifact account from trigger artifact constraints.
* Exclude front50, docker, and k8s artifact accounts from CF deploy stage configuration.


jkschneider:
We now constrain the selection of an artifact account for a *Match Artifact* for certain trigger types. For example, it isn't possible that an Artifactory trigger will produce an artifact that is matchable to a Github artifact account.

image

At some point if we remove `ArtifactDecorator` from Igor, it will no longer be possible that a Jenkins trigger would produce an artifact that matches anything other than a Jenkins artifact account.

app/scripts/modules/core/src/artifact/ArtifactTypes.ts
```diff
@@ -15,3 +19,8 @@ export const ArtifactTypePatterns = {
   HTTP_FILE: /http\/file/,
   FRONT50_PIPELINE_TEMPLATE: /front50\/pipelineTemplate/,
 };
+
+export const excludeAllTypesExcept = (...types: RegExp[]) =>
+  Object.keys(ArtifactTypePatterns)
+    .map(k => (ArtifactTypePatterns as IArtifactTypePatterns)[k])
```
sbwsg: You can achieve the same as `Object.keys(o).map(k => o[k])` with `Object.values(o)`.

I also suggest avoiding `as` casts as much as possible. Giving ArtifactTypePatterns the IArtifactTypePatterns signature should let us leverage the type checker instead of short-circuiting it for this kind of thing:

```typescript
export const ArtifactTypePatterns: IArtifactTypePatterns = {
  // ...
};
Object.keys(ArtifactTypePatterns).map(k => ArtifactTypePatterns[k]) // <- should now infer the type as RegExp[]
```

jkschneider: I think `Object.values()` was introduced in ES2017, but I believe we are still using ES2015? Originally I did the as cast so I found it undesirable to export `IArtifactTypePatterns`, but I guess there is no issue with that, so will change it.

jkschneider: @sbwsg Type change made in b9a2eb5. Can you re-review?

jkschneider:

PR #6712, opened 2019-03-18
fix(artifacts): Fix SpEL text input used in React components

jkschneider:
Fixes SpEL inputs for artifact selectors in React-based components.

image


jkschneider:

PR #6715, opened 2019-03-18
fix(concourse): Fix concourse trigger config

jkschneider:
Previously attempted to split on `undefined` on the happy path here:

image


jkschneider:

PR #6716, opened 2019-03-19
fix(kubernetes): Fix copy from infrastructure link

ezimanyi:
The 'copy from running infrastructure' link in the 'Deploy Manifest' stage is not currently updating the view (but is correctly updating the stage data).  I believe it's just due to a missing `$onChange` handler.

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/wizard/manifestEntry.component.ts
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/wizard/manifestEntry.component.ts
@@ -22,6 +22,8 @@ class KubernetesManifestCtrl implements IController {
     this.rawManifest = yamlDocumentsToString(this.manifests);
   };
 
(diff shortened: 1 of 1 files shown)


ezimanyi:
@maggieneterval : This seemed to fix it and made sense, but let me know if there's a better way of doing this.  Thanks!

maggieneterval:
@ezimanyi Good catch, we should fix this. I think using `$onChanges` here will break typing a manifest in by hand since it will be called on every keystroke and lead to unexpected behavior (for example, when typing a key `key:`, `null` will show up as the value). Give me a couple minutes to look into a better way...

maggieneterval:
Hey this ended up being a bit more of a refactor than expected, so I just pushed up a new PR, going to close this one, if you don't mind.
https://github.com/spinnaker/deck/pull/6719

ezimanyi:
Sounds good, thanks!

PR #6718, opened 2019-03-19
refactor(forms): Creating contexts for Layout and Help

alanmquach:
By having `FormField` and `FormikFormField` *consume* layout from React context, we can now defer to some parent container to *provide* the layout.

This is useful because containers like `StageConfig` can define layouts for all descendants to use without having to explicitly thread the `layout` prop through and allows all `[Formik]FormField`s within a container to have a consistent layout.

The `layout` prop on `FormField` and `FormikFormField` is still available and will take precedence over the context provided layout.

Also adding context to `HelpField` so that `Layout`s can become a context provider allowing layouts to define the details of how a `HelpField` is rendered (currently only used for whether or not to `expand`).


app/scripts/modules/core/src/help/HelpField.tsx
```diff
@@ -54,15 +62,23 @@ export class HelpField extends React.PureComponent<IHelpFieldProps> {
...
     } else {
       const expanded = <div className="help-contents small"> {contents} </div>;
 
       return (
-        <div style={{ display: 'inline-block' }}>
-          {!expand && contents && popover}
-          {expand && contents && expanded}
-        </div>
+        <HelpContextConsumer>
+          {context => (
```
christopherthielen: eagerly anticipating removing this in favor of `useContext`

app/scripts/modules/core/src/help/HelpField.tsx
```diff
@@ -41,6 +42,13 @@ export class HelpField extends React.PureComponent<IHelpFieldProps> {
     }
   };
 
+  private considerContext(fromContext: any, fromProps: any) {
```
christopherthielen: this method and its usage is confusing to me, probably because of the naming.

Might be easier to understand if it was something like 
```
shouldExpandHelpText(expandFromContext, expandFromProps) {}
```

PR #6719, opened 2019-03-19
fix(kubernetes): fix copy manifest from infrastructure button

maggieneterval:
- Lifts `rawManifest` data prop and its change handler from Manifest Entry controller to Deploy Manifest controller so that clicking "Copy from Infrastructure" can update `rawManifest`
- Removes `manifest-entry` Angular component, replaces with `yaml-editor`


maggieneterval:
@sbwsg I'm not sure these styles are still being applied since I can't find the relevant classes anywhere, is this safe to delete (saw you were last committer to this file)? https://github.com/spinnaker/deck/blob/master/app/scripts/modules/kubernetes/src/v2/manifest/wizard/manifestEntry.less

sbwsg:
Yup! was only ever used by `manifestEntry.component.ts` I think.

PR #6722, opened 2019-03-20
feat(provider/azure): Add custom tags support in server group configu…

michaeljqzq:
* Add `tag` page on server group configuration
* Add validation for the tags key/value


michaeljqzq:
This PR is linked with https://github.com/spinnaker/clouddriver/pull/3485

app/scripts/modules/azure/serverGroup/configure/wizard/CloneServerGroup.azure.controller.js
```diff
@@ -4,6 +4,8 @@ const angular = require('angular');
 
 import { SERVER_GROUP_WRITER, TaskMonitor, ModalWizard, FirewallLabels } from '@spinnaker/core';
 
+const Utility = require('../../../utility').default;
```
sbwsg: It's preferable to use `import Utilility from '../../../utility` where possible.  Was there a specific reason not to use that here?

PR #6723, opened 2019-03-20
feat(mptv2): add list screen for pipeline templates

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/3888

Screen Shot 2019-03-20 at 10 28 44 AM

Note: There are two duplicate-looking templates in this screenshot. This is working correctly and is just a side effect of spin's current implementation of template versioning.

- Adds template list screen that allows user to View the JSON of all of their saved templates
- Adds `#/pipeline-templates` route to get to the list screen (behind feature flag)
- Adds button in top nav to get to the list screen (behind feature flag)

I'm not opinionated about the location of the entrypoint (nav button at top of Deck) or the chosen URL (`#/pipeline-templates`). Happy to change these if anyone has suggestions for better placement or naming.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/actions/templateJson/ShowPipelineTemplateJsonModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/actions/templateJson/ShowPipelineTemplateJsonModal.tsx
@@ -1,15 +1,16 @@
 import * as React from 'react';
 
 import { IModalComponentProps, JsonEditor } from 'core/presentation';
-import { IPipeline, IPipelineTemplateV2 } from 'core/domain';
+import { IPipelineTemplateV2 } from 'core/domain';
 import { CopyToClipboard, noop, JsonUtils } from 'core/utils';
-import { PipelineTemplateV2Service } from 'core/pipeline';
 
 import './ShowPipelineTemplateJsonModal.less';
 
 export interface IShowPipelineTemplateJsonModalProps extends IModalComponentProps {
-  ownerEmail: string;
-  pipeline: IPipeline;
+  template: IPipelineTemplateV2;
+  editable?: boolean;
+  modalHeading?: string;
+  descriptionText?: string;
 }
 
 export interface IShowPipelineTemplateJsonModalState {
@@ -22,16 +23,21 @@ export class ShowPipelineTemplateJsonModal extends React.Component<
 > {
   public static defaultProps: Partial<IShowPipelineTemplateJsonModalProps> = {
     dismissModal: noop,
+    editable: true,
+    modalHeading: 'Export as Pipeline Template',
+    descriptionText:
+      'The JSON below is the templated version of your pipeline. Save it by copy/pasting to the Spin CLI tool.',
   };
 
   constructor(props: IShowPipelineTemplateJsonModalProps) {
     super(props);
-
-    const template = PipelineTemplateV2Service.createPipelineTemplate(props.pipeline, props.ownerEmail);
-    this.state = { template };
+    this.state = { template: props.template };
   }
 
-  private onChange = (e: React.ChangeEvent<HTMLInputElement>, property: string) =>
+  private onChange = (e: React.ChangeEvent<HTMLInputElement>, property: string) => {
+    if (!this.props.editable) {
+      return;
+    }
     this.setState({
       template: {
         ...this.state.template,
@@ -41,9 +47,10 @@ export class ShowPipelineTemplateJsonModal extends React.Component<
         },
       },
     });
+  };
 
   public render() {
-    const { dismissModal } = this.props;
+    const { dismissModal, editable, modalHeading, descriptionText } = this.props;
     const { template } = this.state;
     const sortedTemplate = JsonUtils.sortObject(template);
     const templateStr = JsonUtils.makeStringFromObject(sortedTemplate, 0);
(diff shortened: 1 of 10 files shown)


app/scripts/modules/core/src/pipeline/config/templates/v2/pipelineTemplateV2.states.ts
```diff
@@ -0,0 +1,30 @@
+import { module } from 'angular';
+import { SETTINGS } from 'core/config';
+import { INestedState, STATE_CONFIG_PROVIDER, StateConfigProvider } from 'core/navigation/state.provider';
```
louisjimenez: looks like this can be shortened  to `from 'core/navigation';`

app/scripts/modules/core/src/pipeline/config/templates/v2/pipelineTemplateV2.states.ts
```diff
@@ -0,0 +1,30 @@
...
+import { PipelineTemplatesV2 } from './PipelineTemplatesV2';
+
+export const PIPELINE_TEMPLATES_V2_STATES_CONFIG = 'spinnaker.core.pipeline.templates.v2.states.config';
+
+module(PIPELINE_TEMPLATES_V2_STATES_CONFIG, [STATE_CONFIG_PROVIDER]).config([
+  'stateConfigProvider',
+  (stateConfigProvider: StateConfigProvider) => {
+    const pipelineTemplatesList: INestedState = {
+      name: 'mptv2',
+      url: '/mptv2',
```
louisjimenez: Would matching the url to the "pipeline-templates" name used elsewhere make sense here? I'm not sure if /mptv2 will be clear to users

app/scripts/modules/core/src/pipeline/config/templates/v2/PipelineTemplatesV2.tsx
```diff
@@ -0,0 +1,121 @@
...
+    const props = { template, editable: false };
+    ReactModal.show<IShowPipelineTemplateJsonModalProps>(ShowPipelineTemplateJsonModal, props, {
+      dialogClassName: 'modal-lg modal-fullscreen',
+    });
+  };
+
+  public render() {
+    return (
+      <>
+        <div className="infrastructure">
```
louisjimenez: This `infrastructure` markup with a header and body seems like it could be a good candidate for a separate component for the next time we add a route

app/scripts/modules/core/src/pipeline/config/templates/v2/PipelineTemplatesV2.tsx
```diff
@@ -0,0 +1,121 @@
...
+              <h2 className="header-section">
+                <span className="search-label">Pipeline Templates</span>
+              </h2>
+            </div>
+          </div>
+          <div className="infrastructure-section">
+            <div className="container">
+              {this.state.fetchError && (
+                <PipelineTemplatesV2Error
+                  message={`There was an error fetching pipeline templates: ${this.state.fetchError.message}`}
```
louisjimenez: If I make the request fail with a 404 or by blocking the request `message` seems to come back `undefined`. In the case of a 404, the `err` object does have a `data.message` and `data.error`

app/scripts/modules/core/src/pipeline/config/templates/v2/PipelineTemplatesV2.less
```diff
@@ -0,0 +1,13 @@
+.pipeline_templates_error_banner {
+  margin: 1em 0;
+
+  i.fa-exclamation-triangle {
+    padding-right: 8px;
+  }
+}
+
+.pipeline_template_actions {
```
louisjimenez: Just wondering about the convention here, does the project practice underscores instead of dashes? Or is this a similar to BEM with the double underscore to signify an element in a block?

app/scripts/modules/core/src/pipeline/config/actions/templateJson/ShowPipelineTemplateJsonModal.tsx
```diff
@@ -68,6 +70,7 @@ export class ShowPipelineTemplateJsonModal extends React.Component<
                   type="text"
                   value={template.metadata.name}
                   onChange={e => this.onChange(e, 'name')}
+                  disabled={!this.props.editable}
```
louisjimenez: what about adding `editable` to the destructure on line 48

app/scripts/modules/core/src/pipeline/config/templates/v2/PipelineTemplatesV2.less
sbwsg: Just discovered that we use BEM according to our Conventions wiki page.  I'll revisit the names here.

app/scripts/modules/core/src/pipeline/config/templates/v2/PipelineTemplatesV2.tsx
sbwsg: I might save this suggestion for a follow-up refactor PR since both Applications & now templates use this html structure and should ideally both change at once.

sbwsg:
@gcomstock just wanted to keep you posted that this is implemented (behind a flag, so totally open for changes). It's a screen to list managed pipeline templates + a URL and a button to access that screen. I know there's been some initial discussion around the location of the entrypoint and also that you have some designs in-process. If you have any thoughts or suggestions here, would definitely be interested in hearing them.

gcomstock:
Thanks @sbwsg ! I just got out of a meeting for this and will be working up some design in the coming days. I appreciate the heads up.

PR #6733, opened 2019-03-21
feat(tests): Add functional test for manual execution link

sbwsg:
### Example Test Run

manual_execution

### DIFF
--- a/test/functional/tools/MountebankService.ts
+++ b/test/functional/tools/MountebankService.ts
@@ -91,7 +91,6 @@ export class MountebankService {
               {
                 proxy: {
                   to: `http://localhost:${this.options.imposterPort}`,
-                  mode: 'proxyTransparent',
                   predicateGenerators: [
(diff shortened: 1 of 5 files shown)


test/functional/tools/MountebankService.ts
```diff
@@ -91,7 +91,6 @@ export class MountebankService {
               {
                 proxy: {
                   to: `http://localhost:${this.options.imposterPort}`,
-                  mode: 'proxyTransparent',
```
maggieneterval: just curious, what does this do?

sbwsg: The `removeImposters()` function is used to teardown the mock Gate server after a test is finished.  This property isn't one of the required parameters when doing this and was probably left over from an old copy/paste so I just removed it.

sbwsg:
Chip loves functional testing.

PR #6735, opened 2019-03-21
fix(provider/cf): Default clone stage to start on creation

jkschneider:
This is consistent with the Deploy operation and the Clone server group action which both start on creation by default.


jkschneider:

PR #6737, opened 2019-03-21
fix(cf): Map/Unmap LBs and UI cleanup

jasonjchu:
- added 'preventSave' to all required fields in Map / Unmap Load
Balancers pipeline stages
- corrected pluralization
- repaired minor UI glitches in map / unmap
- optionality of route fields in mapping / unmapping screens removed
- removed some Angular wrapping on map / unmap pipeline stages
- repaired minor bugs in AccountRegionClusterSelector in which: a) the
region was not being cleared when a new account is selected; and
b) the cluster list was not being properly initialized for a new stage
when the AccountRegionClusterSelector was specified as single-region
- repaired Rollback Cluster Stage, which was not displaying any of its
CF-specific fields

spinnaker/spinnaker#4146
spinnaker/spinnaker#4147
spinnaker/spinnaker#4169
spinnaker/spinnaker#4171
spinnaker/spinnaker#4172


jkschneider:

PR #6743, opened 2019-03-22
fix(cf): Repair Rollback Cluster pipeline stage

stuart-pollock:
Cloud Foundry's Reactified implementation of AccountRegionClusterSelector did not account for moniker selection.  The moniker field is required by Orca to properly perform the rollback.

spinnaker/spinnaker#4180


jkschneider:

PR #6745, opened 2019-03-25
feat(core): add markdown functionality to custom app banners

maggieneterval:
- Based on the UI for Entity Tag editing, adds support for writing custom application banners in Markdown, allowing for formatted text and links.
- Example: appbannermarkdown

### DIFF
--- a/app/scripts/modules/core/src/application/config/customBanner/CustomBannerConfig.tsx
+++ b/app/scripts/modules/core/src/application/config/customBanner/CustomBannerConfig.tsx
@@ -9,6 +9,8 @@ import {
 } from 'core/application/config/customBanner/customBannerColors';
 
 import { noop } from 'core/utils';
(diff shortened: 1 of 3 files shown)


anotherchrisberry:
Not that anyone would do this, but anyone could do this:
Screen Shot 2019-03-25 at 10 48 56 AM

Wondering if we should restrict what we allow in the custom banners, e.g. don't let people put `<style>` tags in? Maybe it's fine - if people want to hack things up like this, that's up to them?

maggieneterval:
@anotherchrisberry I'm laughing out loud at my desk right now, love the Canadian flags. I don't think we restrict anything in the Entity Tag markdown editor so I guess I'd lean toward doing the same here? Or at least adding the same validation to both :)

christopherthielen:
@anotherchrisberry that looks like a great feature for people to custom style their apps like myspace

PR #6746, opened 2019-03-25
fix(ecs): Fix name of health check grace period attribute

clareliguori:
Fixes spinnaker/spinnaker#4123


clareliguori:

PR #6467, opened 2019-02-01
feat(google): Add support for accelerators when deploying VMs

sbwsg:
* User can add GPU to VMs in GCE server group modals
* The set of GPUs available varies by zone
* The maximum number of GPUs per instance varies by card
* Adds a functional test to verify that the GPU type list renders as expected

#### Here's how the component looks in the server group modal:

screen shot 2019-02-01 at 1 45 06 pm

#### After deployment via Spinnaker, here's the instance with the attached accelerator:

screen shot 2019-02-01 at 1 42 28 pm

#### GIF of functional test run:

accelerators

PR #6750, opened 2019-03-26
fix(google): add better help text around accelerators

sbwsg:
Add a tooltip and slightly stronger guidance on how to correctly attach accelerators in the server group wizard.

In response to comments on the original PR here: https://github.com/spinnaker/deck/pull/6467


maggieneterval:
@sbwsg looks great, maybe we should patch into 1.13 to avoid future confusion?

sbwsg:

ciciya-lal:
@sbwsg , Thanks for adding help text..^^

PR #6751, opened 2019-03-26
fix(google): select all zones by default when deploying regional server group

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/4071

This only affects us-central at the moment, which has 4 zones.

### Before

Screen Shot 2019-03-26 at 9 56 38 AM

### After

Screen Shot 2019-03-26 at 9 55 11 AM


app/scripts/modules/google/src/serverGroup/configure/serverGroupConfiguration.service.js
```diff
@@ -369,7 +369,7 @@ module.exports = angular
         const regions = command.backingData.credentialsKeyedByAccount[command.credentials].regions;
         if (_.isArray(regions)) {
           filteredData.zones = _.find(regions, { name: command.region }).zones;
-          filteredData.truncatedZones = _.takeRight(filteredData.zones.sort(), 3);
+          filteredData.truncatedZones = filteredData.zones.slice().sort();
```
louisjimenez: should this variable be renamed since the list is no longer truncated?

sbwsg: i've renamed these & tested that pipelines saved before/after the rename continue to render as expected.

sbwsg:

sbwsg:

PR #6754, opened 2019-03-26
fix(provider/cf): Fix provider selection for resize stage in pipeline

jasonjchu:
spinnaker/spinnaker#4193


jkschneider:

PR #6755, opened 2019-03-26
fix(google): select all zones by default when deploying a regional gce server group

sbwsg:


sbwsg:
backport of https://github.com/spinnaker/deck/pull/6751 into 1.12 release branch.

PR #6756, opened 2019-03-26
feat(core/execution-parameters): condense parameters/artifacts and make it collapsable

kevinawoo:
This PR address some issues when there's lots of parameters and/or artifacts, making the pipeline page very long. On smaller screen sizes, unusable.
This also addresses https://github.com/spinnaker/spinnaker/issues/1857

Included are some screenshots of what it looks like now and how it looks after applying this PR:

<details><summary>How it currently looks</summary>

<kbd></kbd>

</details>

<details><summary>Let's see a gif in action (download is kind of big, give it a sec)</summary>

<kbd>ezgif com-gif-maker (3)</kbd>

</details>

<details><summary>Collapsed View with some annotations of what's changed</summary>

<kbd></kbd>

</details>

<details><summary>Expanded View with some annotations of what's changed</summary>

<kbd></kbd>

</details>

<details><summary>Standalone View is auto expanded</summary>

<kbd></kbd>

</details>


app/scripts/modules/core/src/pipeline/executions/execution/Execution.tsx
```diff
@@ -379,7 +413,42 @@ export class Execution extends React.Component<IExecutionProps, IExecutionState>
               </Tooltip>
             )}
           </div>
+
+          <div className="execution-parameters-button">
```
christopherthielen: We have a `CollapsibleSection` component to show/hide sections and as a bonus includes an animated chevron

See here for an example usage: 
https://github.com/spinnaker/deck/blob/6e29dacbac867900087c2c169ee971ac084e963a/app/scripts/modules/core/src/presentation/forms/fields/FormikExpressionRegexField.tsx#L63-L68

- `heading` can be a string (or a function component).

note: I suspect this component cannot be used today to also collapse the Details panel because of the unique styling requirements.

app/scripts/modules/core/src/pipeline/status/ExecutionParameters.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import { IExecution } from 'core/domain';
+
+import './executionStatus.less';
+import './executionParameters.less';
+
+export interface IExecutionParametersProps {
+  execution: IExecution;
+  showingParams: boolean;
+  columnFormatAfter: number;
```
christopherthielen: how about `columnLayoutAfter`

app/scripts/modules/core/src/pipeline/status/ExecutionParameters.tsx
```diff
@@ -0,0 +1,75 @@
...
+  constructor(props: IExecutionParametersProps) {
+    super(props);
+
+    // these are internal parameters that are not useful to end users
+    const strategyExclusions = ['parentPipelineId', 'strategy', 'parentStageId', 'deploymentDetails', 'cloudProvider'];
+
+    let parameters: Array<{ key: string; value: any }> = [];
+
+    const { execution } = this.props;
+    if (execution.trigger && execution.trigger.parameters) {
```
christopherthielen: I know you migrated this from existing code, but we should prefer re-evaluating computed values from props in `render` over copying to `state`.  Would you mind refactoring, either in this PR or a separate one?

See https://overreacted.io/writing-resilient-components/#dont-stop-the-data-flow-in-rendering

kevinawoo:
> 2. We (unfortunately) have pipelines that accept large parameter values, such as a base64 encoded blob. This isn't a huge concern, but we'd like to understand what some of these oddball pipelines would look like, i.e., with multi column.

<kbd></kbd>

christopherthielen:
nice, thanks for the screenshot!

gcomstock:
Hi @kevinawoo , thanks for making this fix. I think this could be improved more by counting the string length, and if it goes over a few lines, choose to truncate the value and append a "see all" link that would drop the full value into a modal. Let me know your thoughts. Thanks!

app/scripts/modules/core/src/pipeline/status/ExecutionParameters.tsx
kevinawoo: Man `memoize-one` is an awesome library, also great article!

Too bad we don't have react 16.8 yet, otherwise we could have used hooks.

jkschneider:
@kevinawoo Just want you to be aware that in 1.14 (and optionally in 1.13 with the `artifactsRewrite` feature flag on), the Expected Artifacts section will not be present at all on the Configuration screen.

image

kevinawoo:
Note, I made a PR against this branch so it's a bit easier to review: https://github.com/armory-io/deck/pull/4

kevinawoo:
Note, I made a PR against this branch so it's a bit easier to review: https://github.com/armory-io/deck/pull/4
@christopherthielen 	@gcomstock 	would you guys be able to take a look at it 🙏 ?

kevinawoo:
Note, this PR is only addressing pinning parameters. Artifacts will be handled later.

<details><summary>GIF expanding/collapsing (shows a few parameters pinned)</summary>

<kbd></kbd>

</details>

<details><summary>Pipeline Overview (params/artifacts collapsed) with annotations</summary>

<kbd></kbd>

</details>


<details><summary>Pipeline Overview (params/artifacts expanded) with annotations</summary>

<kbd></kbd>

</details>


<details><summary>Pipeline Config with annotations</summary>

<kbd></kbd>

</details>

<details><summary>Standalone/Permalink View is always expanded with annotations</summary>

<kbd></kbd>

</details>

app/scripts/modules/core/src/pipeline/executions/execution/Execution.tsx
kevinawoo: It's not too easy to wire this one up with pinned parameters, since Pinned parameters are kind of "semi" expanded.

kevinawoo:
> Hi @kevinawoo , thanks for making this fix. I think this could be improved more by counting the string length, and if it goes over a few lines, choose to truncate the value and append a "see all" link that would drop the full value into a modal. Let me know your thoughts. Thanks!

@gcomstock any recommendation on which modal components to use?

app/scripts/modules/core/src/domain/IPipeline.ts
```diff
@@ -28,13 +28,15 @@ export interface IPipeline {
     type: string;
   };
   type?: string;
+  pinAllParameters?: boolean;
```
kevinawoo: there's an outstanding conversation on this line from a prior PR:
https://github.com/armory-io/deck/pull/4#discussion_r270636672
<kbd></kbd>

app/scripts/modules/core/src/pipeline/executions/execution/Execution.tsx
```diff
@@ -73,8 +85,14 @@ export class Execution extends React.Component<IExecutionProps, IExecutionState>
 
     const restartedStage = execution.stages.find(stage => stage.context.restartDetails !== undefined);
 
+    this.displayableParameters = this.getDisplayableParameters(execution);
```
kevinawoo: there's a prior comment on this line:
https://github.com/armory-io/deck/pull/4#discussion_r270638658
<kbd></kbd>

app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
```diff
@@ -15,6 +15,12 @@
             <help-field key="pipeline.config.parallel.cancel.queue"></help-field>
           </label>
         </div>
+        <div class="checkbox">
+          <label>
+            <input type="checkbox" ng-model="pipeline.pinAllParameters" />
+            <strong>Pin all parameters when viewing an execution</strong>
```
kevinawoo: there's a prior comment on this line:
https://github.com/armory-io/deck/pull/4#issuecomment-478275517
<kbd></kbd>

> Would it be fair to say Permalink defaults to expanded, but the user can still click the link to collapse? That's what the screenshot seems to show and is also what I think I would prefer.

YEP!

> I think I'd prefer to see "pin all parameters when viewing an execution" replaced with a "pin all"/"unpin all" toggle in the Parameters section. @gcomstock if you're in the office next week, can you provide some guidance on the direction you'd like to go for check/uncheck all in the forms mockups?

hm.. I can get behind this. A mockup would be great! I'm not sure I have a good idea myself on it.

app/scripts/modules/core/src/domain/IPipeline.ts
kevinawoo: I'm good with removing this if we decide `toggle all pins`/`unpin all`+`pin all` is the best option.

app/scripts/modules/core/src/pipeline/executions/execution/Execution.tsx
gcomstock: Hey @kevinawoo id be very happy to get you a mockup. Im crunching on something else at the moment but will make a note to do this later today.

app/scripts/modules/core/src/pipeline/config/triggers/triggers.html
gcomstock: Hey @kevinawoo id be very happy to get you a mockup. Im crunching on something else at the moment but will make a note to do this later today.

kevinawoo: oh would you also be able to get me a mockup with the `pin/unpin all artifacts` too? you'll may need to set `window.spinnakerSettings.feature.artifactsRewrite = true;` in `settings-local.js`.

But there's some new UI elemenets with it:
<kbd></kbd>
<kbd></kbd>
<kbd></kbd>

kevinawoo: I think we should just leave the old style of artifacts alone for now, since the inline style will be enabled by default later.

kevinawoo: another option is just to merge this and come back to artifacts later, after there's a need. We don't currently have a need to pin artifacts yet, parameters are the ones that are exploding at the moment

gcomstock:
Sorry for the delay, @kevinawoo , I've been slammed with other things. These are pretty quick cut and paste jobs, my apologies. It's on my radar to do some more in-depth thinking around the executions view. 

- I think what you've put together is great, and has really helped solve a big nagging problem. Thank you!

- Having the pinned parameters under the collapsed menu is a great use of space. I think the visual proximity of the pinned parameters to the full parameter count (41) is enough to communicate they are pinned, and having the extra "Pinned parameters" title is unncessary. Please see the mock and let me know if you agree or not.

- These mocks show how it could be broken out so parameters and artifacts are pinned separately. I think this is a safer way to go without knowing the min/max on artifacts, and lets you follow the same pattern on the config page with the pin all checkbox.

- The mocks also show a truncated long value. I don't have a preference of wether clicking "See full value" just reveals it inline or in a modal. inline is probably easier for you to implement so maybe we go with that? Let me know.

- I'm OOO tomorrow but I can try and check in Friday. Thanks again for all the hard work!

params-checkboxes
params-with-truncation

kevinawoo:
> Having the pinned parameters under the collapsed menu is a great use of space. I think the visual proximity of the pinned parameters to the full parameter count (41) is enough to communicate they are pinned, and having the extra "Pinned parameters" title is unncessary. Please see the mock and let me know if you agree or not.

<details>
<summary>👍, removed the extra header. expand to see gif!</summary>

<kbd></kbd>
</details>

> These mocks show how it could be broken out so parameters and artifacts are pinned separately. I think this is a safer way to go without knowing the min/max on artifacts, and lets you follow the same pattern on the config page with the pin all checkbox.

@gcomstock I've been thinking about this and I think it streamlines the process of viewing an execution. We're more likely going to want to see both parameters and artifacts at the same time, rather than just params or artifacts standalone. This avoid doing two actions to expand params and artifacts each time. 

# Comment on `toggle pinned parameters` inside parameters section
<details>
<summary>Area I'm talking about</summary>

<kbd></kbd>
</details>
I'm not too excited about having a random checkbox under parameters in the negative space. I wouldn't expect there to be a checkbox there since the page uses white boxes to indicate actionable sections. I also considered putting a white box, but looks too like a parameter itself. 

# Truncate Blobs
> The mocks also show a truncated long value. I don't have a preference of wether clicking "See full value" just reveals it inline or in a modal. inline is probably easier for you to implement so maybe we go with that? Let me know.

<details>
<summary>Oh! Great Idea.</summary>

<kbd></kbd>
</details>

christopherthielen:
hope this was ready for merge because I just merged it

gcomstock:
Hey all, sorry I didn't get back to you earlier on this.

@kevinawoo thanks for the writeup on putting parameters and artifacts together under one collapsible. Sounds good!

As for the random checkbox, I agree that it looks a bit out of place on its own. We're trying to wrap up some form field systems that bring in more consistency around use cases like this, where there are fields for the whole form, a sub group, and a section. I'll take this example back to the system to think a bit more on how we can create consistency with this hierarchy. Nothing big to change now, but thanks for calling it out.

PR #6761, opened 2019-03-27
feat(templates): add Save button to export pipeline template json modal

sbwsg:
Fixes https://github.com/spinnaker/spinnaker/issues/4190

Adds a button to save a template directly from the pipeline export modal. Clicking the button saves the template and redirects the user to the templates screen where they can see their new template.

mpt_save

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
+++ b/app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
@@ -16,6 +16,7 @@ import { ExecutionsTransformer } from 'core/pipeline/service/ExecutionsTransform
 import { EditPipelineJsonModal } from 'core/pipeline/config/actions/pipelineJson/EditPipelineJsonModal';
 import { ShowPipelineTemplateJsonModal } from 'core/pipeline/config/actions/templateJson/ShowPipelineTemplateJsonModal';
 import { PipelineTemplateV2Service } from 'core/pipeline';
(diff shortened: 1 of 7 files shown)


sbwsg:
@jtk54 Yup, the link to copy to spin command is still there.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -537,5 +543,19 @@ module.exports = angular
...
+        return PipelineTemplateWriter.savePipelineTemplateV2(template).then(
+          response => {
+            console.log('response', response);
+            const id = response.variables.find(v => v.key === 'pipelineTemplate.id').value;
+            $state.go('home.pipeline-templates.pipeline-templates-detail', {
+              templateId: PipelineTemplateV2Service.idForTemplate({ id }),
+            });
+            return true;
+          },
+          err => err,
```
louisjimenez: I think when this errors out, this func will cause the promise to appear as fulfilled to the `then` in `ShowPipelineTemplateJsonModal` line 64 and close the modal without surfacing that something went wrong.

sbwsg: I've added a throw here to propagate to the catch() handler of the modal.

sbwsg:
@louisjimenez updated to respond to route changes and refactored to remove usage of `ReactModal.show()` which, in hindsight, appears to be intended for angular use-cases only.

PR #6774, opened 2019-03-28
feat(gcb): add Google Cloud Build stage

maggieneterval:
- Moves YAML editor from kubernetes module to core in order to re-use in GCB stage
- Adds MVP Google Cloud Build stage

hAnZANhLc89


app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
```diff
@@ -0,0 +1,69 @@
...
+      googleCloudBuildAccounts: [],
+      rawBuildDefinitionYaml: props.stage.buildDefinition ? yamlDocumentsToString([props.stage.buildDefinition]) : '',
+    };
+  }
+
+  public componentDidMount = (): void => {
+    if (this.props.stage.isNew) {
+      this.props.updateStageField({ application: this.props.application.name });
+    }
+    this.fetchGoogleCloudBuildAccounts();
```
louisjimenez: What are your thoughts on catching this when it errors out? Do we have an existing convention for alerting the user / logging? When it 404's it throws an uncaught exception (no existing accounts).

maggieneterval: No convention that I know of; unfortunately looks like many catches are nonexistent or noops. I feel like this would be a great UI SIG discussion though, with the goal of landing on one right way to handle

ezimanyi: I this particular case, the endpoint will return an empty list (rather than 404) when there are no existing accounts, so that's not an error mode to worry about.  But agree with the general sentiment of finding a good way to handle general error cases.

maggieneterval: Added to UI SIG agenda for 4/8 :tada:

sbwsg:
Nice!

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
louisjimenez: @ezimanyi I'm seeing the endpoint return a 404 locally. This is without setting up any GCB accounts:

image

I agree that error handling conventions could be good to cover at the SIG 👍

ezimanyi: It will return a 404 if you don't have the `gcb` feature flag enabled at all in `igor` (as those endpoints don't exist at all)...so I guess this is something that a user could see if they haven't configured properly and might make sense to improve the error message around.

PR #6775, opened 2019-03-28
fix(kubernetes): safe lookups for apiVersion on patch manifest deploy…

maggieneterval:
… status

The manifest on a Patch Manifest stage's execution context will not include `apiVersion`. This safeguards the `apiVersion` lookups so Deck doesn't crash when you try to view a Patch Manifest stage's Deploy Status. Lines 111 and 119 are the only actual changes; the rest are Prettier's.

Closes https://github.com/spinnaker/spinnaker/issues/4204


maggieneterval:
cc @willgorman

maggieneterval:

PR #6777, opened 2019-03-28
fix(kubernetes): fix validation for Find Artifact from Resource stage

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4205

Renames `trafficValidators` to `manifestSelectorValidators` because these validators are for any k8s stage that uses the `ManifestSelector` component. Replaces existing Find Artifact from Resource validators with  `manifestSelectorValidators`.


maggieneterval:

PR #6785, opened 2019-03-29
fix(pipeline): Fix target impedance validator for clone server gro…

jkschneider:
Previously, the comparison logic tested against clone server group's target cluster and not its _destination_.

Validation worked when the server group created at the end of clone operation had the same stack and detail as the source, but failed otherwise.

image


jkschneider:

PR #6796, opened 2019-04-02
fix(provider/cf): Make expected artifacts selectable as clone manifests

jkschneider:
The clone dialog needs to have pipeline and stage state so that it can look up available expected artifacts at this point in the pipeline's execution.

cc / @ciberkleid

image


jkschneider:

PR #6799, opened 2019-04-02
fix(artifacts): Clean pipeline expected artifacts when triggers are removed

jkschneider:
This only applies to the `artifactsRewrite` feature flag for 1.13, which more explicitly requires that pipeline expected artifacts are used as trigger constraints.


jkschneider:

PR #6804, opened 2019-04-02
fix(google): fix autohealing health checks in deploy stages

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4221

We made the following fixes to Deck and Clouddriver to disambiguate health checks of the same name by also specifying the health check kind:
Deck: https://github.com/spinnaker/deck/pull/6382
Clouddriver: https://github.com/spinnaker/clouddriver/pull/3282

This fix worked by using the health check's URL as a unique ID, and extracting the health check's name and kind from the URL prior to submit. However, this only worked for ad-hoc create and clone operations, not the deploy stage, where the field is bound to the stage model. To resolve this, I added an onSelect handler to the health check dropdown that will update the health check's name and kind based on the selected health check URL. Confusingly, Deck receives a server group's `autoHealingPolicy.healthCheck` as the health check URL, but Clouddriver is expecting to receive a payload with `autoHealingPolicy.healthCheck` as the health check name, so to cover the two cases of building the command for stage editing (based on the stage model) and building the command for ad-hoc operations (based on what we receive from the server), we need to check if we have a `healthCheckUrl` field (only true for stages), and otherwise assume that the `healthCheck` is the URL.

Note that if users have worked around this bug by manually editing the deploy stage JSON such that `healthCheck` corresponds to the health check name rather than the URL, they can update these stages to work as expected by adding a `healthCheckUrl` field.

### DIFF
--- a/app/scripts/modules/google/src/domain/autoHealingPolicy.ts
+++ b/app/scripts/modules/google/src/domain/autoHealingPolicy.ts
@@ -1,8 +1,9 @@
 import { IGceHealthCheckKind } from 'google/domain/healthCheck';
 
 export interface IGceAutoHealingPolicy {
-  healthCheck?: string;
-  healthCheckKind?: IGceHealthCheckKind;
+  healthCheck?: string; // received from server as health check URL, but posted as health check name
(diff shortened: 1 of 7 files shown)


maggieneterval:

PR #6808, opened 2019-04-03
fix(google): revert "select all zones by default when deploying a regional gce server group (#6751)"

sbwsg:
This reverts commit ad2e2227442aa2e7abf85bf1a82ef7a5f51b1445.

In hindsight, Deck was already doing the right thing here. The platform, GCP, chooses the zones if the user doesn't specify them.  In some quick testing through the GCP UI, MIGs in us-central1 always seem to get spread across zones -b, -c, and -f and never end up on -a.  So it was incorrect to show in Deck that they might be placed on -a.

Closes https://github.com/spinnaker/spinnaker/issues/4071


sbwsg:

sbwsg:

PR #6816, opened 2019-04-04
feat(kubernetes): feature-flagged support for kubernetes traffic mana…

maggieneterval:
…gement strategies

- Adds initial support for Kubernetes V2 traffic management strategies behind `features.kubernetesTrafficManagementStrategies`
- Enables "dark" deployments and ability to associate a ReplicaSet with one or more Services without having to manually add the Spinnaker traffic management annotation
- Coming soon: a strategies dropdown with options for highlander and red/black strategies!

9rWHRv2dhfd

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/ManifestDeploymentOptions.tsx
+++ b/app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/ManifestDeploymentOptions.tsx
@@ -1,0 +1,151 @@
+import { module } from 'angular';
+import * as React from 'react';
+import { react2angular } from 'react2angular';
+import { cloneDeep, find, get, map, set, split } from 'lodash';
+import Select, { Option } from 'react-select';
+
+import { IAccountDetails, StageConfigField } from '@spinnaker/core';
+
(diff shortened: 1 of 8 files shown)


maggieneterval:
cc @dorbin this is feature-flagged so no rush on this, but any wording input you have for the section labels and tooltip text is most welcome!

app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/ManifestDeploymentOptions.tsx
```diff
@@ -0,0 +1,153 @@
...
+    this.fetchServices();
+  }
+
+  public componentDidUpdate(prevProps: IManifestDeploymentOptionsProps) {
+    // If selected account changes, clear selected namespace.
+    if (prevProps.selectedAccount !== this.props.selectedAccount) {
+      this.onConfigChange('options.namespace', null);
+    }
+
+    // If selected namespace changes, clear selected services, and re-fetch services.
```
sbwsg: Very minor nit: I'm not sure the comments here are necessary; the code below each is self-describing I think.

maggieneterval: You are very right, removed them!

PR #6817, opened 2019-04-04
fix(provider/cf): Clone model so Orca monitors the correct foundation for up instances

jkschneider:
Previously, cross-foundational clone stages never completed because Orca was watching the source foundation for up instances rather than the target.


jkschneider:

PR #6818, opened 2019-04-04
fix(triggers): Remove RunAsUser if pipeline permissions enabled for react triggers

dibyom:
Concourse and Docker are the only two triggers currently in React and are missing the check for `SETTINGS.feature.managedServiceAccounts`.  For the Angular based triggers, we do the check in https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/triggers/runAsUserSelector.component.ts#L11

Addresses https://github.com/spinnaker/spinnaker/issues/4217


christopherthielen:
# YOLO!

christopherthielen:
https://github.com/dibyom/deck/blob/64503f91a2122fb8eb143eca7e7a7a6a2127f1d9/app/scripts/modules/docker/src/pipeline/trigger/DockerTriggerConfig.tsx#L65

PR #6819, opened 2019-04-04
feat(core): Create pipeline from v2 template list

louisjimenez:
https://github.com/spinnaker/spinnaker/issues/3893
Adds support for creating a pipeline from a v2 template in the template list.

image
image
image

### DIFF
--- a/app/scripts/modules/core/src/domain/IPipelineTemplateConfigV2.ts
+++ b/app/scripts/modules/core/src/domain/IPipelineTemplateConfigV2.ts
@@ -1,0 +1,13 @@
+import { IPipeline } from 'core/domain';
+
+export interface IPipelineTemplateConfigV2 extends IPipeline {
+  inherit?: string[];
+  schema: string;
+  template: {
+    artifactAccount: string;
+    reference: string;
+    type: string;
+  };
+  type: string;
+  variables?: { [key: string]: any };
+}
--- a/app/scripts/modules/core/src/domain/index.ts
+++ b/app/scripts/modules/core/src/domain/index.ts
@@ -39,6 +39,7 @@ export * from './IOrchestratedItem';
 
 export * from './IPipeline';
 export * from './IPipelineTemplateV2';
+export * from './IPipelineTemplateConfigV2';
 export * from './IProject';
 export * from './IPubsubSubscription';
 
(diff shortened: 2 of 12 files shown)


app/scripts/modules/core/src/utils/getCancelablePromise.ts
```diff
@@ -0,0 +1,31 @@
+export interface ICancelablePromise<T> {
```
louisjimenez: Using this to handle promises that need to be "cancelled" when a React component unmounts and the promise has yet to settle. Open to any ideas, suggestions, or alternatives. This func is intended for use with native Promises but can support casted IPromises where hooking into the angular digest is not needed..

erikmunson: 👋 If I recall correctly, there have been several PRs with a version of this cancelable promise pattern suggested. We've historically been skeptical of the extra complexity involved (threading it through things with `Promise/IPromise` types, refactoring areas that do/don't use them, etc.), and it hasn't been clear what the value is over using RxJS for cancellation.

Do you think this use case is something that could be solved by an RxJS observable that ends when the component unmounts? that is the pattern we use across the rest of deck and I can send over an example if desired (if you grep for `this.destroy$.next()` you should find some).

If you don't think this particular use case is solvable with the observable approach, then maybe we should consider this pattern again!

louisjimenez: Hey, thanks for taking a look! I hadn't encountered the RxJS pattern in the codebase but that should fit the use case well. I am definitely in favor of sticking to fewer, consistent patterns so I'll give implementing this via an observable a go. 

On a related note, I'd be happy to document the existing pattern if that doc doesn't already exist somewhere. Do you think there would be value in doing so?

erikmunson: Cool! I could maybe see value in adding some written words about it somewhere, perhaps with the intended audience of 'people who submit PRs that could make use of this'. Not sure whether documentation of this kind would end up getting discovered by someone _before_ they submit a PR, but when they do we could shoot them a nice link with a summary and some rationale about why.

If you choose to do that, maybe it could go in the Wiki pages we already have? No strong opinions on my end.

louisjimenez: I could see it living in the [conventions](https://github.com/spinnaker/deck/wiki/Conventions) page or a similar page. I will draft something up on handling async requests and share it.

louisjimenez:
@jtk54 that will follow as part of [adding support](https://github.com/spinnaker/spinnaker/issues/4116) for the configuration page. Similar to how pipelines from v1 templates are provided variables.

jtk54:
👍

PR #6827, opened 2019-04-09
feat(cf): add deployment strategy to clone SG

Jammy-Louie:
spinnaker/spinnaker#4258


Jammy-Louie:
Similar to deploy server group, deployment strategy only appears in pipelines.
55740433-14ab2500-59f9-11e9-9fa5-8b06c574ed50

PR #6829, opened 2019-04-10
 fix(provider/azure): Enable Azure Load Balancer from Azure VM Scale Set

neil-yechenwei:
Background:
Enable Azure Load Balancer which supports inbound NAT port-forwarding rules to connect to azure vm instances.
There is one another related PR (https://github.com/spinnaker/clouddriver/pull/3552) in clouddriver. Please also help to review.


neil-yechenwei:
I've updated all commit title to specified format. Please go ahead to review the code.

PR #6833, opened 2019-04-11
fix(k8s): Fix deploy manifest

jkschneider:
In 1.13, the K8S deploy manifest stage displays the new artifacts UI when the feature flag is enabled. When loading a pipeline defined in this way, artifacts were incorrectly nested one layer too deep.

Also fixed Promise treatment of the React components related to this.

image

See https://github.com/spinnaker/orca/pull/2830 for more details on the JSON structure changes we introduced into 1.13.


jkschneider:

rverma-nikiai:
@jkschneider Can you please share which feature flag is to enable for new artifact UI

PR #6835, opened 2019-04-11
fix(provider/azure): Simplify processed stages of rollback function

neil-yechenwei:
Background:
Simplify processed stages of rollback function.


mybayern1974:
Besides this change working for Azure, there is another reason we make this change: The native roll back action consists of a consecutive steps including Shrink->Disable. We ran into errors between these two stages, whose symptom was that after Shrink succeeded, Disable reported an error complaining it cannot find the newly created Server Group. The roll back action could be normal after waiting for several hours after creating the new Server Group. So we'd like to mitigate this problem first and then continue investigating the reason with keeping the possibility of reverting this change in the future.

PR #6836, opened 2019-04-11
fix(artifacts): HTTP default artifact needs reference field

xavileon:
For the same reason as in a previous PR https://github.com/spinnaker/deck/pull/6679, 
we also need to set the reference field for default http artifacts.

PTAL @jkschneider @jervi


ezimanyi:

PR #6840, opened 2019-04-11
feat(preconfiguredJob): logs for k8s jobs

ethanfrogers:
supports logs for preconfigured kubernetes jobs. titus jobs only link
out to extenal systems for their logs. for kubernetes, logs are stored
in the execution context and can be displayed directly.

## Execution Details
image

## Log display
image

### DIFF
--- a/app/scripts/modules/core/src/pipeline/details/LogsModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/details/LogsModal.tsx
@@ -1,13 +1,13 @@
 import * as React from 'react';
 
-import { IModalComponentProps } from '@spinnaker/core';
+import { IModalComponentProps } from 'core/presentation';
 
-export interface IRunJobLogsModalProps extends IModalComponentProps {
+export interface ILogsModalProps extends IModalComponentProps {
   logs: string;
 }
 
-export class RunJobLogsModal extends React.Component<IRunJobLogsModalProps> {
-  constructor(props: IRunJobLogsModalProps) {
+export class LogsModal extends React.Component<ILogsModalProps> {
+  constructor(props: ILogsModalProps) {
     super(props);
   }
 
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/pipeline/details/LogsModal.tsx
```diff
@@ -1,13 +1,13 @@
...
+import { IModalComponentProps } from 'core/presentation';
 
-export interface IRunJobLogsModalProps extends IModalComponentProps {
+export interface ILogsModalProps extends IModalComponentProps {
   logs: string;
 }
 
-export class RunJobLogsModal extends React.Component<IRunJobLogsModalProps> {
-  constructor(props: IRunJobLogsModalProps) {
+export class LogsModal extends React.Component<ILogsModalProps> {
```
sbwsg: Hey, sorry, just catching up here.  We have a `JobManifestPodLogs` class at `app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/react/JobManifestPodLogs.tsx`. Was that considered and rejected before implementing this one?  It looks somewhat similar - wrapping logs in a \<pre\> tag.  Is there an opportunity to merge these two classes, do you think?

ethanfrogers: @sbwsg Perhaps we could while the Job is executing but not once the Job is completed. `JobManifestPodLogs` depends on the Job (and Pods) existing. Currently, we clean up the Job once it's exited and store the logs in the execution context.

If we wanted to eliminate storing logs in the execution context we _could_ use that component and just stop cleaning up the jobs. The only concern I have is that, if jobs are cleaned up in the future their logs would go away as well.

sbwsg: Discussed on Slack.  Takeaways were:

1. `JobManifestPodLogs` does a bit more than we'd like at the moment.  It reads console output from the job instances and then formats that into a structure for tabbed rendering where each tab corresponds to a container in the job.
2. We'd be interested in implementing a similar tabbed view for the logs in this PR as well but that will need a bit of work in the backend to structure the data in the execution JSON to separate output by container/log.
3. Once we have the logs separated out by container in the data from the b/e we could then factor out the InstanceReader stuff from `JobManifestPodLogs.tsx` and converge with `LogsModal.tsx`.

+1 :shipit:

PR #6845, opened 2019-04-12
feat(preconfiguredJobs): support produce artifacts

ethanfrogers:
adds support for producing artifacts by preconfigured job stages


ethanfrogers:
depends on spinnaker/orca#2935

PR #6847, opened 2019-04-12
fix(kubernetes): safer lookups for deploy stage validators

maggieneterval:
These lookups were previously throwing errors under some conditions


sbwsg:
👍

PR #6848, opened 2019-04-12
 feat(canary) - Add support for Baseline Offsets, bump @kayenta to 0.0.75

perek:
Currently canaries initiated from Orca must always have the same start/end time for all analysis scopes (baseline, canary, etc). This PR adds support for setting an offset value which will time shift the analysis allowing for comparing metrics against a pre-deployment baseline. This PR adds the required new components in deck-kayenta, as well as some new helper fields for enabling support of this new feature.

This PR depends on another PR to be merged and published first due to the dependecy bump of @spinnaker/kayenta to 0.0.75. This PR is required to be merged first: https://github.com/spinnaker/deck-kayenta/pull/422

### DIFF
--- a/app/scripts/modules/canary/acaTask/acaTaskExecutionDetails.html
+++ b/app/scripts/modules/canary/acaTask/acaTaskExecutionDetails.html
@@ -203,6 +203,10 @@
           <div class="col-md-4 sm-label-right compact">Interval</div>
           <div class="col-md-6">{{canaryConfig.canaryAnalysisConfig.canaryAnalysisIntervalMins}} minutes</div>
         </div>
+        <div class="row" ng-if="!canaryConfig.canaryAnalysisConfig.baselineAnalysisOffsetInMins">
+          <div class="col-md-4 sm-label-right compact">Baseline Offset</div>
+          <div class="col-md-6">{{canaryConfig.canaryAnalysisConfig.baselineAnalysisOffsetInMins}}  minutes</div>
+        </div>
         <div class="row">
           <div class="col-md-4 sm-label-right compact">Notification Hours</div>
           <div class="col-md-6">{{canaryConfig.canaryAnalysisConfig.notificationHours.join(', ')}}</div>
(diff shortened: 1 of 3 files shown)


maggieneterval:
I will publish deck-kayenta once your orca and deck-kayenta changes are merged, and then merge this. Thanks @perek!

erikmunson:
@perek Thanks for being thorough enough to find this `acaTask` code and update it! Are you using the `acaTask` internally at all? As far as I know, the `acaTask` code and other stuff in the `canary` folder is actually old, mostly dead code that was really just a front-end for Netflix's internal-only predecessor to Kayenta (through `mine`). The flag you added in spinnaker/orca#2836 lives in the code for the `kayentaCanary` stage, so your `deck-kayenta` PR should be sufficient with no changes needed here.

Unless you're using some kind of configuration internally where the `acaTask` code passes through to a `kayentaCanary` stage, I'd recommend closing this and moving the help text for the baseline offset field into `canary.help.ts` for spinnaker/deck-kayenta#422 — the dependency version bump will get auto-PR'd by a tool we use called [dependabot](https://dependabot.com/) once 0.0.75 is published.

perek:
@erikmunson i needed to edit that acaTask in order to show the offset in the 'Canary Config' in the pipeline stage execution details. I will close and re-open without the autobump

erikmunson:
As far as I know, [the equivalent addition to the execution details for the `kayentaCanary` stage](https://github.com/spinnaker/deck-kayenta/pull/422/files#diff-4a2daff2fb70884b2aefa4d54494a2dcR88) is the thing that will actually do that for you — the only way that `acaTask` execution UI would show up is if you had somehow customized `deck` / `deck-kayenta` to use that execution UI instead of [the one normally registered for `kayentaCanary` stages](https://github.com/spinnaker/deck-kayenta/blob/master/src/kayenta/stages/kayentaStage/kayentaStage.ts#L87).

erikmunson:
@perek and I discussed out-of-band — we confirmed that the changes to the `canary` directory are not needed for spinnaker/deck-kayenta#422 and as a result this PR will be closed. @perek will be moving the help text from this PR into the corresponding deck-kayenta PR for use there.

PR #6856, opened 2019-04-16
feat(cf): Remove React shims from pipeline stages

stuart-pollock:
spinnaker/spinnaker#4284


app/scripts/modules/cloudfoundry/src/presentation/pipeline/stages/CloudfoundryLoadBalancersExecutionDetails.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <StageExecutionLogs stage={props.stage} />
+    </ExecutionDetailsSection>
+  );
+}
+
+// TODO: refactor this to not use namespace
+// eslint-disable-next-line
+export namespace CloudfoundryLoadBalancersExecutionDetails {
+  export const title = 'cloudfoundryLoadBalancersConfig';
+}
```
Jammy-Louie: is this even used? if not we could remove this

stuart-pollock: It appears to be used in core because of https://github.com/spinnaker/deck/blob/1ced961c1e71a8003d4ae93aa04dc44932164b07/app/scripts/modules/core/src/domain/IStageTypeConfig.ts#L6.  Removing this from core is non-trivial and we feel that it is beyond the scope of this story.

jkschneider: Agree with @stuart-pollock. I've also looked into whether it's optional on prior stages and concluded it isn't yet.

PR #6880, opened 2019-04-23
feat(core): UI for configuring pipelines from mptv2

louisjimenez:
This adds UI support for configuring pipelines from MPTv2. It works similar to MPTv1. When a user creates a new pipeline from a template, they are routed to the following modal:

image

This allows the pipeline to inherit templated keys (params, triggers, notifications) and configure any variables. Inherited values can also be modified after creation by following the "Configure" link.


app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.ts
```diff
@@ -0,0 +1,251 @@
+import { module, IController, IScope, IHttpPromiseCallbackArg, IPromise } from 'angular';
```
louisjimenez: Rather than add complexity to the V1 `configurePipelineTemplateModal.controller.ts` and risk breaking support for v1 templates, I have split the v2 logic into its own controller and test files. 

This will make it easier to remove V1 files once the templates are fully deprecated and migrated over. Much of this controller is similar to V1 but I have annotated major differences.

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.ts
```diff
@@ -0,0 +1,251 @@
...
+  ) {}
+
+  public $onInit(): void {
+    this.initialize();
+  }
+
+  public initialize(): void {
+    const { exclude, name, template } = this.pipelineTemplateConfig;
+
+    const excluded = exclude || [];
```
louisjimenez: triggers, notifications, and parameters are now automatically inherited from templates unless the pipeline opts out by defining an exclude array

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.ts
```diff
@@ -0,0 +1,251 @@
...
+      ...(inheritTemplateTriggers ? [] : ['triggers']),
+    ];
+
+    return {
+      ...(pipelineTemplateConfig || {}),
+      name: pipelineName,
+      application: appName,
+      variables: this.transformVariablesForPipelinePlan(),
+      exclude: excludeConfig,
+      ...PipelineTemplateV2Service.getPipelineTemplateConfigV2(source),
```
louisjimenez: The config structure for V2 is different from V1 and relies on the V2 service to get the correct template structure

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.ts
```diff
@@ -0,0 +1,251 @@
...
+      },
+    );
+  }
+
+  private transformVariablesForPipelinePlan(): { [key: string]: any } {
+    return chain(this.variables || [])
+      .cloneDeep()
+      .map(v => {
+        if (v.type === 'object') {
+          v.value = JSON.parse(v.value);
```
louisjimenez: YAML has been replaced with JSON for MPTv2

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.html
```diff
@@ -0,0 +1,54 @@
...
+  </div>
+  <div ng-if="ctrl.state.loadingError" class="alert alert-danger">
+    <p>Could not load pipeline template.</p>
+  </div>
+  <div ng-if="ctrl.state.noVariables && !ctrl.state.error" class="alert alert-info">
+    <p>This template has no variables to configure.</p>
+  </div>
+  <div ng-if="!ctrl.state.error" class="alert alert-info">
+    <strong>Inherit the following configuration from the template</strong>
+    <p><input type="checkbox" ng-model="ctrl.state.inheritTemplateNotifications" /> Notifications</p>
```
louisjimenez: Notifications can now be inherited

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.spec.ts
```diff
@@ -0,0 +1,306 @@
+import { mock, IScope, IQService } from 'angular';
```
louisjimenez: These tests have been brought over from the V1 controller and modified for the MPTv2 controller

app/scripts/modules/core/src/pipeline/config/pipelineConfig.controller.js
```diff
@@ -45,7 +47,14 @@ module.exports = angular
 
         if (this.pipelineConfig && this.pipelineConfig.type === 'templatedPipeline') {
           this.isTemplatedPipeline = true;
-          this.hasDynamicSource = this.containsJinja(this.pipelineConfig.config.pipeline.template.source);
+          this.hasDynamicSource = isV2PipelineConfig
```
maggieneterval:
```suggestion
          this.hasDynamicSource = !isV2PipelineConfig && this.containsJinja(this.pipelineConfig.config.pipeline.template.source);
```

maggieneterval: Might be more readable?

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.ts
maggieneterval: I probably missed this but are there a defined set of possible types somewhere and are they just object, int, and float?

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.spec.ts
maggieneterval: Btw thanks for all these really helpful comments, it would have been much more difficult to review otherwise without any context around this code :)

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.ts
louisjimenez: Great question. There is MPTV2 documentation [here](https://www.spinnaker.io/guides/user/pipeline/pipeline-templates/) but I don't believe the variable types have been documented, looping in @jtk54 to confirm.

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.spec.ts
louisjimenez: glad it is helpful 👍

PR #6885, opened 2019-04-24
fix(core): initialize default artifact with type

jmelchio:
spinnaker/spinnaker#4325


jkschneider:

PR #6890, opened 2019-04-24
fix(kubernetes): fix discrepancy between orca deploy stage model and …

maggieneterval:
…artifacts rewrite stage model

Orca expects `requiredArtifactIds` to be a list of strings ([source](https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/manifest/DeployManifestContext.java#L37)), which is what Deck sends when the artifacts rewrite flag is off.

Before:

before1

before2

before3

After:

after1

after2


maggieneterval:
@ezimanyi @jkschneider while I was testing this I realized I don't precisely understand when we need to explicitly set a "required artifact to bind" from the UI -- it seems like Spinnaker will inject any matching image or ConfigMap in the context into the manifest per this [documentation](https://www.spinnaker.io/reference/artifacts/in-kubernetes-v2/#binding-artifacts-in-manifests), so is setting it as required more to make sure the stage fails if a matching artifact _isn't_ present in the context?

ezimanyi:
Yes, that's my understanding of that field---it basically allows to you to assert that the artifacts listed below need to be bound into the manifest (so you don't accidentally send it along without binding in a tag).

maggieneterval:
:+1: thank you!

PR #6894, opened 2019-04-24
fix(core): Restore config button on exec view and dropdown for mptv2

louisjimenez:
This adds support for the config button when viewing the execution of a mptv2 pipeline:
image

 and also allows mptv2 pipelines to appear in the config dropdown:
image


app/scripts/modules/core/src/pipeline/config/CreatePipeline.tsx
```diff
@@ -23,9 +24,11 @@ export class CreatePipeline extends React.Component<ICreatePipelineProps> {
   public render() {
     const { application } = this.props;
 
-    const pipelineConfigs = get(application, 'pipelineConfigs.data', []).filter(
-      (pipelineConfig: IPipeline) => !PipelineTemplateV2Service.isV2PipelineConfig(pipelineConfig),
-    );
+    const pipelineConfigs = SETTINGS.feature.managedPipelineTemplatesV2UI
+      ? get(application, 'pipelineConfigs.data', [])
+      : get(application, 'pipelineConfigs.data', []).filter(
+          (pipelineConfig: IPipeline) => !PipelineTemplateV2Service.isV2PipelineConfig(pipelineConfig),
```
maggieneterval: maybe replace the ternary with moving the `SETTINGS` check into the filter so that if one were to have MPTv2 enabled at some point but then later disable it, the MPTv2 configs wouldn't show up?

louisjimenez: I was thinking that toggling the flag on and off would cause the next component render to check for the flag and if true just return all configs or if not return configs with mptv2 ones filtered out. Is that off though, does it work differently in spinnaker?

maggieneterval: ah sorry i'm being dumb, all good :+1:

PR #6801, opened 2019-04-02
feat(provider/azure): Add redblack strategy for azure

neil-yechenwei:
Background:
Add redblack strategy for azure provider.
There is another related PR (https://github.com/spinnaker/clouddriver/pull/3533) with same name in clouddriver for azure redblack strategy. Please also help to review.

PR #6896, opened 2019-04-25
refactor(provider/azure): convert rollback function from angularjs to react

neil-yechenwei:
Background:
Previously we added four newly files for new azure feature "redblack". At last time, the newly files are written by angularjs (Previous PR: https://github.com/spinnaker/deck/pull/6801 ). According by the suggestion from @sbwsg , so we converted the previous newly files from angularjs to React in this PR. Please help to review it.


app/scripts/modules/azure/serverGroup/details/rollback/RollbackServerGroupModal.tsx
```diff
@@ -0,0 +1,162 @@
...
+  };
+
+  private isValid = () => {
+    const restoreServerGroupName = this.state.command.rollbackContext.restoreServerGroupName;
+    return restoreServerGroupName !== undefined;
+  };
+
+  private handleServerGroupChange = (restoreServerGroupOption: any) => {
+    const { disabledServerGroups } = this.props;
+    const newCommand = this.state.command;
```
sbwsg: It is more typical in React to make a copy of the object when making changes to state, like this:

```jsx
const newCommand = { ...this.state.command };
// make changes to newCommand
this.setState({
  command: newCommand
});
```

I believe this is for performance - it is faster for the library to check equality between two different objects than it is to check equality between all the primitive fields of two objects.  I'm not 100% sure how Deck approaches this however and might be wrong about this.

sbwsg:
I'm going to merge as-is.  The feedback I've left above can be tackled in a separate commit.

PR #6901, opened 2019-04-26
feat(roles): Add execute permission type to create new application.

dibyom:
Applications can now have a new permission type called execute along with read and write. The application creation via the UI ensures that execute always implies read access and write always includes execute (and read).

Part of spinnaker/spinnaker#3554


christopherthielen:
closing and reopening to see if travis picks it up

PR #6916, opened 2019-05-01
refactor(runJob/kubernetes): use joblogviewer

ethanfrogers:
instead of reading logs from execution context (which will be removed
soon), use the job log viewer modal to grab logs directly from the
infrastructure. large portions of code stolen directly from the
DeployStatus component.

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/manifest.service.ts
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/manifest.service.ts
@@ -4,6 +4,15 @@ export interface IManifestContainer {
   manifest: IManifest;
 }
 
+export interface IStageManifest {
+  kind: string;
+  apiVersion: string;
+  metadata: {
+    namespace: string;
+    name: string;
+  };
+}
+
 export interface IManifestParams {
   account: string;
   location: string;
@@ -31,5 +40,17 @@ export class KubernetesManifestService {
 
   private static updateManifest(params: IManifestParams, fn: IManifestCallback) {
     ManifestReader.getManifest(params.account, params.location, params.name).then(manifest => fn(manifest));
+  }
+
+  public static manifestIdentifier(manifest: IStageManifest) {
+    const kind = manifest.kind.toLowerCase();
+    // manifest.metadata.namespace doesn't exist if it's a namespace being deployed
+    const namespace = (manifest.metadata.namespace || '_').toLowerCase();
+    const name = manifest.metadata.name.toLowerCase();
+    const apiVersion = (manifest.apiVersion || '_').toLowerCase();
+    // assuming this identifier is opaque and not parsed anywhere. Including the
+    // apiVersion will prevent collisions with CRD kinds without having any visible
+    // effect elsewhere
+    return `${namespace} ${kind} ${apiVersion} ${name}`;
   }
 }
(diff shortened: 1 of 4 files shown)


app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/RunJobExecutionDetails.tsx
```diff
@@ -1,43 +1,131 @@
...
-
-      const modalProps = { dialogClassName: 'modal-lg modal-fullscreen' };
-      ReactModal.show(
-        LogsModal,
-        {
-          logs: get(fullStage, 'context.jobStatus.logs', 'No log output found.'),
-        } as ILogsModalProps,
-        modalProps,
-      );
+  constructor(props: IExecutionDetailsSectionProps) {
```
maggieneterval: can replace the constructor with the following since `props` is not used to initialize state:
```suggestion
    public state = { subscription: { id: '', unsubscribe: null, manifest: null }, manifestId: '' };
```

app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/RunJobExecutionDetails.tsx
```diff
@@ -1,43 +1,131 @@
...
+
+  public componentDidMount() {
+    this.componentDidUpdate(this.props, this.state);
+  }
+
+  public componentWillUnmount() {
+    this.unsubscribe();
+  }
+
+  public componentDidUpdate(_prevProps: IExecutionDetailsSectionProps, prevState: IRunJobExecutionDetailsState) {
```
maggieneterval: Perhaps rename the logic in `componentDidUpdate` to a more descriptive private method to call from both `componentDidMount` and `componentDidUpdate`?

maggieneterval: Ah I see that this is just copied from `DeployStatus.tsx`.

ethanfrogers: I did :) I can refactor this, though, if it makes more sense.

ethanfrogers:
@sbwsg I moved `manifestIdentifier` and `IStageManifest` over to `KubernetesManifestService` but left `subscribeToManifestUpdates` in both components since the body of each is different enough (getting the name, updating the subscription).

app/scripts/modules/kubernetes/src/v2/manifest/manifest.service.ts
```diff
@@ -12,6 +21,18 @@ export interface IManifestParams {
 
 export type IManifestCallback = (manifest: IManifest) => void;
 
+export function manifestIdentifier(manifest: IStageManifest) {
```
sbwsg: v minor nit: It's possibly more Deck-like to include this as a static method on the service class.

ethanfrogers: Updated!

sbwsg: Nice. :+1:

ethanfrogers:
Travis doesn't seem to have picked up this PR. I'm going to close and resubmit.

PR #6917, opened 2019-05-01
refactor(runJob/kubernetes): use joblogviewer

ethanfrogers:
instead of reading logs from execution context (which will be removed
soon), use the job log viewer modal to grab logs directly from the
infrastructure. large portions of code stolen directly from the
DeployStatus component.


ethanfrogers:
@maggieneterval @sbwsg this is the same as the last PR but resubmitted so that the build got picked up.

PR #6924, opened 2019-05-02
refactor(runJob/kubernetes): refactor exec details

ethanfrogers:
refactor some of the manifest components into the core module and use
those instead of the kubernetes specific ones. doing this because
`StageExecutionLogs` in the `core` module needs to use
`JobManifestPodLogs`.


ethanfrogers:
I essentially pulled all of the logic that I added in #6917 into a common component that could be used across both places. I also was able to revert `StageExecutionLogs` back to its original state because I had to pull some of the logic up a layer in order to avoid changing a ton of Cloud Foundry usages of `StageExecutionLogs`.

maggieneterval:
:100: thanks Ethan!

PR #6926, opened 2019-05-03
refactor(core): add React Components for PageSection and PageNavigator

Jammy-Louie:
spinnaker/spinnaker#4337

create React Components for PageSection and PageNavigator to replace the angular usages in Stages and Triggers


Jammy-Louie:
@christopherthielen any thoughts?

PR #6927, opened 2019-05-03
fix(core/api): normalize API request urls

cfieber:
This only handles the case of a consumer of ApiService requesting with a leading
slash character. Full uri normalization may be a useful thing to add to handle
a request containing directory traversal characters but I don't suspect we have
any of those whereas the double-slash case is pretty easy to accidentally do

### DIFF
--- a/app/scripts/modules/core/src/api/ApiService.spec.ts
+++ b/app/scripts/modules/core/src/api/ApiService.spec.ts
@@ -11,13 +11,46 @@ describe('API Service', function() {
   beforeEach(
     mock.inject(function(_$httpBackend_: ng.IHttpBackendService) {
       $httpBackend = _$httpBackend_;
-      baseUrl = SETTINGS.gateUrl;
+      baseUrl = API.baseUrl;
     }),
   );
 
   afterEach(function() {
+    SETTINGS.resetToOriginal();
     $httpBackend.verifyNoOutstandingExpectation();
     $httpBackend.verifyNoOutstandingRequest();
+  });
+
+  describe('ensure api requests generate normalized urls', function() {
+    let expected: ng.IRequestConfig;
+    beforeEach(() => {
+      expected = {
+        method: '',
+        url: '',
+      };
+    });
+
+    it('trims leading slashes from urls', function() {
+      const result = API.one('/foo');
+      expected.url = `${baseUrl}/foo`;
+      expect(result.config).toEqual(expected);
+    });
+
+    it('trims repeated leading slashes from urls', function() {
+      const result = API.one('/////foo');
+      expected.url = `${baseUrl}/foo`;
+      expect(result.config).toEqual(expected);
+    });
+
+    it('trims trailing slashes from baseUrl', function() {
+      SETTINGS.gateUrl = 'http://localhost/';
+      expect(API.baseUrl).toEqual('http://localhost');
+    });
+
+    it('trims repeated trailing slashes from baseUrl', function() {
+      SETTINGS.gateUrl = 'http://localhost/////';
+      expect(API.baseUrl).toEqual('http://localhost');
+    });
   });
 
   describe('validate response content-type header', function() {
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/api/ApiService.spec.ts
```diff
@@ -20,6 +20,38 @@ describe('API Service', function() {
...
+      const result = API.one('/////foo');
+      expected.url = `${baseUrl}/foo`;
+      expect(result.config).toEqual(expected);
+    });
+
+    it('trims trailing slashes from baseUrl', function() {
+      const originalUrl = SETTINGS.gateUrl;
+      SETTINGS.gateUrl = 'http://localhost/';
+      const result = API.one('foo');
+      SETTINGS.gateUrl = originalUrl;
```
cfieber: feels like there is a better way to do this, but this test would side-effect all the other tests unless I cleaned up this change to SETTINGS.gateUrl but I don't really understand the mocking/testing framework well enough to figure out how to customize the top level beforeEach that is setting the config to the value of SETTINGS.gateUrl just for this one test....

app/scripts/modules/core/src/api/ApiService.ts
```diff
@@ -164,9 +164,9 @@ export class API {
   private static init(urls: string[]) {
```
cfieber: you'd think the `toString()` on these was optional, but there were a few tests that found callers that passing in url objects instead of Strings

cfieber:
context for this: newer spring-boot / spring-security installs a http request filter that rejects fishy request URLs do prevent a reported CVE against the framework and we had a case where we were generating a `//` in one of our internal integrations. I could fix the source but it does seem like an easy mistake to make and fixing it here seems better?

app/scripts/modules/core/src/api/ApiService.ts
anotherchrisberry: We should fix those tests and not call `toString` here. The app won't work if `gateUrl` is not a string; this kind of implies it could.

anotherchrisberry: Okay, we talked through this, Cam is half right

app/scripts/modules/core/src/api/ApiService.ts
```diff
@@ -164,9 +164,9 @@ export class API {
...
     const config: IRequestConfig = {
       method: '',
-      url: SETTINGS.gateUrl,
+      url: SETTINGS.gateUrl.toString().replace(/\/+$/, ''),
```
anotherchrisberry: omit this one, the one below is great!

app/scripts/modules/core/src/api/ApiService.spec.ts
anotherchrisberry: Slightly better way: in the `afterEach` method of this test, call `SETTINGS.resetToOriginal();`

app/scripts/modules/core/src/api/ApiService.ts
cfieber: done (and moved down to the baseUrl getter)

app/scripts/modules/core/src/api/ApiService.ts
cfieber: I'm a glass-is-half-right kind of guy

app/scripts/modules/core/src/api/ApiService.spec.ts
cfieber: I added that, though I also change the mock's baseUrl to `API.baseUrl` so this becomes unnecessary in this test for it to continue to pass (but feels like still the right thing to do if I'm mucking with SETTINGS)

app/scripts/modules/core/src/api/ApiService.ts
christopherthielen: > Okay, we talked through this, Cam is half right

What was the secret discussion?

anotherchrisberry: lol that was so long ago

PR #6928, opened 2019-05-03
feat(cloudformation): support cfn tags

xavileon:
This patch allows specifying tags through the UI. Backend support in
clouddriver for tags is added in this [PR].

[PR]: https://github.com/spinnaker/clouddriver/pull/3625


xavileon:
@spinnaker/ui-ux-team PTAL

PR #6929, opened 2019-05-03
feat(provdier/google): Display serviceAccount properties for firewall rules

seanpeters86:
This PR adds basic support for displaying target and source service accounts to firewall rules. These function similarly as the existing tags display, but cannot be correlated to other rules like tags can. 

Initial support for https://github.com/spinnaker/spinnaker/issues/4332

Screen Shot 2019-05-03 at 4 11 42 PM


app/scripts/modules/google/src/securityGroup/details/securityGroupDetail.controller.js
```diff
@@ -89,6 +89,20 @@ module.exports = angular
                 $scope.securityGroup.sourceTags = sourceTags.substring(1, sourceTags.length - 1).split(', ');
               }
 
+              // These come back from the global firewall endpoint as '[account-a@project.iam.gserviceaccount.com, account-b@project.iam.gserviceaccount.com]'
+              if (typeof $scope.securityGroup.targetServiceAccounts === 'string') {
```
maggieneterval: if these don't come back as a string, are they `undefined`? asking because if so we'll get an error when we try to read `.length` off them in the template

seanpeters86: @maggieneterval this block is basically a duplicate of the work done in the `sourceTags` and `targetTags` block above, so I'm not totally sure on this. I believe in the case of both tags and serviceAccounts if they don't exist they ARE undefined. Couple of examples in gate:

# Tag rule
```
{
id: "host/platform-vault",
moniker: {
app: "platform",
cluster: "platform-vault-egress-allow",
detail: "vault-egress-allow",
stack: "services"
},
name: "platform-vault-egress-allow",
network: "host/host-network",
selfLink: "https://www.googleapis.com/compute/beta/projects/prod-host/global/firewalls/platform-vault-egress-allow",
targetTags: "[vault-server]"
},
```

# Service Account rule
```
{
id: "host/platform-vault",
moniker: {
app: "platform",
cluster: "platform-vault-egress-allow",
detail: "vault-egress-allow",
stack: "services"
},
name: "platform-vault-egress-allow",
network: "host/host-network",
selfLink: "https://www.googleapis.com/compute/beta/projects/prod-host/global/firewalls/platform-vault-egress-allow",
targetServiceAccounts: "[vault@prod-app-vault.iam.gserviceaccount.com]"
}
```

Any suggestions on how to go forward would be great! Pretty new to this stuff :)

app/scripts/modules/google/src/securityGroup/details/securityGroupDetail.html
```diff
@@ -144,6 +144,22 @@ <h3 class="horizontal middle space-between flex-1" select-on-dbl-click>
       </ul>
       <div ng-if="!securityGroup.sourceTags.length">None</div>
     </collapsible-section>
+    <collapsible-section heading="Target Service Accounts ({{securityGroup.targetServiceAccounts.length || 0}})">
```
jtk54: Is there ever a case where both tags and service accounts are present on a security group? My understanding is that it's an either-or or not present for 'default' security groups. Would it be better to have some conditional rendering instead of showing at least one empty accordion?

app/scripts/modules/google/src/securityGroup/details/securityGroupDetail.controller.js
maggieneterval: Ahh I'm going to chalk this up to angular magic because you're right, this works fine even when each is undefined!

app/scripts/modules/google/src/securityGroup/details/securityGroupDetail.html
jtk54: Discussed with @maggieneterval and actually keeping both present (even if they are empty) is clearer from a UX perspective.

maggieneterval: @jtk54 and I just chatted offline and decided to hold off on this conditional rendering for now in case it causes more confusion than it's worth.

Thanks for the PR, @seanpeters86!

PR #6930, opened 2019-05-06
feat(runJob/kubernetes): render external link

ethanfrogers:
if `context.execution.logs` is present (pulled from manifest annotation
`jobs.spinnaker.io/logs`), render the content as a link, templated with
manifest data.

Depends on spinnaker/orca#2893.

### DIFF
--- a/app/scripts/modules/core/src/manifest/stage/JobStageExecutionLogs.tsx
+++ b/app/scripts/modules/core/src/manifest/stage/JobStageExecutionLogs.tsx
@@ -5,7 +5,7 @@ import { IStageManifest, ManifestService } from '../ManifestService';
 import { JobManifestPodLogs } from './JobManifestPodLogs';
 import { IManifest } from 'core/domain/IManifest';
 
-import { get } from 'lodash';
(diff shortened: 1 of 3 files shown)


ethanfrogers:
@sbwsg @maggieneterval would love your input on this. Like the Annotation Driven UI, if Orca pulls `jobs.spinnaker.io/logs` off of the manifest, we display the link in the UI for log output. 

I'm making a couple of assumptions here:
1. If you are using that annotation you prefer to view logs from an external system. 
2. You can build a link to this external system purely based off of the data in the manifest.

Also, I went with `lodash.template` since this is React (the Annotation Driven UI uses Angular interpolation) but used `{{ }}` as delimiters. Not sure if we have another preference here or not.

maggieneterval:
> Finally, I implemented pulling the annotation off in Orca to align with what Titus is doing, but, on second thought, I could probably just do that here. Thoughts?

I'm in favor of keeping the UI as dumb as possible so slightly prefer keeping it the way you've done but don't have a super strong opinion!

PR #6935, opened 2019-05-06
feat(gcb): accept gcb definition file as artifact

maggieneterval:
- Allows GCB stage config to accept build definition as an artifact when `ARTIFACTS_REWRITE_ENABLED=true`. Otherwise, only accepts build definition as text and hides artifact selector. Accepting pre-rewrite artifacts in a React component without introducing Angular dependencies is non-trivial and will be addressed in a separate PR.
- Modifies function returned by `ExpectedArtifactService.accumulateArtifacts` to accept nested fields (i.e. `keyA.keyB`) in order to display consumed artifacts nested in context
- Refactors GCB stage to more closely adhere to new `FormikStageConfig` pattern and cancelable promise convention

Artifact defined inline:
96ZABYr4DGX

Artifact defined in pipeline config:
5UFvcdFBPtm

Consumed artifact (when artifact is defined in pipeline config):
umKTphNGnn8

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
@@ -1,0 +1,129 @@
+import * as React from 'react';
+import Select, { Option } from 'react-select';
+import { get } from 'lodash';
+
+import {
+  ArtifactTypePatterns,
+  IArtifact,
+  IExpectedArtifact,
+  IFormikStageConfigInjectedProps,
+  RadioButtonInput,
+  SETTINGS,
+  StageArtifactSelector,
+  StageConfigField,
+  yamlDocumentsToString,
+  YamlEditor,
+} from '@spinnaker/core';
+
+export enum buildDefinitionSources {
+  ARTIFACT = 'artifact',
+  TEXT = 'text',
+}
+
+interface IGoogleCloudBuildStageFormProps {
+  googleCloudBuildAccounts: string[];
+}
+
+interface IGoogleCloudBuildStageFormState {
+  rawBuildDefinitionYaml: string;
+}
+
+export class GoogleCloudBuildStageForm extends React.Component<
+  IGoogleCloudBuildStageFormProps & IFormikStageConfigInjectedProps,
+  IGoogleCloudBuildStageFormState
+> {
+  public constructor(props: IGoogleCloudBuildStageFormProps & IFormikStageConfigInjectedProps) {
+    super(props);
+    const stage = props.formik.values;
+    this.state = {
+      rawBuildDefinitionYaml: stage.buildDefinition ? yamlDocumentsToString([stage.buildDefinition]) : '',
+    };
+  }
+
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/googleCloudBuildStage.ts
```diff
@@ -12,5 +11,6 @@ Registry.pipeline.registerStage({
   producesArtifacts: true,
   component: GoogleCloudBuildStageConfig,
   executionDetailsSections: [GoogleCloudBuildExecutionDetails, ExecutionDetailsTasks, ExecutionArtifactTab],
-  validators: [{ type: 'requiredField', fieldName: 'account' }],
+  validateFn: validate,
+  artifactExtractor: ExpectedArtifactService.accumulateArtifacts(['buildDefinitionArtifact.artifactId']),
```
maggieneterval: @ezimanyi do we need an accompanying `artifactRemover` here? I wasn't actually sure why we would need to remove certain artifacts from the context but it seemed to be defined on the stages that included an extractor.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -0,0 +1,129 @@
...
+      { value: buildDefinitionSources.TEXT, label: 'Text' },
+      { value: buildDefinitionSources.ARTIFACT, label: 'Artifact' },
+    ];
+  };
+
+  private onSourceChange = (e: any) => {
+    this.props.formik.setFieldValue('buildDefinitionSource', e.target.value);
+  };
+
+  private onExpectedArtifactSelected = (expectedArtifact: IExpectedArtifact) => {
```
maggieneterval: @jkschneider I tagged you on this to make sure I'm using your lovely new component as intended -- we need to post either 1) just an artifact ID if a pre-existing artifact is selected or 2) a complete artifact if defined inline, correct?

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/googleCloudBuildStage.ts
ezimanyi: The remover handles cleaning up references to artifacts if you remove the artifact from the pipeline so that stages don't have references to artifact id's that don't exist.  The thing to test here would be deleting the expected github artifact in the trigger; ideally this would clear out the artifact in the GCB stage rather than leave an invalid one.

maggieneterval: Ah, that makes sense, thanks Eric. Added an extractor which won't work yet because the extractors don't seem to be called post-artifacts-rewrite, so I'm going to address that in a followup PR as well as supporting nested references.

maggieneterval: Followup PR: https://github.com/spinnaker/deck/pull/6939

PR #6937, opened 2019-05-07
fix(bake/manifest): Preserve artifact account selection

german-muzquiz:
This fixes a bug in Bake Manifest stage configuration. If the input artifacts have more than one account, UI shows the user a dropdown for selecting the desired account. However this selection is lost every time the user navigates to another screen/stage and comes back to Bake Manifest stage configuration, falling back always to the first artifact account shown in the dropdown.

Screen Recording 2019-05-07 at 10 55 AM

The bug is caused because `InputArtifact` objects are created before having the list of artifact accounts ready. This has the effect of clearing the selected account, and when the account list is initialized the selected account fallbacks to the first in the list.


maggieneterval:
Woohoo! I've been meaning to address this, @german-muzquiz, nice fix! Going to patch into 1.12 and 1.13.

Closes https://github.com/spinnaker/spinnaker/issues/4339

maggieneterval:

maggieneterval:

PR #6948, opened 2019-05-08
feat(gcb): allow pre-artifacts-rewrite expected artifacts to be selec…

maggieneterval:
…ted in gcb stage

- Abstracts pre-rewrite artifact selection and inline creation into new React component, `PreRewriteStageArtifactSelector`. In order to avoid introducing angular dependencies to this component, expose `updatePipeline` prop to React stage components. Saving an artifact inline needs to add it to the parent pipeline's list of `expectedArtifacts`. This approach feels "bad" so if anyone has a more elegant solution please let me know. (:cry:) Existing components handle this by mutating `$scope.parent.pipeline` directly via angular delegates/controllers. 
- We will be able to remove `updatePipeline` once we deprecate pre-rewrite artifact handling. In the meantime, it will unblock us from removing the angular artifact delegates and controllers as dependencies from all Kubernetes stages that handle artifacts.

xBnbHHZHy4A

RcipPGWcRiu


maggieneterval:
@ezimanyi FYI this works for me both pre- and post-artifacts rewrite without ever defining `buildDefinitionArtifact.artifactAccount` -- seems the ID is sufficient?

PR #6959, opened 2019-05-09
fix(core/executions): make text on truncated params selectable

erikmunson:
In the new parameter UI for executions, we truncate parameter values over a certain size. Turns out that the toggle for showing/hiding the truncated value was setup to fire when clicking _any part of the value_, not just the 'View Full' link, and that makes it hard/impossible to select text out of a long value (like a JSON object) to copy-paste.

This changes to an always-visible 'Show more' / 'Show less' toggle, and removes click handling on the text itself. I had to do some weird layout hacking with `line-height` to get the link to work, which as far as I can tell is caused by the table layout these parameters are using (the clickable area seems to get truncated by the table row element somehow).

In my heart of hearts I'd like to move to using flex layouts here as I think that would fix these weird layout problems, but don't really want to go down the rabbit hole right now.

Before:
Screen Shot 2019-05-09 at 11 03 57 AM

After:
Screen Shot 2019-05-09 at 11 05 13 AM

### DIFF
--- a/app/scripts/modules/core/src/pipeline/status/ExecutionParameters.tsx
+++ b/app/scripts/modules/core/src/pipeline/status/ExecutionParameters.tsx
@@ -16,11 +16,7 @@ export interface IDisplayableParameter {
   valueTruncated?: string;
 }
 
-interface IExecutionParametersState {
-  toggle: boolean;
-}
-
-export class ExecutionParameters extends React.Component<IExecutionParametersProps, IExecutionParametersState> {
+export class ExecutionParameters extends React.Component<IExecutionParametersProps> {
   constructor(props: IExecutionParametersProps) {
     super(props);
   }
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/status/ExecutionParameters.tsx
```diff
@@ -29,7 +25,7 @@ export class ExecutionParameters extends React.Component<IExecutionParametersPro
     if (parameter.valueTruncated) {
       return () => {
         parameter.showTruncatedValue = !parameter.showTruncatedValue;
-        this.setState({ toggle: parameter.showTruncatedValue });
+        this.forceUpdate();
```
erikmunson: AFAICT this `setState` call was used exclusively to force a render when this parameter object was mutated, which gets confusing if you don't know that already. This is behaviorally the same but makes the intent more obvious.

PR #6961, opened 2019-05-09
fix(artifacts): Allow defaulting in execution artifact

ezimanyi:
Fixes spinnaker/spinnaker#4344.  (This will also require a small change to orca a the orca task does not support looking up a default artifact.)

Find artifact from execution doesn't use the full expected artifact component, which means users can't assign a name or use a default artifact. Upgrade it to the full expected artifact component so
it gets these features.

Also fix the field defaulting; currently the fields will reset if you navigate away and back to the stage before saving.

Also add the "Artifact Status" tab to this stage so it's easier to see the artifact that was bound.


ezimanyi:
Pushed a 1-line fix to remove the trash can icon which doesn't do anything in this stage

PR #6971, opened 2019-05-10
fix(executions): URI Encode pipeline names

alanmquach:
Some people (I won't mention who), don't like bracket so we need to encode them or Gate will blow up.


cfieber:
it was me

cfieber:
Any reason not to just stick the URI encoding in ApiService and encode all the path segments there for all requests?

alanmquach:
Haha, I meant boot2?

Yeah, unfortunately there is already some encoding going on in random spots so if I were to add it in ApiService, we'd end up with some double encoding which would be bad.

This was more of a quick fix. We should probably do a proper sweep and push the encoding down to ApiService and remove it everywhere else.

PR #6978, opened 2019-05-13
feat(runjob): capture output ui

ethanfrogers:
UI for v2 run job capture output. there are 2 options - logs and
artifact. setting to `logs` will set the `propertyFile` option to
trigger the runJob stage to read from the container logs. setting to
artifact will trigger the `ConsumeArtifactTask` to fetch and inject the
defined artifact.

## Output selection (logs)
image

## Output selection (artifact)
image

## Output selection (none)
image

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
+++ b/app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
@@ -1,6 +1,19 @@
 import * as React from 'react';
+import Select from 'react-select';
 
-import { IStageConfigProps, AccountService, YamlEditor, yamlDocumentsToString, IAccount } from '@spinnaker/core';
+import {
+  IStageConfigProps,
+  AccountService,
+  YamlEditor,
+  yamlDocumentsToString,
+  IAccount,
+  StageArtifactSelector,
+  SETTINGS,
+  IExpectedArtifact,
+  IArtifact,
+  PreRewriteStageArtifactSelector,
+  StageConfigField,
+} from '@spinnaker/core';
 
 import { ManifestBasicSettings } from 'kubernetes/v2/manifest/wizard/BasicSettings';
 
@@ -13,6 +26,12 @@ export class KubernetesV2RunJobStageConfig extends React.Component<IStageConfigP
   public state: IKubernetesRunJobStageConfigState = {
     credentials: [],
   };
+
+  public outputOptions = [
+    { label: 'None', value: 'none' },
+    { label: 'Logs', value: 'propertyFile' },
+    { label: 'Artifact', value: 'artifact' },
+  ];
 
   public accountChanged = (account: string) => {
     this.props.updateStageField({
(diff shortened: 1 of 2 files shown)


app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
```diff
@@ -14,6 +27,11 @@ export class KubernetesV2RunJobStageConfig extends React.Component<IStageConfigP
     credentials: [],
   };
 
+  public findSelectedArtifact(id: string): IExpectedArtifact {
```
maggieneterval: where is this used?

app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
```diff
@@ -43,9 +61,94 @@ export class KubernetesV2RunJobStageConfig extends React.Component<IStageConfigP
     this.initRawManifest();
   }
 
+  public sourceChanged = (event: any) => {
```
maggieneterval: I think all these public methods can be private?

app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
```diff
@@ -43,9 +61,94 @@ export class KubernetesV2RunJobStageConfig extends React.Component<IStageConfigP
...
+  }
+
   public render() {
     const { application, stage } = this.props;
 
+    let outputSource = <div />;
+    if (stage.consumeArtifactSource === 'propertyFile') {
+      outputSource = this.logSourceForm();
+    } else if (stage.consumeArtifactSource === 'artifact') {
+      outputSource = this.checkFeatureFlag('artifactsRewrite') ? this.artifactRewriteForm() : this.artifactForm();
```
maggieneterval: This is more a note for me, but wondering if you think it would be useful to create a component that abstracts away the feature flag check and conditional rendering of pre and post rewrite artifacts?

maggieneterval: (Not at all a note for this PR, just something I might add later)

app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
```diff
@@ -57,6 +160,17 @@ export class KubernetesV2RunJobStageConfig extends React.Component<IStageConfigP
         />
         <h4>Manifest Configuration</h4>
         <YamlEditor value={this.state.rawManifest} onChange={this.handleRawManifestChange} />
+        <h4>Output</h4>
+        <StageConfigField label="Capture Output From" helpKey="kubernetes.runJob.captureSource">
+          <div>
+            <select className="form-control input-sm" value={stage.consumeArtifactSource} onChange={this.sourceChanged}>
```
maggieneterval: just for consistency with the other React selects, would you mind using the `Select` component from `react-select`?

app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
ethanfrogers: A remnant of a previous approach. Will remove.

app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
ethanfrogers: yes :)

app/scripts/modules/kubernetes/src/v2/pipelines/stages/runJob/KubernetesV2RunJobStageConfig.tsx
ethanfrogers: can do!

PR #6991, opened 2019-05-14
fix(*): Fix imports

ezimanyi:
The deck build is currently failing as non-allowed imports are present; fix these to import submodules


ezimanyi:
Fixes issue from #6970.  It looks like we don't run the full build when testing a PR, so that's why travis passed on the PR but our build process is failing.

ezimanyi:
It looks like we skip for performance reasons:
https://github.com/spinnaker/deck/blob/ea23876e4a02a5d8f3418bae93e8aca1fad08fbf/gradle/buildViaTravis.sh#L5
though maybe it makes sense to re-enable the full build to avoid only catching things after they get merged?

christopherthielen:
Weird, the travis PR build actually did fail.

https://travis-ci.org/spinnaker/deck/builds/532531524

christopherthielen:
oh, got it!  This was the `master` build ☝️

ezimanyi:
Yeah, it passed on the PR itself (though did output a warning about this somewhere deep in the hundreds of lines of output)

PR #7012, opened 2019-05-16
refactor(kubernetes): namespace selector component

ethanfrogers:
create a namespace selector component which can be reused. optional
creatable functionality to support cases where SpEL is allowable.


app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/NamespaceSelector.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  private getNamespaceOptions(): Array<Option<string>> {
+    const { accounts, selectedAccount } = this.props;
+    const selectedAccountDetails = find(accounts, a => a.name === selectedAccount);
+    const namespaces = get(selectedAccountDetails, 'namespaces', []);
+    const options = map(namespaces, n => ({ label: n, value: n }));
+    if (this.props.createable && !includes(namespaces, this.props.selectedNamespace)) {
+      options.push({ label: this.props.selectedNamespace, value: this.props.selectedNamespace });
+    }
+    return options;
```
ethanfrogers: This block of code isn't very elegant. I'm sure there's a cleaner way to do this.

app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/NamespaceSelector.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export class NamespaceSelector extends React.Component<INamespaceSelectorProps> {
+  public defaultProps = { createable: false };
+
+  private getNamespaceOptions(): Array<Option<string>> {
+    const { accounts, selectedAccount } = this.props;
+    const selectedAccountDetails = find(accounts, a => a.name === selectedAccount);
+    const namespaces = get(selectedAccountDetails, 'namespaces', []);
+    const options = map(namespaces, n => ({ label: n, value: n }));
+    if (this.props.createable && !includes(namespaces, this.props.selectedNamespace)) {
```
maggieneterval: In the situation where a user has selected a namespace available in a particular account, and then changes the account to one where that namespace is no longer an option, we will keep that namespace available as an option, which might be confusing. Wonder if there's any way to designate a particular option as being user-created and only persist those between account re-selects?

ethanfrogers: Oh, that's a good catch. I didn't consider that. What if we just limited it to options containing `${`? That should satisfy the use case without having to worry about the above. The option would only persist if you're using SPeL.

maggieneterval: love it, ship it

ethanfrogers:
@maggieneterval updated the check to `this.props.creatable && selectedNamespace.includes('${')`

app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/NamespaceSelector.tsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  public render() {
+    const componentProps = {
+      clearable: false,
+      options: this.getNamespaceOptions(),
+      value: this.props.selectedNamespace,
+      onChange: (option: Option) => this.props.onChange(option.value.toString()),
+    };
+    return <>{this.props.createable ? <Creatable {...componentProps} /> : <Select {...componentProps} />}</>;
```
maggieneterval:
```suggestion
    return this.props.createable ? <Creatable {...componentProps} /> : <Select {...componentProps} />;
```
I don't think wrapping in a fragment is necessary here

PR #7016, opened 2019-05-16
feat(provider/kubernetes): namespace deployManifest

ethanfrogers:
adds fields for namespaceOverride to allow users to define the namespace
their manifests are deployed into (particularly for manifests baked from
Helm charts).

## Default - no override
image

## Namespace override enabled
image

## Override enabled w/ SpEL expression
image

### DIFF
--- a/app/scripts/modules/kubernetes/src/v2/manifest/wizard/BasicSettings.tsx
+++ b/app/scripts/modules/kubernetes/src/v2/manifest/wizard/BasicSettings.tsx
@@ -19,7 +19,7 @@ export function ManifestBasicSettings({
   selectedAccount,
 }: IManifestBasicSettingsProps) {
   return (
-    <div className="container-fluid form-horizontal">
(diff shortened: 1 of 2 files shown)


app/scripts/modules/kubernetes/src/v2/manifest/wizard/BasicSettings.tsx
```diff
@@ -19,7 +19,7 @@ export function ManifestBasicSettings({
   selectedAccount,
 }: IManifestBasicSettingsProps) {
   return (
-    <div className="container-fluid form-horizontal">
```
maggieneterval: was this class just not doing anything?

ethanfrogers: It was causing some alignment issues if things were added underneath it in other forms.

PR #7015, opened 2019-05-16
fix(core): Backward compatibility for old application permissions

jervi:
Applications saved before #6901 and spinnaker/fiat#373 does not have EXECUTE permission set, and in some cases this breaks Deck. This commit fixes so that the old READ/WRITE permission is converted to a READ/EXECUTE/WRITE permission.
Also adds EXECUTE support to the applicationAttributes-ng-directive.

When this PR is merged, Deck will support old style permissions without EXECUTE, and the next time one saves the application, EXECUTE will be added (if the old permissions were READ/WRITE).

Current behaviour:

Application saved with read/write:
image
And the permissions are set as follows in the application definition:
```json
        "permissions": {
            "READ": [
                "spt-delivery-team"
            ],
            "WRITE": [
                "spt-delivery-team"
            ]
        }
```

When opening the edit modal, the rendering fails and Permissions are not displayed at all:
image

The stacktrace:
```
[Error] TypeError: undefined is not an object (evaluating 'permissions.EXECUTE.forEach')
	getPermissionRows (PermissionsConfigurer.tsx:48)
	getState (PermissionsConfigurer.tsx:36)
	PermissionsConfigurer (PermissionsConfigurer.tsx:26)
	constructClassInstance (react-dom.development.js:11361)
	updateClassComponent (react-dom.development.js:14687)
	performUnitOfWork (react-dom.development.js:19312)
	workLoop (react-dom.development.js:19352)
	callCallback (react-dom.development.js:149)
	dispatchEvent
	invokeGuardedCallbackDev (react-dom.development.js:199)
	invokeGuardedCallback (react-dom.development.js:256)
	replayUnitOfWork (react-dom.development.js:18578)
	renderRoot (react-dom.development.js:19468)
	performWorkOnRoot (react-dom.development.js:20342)
	performWork (react-dom.development.js:20254)
	performSyncWork (react-dom.development.js:20228)
	requestWork (react-dom.development.js:20097)
	scheduleWork (react-dom.development.js:19911)
	scheduleRootUpdate (react-dom.development.js:20572)
	render (react-dom.development.js:20953)
	(anonym funksjon) (react-dom.development.js:21090)
	legacyRenderSubtreeIntoContainer (react-dom.development.js:21086)
	render (index.js:60)
	$onChanges (index.js:28)
	(anonym funksjon) (angular.js:10022)
	forEach (angular.js:426)
	nodeLinkFn (angular.js:10018)
	compositeLinkFn (angular.js:9311)
[...]

PR #7017, opened 2019-05-16
fix(authz): Handle apps without execute permissions

dibyom:
PermissionsConfigurer always assumed apps had EXECUTE permissions set
which is not true for applications created before EXECUTE was a thing.

A more detailed description is here: https://github.com/spinnaker/deck/pull/7015#issue-279537064

Thanks @jervi !!

### DIFF
--- a/app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
+++ b/app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
@@ -35,7 +35,7 @@ export class PermissionsConfigurer extends React.Component<IPermissionsConfigure
     { value: 'READ,EXECUTE', label: 'Read and execute' },
     { value: 'READ,EXECUTE,WRITE', label: 'Read, execute, write' },
   ];
-  private static legacyAccessTypes: Option[] = [{ value: 'READ,WRITE', label: 'Read, and write' }];
+  private static legacyAccessTypes: Option[] = [{ value: 'READ,WRITE', label: 'Read and write' }];
 
   constructor(props: IPermissionsConfigurerProps) {
     super(props);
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
```diff
@@ -188,8 +192,9 @@ export class PermissionsConfigurer extends React.Component<IPermissionsConfigure
     return (
       <div className="permissions-configurer">
         {this.state.permissionRows.map((row, i) => {
-          const permissionTypeLabel = PermissionsConfigurer.accessTypes.find(type => type.value === row.access).label;
-
+          const permissionTypeLabel = []
+            .concat(PermissionsConfigurer.accessTypes, PermissionsConfigurer.legacyAccessTypes)
```
louisjimenez: nit: I think the es6 syntax is a little cleaner here but what do you think? `[...PermissionsConfigurer.accessTypes, ...PermissionsConfigurer.legacyAccessTypes]`

app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
```diff
@@ -188,8 +192,9 @@ export class PermissionsConfigurer extends React.Component<IPermissionsConfigure
...
+            .find(type => type.value === row.access).label;
```
louisjimenez: are there cases where we would not expect to find anything and get `undefined`, if so reading `label` could throw an  error

app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
```diff
@@ -35,6 +35,7 @@ export class PermissionsConfigurer extends React.Component<IPermissionsConfigure
     { value: 'READ,EXECUTE', label: 'Read and execute' },
     { value: 'READ,EXECUTE,WRITE', label: 'Read, execute, write' },
   ];
+  private static legacyAccessTypes: Option[] = [{ value: 'READ,WRITE', label: 'Read, and write' }];
```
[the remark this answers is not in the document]
dibyom: Done!

app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
dibyom: I concur! Thanks!

app/scripts/modules/core/src/application/modal/PermissionsConfigurer.tsx
dibyom: No, the label should always be something that is within the `accessTypes` or `legacyAccessTypes`

dibyom:

PR #7018, opened 2019-05-17
fix(core/pipeline): fix type mismatch in pipeline trigger, broken webhook trigger

erikmunson:
few fixes for the new react-powered pipeline and webhook triggers:

- the status chcecklist for for pipeline triggers wasn't transforming the `Set`s coming from `Checklist` to arrays, which got masked by the fact that the `onUpdateTrigger` method is typed with `any`. End result was that once a status was changed the underlying JSON was malformed and made the trigger component un-renderable on the next navigation to the same view
- fixed up some minor things in `PipelineTrigger` like not triggering the `destroy$` subject on unmount and not using the pre-existing `status` list if one existed
- webhook triggers were un-renderable in all cases due to using a standard instance method instead of an arrow function

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/triggers/pipeline/PipelineTrigger.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/pipeline/PipelineTrigger.tsx
@@ -1,49 +1,41 @@
 import * as React from 'react';
 import Select, { Option } from 'react-select';
 
-import { Observable } from 'rxjs';
+import { Observable, Subject } from 'rxjs';
 
-import { ApplicationReader } from 'core/application/service/ApplicationReader';
-import { BaseTrigger } from 'core/pipeline';
+import { Application, ApplicationReader } from 'core/application';
+import { BaseTrigger, PipelineConfigService } from 'core/pipeline';
 import { IPipeline, IPipelineTrigger } from 'core/domain';
-import { PipelineConfigService } from 'core/pipeline/config/services/PipelineConfigService';
+import { Omit } from 'core/presentation';
 import { Checklist } from 'core/forms';
-import { Application } from '@spinnaker/core';
-import { Subject } from 'rxjs/Subject';
+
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/pipeline/config/triggers/pipeline/PipelineTrigger.tsx
```diff
@@ -1,64 +1,60 @@
...
+import { Application, ApplicationReader } from 'core/application';
+import { BaseTrigger, PipelineConfigService } from 'core/pipeline';
 import { IPipeline, IPipelineTrigger } from 'core/domain';
-import { PipelineConfigService } from 'core/pipeline/config/services/PipelineConfigService';
+import { Omit } from 'core/presentation';
 import { Checklist } from 'core/forms';
-import { Application } from '@spinnaker/core';
-import { Subject } from 'rxjs/Subject';
+
+type IPipelineTriggerConfig = Omit<IPipelineTrigger, 'parentExecution' | 'parentPipelineId' | 'user' | 'rebake'>;
```
erikmunson: this was a maybe-temporary way of accurately typing what's really an acceptable input to `onTriggerUpdated`. I'd like to investigate how feasible having config-only trigger interfaces would be so we can use those in configs rather than the interfaces for triggers that show up in executions.

app/scripts/modules/core/src/pipeline/config/triggers/pipeline/PipelineTrigger.tsx
```diff
@@ -1,64 +1,60 @@
...
 
 export interface IPipelineTriggerConfigProps {
   status: string[];
   trigger: IPipelineTrigger;
   application: Application;
   pipelineId: string;
-  triggerUpdated: (trigger: IPipelineTrigger) => void;
+  triggerUpdated: (trigger: IPipelineTriggerConfig) => void;
```
erikmunson: this is fine for now because the thing that passes in `triggerUpdated` is a vanilla JS file 🤷‍♂

PR #7022, opened 2019-05-17
fix(kubernetes): Fix NPE in bake manifest details

ezimanyi:
If the bake manifest stage doesn't output any artifacts, the stage fails to render due to an NPE. This is probably rare as you'd generally want at least one artifact, but still worth fixing.


ezimanyi:

maggieneterval:
Nice catch, thanks Eric! :smile:

PR #7027, opened 2019-05-17
fix(cf): server group header build links should precede images

claymccoy:
In the (currently unlikely) event that images and build info are both available, the images should not be first.


jkschneider:

PR #7029, opened 2019-05-20
fix(kubernetes): fix runjob stage init

ethanfrogers:
fix stage initialization by setting the stage's application config if it
doesn't exist. this field is required to be set for the stage to work.
initially i had assumed the that BasicSettings component would configure
it.


ethanfrogers:

PR #7030, opened 2019-05-20
fix(artifact/helm): fix version list

ethanfrogers:
helm chart versions weren't being listed properly because the chart name
was being pulled from the previous value. so, the list of versions you
got back were for a previously selected chart name instead of the newly
selected chart name.


ethanfrogers:
cc @jkschneider @erikmunson

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/helm/HelmArtifactEditor.tsx
```diff
@@ -54,7 +54,7 @@ class HelmEditor extends React.Component<IArtifactEditorProps, IHelmArtifactEdit
             value={artifact.name || ''}
             onChange={(e: Option) => {
               this.onChange(e, 'name');
-              this.onNameChange();
+              this.onNameChange(e.value.toString());
```
maggieneterval: Is the value not already a string?

ethanfrogers: Probably? I was just being consistent with other places in the file.

PR #7033, opened 2019-05-20
fix(artifacts/helm): support regex/spel in version

ethanfrogers:
the version field in an artifact supports regex but the previous UI
forced a hardcoded version. this made triggers really confusing because
users are forced to set a version on the artifact when they really
wanted something like `.*` to indicate that this chart should be matched
for all versions.


ethanfrogers:
This only handles _most_ of the cases where this can be set. I'm not sure what to do about `helm.artifact.ts` under `Expected Artifacts` w/ the artifact rewrite disabled. That bit is still Angular but I need a creatable select. Open to suggestions.

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/helm/HelmArtifactEditor.tsx
```diff
@@ -4,71 +4,98 @@ import { Option } from 'react-select';
 import { ArtifactTypePatterns } from 'core/artifact';
 import { IArtifact, IArtifactEditorProps, IArtifactKindConfig } from 'core/domain';
 import { StageConfigField } from 'core/pipeline';
-import { TetheredSelect } from 'core/presentation';
-
+import { TetheredSelect, TetheredCreatable } from 'core/presentation';
 import { ArtifactService } from '../ArtifactService';
 import { cloneDeep } from 'lodash';
```
maggieneterval: I know this wasn't you but while you're in here would you mind moving this import up into the same block as the other third-party imports?

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/helm/HelmArtifactEditor.tsx
```diff
@@ -4,71 +4,98 @@ import { Option } from 'react-select';
...
+    names: [],
+    versions: [],
+    versionsLoading: true,
+    namesLoading: true,
+  };
+
   constructor(props: IArtifactEditorProps) {
     super(props);
-    if (props.artifact.type !== TYPE) {
+    const { artifact } = this.props;
```
maggieneterval: If we're going to destructure this, let's go ahead and replace `props.artifact` with `artifact` below on line 35

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/helm/HelmArtifactEditor.tsx
```diff
@@ -4,71 +4,98 @@ import { Option } from 'react-select';
...
   }
 
+  public componentDidMount() {
+    const { artifact } = this.props;
+    if (artifact.name) {
+      this.getChartVersionOptions(artifact.name);
+    }
+  }
+
   public componentWillReceiveProps(nextProps: IArtifactEditorProps) {
```
maggieneterval: I know you didn't add this but just as an FYI for future, let's avoid `componentWillReceiveProps` since it will be deprecated with React 17: https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

app/scripts/modules/core/src/pipeline/config/triggers/artifacts/helm/HelmArtifactEditor.tsx
```diff
@@ -80,11 +107,21 @@ class HelmEditor extends React.Component<IArtifactEditorProps, IHelmArtifactEdit
...
-  private onNameChange = (chartName: string) => {
-    ArtifactService.getArtifactVersions(TYPE, this.props.account.name, chartName).then(versions => {
-      this.setState({ versions });
+  private getChartVersionOptions(chartName: string) {
+    const { artifact, account } = this.props;
+    this.setState({ versionsLoading: true });
+    ArtifactService.getArtifactVersions(TYPE, account.name, chartName).then((versions: string[]) => {
+      // if the version doesn't match SEMVER we assume that it's a regular expression or SpEL expression
+      // and add it to the list of valid versions
+      if (artifact.version && !new RegExp(SEMVER).test(artifact.version)) {
```
maggieneterval: Might possibly be cleaner to create the RegExp only once as a private member of this class?

maggieneterval:
> This only handles _most_ of the cases where this can be set. I'm not sure what to do about `helm.artifact.ts` under `Expected Artifacts` w/ the artifact rewrite disabled. That bit is still Angular but I need a creatable select. Open to suggestions.

My suggestion would be converting it to React :angel:

ethanfrogers:
@maggieneterval Feedback addressed! Unfortunately, refactoring that bit to React is more work than I have time for at the moment. At least from what I can tell. I'll revisit again, later.

PR #7042, opened 2019-05-21
feat(cf): Accepting both JSON and YAML in the Deploy Service Configuration stage `parameters` field

pdelagrave:
Implements: spinnaker/spinnaker#4361
Corresponding clouddriver patch: spinnaker/clouddriver#3686


app/scripts/modules/cloudfoundry/src/pipeline/stages/deployService/cloudfoundryDeployServiceStage.module.ts
```diff
@@ -52,30 +53,20 @@ PipelineConfigValidator.registerValidator(
...
         return true;
+      } catch (e) {
+        return false;
       }
+    }
 
+    private isYaml(value: string): boolean {
       try {
-        JSON.parse(content);
+        safeLoad(value, null);
```
Jammy-Louie: nit picky - you don't need to specify a second parameter
`safeLoad(value);` 
would work fine

PR #7060, opened 2019-05-24
fix(runJob/kubernetes): reliably display logs

ethanfrogers:
* fix(runJob/kubernetes): fix log display

  fix a case where the log link was being displayed before all the data
  needed to render it was available. this caused some confusing cases where
  the link would be rendered but users would be met with a Not Found error in
  the popup. ultimately this was due to a missing namespace. resolved by
  pulling the namespace off of the job status instead of depending on the
  manifest subscription.
 
* fix(runJob/kubernetes): fetch manifest once

  jobs collect owned pods when collecting job status so we are no longer
  dependent on subscribing to manifest events to get the pod names created by
  the job. instead, simply fetch the manifest once. this is more reliable
  than subscribing to manifest updates.

### DIFF
--- a/app/scripts/modules/core/src/manifest/stage/JobStageExecutionLogs.tsx
+++ b/app/scripts/modules/core/src/manifest/stage/JobStageExecutionLogs.tsx
@@ -15,6 +15,7 @@ interface IJobStageExecutionLogsProps {
   application: Application;
   externalLink: string;
   podNameProvider: IPodNameProvider;
+  location: string;
 }
 
 interface IJobStageExecutionLogsState {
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/manifest/stage/JobStageExecutionLogs.tsx
```diff
@@ -15,77 +15,25 @@ interface IJobStageExecutionLogsProps {
...
-      account: this.props.account,
-      name: this.props.deployedName,
-      location: manifest.metadata.namespace == null ? '_' : manifest.metadata.namespace,
-    };
-    return ManifestService.subscribe(this.props.application, params, (updated: IManifest) => {
-      const subscription = { ...this.state.subscription, manifest: updated };
-      this.setState({ subscription });
+    const { account, location, deployedName } = this.props;
+    ManifestReader.getManifest(account, location, deployedName).then(manifest => {
+      this.setState({ manifest });
```
maggieneterval: perhaps use the cancelable promise pattern here? https://github.com/spinnaker/deck/wiki/Conventions#setting-state-in-promises

ethanfrogers:

PR #7064, opened 2019-05-28
fix(core): Disable rewriteLinks to allow proper event handling

alanmquach:
Otherwise, in certain scenarios Angular ends up preventing react from being able to properly handle the link's `click` event, causing some unexpected behavior.

This should cause any issues, but if it does, we should try to fix those issues if we can instead of simply reverting this change.


christopherthielen:
here is the click handler in question: 

https://github.com/angular/angular.js/blob/864c7f0/src/ng/location.js#L894-L937

PR #7065, opened 2019-05-28
feat(provider/azure): Enable dns label

neil-yechenwei:
Background:
Currently dns label is specified with default value while creating and updating dns label. So we add an entry to specify the custom dns label while creating and updating dns label. There is a related PR in clouddriver (https://github.com/spinnaker/clouddriver/pull/3722).


neil-yechenwei:
@sbwsg , a kindly ping, please help to review it. Thanks.

app/scripts/modules/azure/loadBalancer/loadBalancer.transformer.js
```diff
@@ -62,6 +62,9 @@ module.exports = angular
           }
 
           toEdit.probes = elb.probes;
+          if (elb.dnsName && elb.dnsName != 'dns-not-found') {
```
louisjimenez: can this use strict inequality instead? `!==`

neil-yechenwei: Sure. Updated code.

neil-yechenwei: Sure. Updated code.

PR #7067, opened 2019-05-29
fix(triggers): Poll on execution id instead of event id

ezimanyi:
The V2 manual trigger endpoint only returned an event id, so deck had to poll for the appearance of a new trigger by polling on the event id.

Now that the endpoint also returns the execution id, we can poll on the execution id which should be significantly less expensive (for both orca and the redis/sql back-end).


ezimanyi:
For some context, this is mostly restoring the old code path that was removed in #6680.

This PR depends on spinnaker/gate#811.

PR #7071, opened 2019-05-31
feat(cfn/changesets): Introduce support for CFN changesets

xavileon:
This patch introduces a flag into the deploy cloud formation UI to
chose wether the deployment operation will create/update a full CFN
stack or create a change set instead. If the checkbox is ticked, a
text form appear to allow the user specify the change set name.

Depends on:
https://github.com/spinnaker/clouddriver/pull/3731
https://github.com/spinnaker/orca/pull/2950


app/scripts/modules/amazon/src/pipeline/stages/deployCloudFormation/deployCloudFormationStackConfig.html
```diff
@@ -6,67 +6,90 @@ <h4>Basic Settings</h4>
...
       <stage-config-field label="Source" help-key="aws.cloudformation.source">
         <label class="sm-label-right">
-          <input type="radio" ng-model="ctrl.$scope.stage.source" value="{{ctrl.textSource}}">
+          <input type="radio" ng-model="ctrl.$scope.stage.source" value="{{ctrl.textSource}}" />
           {{ ctrl.textSource | robotToHuman }}
-          <span ng-if="stage.source === ctrl.textSource">
-          </span>
-        </label><br/>
+          <span ng-if="stage.source === ctrl.textSource"> </span> </label
+        ><br />
```
maggieneterval: curious if this was prettier changing the spacing here? looks kinda odd :man_shrugging:

xavileon: yes, it was intellij doing some aesthetics on my behalf without me noticing. I fixed that odd spacing 👍

PR #7077, opened 2019-06-03
refactor(core/presentation): Consolidate Checklist and ChecklistInput components

christopherthielen:
This PR consolidates two react components that do basically the same thing.  `ChecklistInput` is compatible with `FormikFormField` so I used that one.  I added `inline` rendering mode and a select/deselect all button from the `Checklist` component.  I refactored all usages of `Checklist` that I found


app/scripts/modules/core/src/presentation/forms/inputs/ChecklistInput.spec.tsx
```diff
@@ -0,0 +1,126 @@
+import * as React from 'react';
+import { mount } from 'enzyme';
+
+import { ChecklistInput } from './ChecklistInput';
+
+const noop = () => {};
+
+fdescribe('<ChecklistInput />', () => {
```
christopherthielen: whops.

app/scripts/modules/core/src/presentation/forms/inputs/ChecklistInput.spec.tsx
```diff
@@ -0,0 +1,126 @@
+import * as React from 'react';
+import { mount } from 'enzyme';
+
+import { ChecklistInput } from './ChecklistInput';
+
+const noop = () => {};
+
+describe('<ChecklistInput />', () => {
```
christopherthielen: thank Justin for these tests

PR #7081, opened 2019-06-04
fix(provider/azure): Fail to run pipeline at WaitForUpInstance task

neil-yechenwei:
Background:
Currenly, it failed to run pipeline at WaitForUpInstance task and threw exception "Cannot cast object 'false' with class 'java.lang.Boolean' to class 'java.lang.Integer'". After investigated, I found the root cause is that orca will convert the capacity object to Map<String,Integer> but the azure code in deck would pass the property "useSourceCapacity" which is boolean value type under capacity object so that it failed. After checked, the property "useSourceCapacity" shouldn't be setted up under capacity object like other providers. And this property is unused in azure clouddriver.

Fix:
So I submitted this PR to remove the property "useSourceCapacity" from capacity object and add to outer layer of capacity under stage data.

### DIFF
--- a/app/scripts/modules/azure/serverGroup/serverGroup.transformer.js
+++ b/app/scripts/modules/azure/serverGroup/serverGroup.transformer.js
@@ -74,8 +74,8 @@ module.exports = angular
         vnet: command.vnet,
         vnetResourceGroup: command.selectedVnet.resourceGroup,
         subnet: command.subnet,
+        useSourceCapacity: false,
         capacity: {
-          useSourceCapacity: false,
           min: command.sku.capacity,
           max: command.sku.capacity,
         },


mybayern1974:
> Would you like me to cherry-pick this fix into 1.14?

Thank you @maggieneterval for this proactive reminder! Yes please help on that if the required effort is not heavy. We did not call out this initially because this issue had been hidden until below commit with its PR got merged targeting at v1.15. 
- commit: https://github.com/spinnaker/orca/commit/5fe370db9f632ef062f8d89a35a1618b02c07bac
- PR: https://github.com/spinnaker/orca/pull/2923

So if the above won't enter into v1.14 hotfix, then things will be safe. While whatever assume making this change here to v1.14 hotfix will do no harm.

maggieneterval:
@mybayern1974 got it, thanks for explaining! If I'm understanding correctly, this fix addresses a bug introduced by the feature in the PR you linked, so since we generally do not cherry-pick features, I am going to leave both the feature and this fix on master. Thanks!

PR #7084, opened 2019-06-05
fix(deps): hoist appropriate version of d3-format

erikmunson:
Turns out that prior to this change yarn implicitly hoisted an older version of `d3-format` (1.2.0). `deck-kayenta` doesn't really like a version that low, which is why it asked for a higher one, so as a result reports have been blowing up whenever they want to render a metric graph.

I don't see a good path forward today for coercing yarn to hoist the more desirable version without explicitly asking for it, so here we are.


maggieneterval:

PR #7095, opened 2019-06-06
fix(kubernetes): fix req. artifacts to bind selector in patch manifes…

maggieneterval:
…t stage
Closes https://github.com/spinnaker/spinnaker/issues/4496

This has been broken for at least the last two releases, probably went undetected because requiring an artifact to bind in the patch stage is relatively rare? Will patch back to 1.13 and 1.14.

Before:
OLRqbqoLLy8

After:
DsVgXv33Wjr


maggieneterval:

maggieneterval:

PR #7102, opened 2019-06-10
fix(core): Display template inherited items (mptv2) as read only

louisjimenez:
MPTv1 saved inherited items (notifications, triggers, parameters) directly to the pipeline config which allowed them to be displayed while configuring in deck. MPTv2 pipelines saved in spin do not save them to the pipeline config and instead rely on them being inherited during the plan phase. This PR renders items inherited from the plan and marks them as read only since they cannot be edited or deleted directly through the pipeline config.

https://github.com/spinnaker/spinnaker/issues/4451

Inherited Trigger
image

Inherited Notification
image

Inherited Parameter
image

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/parameters/Parameter.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/parameters/Parameter.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as classNames from 'classnames';
 
 import { HelpField } from 'core/help';
 import { Tooltip } from 'core/presentation';
@@ -14,6 +15,7 @@ export interface IParameter {
   default: string;
   hasOptions: boolean;
   options: IParameterOption[];
+  inherited: boolean;
 }
 
 export interface IParameterOption {
@@ -89,7 +91,18 @@ export class Parameter extends React.Component<IParameterProps> {
   };
 
   public render(): JSX.Element {
-    const { name, label, required, pinned, description, hasOptions, options, isMultiple, removeParameter } = this.props;
+    const {
+      name,
+      label,
+      required,
+      pinned,
+      description,
+      hasOptions,
+      options,
+      isMultiple,
+      removeParameter,
+      inherited,
+    } = this.props;
 
     const {
       addOption,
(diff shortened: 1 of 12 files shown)


louisjimenez:
screenshots incoming

app/scripts/modules/core/src/pipeline/config/parameters/Parameter.tsx
```diff
@@ -107,113 +121,124 @@ export class Parameter extends React.Component<IParameterProps> {
...
-                    </div>
-                  </div>
-                </div>
-              </div>
-
-              <StageConfigField label="Required">
-                <div className="checkbox">
-                  <label>
-                    <input type="checkbox" checked={required} onChange={handleRequiredChange} />
+            <fieldset disabled={inherited} className={classNames({ 'templated-pipeline-item': inherited })}>
```
louisjimenez: this large diff comes from indenting pre-existing elements in the added `<fieldset>` container and from making the delete button conditional to inheritance (line 154)

app/scripts/modules/core/src/pipeline/config/templates/v2/configurePipelineTemplateModalV2.controller.ts
```diff
@@ -115,19 +112,14 @@ export class ConfigurePipelineTemplateModalV2Controller implements IController {
...
-        } = this.state;
+        const { inheritTemplateParameters, inheritTemplateNotifications, inheritTemplateTriggers } = this.state;
 
         const configWithInheritedValues = {
           ...config,
           ...(inheritTemplateParameters && parameterConfig ? { parameterConfig } : {}),
           ...(inheritTemplateNotifications && notifications ? { notifications } : {}),
           ...(inheritTemplateTriggers && triggers ? { triggers } : {}),
-          ...(inheritTemplateExpectedArtifacts && expectedArtifacts ? { expectedArtifacts } : {}),
+          ...(expectedArtifacts ? { expectedArtifacts } : {}),
```
louisjimenez: As of this [PR](https://github.com/spinnaker/orca/commit/ca1a5269bf20ad5807dcadce20611fedeb6202f8), inheritance of expectedArtifacts changed from opt in to always inherited. Currently only parameters, triggers, and notifications can be marked for exclusion.

app/scripts/modules/core/src/pipeline/config/triggers/trigger.html
```diff
@@ -1,102 +1,111 @@
...
-              <select
-                class="input-sm"
-                ng-change="triggerCtrl.loadTrigger()"
-                ng-options="option.key as option.label for option in options"
-                ng-model="trigger.type"
-              >
-                <option value="">Select...</option>
-              </select>
-              {{description}}
+    <fieldset ng-disabled="trigger.inherited" ng-class="{'templated-pipeline-item': trigger.inherited}">
```
louisjimenez: Similar to Parameter.tsx changes, much of this is shifting elements and Prettier. Inheritance related changes are on lines 3, 22-32, 55

louisjimenez:

PR #7104, opened 2019-06-11
refactor(core): reactify pipelineRoles component

Jammy-Louie:


Jammy-Louie:
Refactored pipeline roles to React:

Screen Shot 2019-06-11 at 10 30 15 AM

Much smaller change then my last PR

app/scripts/modules/core/src/pipeline/config/triggers/PipelineRoles.tsx
```diff
@@ -0,0 +1,45 @@
...
+export interface IPipelineRolesConfigProps {
+  roles: any[];
+  updateRoles: (roles: string[]) => void;
+}
+
+export const PipelineRoles = (props: IPipelineRolesConfigProps) => {
+  const [allowedRoles, setAllowedRoles] = useState([]);
+
+  useEffect(() => {
+    setAllowedRoles(AuthenticationService.getAuthenticatedUser().roles);
```
maggieneterval: If my understanding is correct, `useEffect` will be called after each render -- do we need to fetch the allowed roles repeatedly? I only ask because it appears that in the controller we're replacing, this is only called on init.

maggieneterval: Oh, perhaps we could use this pattern to only conditionally call `useEffect`: https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect
I also might be misunderstanding, just trying to read up on this stuff recently!

Jammy-Louie: Good Idea, I set the conditional to an empty array which should only trigger once.

app/scripts/modules/core/src/pipeline/config/triggers/PipelineRoles.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export interface IPipelineRolesConfigProps {
+  roles: any[];
+  updateRoles: (roles: string[]) => void;
+}
+
+export const PipelineRoles = (props: IPipelineRolesConfigProps) => {
+  const [allowedRoles, setAllowedRoles] = useState([]);
+
+  useEffect(() => {
```
christopherthielen: I don't think you need a `useEffect` at all...  since getAuthenticatedUser() is synchronous you can either:

1) simply grab the value each time
2) cache it using `useMemo`
```const allowedRoles = useMemo(() => AuthenticationService.getAuthenticatedUser().roles, []);```

PR #7105, opened 2019-06-11
fix(google): allow SpEL in stack/detail in deploy stage

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4324

- You *can* use SpEL in stack/detail fields in Deploy stages, but you *cannot* use SpEL in stack/detail fields in ad-hoc Deploy/Clone operations. We set `command.viewState.templatingEnabled = true` in deploy state initialization methods shared by other providers but not by GCE, so previously we were incorrectly disallowing SpEL in stack/detail fields in Deploy stages.


maggieneterval:
@plumpy Not sure why I can't add you as a reviewer :sob:

maggieneterval:

PR #7107, opened 2019-06-11
refactor(core): reactify overrideFailure component

Jammy-Louie:
* Reactify overrideFailure component
* Updated `RadioButtonInput` to allow a `HelpField` to be added next to the label to preserve existing UI


app/scripts/modules/core/src/pipeline/config/stages/overrideFailure/OverrideFailure.tsx
```diff
@@ -0,0 +1,96 @@
...
+        completeOtherBranchesThenFail: false,
+      });
+    } else if (value === 'faileventual') {
+      props.updateStageField({
+        failPipeline: false,
+        continuePipeline: true,
+        completeOtherBranchesThenFail: true,
+      });
+    }
+    setFailureOption(value);
```
maggieneterval: Curious why we chose to store the failure option on state when it can be derived from props?

Jammy-Louie: You're are right

only reason I stored it state was because there was a state object in the old angular code https://github.com/spinnaker/deck/pull/7107/commits/60259bca4983a3df82db6f6e1effdcc451175280#diff-d25eb78f41d61032957a3b187bffed50L13

kept the same logic as I was continuing to refactor. 

I've made the change to derive it from props. Thanks for the suggestion

PR #7111, opened 2019-06-12
fix(core): Make template table list scrollable

louisjimenez:
https://github.com/spinnaker/spinnaker/issues/4519
Previously, the table did not scroll and hid table entries.

image


louisjimenez:

PR #7115, opened 2019-06-13
perf(google): avoid unnecessary fetching and filtering of gce images

maggieneterval:
Addresses https://github.com/spinnaker/spinnaker/issues/4510

- [perf] Removes unnecessary client-side alphabetization of image names: Clouddriver already alphabetizes image results [here](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/controllers/GoogleNamedImageLookupController.java#L118).
- [perf] Removes client-side filtration of images by account; adds account as image request parameter.
- [ux enhancement] Filters options by search text in image select dropdowns rather than only highlighting the matching text without actually removing non-matching options.
- [perf] Since we have already fetched all available account images to populate the persistent disk configuration, do not re-fetch images for Basic Settings component. Also do not re-fetch images as user types a search query; rely on client-side filtration only.

Upcoming work: server-side performance enhancements

### DIFF
--- a/app/scripts/modules/google/src/serverGroup/configure/serverGroupConfiguration.service.js
+++ b/app/scripts/modules/google/src/serverGroup/configure/serverGroupConfiguration.service.js
@@ -90,7 +90,7 @@ module.exports = angular
         } else {
           imageLoader = command.viewState.imageId
             ? loadImagesFromImageName(command)
-            : loadImagesFromApplicationName(application, command.selectedProvider);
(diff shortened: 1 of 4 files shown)


maggieneterval:
cc @plumpy :smile_cat:

maggieneterval:
Updated this PR to ensure images are re-fetched when a new account is selected, and opened corresponding Clouddriver PR here: https://github.com/spinnaker/clouddriver/pull/3785
:smile:

PR #7125, opened 2019-06-18
fix(pipeline): fix invisible parameter when default is not in options

alanmquach:
In the pipeline stage's stage config, if a parameter has options and the parameter's default is not part of the options, it will be invisible.

This fix injects the default into the list of options and also fixes the invisibility (because it is an option, it no longer is invisible).

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/pipeline/pipelineStage.js
+++ b/app/scripts/modules/core/src/pipeline/config/stages/pipeline/pipelineStage.js
@@ -93,6 +93,11 @@ module.exports = angular
               $scope.stage.pipelineParameters = {};
             }
             $scope.pipelineParameters = config.parameterConfig;
+            $scope.pipelineParameters.forEach(parameterConfig => {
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/pipeline/pipelineStage.js
```diff
@@ -93,6 +93,11 @@ module.exports = angular
               $scope.stage.pipelineParameters = {};
             }
             $scope.pipelineParameters = config.parameterConfig;
+            $scope.pipelineParameters.forEach(parameterConfig => {
+              if (parameterConfig.default && parameterConfig.options && !parameterConfig.options.some(option => option.value === parameterConfig.default)) {
+                parameterConfig.options.unshift({ value: parameterConfig.default });
```
christopherthielen: doesn't need a `label`?

alanmquach: Nope. `ui-select` deals with objects and we decide however we want to render the object.

In this particular case, we're just doing
```
<div><strong>{{option.value}}</strong></div>
```
so no need to duplicate `value` and `label`.

PR #7126, opened 2019-06-18
refactor(core): Reactify overrideTimeout

Jammy-Louie:


Jammy-Louie:
Converting the `Override Timeout` to react

Screen Shot 2019-06-18 at 9 46 58 AM

app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
```diff
@@ -0,0 +1,138 @@
...
+  defaultTimeoutMs: number;
+}
+
+export const OverrideTimeout = (props: IOverrideTimeoutConfigProps) => {
+  const [hours, setHours] = useState(0);
+  const [minutes, setMinutes] = useState(0);
+  const [overrideTimeout, setOverrideTimeout] = useState(false);
+  const [configurable, setConfigurable] = useState(false);
+  const [defaults, setDefaults] = useState({ hours: 0, minutes: 0 });
+  const helpContent = HelpContentsRegistry.getHelpField('pipeline.config.timeout');
```
maggieneterval: Thanks so much for doing this, Jammy!

I am having a little bit of a hard time understanding the data flow here -- is there a reason we need to store all these values on state and handle updating the stage config with `useEffect`? My (naive) understanding is that `useEffect` is mainly intended for more side-effect-y operations previously done in `componentDidMount` / `componentDidUpdate`, and that whenever possible it's a bit more straightforward to directly call methods to update the model in response to whatever action should change the data. In general I feel like components like this should be as dumb as possible, receive the minimum props necessary to display what they need to display, and not store anything on state that can be derived from the props they receive. But very open to new paradigms now that we have hooks and super curious for @christopherthielen's wisdom here!

Jammy-Louie: The reason I needed `useEffect` was when a user chooses a different stage type in a different component it would result in different `stageConfig` (https://github.com/spinnaker/deck/pull/7126/commits/b8bd46c215bce9944ccdef6a0a28d7ad38511f8c#diff-f0f41e28f10844cbe8077af8dfc669edR255)

which would cause the trigger of `setOverrideValues`

You are right, I think I could rework it a little bit to have it store fewer things in state, I was just focused on lifting the previous logic implemented and porting it over.

christopherthielen: > I was just focused on lifting the previous logic implemented and porting it over.

I think this is a good approach, when possible.  There are definitely different paradigms when using angular and react (and the libraries within each ecosystem).  I'd like to shift our codebase to use more react-friendly paradigms.  However, if we do both things in one commit (port from angular to react *and* refactor), it can be very difficult to later track down regressions in behavior. I believe it's safer long term to migrate the existing logic first in one commit, and then later simplify in another.

maggieneterval: Got it, appreciate the explanation Jammy! :+1:

christopherthielen: In this particular case, the usage of `useEffect` looks appropriate to me.  We want to be able to respond to changes to the stageConfig` and perform some "expensive" computations.  This aligns with both [Don't stop the data flow](https://overreacted.io/writing-resilient-components/#principle-1-dont-stop-the-data-flow) and [Lazy initial state](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state).

maggieneterval: > I believe it's safer long term to migrate the existing logic first in one commit, and then later simplify in another.

Makes sense to me! And thanks for the links, will read :book:

PR #7131, opened 2019-06-19
chore(core): clarify clone stage help text

maggieneterval:
Addresses half of https://github.com/spinnaker/spinnaker/issues/4354

- Adds clarifying help text to "Clone Stage" dialog to prevent user confusion surrounding cloning stages that include artifacts.
- A future PR will allow selection of multiple artifacts in the Find Artifact from Execution stage to address root cause of users using "clone" to create multiple similar stages targeting different artifacts from the same execution.


maggieneterval:
@dorbin Feel free to suggest alternate text!

PR #7134, opened 2019-06-20
feat(core/presentation): Always call onBlur in Checklist to "mark as touched"

christopherthielen:
Also: mark `stringOptions` as `readonly` to allow read only arrays of options to be provided.


christopherthielen:
Validation errors don't surface until a form field is touched.  Since this is a composite input component, we can't easily use the `onBlur` paradigm that formik relies on.  

I don't know if this is a good idea or not.  I can think of two alternate options:

- use the `FormikFormField` `touched` and force it to true at a higher level
- Keep track of the focus state of all the separate checkboxes and roll them up to a meta-focus state, and trigger `onBlur` when appropriate.

PR #7139, opened 2019-06-21
feat(artifacts): find multiple artifacts from single execution

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4354
Orca PR: https://github.com/spinnaker/orca/pull/3006

- Enables the selection of multiple artifacts in the Find Artifact from Execution stage.
- Makes minor UI changes to Find Artifact execution details tab.

Before:
before

After:
after


maggieneterval:
Actually I'm also very open to just deleting the first execution details tab, since there's also the Artifact Status tab and if you need to see the JSON you can always open the full execution source :woman_shrugging:

maggieneterval:
Okay I'll leave it in for now! Curious for @thecodejunkie's thoughts on the UI!

PR #7144, opened 2019-06-24
fix(webhooks): addresses issue 3450 - introduce a delay before polling webhook

marchello2000:
Add additional parameters to the monitored webhook to allow:
* waiting some number of seconds before polling starts
* retrying on specific HTTP status codes

see https://github.com/spinnaker/spinnaker/issues/3450

orca counterpart (https://github.com/spinnaker/orca/pull/2984)

### DIFF
--- a/app/scripts/modules/core/src/help/help.contents.ts
+++ b/app/scripts/modules/core/src/help/help.contents.ts
@@ -412,6 +412,10 @@ const helpContents: { [key: string]: string } = {
     "Pick the status url from the JSON returned by the webhook's response call.",
   'pipeline.config.webhook.statusUrlJsonPath':
     "JSON path to the status url in the webhook's response JSON. (i.e. <samp>$.buildInfo.url</samp>)",
+  'pipeline.config.webhook.retryStatusCodes':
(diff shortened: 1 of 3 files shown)


marchello2000:
CC: @ezimanyi

app/scripts/modules/core/src/help/help.contents.ts
```diff
@@ -412,6 +412,10 @@ const helpContents: { [key: string]: string } = {
     "Pick the status url from the JSON returned by the webhook's response call.",
   'pipeline.config.webhook.statusUrlJsonPath':
     "JSON path to the status url in the webhook's response JSON. (i.e. <samp>$.buildInfo.url</samp>)",
+  'pipeline.config.webhook.retryStatusCodes':
+    'Normally, webhook stage only retries on 429 and 5xx status codes. <br>You can specify additional status codes here that will cause the monitor to retry (e.g. <samp>404, 418</samp>)',
```
ezimanyi:
```suggestion
    'Normally, webhook stages only retry on 429 and 5xx status codes. <br>You can specify additional status codes here that will cause the monitor to retry (e.g. <samp>404, 418</samp>)',
```

PR #7145, opened 2019-06-25
fix(core): Display latest template in pipeline template list

louisjimenez:
https://github.com/spinnaker/spinnaker/issues/4561

image

When templates are saved, they create 3 versions in front50. The first version represents the template addressed by id, the second version is the template addressed by the "latest" tag, and the third version is the template addressed by its content (digest). 

Displaying three versions in the UI was causing confusion as to which version was which. This change displays solely the `latest` version. The `latest` version allows a pipeline to automatically receive updates when a new version is published. Pinning to a version can be done through Spin CLI.


louisjimenez:

PR #7147, opened 2019-06-25
feat(appengine): Enable new artifacts workflow in deploy SG

Jammy-Louie:


Jammy-Louie:
the previous modal was displaying an option to grab the expected artifact from the pipeline even when it was being executed outside of a pipeline, which is now corrected.

new expected artifacts:
Screen Shot 2019-06-25 at 3 04 37 PM

if the rewrite flag is turned off:
Screen Shot 2019-06-25 at 3 03 11 PM

app/scripts/modules/appengine/src/serverGroup/configure/wizard/basicSettings.controller.ts
```diff
@@ -104,6 +108,54 @@ class AppengineServerGroupBasicSettingsCtrl implements IController {
...
+    });
+  };
+
+  public onExpectedArtifactAccountSelected = (accountName: string): void => {
+    this.$scope.$applyAsync(() => {
+      this.$scope.command.storageAccountName = accountName;
+    });
+  };
+
+  public updatePipeline = (_pipeline: IPipeline): void => {};
```
maggieneterval: Curious why we need this noop?

app/scripts/modules/appengine/src/serverGroup/configure/wizard/basicSettings.controller.ts
```diff
@@ -104,6 +108,54 @@ class AppengineServerGroupBasicSettingsCtrl implements IController {
     }
   }
 
+  public readonly excludedGcsArtifactTypes = [
```
ezimanyi: It looks like this predates this PR, but I wonder if it would make more sense if each artifact component just explicitly stated which artifact types it wanted rather than enumerated a list of all the ones to exclude.  This seems reasonably fragile, as any new artifact type will now show up in this stage unless the code here is updated to actively exclude it.

ezimanyi: Actually it looks like there's an `excludeAllTypesExcept` in `ArtifactTypePatterns` that would eliminate the two hard-coded lists below (though arguably just having an `include-types` in the artifact selector might be cleaner).

app/scripts/modules/appengine/src/serverGroup/configure/wizard/basicSettings.controller.ts
Jammy-Louie: I missed something in my git stash. But It apparently worked as a no-op. I've added the code from my stash back just in case.
https://github.com/spinnaker/deck/pull/7147/commits/5e4de5f960dd5c228ce2d7b2765b73d7e4ca4b26#diff-b17ed7a30567ee09f9205d401ee16cd9R157

app/scripts/modules/appengine/src/serverGroup/configure/wizard/basicSettings.controller.ts
Jammy-Louie: You are right, Thanks!

PR #7151, opened 2019-06-26
fix(amazon): Fix SpEL support for load balancers

alanmquach:
Looks like we were actually 99% of the way there for supporting SpEL load balancers.

The only thing we were missing was to append the expression'ed load balancers into the list of valid options so that they don't get filtered out when we run `_.intersection` to filter out invalid load balancers.


app/scripts/modules/amazon/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -474,10 +474,10 @@ export class AwsServerGroupConfigurationService {
     const vpcLoadBalancers = this.getVpcLoadBalancerNames(command);
     const allTargetGroups = this.getTargetGroupNames(command);
 
-    if (currentLoadBalancers && command.loadBalancers && !currentLoadBalancers.includes('${')) {
```
alanmquach: `!currentLoadBalancers.includes('${')` was always true because `currentLoadBalancers`, an array, almost never includes `'${'` as an item. Probably should've been something like `!currentLoadBalancers.some(lb => lb.includes('${'))` but I don't think we want that now either.

PR #7161, opened 2019-06-28
 feat(core/amazon): add a visual indicator to infra managed by Keel

erikmunson:
The first PR for bringing [keel](http://github.com/spinnaker/keel) into the UI — this is a quick (largely experimental) attempt at indicating resources are managed. Currently applies to clusters, load balancers/target groups, and security groups.

Right now this is implemented via entity tags, but may change in the future. I'm not particularly tied to any of the specifics (including new directory structures, etc.).

Screen Shot 2019-06-27 at 9 13 23 PM

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/details/targetGroupDetails.html
+++ b/app/scripts/modules/amazon/src/loadBalancer/details/targetGroupDetails.html
@@ -22,42 +22,21 @@
         {{ctrl.targetGroup.name}}
       </h3>
     </div>
-    <div>
-      <div class="actions">
-        <div class="dropdown" uib-dropdown dropdown-append-to-body>
-          <button
-            type="button"
-            class="btn btn-sm btn-primary dropdown-toggle"
-            style="visibility: hidden"
-            disabled
-            uib-dropdown-toggle
-          >
-            Target Group Actions <span class="caret"></span>
-          </button>
-          <ul class="dropdown-menu" uib-dropdown-menu role="menu">
-            <li><a href ng-click="ctrl.editLoadBalancer()">Edit Target Group</a></li>
-            <li ng-if="!ctrl.loadBalancer.instances.length">
-              <a href ng-click="ctrl.deleteLoadBalancer()">Delete Target Group</a>
-            </li>
-            <li
-              ng-if="ctrl.loadBalancer.instances.length"
-              class="disabled"
-              uib-tooltip="You must detach all instances before you can delete this load balancer."
-            >
-              <a href ng-click="ctrl.deleteLoadBalancer()">Delete Target Group</a>
-            </li>
-          </ul>
-        </div>
-      </div>
-    </div>
   </div>
+  <managed-resource-details-indicator
+    ng-if="!ctrl.state.loading && ctrl.loadBalancer.entityTags"
+    entity-tags="ctrl.loadBalancer.entityTags"
+  >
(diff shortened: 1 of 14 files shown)


app/scripts/modules/amazon/src/loadBalancer/details/targetGroupDetails.html
```diff
@@ -22,42 +22,21 @@ <h3 class="horizontal middle space-between flex-1" select-on-dbl-click>
...
-              ng-if="ctrl.loadBalancer.instances.length"
-              class="disabled"
-              uib-tooltip="You must detach all instances before you can delete this load balancer."
-            >
-              <a href ng-click="ctrl.deleteLoadBalancer()">Delete Target Group</a>
-            </li>
-          </ul>
-        </div>
-      </div>
-    </div>
```
anotherchrisberry: was this on purpose?

app/scripts/modules/core/src/managed/managedResourceDetailsIndicator.component.ts
```diff
@@ -0,0 +1,10 @@
+import { module } from 'angular';
+import { react2angular } from 'react2angular';
+
+import { ManagedResourceDetailsIndicator } from './ManagedResourceDetailsIndicator';
+
+export const MANAGED_RESOURCE_DETAILS_INDICATOR = 'spinnaker.amazon.managed.resourceDetailsIndicator.component';
```
anotherchrisberry:
```suggestion
export const MANAGED_RESOURCE_DETAILS_INDICATOR = 'spinnaker.core.managed.resourceDetailsIndicator.component';
```

app/scripts/modules/amazon/src/loadBalancer/details/targetGroupDetails.html
erikmunson: oh yeah, forgot I did this otherwise would've commented it. all this markup is all hidden and seemingly has been forever — we don't actually let you mutate target groups directly from the details pane. Effectively the only noticeable change is that this silly whitespace goes away:

Screen Shot 2019-06-27 at 9 34 19 PM

app/scripts/modules/core/src/managed/ManagedResourceDetailsIndicator.tsx
```diff
@@ -0,0 +1,41 @@
...
+  if (!isManaged) {
+    return null;
+  }
+
+  const helpText = (
+    <>
+      <p>
+        <b>Spinnaker is continuously managing this resource.</b>
+      </p>
+      <p>Changes made in the UI will be stomped in favor of the existing declarative configuration.</p>
```
anotherchrisberry: Is there anything we can link to here? Unless you already know about keel and declarative delivery, it's not really clear what any of this means.

erikmunson: Indeed, a 'what on earth is this referring to?' doc on spinnaker.io for this tooltip is on the list to write in the near-term. For now anything keel-related is restricted to a few test apps as we work out documentation and some other rough edges.

app/scripts/modules/core/src/managed/managedResourceDetailsIndicator.component.ts
erikmunson: danke

PR #7177, opened 2019-07-02
fix(core): Fix link to strategic merge patch docs 

brianstorti:
The file was moved in this PR https://github.com/kubernetes/community/pull/3145

### DIFF
--- a/app/scripts/modules/kubernetes/src/help/kubernetes.help.ts
+++ b/app/scripts/modules/kubernetes/src/help/kubernetes.help.ts
@@ -221,7 +221,7 @@ const helpContents: { [key: string]: string } = {
   'kubernetes.manifest.patch.record': `Record the applied patch in the <i>kubernetes.io/change-cause</i> annotation. If the annotation already exists, the contents are replaced.`,
   'kubernetes.manifest.patch.mergeStrategy': `
     <p>The merge strategy to use when patching Kubernetes objects:</p>
-    <p><b>strategic</b>: <i>(Default)</i> <a href="https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md" target="_blank">Kubernetes Strategic merge patch.</a></p>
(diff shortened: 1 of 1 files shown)


app/scripts/modules/kubernetes/src/help/kubernetes.help.ts
```diff
@@ -221,7 +221,7 @@ const helpContents: { [key: string]: string } = {
   'kubernetes.manifest.patch.record': `Record the applied patch in the <i>kubernetes.io/change-cause</i> annotation. If the annotation already exists, the contents are replaced.`,
   'kubernetes.manifest.patch.mergeStrategy': `
     <p>The merge strategy to use when patching Kubernetes objects:</p>
-    <p><b>strategic</b>: <i>(Default)</i> <a href="https://github.com/kubernetes/community/blob/master/contributors/devel/strategic-merge-patch.md" target="_blank">Kubernetes Strategic merge patch.</a></p>
+    <p><b>strategic</b>: <i>(Default)</i> <a href="https://github.com/kubernetes/community/blob/90e408affda2f6c97fec1195ab3b7aa554432227/contributors/devel/sig-api-machinery/strategic-merge-patch.md" target="_blank">Kubernetes Strategic merge patch.</a></p>
```
louisjimenez: why `90e408affda2f6c97fec1195ab3b7aa554432227` instead of master?

brianstorti: This should prevent the link from breaking in case the file path changes again. The tradeoff is that it won't point to the latest version of the file. Would you rather keep using `master`?

louisjimenez: Agreed, there is a trade-off here. We can tie to master and get latest with the risk of breakage or pin and risk serving outdated docs. I think it's best to be consistent throughout deck and serve the latest, examples [1](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/amazon/src/help/amazon.help.ts#L92) and [2](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/chaosMonkey/chaosMonkey.help.ts#L43)

brianstorti: Yeah, that makes sense.

PR #7178, opened 2019-07-02
feat(gce): Support new artifact model for deploy SG

Jammy-Louie:


Jammy-Louie:
refactored the deploy SG to use the new artifacts rewrite and added the support for custom artifacts.

corresponding PRS:
https://github.com/spinnaker/orca/pull/3018
https://github.com/spinnaker/clouddriver/pull/3835

artifacts rewrite:
Screen Shot 2019-07-02 at 11 32 23 AM

prewrite:

Jammy-Louie:
yes that is correct. according to the comments in the code it only used the custom artifact type

PR #7180, opened 2019-07-03
feat(aws): Support new artifact model for deploy cloudformation

Jammy-Louie:


Jammy-Louie:
refactored to use the new artifact component as seen here:

relies on: https://github.com/spinnaker/orca/pull/3022

Screen Shot 2019-07-03 at 10 47 28 AM

PR #7188, opened 2019-07-04
chore(core): add license finder plugin

Jammy-Louie:


christopherthielen:
can you provide some context on this?

Jammy-Louie:
sure thing @christopherthielen, we need to have the repos scanned by our legal team here and I tried to have this repo scanned but it failed to scan because our legal team's scripts would try to inject the following if not present

```
plugins {
  id "com.github.hierynomous.license version "0.15.0"
}
```

some of the existing plugins are not compatible with this style of `plugins` block. If I apply the license finder plugin their script would no longer fail.

christopherthielen:
I have no problem with this.  However, if this license checker is gradle based, I doubt it's scanning the right things. Also note I'd like to get rid of as much gradle usage as possible (not now, but in the future "when I have time")

Jammy-Louie:
Thats fine, I think the legal team's script just spots the `build.gradle` file and tries to run a scan through gradle in addition to their other scans. If the gradle usage eventually goes away that would be great and it would stop trying to run the scan with gradle

PR #7196, opened 2019-07-08
feat(google): support stateful MIG operations

maggieneterval:
- Adds `feature.gceStatefulMigsEnabled` flag so alpha stateful MIG operations are hidden by default.
- Converts server group disk descriptions components to React, moves all disk data-massaging logic out of controller into relevant component.
- Makes minor UI adjustments to separate multiple disks of same image into own rows, and label boot disk as such.
- Adds "Mark Disk Stateful" button to non-boot disks (and adds "Marked Stateful" label if already marked stateful).
- Adds "Statefully Update" button to boot disks, allowing a new boot image to be entered. A future PR will add an image selection component so users do not need to manually type in the image name, and remove the unnecessary left sidebar from the image selection modal (I think a hide-nav prop was added in a pending PR).

7e2331b8-6dac-4ca8-8191-78bfcb923af5


maggieneterval:
cc @plumpy :dog2: :dog: :smile_cat:

PR #7197, opened 2019-07-08
fix(amazon): Disallow enable instance in a disabled server group

alanmquach:
Attempting to enable a single instance in a disabled server group will never work so we should remove it as an option.

Currently we'll submit the task, and it will eventually just fail, timing out waiting for the instance to become enabled, which will not happen when the server group is disabled.


alanmquach:
Looks like travis has passed but it did not report back to github

PR #7208, opened 2019-07-10
perf(google): improve performance of GCE image selection

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4510

- The problem that this PR is trying to solve is a very slow or totally unresponsive server group modal UI for users with tens of thousands of images per account.
- Replaces Angular `ui-select` component with new React `ImageSelect` component, which uses an `Async` dropdown from the `react-select` library with several custom performance enhancements: 1) Only loads options when three or more characters have been typed 2) Only loads the first 20 matches for a given search string. (I also tested `react-virtualized`, but with tens of thousands of images the virtualization seemed ineffective.)
- Removes all the logic for toggling between searching all images and images that contain the current application's name. Since we are already making one fetch for all an account's images, we should not also make a separate fetch for a subset of those images.

Before (can't really get a good gif with ~25k images because the UI freezes for about ten seconds):
:sob: 

After:
b0f3b0f3-1236-4ddf-9c75-4384e1d75cb4

### DIFF
--- a/app/scripts/modules/google/src/image/ImageSelect.tsx
+++ b/app/scripts/modules/google/src/image/ImageSelect.tsx
@@ -1,0 +1,61 @@
+import { module } from 'angular';
+
+import { chain } from 'lodash';
+import * as React from 'react';
+import { Async, AutocompleteResult, Option } from 'react-select';
+import { react2angular } from 'react2angular';
+
+interface IImage {
+  imageName: string;
+}
+
+interface IImageSelectProps {
+  availableImages: IImage[];
+  selectedImage: string;
+  selectImage: (image: string, target?: any) => void;
+  target?: any;
+}
+
+export class ImageSelect extends React.Component<IImageSelectProps> {
+  private loadOptions = (inputValue: string): Promise<AutocompleteResult<string>> => {
+    return new Promise(resolve => {
+      if (!inputValue || inputValue.length < 3) {
+        resolve({
+          options: [],
+          complete: false,
+        });
+      } else {
+        const filteredOptions = chain(this.props.availableImages)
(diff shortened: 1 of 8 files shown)


maggieneterval:
cc @seanpeters86 Thank you so much for the helpful feedback on this solution :1st_place_medal:

app/scripts/modules/google/src/image/ImageSelect.tsx
```diff
@@ -0,0 +1,61 @@
...
+  private loadOptions = (inputValue: string): Promise<AutocompleteResult<string>> => {
+    return new Promise(resolve => {
+      if (!inputValue || inputValue.length < 3) {
+        resolve({
+          options: [],
+          complete: false,
+        });
+      } else {
+        const filteredOptions = chain(this.props.availableImages)
+          .filter(i => i.imageName.toLowerCase().includes(inputValue))
```
ezimanyi: Does `inputValue` need to also be lower-cased (or does that happen somewhere prior to this function)?

maggieneterval: It happens mAgIcAlLy at some point before this callback receives the input value. This Async select component has an `ignoreCase` prop that defaults to true. I'm setting the also-defaulted-to-true `ignoreAccents` prop to false to [improve perf](https://github.com/JedWatson/react-select/issues/3128#issuecomment-487256349).

maggieneterval:

app/scripts/modules/google/src/image/ImageSelect.tsx
ezimanyi: Awesome, thanks for the explanation!

PR #7210, opened 2019-07-10
fix(google): replace stateful MIG image input with dropdown

maggieneterval:
- Replaces yucky text input with a better image select component because typing out a long image name every time you need to statefully update your boot disk image would be very annoying!
- Also replaces the yucky non-type-checked angular image reader with a better react/typescript class with static methods!


maggieneterval:
@plumpy we can :cherries: pick this into 1.15 is at your discretion

plumpy:
;cake;

maggieneterval:

PR #7215, opened 2019-07-11
fix(tests): functional test clone form locators to match new reality

maggieneterval:


maggieneterval:

PR #7222, opened 2019-07-14
Simplify filtering logic

christopherthielen:
This refactor tries to simply the logic that handles saving and restoring of filters when switching between the tabs of an app.    It does away with most of the manual reading and writing of the URL's query parameters in favor of some router hooks.  I believe the URL manipulation was responsible for most of the "crazy celery" bugs we have witnessed so far, although I think we've addressed most of those bugs separately.

There is more than can be done to simply further (such as avoiding statefulness in FilterModel altogether, and reducing collaborators that try to manually drive the FilterModel lifecycle), but we are currently held back by the AngularJS code that mutates objects and responds to object mutations.

Additionally, there is more work to be done regarding saving and restoring "display options" (Executions grouping, # of executions per pipeline, etc).  They currently have some special cases that aren't necessary any more.  In the future I'd like to save these as global preferences, and remove all special cases.

### DIFF
--- a/app/scripts/modules/core/src/filterModel/FilterModelService.ts
+++ b/app/scripts/modules/core/src/filterModel/FilterModelService.ts
@@ -1,6 +1,6 @@
 import { cloneDeep, size, some, isNil, reduce, forOwn, includes, pick } from 'lodash';
 
-import { IFilterModel, IFilterConfig, ISortFilter } from './IFilterModel';
+import { IFilterModel, IFilterConfig } from './IFilterModel';
 import { ReactInjector } from 'core/reactShims';
 
 export class FilterModelService {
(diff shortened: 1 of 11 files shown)


app/scripts/modules/core/src/filterModel/FilterModelService.ts
```diff
@@ -175,37 +27,91 @@ export class FilterModelService {
...
+      const savedParams = savedParamsForScreen;
+      const hasSavedFilters = filterParams.some(key => !isNil(savedParams[key]));
+
+      // Don't restore the saved filters if there are already filters specified (via url, ui-sref, etc)
+      const shouldRedirectWithSavedParams = !hasFilters && hasSavedFilters;
+      return shouldRedirectWithSavedParams ? trans.targetState().withParams(savedParams) : null;
+    });
+
+    // When switching apps, clear the saved state
+    transitionService.onStart({ exiting: '**.application' }, () => {
```
erikmunson: forgive my ignorance — is this `exiting: '**.application'` bit also triggered when going from one app's states to another app's states directly? or only when going from an application state to a non-application state like `search`?

PR #7228, opened 2019-07-15
fix(core): Render templated pipeline params in the pipeline run stage

louisjimenez:
The pipeline run stage needs to do a plan for templated v2 pipelines in order to allow configuration of any inherited parameters.

image

https://github.com/spinnaker/spinnaker/issues/4623


louisjimenez:
@spinnakerbot cherry-pick 1.15

louisjimenez:

PR #7241, opened 2019-07-18
fix(core): save and validate trigger type in pipeline config

maggieneterval:
- I think the conversion of the trigger component to React may have introduced an issue where we fail to save the `type` of a trigger. This manifested as the validation being missing from the pipeline config, and the trigger not being visible if you were to save/exit the pipeline config, and then click back into the pipeline config.
- Resolved this by updating the trigger model on type change instead of only the state of this component. Also ended up removing the component state altogether to prevent other bugs that could result from it getting out of sync. LMK what you think, thanks!


christopherthielen:
I tested this locally.  There's still a problem regarding the mutation/object merging.  It manifests when you configure a trigger, then change trigger type and configure the second trigger.  Properties from both trigger types exist in the saved JSON for the trigger.

That said, I think this PR should be merged ASAP because I believe the merged JSON is an existing problem and this PR only improves things.

PR #7242, opened 2019-07-18
feat(pager): Allow providing pager duty subject/details in URL

alanmquach:
This change allows more than just a single default subject/details combination to pre-populate the Send Page modal.

I'm not loving the URL reading/syncing but still settled on this approach because it is the most frictionless for users.

### DIFF
--- a/app/scripts/modules/core/src/pagerDuty/PageModal.tsx
+++ b/app/scripts/modules/core/src/pagerDuty/PageModal.tsx
@@ -4,7 +4,7 @@ import { get } from 'lodash';
 
 import { Application, ApplicationModelBuilder } from 'core/application';
 import { IPagerDutyService, PagerDutyWriter } from 'core/pagerDuty';
-import { NgReact } from 'core/reactShims';
+import { NgReact, ReactInjector } from 'core/reactShims';
 import { SETTINGS } from 'core/config';
 import { SubmitButton } from 'core/modal';
 import { TaskMonitor } from 'core/task';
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pagerDuty/PageModal.tsx
```diff
@@ -40,8 +40,13 @@ export class PageModal extends React.Component<IPageModalProps, IPageModalState>
   }
 
   private getDefaultState(props: IPageModalProps): IPageModalState {
-    const defaultSubject = get(SETTINGS, 'pagerDuty.defaultSubject', 'Urgent Issue');
-    const defaultDetails = get(SETTINGS, 'pagerDuty.defaultDetails', '');
+    const {
```
christopherthielen: I prefer that you access via `ReactInjector.$uiRouter.globals.params`

app/scripts/modules/core/src/pagerDuty/PageModal.tsx
```diff
@@ -61,11 +66,13 @@ export class PageModal extends React.Component<IPageModalProps, IPageModalState>
 
   private handleSubjectChanged = (event: any): void => {
     const value = event.target.value;
+    ReactInjector.$state.go('.', { subject: value });
     this.setState({ subject: value });
   };
 
   private handleDetailsChanged = (event: any): void => {
     const value = event.target.value;
+    ReactInjector.$state.go('.', { details: value });
```
christopherthielen:
```suggestion
    ReactInjector.$state.go('.', { details: value }, { replace: true });
```
?

erikmunson:
👍 not sure this router business is all that bad? seems straightforward enough 🤷‍♂

app/scripts/modules/core/src/pagerDuty/PageModal.tsx
alanmquach: 👍

app/scripts/modules/core/src/pagerDuty/PageModal.tsx
alanmquach: TIL. Thanks!

PR #7243, opened 2019-07-19
fix(k8s/runJob): null property file if value none

ethanfrogers:
if the logs capture selector changes to `none` we should null out the
`propertyFile` field. the `consumeArtifactSource` field is used as a
flag for deck to determine which option is selected but it isn't taken
into account in orca. so, if `propertyFile` is non-null, orca will still
try to find properties even if we have `consumeArtifactSource = none`.

Fixes spinnaker/spinnaker#4642


ethanfrogers:
@maggieneterval is there something like `removeStageField` that I can use instead of nulling out the field? That would actually be more ideal than what I'm doing here.

ethanfrogers:

PR #7244, opened 2019-07-19
fix(core): ensure trigger type is set on the model

Jammy-Louie:


Jammy-Louie:
Angular previously automatically handled this for and this was missed in the refactor

maggieneterval:
Hey @Jammy-Louie, I think I already resolved the issue of setting the trigger type on the model with this PR: https://github.com/spinnaker/deck/pull/7241

However, as @christopherthielen mentioned in a comment on that PR, he found a separate issue that occurs when you add a second trigger after configuring the type of the first (I think?) -- are you looking into that? If not let me know and I can take a look. Thanks!

christopherthielen:
I'm going to finish updating the netflix internal trigger, then I'll take a look at the trigger JSON merge issue

Jammy-Louie:
@christopherthielen @maggieneterval 
I noticed that problem too. I was going to fix that and convert the last part `triggers.directive.js` to React

but first I need to fix some problem with manual executions

PR #7247, opened 2019-07-22
fix(core/pipeline): When changing trigger type, only retain the fields common to all trigger types

christopherthielen:
This change moves ownership of managing the trigger state from angular `triggers` directive to the React `Trigger` component

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/triggers/Trigger.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/Trigger.tsx
@@ -1,31 +1,37 @@
 import * as React from 'react';
 import Select, { Option } from 'react-select';
 import * as classNames from 'classnames';
+import { pick } from 'lodash';
 
 import { Application } from 'core/application';
-import { IExpectedArtifact, IPipeline, ITrigger, ITriggerTypeConfig } from 'core/domain';
-import { Registry } from 'core/registry';
 import { SETTINGS } from 'core/config/settings';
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/pipeline/config/triggers/Trigger.tsx
```diff
@@ -34,20 +40,32 @@ export class Trigger extends React.Component<ITriggerProps> {
...
+      'expectedArtifactIds',
+      'runAsUser',
+      'excludedArtifactTypePatterns',
+    ];
+
+    // Clear out all non-common fields when the type is changed
+    const trigger = pick(this.state.trigger, commonFields) as ITrigger;
+    trigger.enabled = this.disableAutoTriggering.includes(option.valuetype) ? false : trigger.enabled;
+    trigger.type = option.value;
+    this.setState({ trigger }, () => this.updateTriggerFields({}));
```
maggieneterval: It seems like we are creating two sources of truth for the trigger, one on this component's state and one in the angular model that is updated when we call `updateTriggerFields`. I wonder if we could eliminate this component's local state and just call `updateTriggerFields` here? Otherwise, I think it's a bit confusing to have an `initialTrigger` prop that is not actually guaranteed to always be the initial data, since it will presumably stay in sync with any changes we make to the angular model? Let me know if that makes sense, I may be totally off base here.

christopherthielen: you're absolutely right.  I'll update this

PR #7249, opened 2019-07-23
fix(k8s): fix bake manifest selector

ethanfrogers:
Fixes spinnaker/spinnaker#4656

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { Option } from 'react-select';
 
 import { StageArtifactSelectorDelegate } from 'core/artifact';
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
```diff
@@ -112,7 +112,7 @@ export class BakeManifestStageForm extends React.Component<
             <ReactSelectInput
               clearable={false}
               onChange={(o: Option<string>) => {
```
maggieneterval:
```suggestion
              onChange={(e: React.ChangeEvent<HTMLSelectElement>) => {
```

maggieneterval:
@Jammy-Louie I did a quick search for the `ReactSelectInput` component elsewhere in the codebase and noticed a few other spots where I think we are mis-labelling `React.ChangeEvent` as `Option` types in the `onChange` handler -- Typescript isn't yelling at us for trying to read `target.value` off an `Option` because it can accept arbitrary key-value pairs. Just something to be aware of as we convert more forms to React!

ethanfrogers:

PR #7252, opened 2019-07-23
fix(k8s/runJob): External logs URL support manifest with implicit default namespace

pdelagrave:
When the manifest had no namespace explicitly provided, the external logs URL couldn't be prepared.
Now sourcing the namespace from `context.jobStatus.location`, reusing the existing variable.
Also removed the `manifest` property from `JobStageExecutionLogs` because unused.


pdelagrave:
@ethanfrogers

PR #7254, opened 2019-07-23
feat(ui): add spinnaker version to UI, addresses #4383

milesmatthias:
Would really love to see the Spinnaker version in the UI, and the help menu seems like a decent place to do it.

https://github.com/spinnaker/spinnaker/issues/4383


maggieneterval:
Thanks so much for the PR, @milesmatthias!

Sorry if this is a dumb question but is there a separate PR to write the current version to `SETTINGS` or is that already being added by Halyard? I think we'll probably want to hide this link if the version is undefined, since we don't want to assume the user is running a supported version of Spinnaker or even using Halyard (cc @christopherthielen for thoughts on whether the versions link would be useful for orgs like Netflix?). 

Additionally, we'll want to add your new `MenuItem` to the `HelpMenu` that is rendered for large screens (right below the code you edited, but with different class names).

milesmatthias:
I was looking at how the version seems to be set here: https://github.com/spinnaker/deck/blob/f340b0268c1c85aae96c54acee77aa39ef9770d6/halconfig/settings.js#L44

and then picked up as `SETTINGS` here: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/config/settings.ts#L141

 I originally was doing `window.spinnakerSettings.version`, which is correctly set in my Spinnaker instance (I can open the Chrome console and type this to see the version), but maybe that's not set in some other cases?

maggieneterval:
Oh awesome, thanks for explaining! So it seems like if you're using Halyard the version will be set, and otherwise it won't be. For now maybe let's only show your new link if the version is there, since many Spinnaker users do not use Halyard?

maggieneterval:
Thanks for adding the `SETTINGS.version` check, @milesmatthias! Sorry if my previous comment about this was confusing, but could you also add the same codeblock to the help dropdown component that is basically a duplicate of the one you edited, but with class names that make it only render on larger screens? Right above [here](https://github.com/spinnaker/deck/pull/7254/files#diff-16e7d3681cf7e7739598d488eea26c54R84). Thanks!

milesmatthias:
You're not confusing. It's just still morning time out here in LA... ☕️ 😴

christopherthielen:
neat :)

PR #7256, opened 2019-07-24
fix(k8s): fix job log modal overflow

ethanfrogers:
adding a couple flex fills to ensure that the model fills the screen
without overflowing. now the log output scrolls when it overflows and
the rest of the modal stays static.


ethanfrogers:

PR #7258, opened 2019-07-24
fix(k8s/runJob): Allowing only v1 accounts for v1 runJob

pdelagrave:
Using a v2 account would have `com.netflix.spinnaker.clouddriver.kubernetes.v2.op.job.KubernetesRunJobOperation` be called and throw an NPE at line 70, because there's no manifest.


pdelagrave:
@ethanfrogers

ethanfrogers:

PR #7261, opened 2019-07-24
Remove base trigger and legacy layout components

christopherthielen:
Remove the "Legacy" trigger layout (Bootstrap Grid) and replace it with the `StandardFieldLayout`.
Remove the `<BaseTrigger/>` component in favor of using `<RunAsUser/>` directly inside the `<Trigger/>` component.

Also:
- switch to rendering Trigger description inside the dropdown
- Fix `useLatestPromise` mount bug
- Fix bugs in `StandardFieldLayout` and add css classes to target from parent component css


app/scripts/modules/core/src/pipeline/config/triggers/Trigger.tsx
```diff
@@ -143,83 +149,78 @@ export class Trigger extends React.Component<ITriggerProps> {
...
-                label=""
-                value={trigger.enabled}
-                onChange={this.handleTriggerEnabled}
-                input={props => <CheckboxInput {...props} text="Trigger Enabled" />}
-              />
-            )}
-          </div>
-        </fieldset>
-      </LayoutProvider>
+        {/*<pre>{JSON.stringify(trigger, null, 2)}</pre>*/}
```
alanmquach: Leftover from debugging?

app/scripts/modules/core/src/presentation/forms/inputs/ReactSelectInput.tsx
```diff
@@ -81,20 +81,15 @@ export class ReactSelectInput extends React.Component<IReactSelectInputProps> {
       onChange: reactSelectOnChangeAdapter(name, onChange),
     };
 
+    const commonProps = { className, style, ...fieldProps, ...otherProps };
```
alanmquach: Nice

PR #7263, opened 2019-07-25
feat(cf): Logs for Run Job stage

pdelagrave:
* Recent logs (Console Output like)
* External logs URL


app/scripts/modules/cloudfoundry/src/pipeline/stages/runJob/CloudfoundryRunJobStageConfig.tsx
```diff
@@ -52,21 +53,15 @@ export class CloudfoundryRunJobStageConfig extends React.Component<
...
-  };
-
-  private JobNameUpdated = (event: React.ChangeEvent<HTMLInputElement>) => {
-    this.props.updateStageField({ jobName: event.target.value });
-  };
-
-  private targetUpdated = (target: string) => {
-    this.props.updateStageField({ target });
+  private stageFieldChanged = (fieldName: string, value: any) => {
+    set(this.props.stage, fieldName, value);
```
Jammy-Louie: we shouldn't be modifying property values directly any more should be using call backs such as `updateStageField`

PR #7265, opened 2019-07-25
fix(appengine): handle multiple cluster accounts in server group modal

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4676

- Fixes longstanding bug that manifests as the default appengine account always being displayed in server group modals, even if another account is actually associated with that server group.
- Will patch into 1.13, 1.14, and 1.15.


maggieneterval:

maggieneterval:

maggieneterval:

PR #7272, opened 2019-07-25
feat(core/presentation): Add a <Formik/> wrapper which applies fixes and opinions that we want in Spinnaker

christopherthielen:
This is a drop-in replacement for the `<Formik/>` component

cc: @erikmunson @maggieneterval

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/SpinFormik.tsx
+++ b/app/scripts/modules/core/src/presentation/forms/SpinFormik.tsx
@@ -1,0 +1,20 @@
+import * as React from 'react';
+import { Formik, FormikConfig } from 'formik';
+
+/**
+ * This component wraps the <Formik/> component, applying fixes and spinnaker opinions
+ * Use this component like you would use the <Formik/> component
+ */
+function SpinFormikImpl<Values extends {}>(props: FormikConfig<Values>, ref?: React.MutableRefObject<Formik<Values>>) {
+  const formikRef = ref || React.useRef<Formik<Values>>();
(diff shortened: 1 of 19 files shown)


app/scripts/modules/core/src/presentation/forms/SpinFormik.tsx
```diff
@@ -0,0 +1,20 @@
...
+function SpinFormikImpl<Values extends {}>(props: FormikConfig<Values>, ref?: React.MutableRefObject<Formik<Values>>) {
+  const formikRef = ref || React.useRef<Formik<Values>>();
+  const defaultIsInitialValid = () => formikRef.current && Object.keys(formikRef.current.state.errors).length === 0;
+
+  // Run initial validation when the form is first rendered
+  React.useEffect(() => {
+    formikRef.current && formikRef.current.getFormikActions().validateForm();
+  }, [formikRef.current]);
+
+  return <Formik<Values> ref={formikRef} isInitialValid={props.isInitialValid || defaultIsInitialValid} {...props} />;
```
alanmquach: Would there ever be cases where someone would want to opt out of this initial validation? And if so, would `isInitialValid={true}` be the way to do that?

christopherthielen: 1) I don't know
2) I don't know

christopherthielen: I believe that formik v2 will deprecate `isInitialValid` paradigm but leave in backwards compat code

PR #7075, opened 2019-06-03
refactor(core): reactify notification and manual execution modal

Jammy-Louie:
spinnaker/spinnaker#4303

* Refactor `Notifications` from angular components to React components.
* `Notifications` now uses `formik`
* Refactoring `Notifications` to `formik` forced me to refactor `Manual Execution Modal` to React and `formik`
* ~~Updated the Wizard Modal to hide the side navigation and section headers for reusability.~~
* ~~Enhance `CheckListInput` to also contain additional inputs (text inputs) when a checkbox is checked.~~
* Refactored Stages, Manual Judgment, Triggers and Application Config to use new `Notifications`.
* Updated previously implemented manual execution components to no longer update the `command` in their props but to use a callback instead.
* Leveraged the Registry Service to add Notifications types similar to triggers

PR #7285, opened 2019-07-30
fix(tests): fix tests for the refactored manual execution

Jammy-Louie:
fixed tests caused by https://github.com/spinnaker/deck/pull/7075


Jammy-Louie:
sorry for the delay, this should fix the tests @maggieneterval

plumpy:
Thanks so much for taking care of this, I really appreciate it! Tests are back to green again! :1st_place_medal:

PR #7286, opened 2019-07-30
feat(*/pipeline): Remove the concept of default stage timeouts, rename option

erikmunson:
## Why though?
For, well, basically forever we've had the concept of 'default stage timeouts' — the idea that by default, a stage has a maximum amount of time it will run regardless of the tasks it contains.

Turns out that for some pretty significant period of time (maybe since the last orca rewrite?) this isn't even close to true. In reality, stages have no concept of a timeout besides the what we call the 'override timeout' in the UI (`stageTimeoutMs`). _Tasks_ have timeouts, so a stage's maximum running time is actually the sum of all task timeouts in the stage. In some ways this actually makes a lot of sense — if a stage adds or removes synthetic stages/tasks due to the way it's configured (or even some external side-effect) it could theoretically require significantly more time to run.

A good example of why default timeouts are tricky is running pipelines in between steps in a Rolling Red Black strategy — how long should the parent deploy stage take by default? It certainly depends on the strategy you choose, but in the RRB case it also depends on the steps and what the pipeline being run is actually doing.

This gets really confusing for users when they're configuring things, because we _actually do_ show a default timeout for many stages in the UI. This gives people a very misleading impression about the behavior of their stages when run, and (at least at Netflix) has caused various kinds of operational pain as a result.

## Changes
This PR effectively removes the whole concept of a 'default stage timeout' in the UI, and changes the language to 'Fail stage after a specific amount of time'. This means that existing stage timeouts in configs are left totally unchanged, and the actual user flow is identical. The only difference is in how the meaning of the option is communicated.

The biggest switch (and the one that has breakage potential) is a switch in the stage config model/interface from `defaultTimeoutMs` (number) to `supportsCustomTimeout` (boolean). This means that some stages which are loaded in outside of the `deck` repo may need to change the config they register.

**Before:**
stage-timeouts-before

**After:**
stage-timeouts-after

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
@@ -40,7 +40,7 @@ export const OverrideTimeout = (props: IOverrideTimeoutConfigProps) => {
 
   useEffect(() => {
     stageChanged();
-  }, [!!get(props.stageConfig, 'supportsCustomTimeout'), props.stageTimeoutMs]);
+  }, [props.stageTimeoutMs]);
 
   const stageChanged = () => {
     if (props.stageTimeoutMs !== undefined) {
--- a/app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { get } from 'lodash';
+import { Duration } from 'luxon';
 
 import { IStage } from 'core/domain';
 import { CheckboxInput, NumberInput } from 'core/presentation';
@@ -21,10 +22,13 @@ const toHoursAndMinutes = (ms: number) => {
   if (!ms) {
     return { hours: 0, minutes: 0 };
   } else {
-    const seconds = ms / 1000;
+    const { hours, minutes } = Duration.fromMillis(ms)
+      .shiftTo('hours', 'minutes')
+      .toObject();
+
     return {
-      hours: Math.floor(seconds / 3600),
-      minutes: Math.floor(seconds / 60) % 60,
+      hours: Math.floor(hours),
+      minutes: Math.floor(minutes),
     };
   }
 };
--- a/app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { get } from 'lodash';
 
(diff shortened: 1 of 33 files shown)


app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
```diff
@@ -13,26 +14,32 @@ export interface IOverrideTimeoutConfigProps {
...
 interface IStageConfig {
-  defaultTimeoutMs: number;
+  supportsCustomTimeout: boolean;
 }
 
+const toHoursAndMinutes = (ms: number) => {
+  if (!ms) {
+    return { hours: 0, minutes: 0 };
+  } else {
+    const seconds = ms / 1000;
```
marchello2000: super minor nit: you could make it `minutes = ms / 60000` to make the subsequent lines slightly more readable.

marchello2000:
❤️

app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
marchello2000: just noticed this was moved...

erikmunson: eh, i was actually just moving this from down below in the file so intentionally didn't touch anything. since you mentioned i'll actually move it over to `luxon` calls which will (hopefully?) be even more straightforward

erikmunson: i do these things because i care about you mark 35530d0

app/scripts/modules/core/src/pipeline/config/stages/overrideTimeout/OverrideTimeout.tsx
```diff
@@ -13,26 +15,35 @@ export interface IOverrideTimeoutConfigProps {
...
   const [hours, setHours] = useState(0);
   const [minutes, setMinutes] = useState(0);
   const [overrideTimeout, setOverrideTimeout] = useState(false);
-  const [configurable, setConfigurable] = useState(false);
-  const [defaults, setDefaults] = useState({ hours: 0, minutes: 0 });
 
   useEffect(() => {
     stageChanged();
-  }, [props.stageConfig, props.stageTimeoutMs]);
+  }, [!!get(props.stageConfig, 'supportsCustomTimeout'), props.stageTimeoutMs]);
```
alanmquach: Does hook syntax allow for this? And I guess if it won't change at runtime (a specific stage's ability to support a custom timeout), does it need to be present in this array?

erikmunson: yeah, i'm kinda skeptical this even needs to be in here and should probably just remove it entirely. i suppose it partly kept it out of some fear that we were watching changes to the stage config in here for a reason i didn't understand. AFAICT the only thing that should trigger this effect is `stageTimeoutMs` changing.

PR #7290, opened 2019-08-01
feat(cloudformation): Make the template editor more lenient

jervi:
When paired with the changes in https://github.com/spinnaker/orca/pull/3024 and https://github.com/spinnaker/clouddriver/pull/3849, this makes it possible to use Yaml instead of Json in the editor, and it will also accept CloudFormation specific functions like `!Ref` and `!Join`. Since Json is a subset of Yaml, Json will still be accepted and parsed correctly.

Cloudformation specific YAML:
image

This will be parsed as a string because of the `!Join`:
image

"Valid" Json and Yaml will be parsed into a map/list as before.
image


jervi:
@spinnaker/reviewers PTAL

PR #7293, opened 2019-08-02
feat(executions): Allow rerun on active executions

alanmquach:
Felt like a weird ask but the use-case actually behind this is users just want to resubmit a new execution with the same parameters as an existing one _regardless of whether the existing one is active or not_.

Most of the time this is not necessary and we wouldn't want to globally add clutter by showing the rerun button even on active executions so hiding it behind an application attribute seemed like a good compromise.


christopherthielen:
What erik said.

alanmquach:
I disliked the visual clutter enough to add the config. ¯\\\_(ツ)_/¯

PR #7299, opened 2019-08-06
Migrate Map Editor to be a React controlled input

christopherthielen:
This PR migrates the `MapEditor` component to a `MapEditorInput` component which roughly conforms to the controlled input api.  As an `Input`, this component can now be used with the `FormField`/`FormikFormField` API.  The state has been removed and pulled up.  For backwards compatibility, a new `MapEditor` component that is compatible with non-controlled usage has been added.

Usage of this `MapEditorInput` component (see the Triggers PR) highlights a gap in the `FormikFormField` api.  This input returns an object with multiple keys/values, not a primitive.  In order for this to work with the current api, the `errors` object has to be passed into the input separately as `formik.errors.[property]`.  I think this may prompt a change to the API where a "Composite Input" (something that manages a complex value, not a primitive) is supported and/or we change the way we think about `validationMessage` and `validationStatus`, which currently only understand primitive values.

### DIFF
--- a/app/scripts/modules/core/src/forms/mapEditor/MapEditorInput.tsx
+++ b/app/scripts/modules/core/src/forms/mapEditor/MapEditorInput.tsx
@@ -1,0 +1,158 @@
+import * as React from 'react';
+import { isNil, isFunction, isString } from 'lodash';
+
+import { IPipeline } from 'core/domain';
+import { createFakeReactSyntheticEvent, IFormInputProps, IValidator } from '../../presentation/forms';
+import { IMapPair, MapPair } from './MapPair';
+
+export interface IMapEditorInputProps extends IFormInputProps {
+  addButtonLabel?: string;
+  hiddenKeys?: string[];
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/forms/mapEditor/MapEditorInput.tsx
```diff
@@ -0,0 +1,158 @@
...
+  });
+}
+
+// Convert a list of IMapPairs to an object with keys and values
+// Prepend any duplicate keys with a magic string
+function tuplesToObject(pairs: IMapPair[]): IMapEditorModel {
+  return pairs.reduce(
+    (acc, pair, idx) => {
+      // Cannot have duplicate keys in an object, so prepend a magic string to the key
+      const key = isNil(acc[pair.key]) ? pair.key : `__MapEditorDuplicateKey__${idx}__${pair.key}`;
```
alanmquach: Neat trick!

PR #7304, opened 2019-08-06
fix(ui): require type on the stage

milesmatthias:
It shouldn't be possible to create a new stage in the UI that doesn't have a type configured.

If you have a pipeline with even 1 of the stages without a type, you can't manually execute it or re-run a previous execution.

If you do have a stage without a type and try to re-run it in the UI, [this function](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/PipelineRegistry.ts#L279) returns null.

and then [this line](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/PipelineRegistry.ts#L349) throws an error about no such attribute `manualExecutionComponent` on `null`.


milesmatthias:
The only thing this PR is missing is a warning that your stage needs a type. With this PR, the following could cause confusion:

1. Create new stage
2. Give stage a name
3. Try to press "Save changes" button, but you can't.

Since the name field is just a regular text input, we can put the `required` HTML attribute on there and your browser + CSS will red outline it until you put in a value, but the angular `ui-select` directive doesn't have a `required` attribute: https://github.com/angular-ui/ui-select/wiki/ui-select

So not sure the best way to go about that.

milesmatthias:
BTW, it's also possible to create a pipeline trigger without selecting a pipeline to trigger from. Will open a PR about that too.

maggieneterval:
@milesmatthias I think ideally we'd have the same red validation triangle we do for invalid stage configs on this screen, but this change at least makes things better so I am happy to merge as-is! Could you please reformat the commit title so spinnakerbot stops yelling at us? :grin:

milesmatthias:
Hmmm I thought I did? https://github.com/spinnaker/deck/pull/7304/commits/64d90b90441525c6797817cf03997f055d88f763

maggieneterval:
Ah sorry yes you did change the commit message, I just updated the PR title to match it since I think that's what it'll default to when we squash and merge!

PR #7305, opened 2019-08-07
fix(tasks): Not bothering with null user when we have an authed one

alanmquach:
`null ( authenticated user )` seems silly when we can just display `authenticated user`

Especially since the authenticated user is the stronger identity.


app/scripts/modules/core/src/task/tasks.html
```diff
@@ -147,8 +147,9 @@ <h4>
                 </span>
               </td>
               <td>
-                {{ task.execution.authentication.user === task.getValueFor('user') ? task.execution.authentication.user
-                : (task.getValueFor('user') + ' (' + (task.execution.authentication.user || 'unknown user') + ')') }}
+                {{ task.getValueFor('user') ? ( (task.execution.authentication.user === task.getValueFor('user')) ?
```
christopherthielen: I know this is angular code and we're not investing in it, but I find this particularly difficult to read

alanmquach: Agree. It was prettier before prettier

emjburns:
Thanks @alanmquach!

PR #7307, opened 2019-08-08
feat(tasks): Adding redirect for task by id without application

alanmquach:
Similar to #7076 

Redirects from `/#/tasks/8675309` to `/#/applications/sandgnat/tasks/8675309`

### DIFF
--- a/app/scripts/modules/core/src/task/task.states.ts
+++ b/app/scripts/modules/core/src/task/task.states.ts
@@ -1,12 +1,15 @@
 import { module } from 'angular';
 
-import { INestedState } from 'core/navigation/state.provider';
+import { INestedState, StateConfigProvider } from 'core/navigation/state.provider';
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/task/task.states.ts
```diff
@@ -40,6 +43,36 @@ module(TASK_STATES, [APPLICATION_STATE_PROVIDER]).config([
...
+        taskId: { dynamic: true },
+      },
+      redirectTo: transition => {
+        const { taskId } = transition.params();
+
+        if (!taskId) {
+          return undefined;
+        }
+
+        return Promise.resolve()
```
louisjimenez: nit: Could this be shortened to `return Promise.resolve(TaskReader.getTask(taskId))`?

alanmquach: Yup. Thanks! Dunno why I did that..

PR #7309, opened 2019-08-08
feat(provider/k8s): support artifacts in run job

ethanfrogers:
adds support for artifacts in the v2 run job stage. very similar to
deploy manifest.


ethanfrogers:
@maggieneterval I totally realize this is a lot of copy/pasta from Deploy Manifest. I'll circle back up once I finish this series of Run Job improvements and see if I can come up with a way to DRY this up a bit.

PR #7310, opened 2019-08-08
fix(gce): remove use of ONLY_DOWN deprecated autoscaler policy

plumpy:


maggieneterval:
:x: :no_good_man:  :arrow_down:

plumpy:

PR #7312, opened 2019-08-08
feat(helm): allow manual execution overrides for helm charts

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4628

Allows helm chart artifact names and version to be overridden during manual execution under the following conditions:
- The pipeline has exactly one Bake (Manifest) stage
- The pipeline has exactly one expected artifact of type `helm/chart`

The specific type of workflow this unblocks is users setting either the helm chart Match Artifact name and/or version as a regular expression (for example, `.*` to match all versions) who want to manually test and select a certain name and/or version at pipeline run time. I only tested this with artifacts rewrite _off_ but since it's supposed to be backwards compatible this should work with it on as well.

20c65c8f-b13d-4e3b-9efe-1137529c452b


maggieneterval:
cc @dbyron0 if you could test this out at some point and let me know if it solves some of your problems that would be :100:

PR #5844, opened 2018-10-10
refactor(core/presentation): Refactor form inputs and layouts for better composability and reuse.

christopherthielen:
This is an iteration of the Spinnaker Forms API.  If I pinged you on this PR you expressed interest in this update.  Review if you want.

- Introduce `<FormikForm render={formikProps => { render the form values }} />` component
- Introduce `<FormField/>` and `<FormikFormField/` components with render props
- Remove `formikField` HOC in favor of `<FormikFormField/>`
- Remove `TextField` in favor of using `<FormikFormField input={TextInput} />`
- Rename `BasicLayout` to `StandardFieldLayout`
- Refactor Expression Form Fields to use new API/Components

# React Form Components

# FormikFormField

`FormikFormField` is a component used to render form fields in Spinnaker
using Formik for form state management and data validation.

basic formik form

This example sets up a Formik form and renders two `FormikFormField`s.
Validation errors on each field are surfaced when the field is blurred.
It has a submit button which is disabled when the form is invalid or being submitted.

```js
<Formik
  initialValues={{ name: '', email: '' }}
  onSubmit={(_values, formik) => setTimeout(() => formik.setSubmitting(false), 2000)}
  render={formik => {
    return (
      <Form>
        <FormikFormField
          name="name"
          label="Your Name"
          input={TextInput}
          validate={value => (!value ? 'Please enter your name' : undefined)}
        />

        <FormikFormField
          name="email"
          label="Your Email"
          input={TextInput}
          validate={value => {
            if (!value) return 'Please enter your email';
            if (!/[^@]+@[^@]+/.exec(value)) return 'Please enter a valid email';
          }}
        />

        <button disabled={!formik.isValid || formik.isSubmitting} className={`pull-right primary`}>
          {formik.isSubmitting ? 'Submitting...' : 'Submit Form'}
        </button>
      </Form>
    );
  }}
/>
```

The `FormikFormField` component should be a descendent of a `Formik` component.
It accepts "all the props", organizes them, and passes them down to the Input and Layout in the correct spots.

- `input`: the Input component to use (see Input section below for details)
- `layout`: the (optional) Layout component to use. (see Layout section for details. `StandardFieldLayout` is used by default)
- `name`: the path to the field's value in the formik `values`
- `label`, `help`, `required`, `actions` (see `IFieldLayoutPropsWithoutInput`)
- `touched`, `validationMessage`, `validationStatus` (see: `IValidationProps`)

[...]

PR #7313, opened 2019-08-08
refactor(core/presentation): in renderContent helper, always render Component as JSX, not as a bare function call.

christopherthielen:
This helper function is used in `FormField` and `FormikFormField` to render the render props (`input`, `label`, etc).  If a functional component is provided, it should be rendered as JSX (`<Component {...props} />`), not simply as a function call (`Component(props)`).  This is especially apparent as we have now introduced components with react hooks into the codebase.

I tried to recall why this code originally differentiated between Class components and Functional components in #5844.  My current assumption is that they should both be rendered in the same way, using JSX.  I believe this change will not cause any regressions in behavior, but we should keep a close eye on newly reported problems in deck forms.

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/fields/renderContent.tsx
+++ b/app/scripts/modules/core/src/presentation/forms/fields/renderContent.tsx
@@ -6,19 +6,13 @@ import * as React from 'react';
  * Supports:
  *
  * - React Class
- * - Arrow Function
- * - Render Function
+ * - Functional component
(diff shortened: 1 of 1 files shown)


christopherthielen:
OK... I now recall why we did it this way: if the functional component is defined inline (as is common with a render prop), the function reference changes each time and the component remounts on each render.

erikmunson:
It sure was good while it lasted

PR #7314, opened 2019-08-09
chore(build): Update cloudbuild.yaml file

ezimanyi:
Remove the $REPO_NAME variable from the cloudbuild.yaml file; this is being used to decide the name of the image to push which will not always correspond to the image name. In particular, if we start publishing both alpine and ubuntu images we'll want the image name to have a suffix reflecting that.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -4,10 +4,10 @@ steps:
     entrypoint: 'bash'
     args: ['-c', './gradlew build -PskipTests']
   - name: gcr.io/cloud-builders/docker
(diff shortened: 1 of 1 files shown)


WillBeebe:
FWIW images with different OSs, I typically see that in the tag, not the image name.

```
deck:1.15-alpine
deck:1.15
```

ezimanyi:
@WillBeebe : Thanks!  That makes sense, and is what we're planning to do when we support different OSes.  (There's a design doc [here](https://docs.google.com/document/d/1ig1vqAR-DjIjjJYOayB8oGjEwjH_eb5WH0J6lIiCf2c/edit), comments welcome!)

PR #7317, opened 2019-08-12
fix(titus): Removing NLB as it is not compatible with Titus

alanmquach:


app/scripts/modules/amazon/src/loadBalancer/configure/AmazonLoadBalancerChoiceModal.tsx
```diff
@@ -81,6 +86,14 @@ export class AmazonLoadBalancerChoiceModal extends React.Component<
                 </div>
               ))}
             </div>
+            {selectedChoice.type === 'classic' && cloudProviders.includes('titus') && (
+              <div className="alert alert-warning">
+                <p>
+                  <i className="fa fa-exclamation-triangle" /> Note: Classic Load Balancers cannot be used with Titus as
+                  they do not have <em>ip</em> based target groups.
```
erikmunson:
```suggestion
                  they do not have <em>IP</em> based target groups.
```

alanmquach:
Yeah, I don't like that either. Looking to see if there's an easy way to fork out and have this live in Titus instead.

app/scripts/modules/titus/src/titus.module.ts
```diff
@@ -61,10 +61,12 @@ module(TITUS_MODULE, [
...
     },
-    loadBalancer: {
-      LoadBalancersTag: AmazonLoadBalancersTag,
-      useProvider: 'aws',
-    },
+    loadBalancer: defaultsDeep(
+      {
+        CreateLoadBalancerModal: TitusLoadBalancerChoiceModal,
+      },
+      CloudProviderRegistry.getProvider('aws').loadBalancer,
```
alanmquach: @erikmunson I like this better, but I have mixed feelings about getting the the aws loadBalancer object from the `CloudProviderRegistry` vs importing it directly

alanmquach: ugghhhh, this is not great either because now for applications with both `aws` and `titus` configured, users are now asked to pick a provider because `titus` is not 100% delegated to `aws`

app/scripts/modules/amazon/src/loadBalancer/configure/AmazonLoadBalancerChoiceModal.tsx
```diff
@@ -5,21 +5,29 @@ import { ILoadBalancerModalProps, ModalClose, ReactModal, noop } from '@spinnake
 
 import { IAmazonLoadBalancerConfig, LoadBalancerTypes } from './LoadBalancerTypes';
 
+export interface IAmazonLoadBalancerChoiceModalDetailsInectedProps {
```
erikmunson: s/Inected/Injected

PR #7320, opened 2019-08-13
feat(kubernetes): permit creatable options for service and namespace …

maggieneterval:
…in rollout strategy config

Closes https://github.com/spinnaker/spinnaker/issues/4687

Example:

nyHrsTjDgGG

P.S. Thanks for the awesome reusable creatable namespace selector, @ethanfrogers! If we get requests to modify any more otherwise-prepopulated inputs to accept arbitrary SpEL than I might put some thought into an even more general purpose component.


ezimanyi:
🎳📄🔠🎉🏅

PR #7318, opened 2019-08-13
refactor(core): Reactify the pipeline trigger configuration

Jammy-Louie:
Refactored pipeline config from Angular to React

Screen Shot 2019-08-13 at 11 58 19 AM

Screen Shot 2019-08-13 at 11 58 40 AM

PR #7324, opened 2019-08-14
fix(artifacts): fix artifact account default logic

Jammy-Louie:
Fixes the logic to default an artifact account to the correct type. The previous logic always set the first one from the list.


Jammy-Louie:
@maggieneterval this should fix https://github.com/spinnaker/deck/pull/7318#issuecomment-521314169

PR #7327, opened 2019-08-15
fix(core/pipeline): Fix github trigger manual execution missing 'hash' property

christopherthielen:
Adding `key={trigger.description}` to `Triggers.tsx` in a recent PR had an unexpected side effect of remounting the component when this submit method deletes the trigger 'description' field.
Clone the trigger before mutating to stop this.

Also pass the correct trigger object reference to `triggerChanged` invoke in `Triggers.tsx`


christopherthielen:
Here's the delete I mentioned: https://github.com/spinnaker/deck/blob/8f786c550d244ca95e29cbab940ce968c97841ef/app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx#L157

PR #7334, opened 2019-08-15
fix(execution): Parameter names should be treated as strings

alanmquach:
Otherwise we'll get unexpected behavior on crazy pipelines that have things like `foo.bar.baz` as the name.


alanmquach:
That's a good idea. I'll drop a comment in here. But I think this might be an appropriate case were we could probably add a few unit tests with weird strings and make sure they all work properly

PR #7336, opened 2019-08-16
fix(helptext): clarify text for stage timeout

marchello2000:


ajordens:
Not sure if we want to have the examples listed as bullet points ... but I appreciate the wording change to remove "task" and introduce "sensible". 

That strikes me as a happy medium without investing in better (server-side) calculation.

marchello2000:
tweaked to this:
image

erikmunson:
After discussing this at length when we originally made the wording change, we moved away from stage-specific examples like these because it causes similar problems to implying there's a predictable default timeout for stages. AFAICT talking about how long stages typically take is either 1) vague and doesn't build confidence/predictability or 2) too nuanced and specific and causes more confusion (e.g. I am pretty confident the 2-hour deploy stage example would've still led to the internal issues we saw with RRB-based deploys where folks didn't expect a multi-hour timeout due to pipeline runs).

I'd advocate for continuing to avoid specific examples but tweaking the wording about task timeouts to be closer to this proposal re: sensible defaults, as I 100% agree that part is a clearer articulation of the real behavior. For example: `Note: By default, Spinnaker will use sensible timeouts that depend on the stage type and the operations the stage needs to perform at runtime. These defaults can vary based on chosen configuration and other external factors.`

marchello2000:
@erikmunson : i (after months of dilly-dallying) incorporated your suggestion.

erikmunson:
Love it, thanks Mark!

PR #7337, opened 2019-08-16
fix(core/pipeline): fix side menu rendering for trigger config

Jammy-Louie:
Shouldn't rely on `child.type.name` when adding the menu items because the nightly build does not contain the name `PageSection` but `PageSection_ PageSection` instead.


Jammy-Louie:
Issue happens on the nightly build and not locally

Screen Shot 2019-08-16 at 3 14 53 PM

PR #7342, opened 2019-08-19
feat(bakeManifest): add kustomize support

ethanfrogers:
adds support for kustomize in the UI. the main user input is a single
kustomize file which tells the bakery what it needs to do to render the
manifests. broke the helm config form into a new component as well.
putting this behind a feature flag because the backend isn't complete
yet. will remove once we merge the backend.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
@@ -1,10 +1,11 @@
 import * as React from 'react';
 
-import { StageArtifactSelectorDelegate } from 'core/artifact';
-import { IArtifact, IExpectedArtifact, IPipeline } from 'core/domain';
-import { MapEditor } from 'core/forms';
+import { IPipeline } from 'core/domain';
 import { IFormikStageConfigInjectedProps, StageConfigField } from 'core/pipeline';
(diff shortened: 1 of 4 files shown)


ethanfrogers:
## Brand new Bake (Manifest) stage
image

## With Kustomize Selected
image

## With Helm Selected
image

app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
```diff
@@ -13,89 +14,22 @@ interface IBakeManifestStageFormProps {
 export class BakeManifestStageForm extends React.Component<
   IBakeManifestStageFormProps & IFormikStageConfigInjectedProps
 > {
-  public readonly templateRenderers = ['HELM2'];
+  public HELM_RENDERER = 'HELM2';
+  public KUSTOMIZE_RENDERER = 'KUSTOMIZE';
+  // public readonly templateRenderers = [this.HELM_RENDERER, this.KUSTOMIZE_RENDERER];
```
maggieneterval: guessing we will uncomment this when we remove the feature flag?

ethanfrogers: Nice catch. I meant to remove that, so I'll pull it out. It'll make another appearance once the feature flag is removed. :)

PR #7349, opened 2019-08-20
fix(notifications): Fixing custom message text shape

alanmquach:
Echo expects the custom message structure to be:
```
message: {
  "pipeline.starting": {
    text: 'foo'
  }
}
```
So we have to set the path to be `message["pipeline.starting"].text` instead of `message.pipeline.starting.text`, which would result in:
```
message: {
  pipeline: {
    starting: {
      text: 'foo'
    }
  }
}
```

### DIFF
--- a/app/scripts/modules/core/src/notification/modal/NotificationDetails.tsx
+++ b/app/scripts/modules/core/src/notification/modal/NotificationDetails.tsx
@@ -57,7 +57,7 @@ export class NotificationDetails extends React.Component<INotificationDetailsPro
     if (whenOption !== 'manualJudgment' && ['email', 'slack', 'googlechat'].includes(type)) {
       return (
         <FormikFormField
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/notification/modal/NotificationDetails.tsx
```diff
@@ -57,7 +57,7 @@ export class NotificationDetails extends React.Component<INotificationDetailsPro
     if (whenOption !== 'manualJudgment' && ['email', 'slack', 'googlechat'].includes(type)) {
       return (
         <FormikFormField
-          name={'message.' + whenOption + '.text'}
+          name={'message["' + whenOption + '"].text'}
```
louisjimenez: nit: this may read a little more easily with a template string:
```javascript
{`message["${whenOption}"].text`}
```

PR #7356, opened 2019-08-22
refactor(core): use a fieldLayout for manualExecution

Jammy-Louie:
created a fieldLayout specifically for manualExection so that `FormikField` components will be aligned correctly.

**StandardField Layout When Adding a Notification:**

Screen Shot 2019-08-22 at 2 16 02 PM

**ManualExecutionFieldLayout When Adding a Notification to the Manual Execution Modal:**

Screen Shot 2019-08-22 at 2 15 50 PM


Jammy-Louie:
Applying what was discussed in the UI-Sig

app/scripts/modules/core/src/pipeline/manualExecution/layout/ManualExecutionFieldLayout.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { IValidationMessageProps, ValidationMessage } from 'core/validation';
+import { IFieldLayoutProps } from 'core/presentation';
+import { isUndefined } from 'util';
+
+export class ManualExecutionFieldLayout extends React.Component<IFieldLayoutProps> {
+  public render() {
+    const { label, help, input, actions, touched, validationMessage, validationStatus } = this.props;
+
+    const showLabel = !isUndefined(label) || !isUndefined(help);
```
maggieneterval: Not sure if there's any chance of these being `null` or empty strings, if there is maybe `_.isEmpty` would be safer?

Jammy-Louie: 👍

Jammy-Louie:
@christopherthielen I think it depends on your preference for aesthetics.

without the new layout and usage of the ` standard layout `

you would see:

Screen Shot 2019-09-03 at 9 56 06 AM

the label tags become slightly misaligned.

PR #7358, opened 2019-08-23
refactor(core): Reactify disable pipeline modal

Jammy-Louie:
Reactify Disable Pipeline modal:

Screen Shot 2019-08-23 at 10 45 09 AM


app/scripts/modules/core/src/pipeline/config/actions/delete/DeletePipelineModal.tsx
```diff
@@ -67,7 +67,9 @@ export function DeletePipelineModal(props: IDeletePipelineModalProps) {
                 {errorMessage}
               </p>
               <p>
-                <a onClick={() => setDeleteError(false)}>[dismiss]</a>
+                <a href="javascript:void(0)" onClick={() => setDeleteError(false)}>
```
maggieneterval: If the purpose of using an `a` element for this is to style a button like a link, maybe we can avoid the `href="javascript:void(0)"` if we just use a div but style it like a link (I think we have a `btn-link` class)? I also could be missing something here so feel free to ignore my friday brain!

louisjimenez: Great point! Would adding a `e.preventDefault()` to the `setDeleteError` func wrapper be of use here?

Jammy-Louie: thanks @maggieneterval, I would definitely rather use `btn btn-link`.  I'll remember as I am doing the conversion. A lot of the angular code just uses an anchor tag with an empty `href`

PR #7363, opened 2019-08-27
feat(ecs): Add security group view for ECS

clareliguori:
Shows security groups for ECS server groups, but not yet for ECS load balancers.  Related to spinnaker/spinnaker#4798

Mostly copying directly from the amazon security group details implementation:

Screenshot from 2019-08-27 12-20-52


app/scripts/modules/ecs/src/securityGroup/securityGroup.reader.ts
```diff
@@ -0,0 +1,16 @@
+import { module } from 'angular';
+
+import { ISecurityGroupsByAccount, ISecurityGroup } from '@spinnaker/core';
+
+export class EcsSecurityGroupReader {
+  public resolveIndexedSecurityGroup(
```
maggieneterval: Can this be static? And then can just access this method statically instead of angular-module-ing the class?
I'm assuming the rest of the angular code you're copying from the amazon module like you said, but any new code would be :100: if it's in React!

clareliguori: I'm not sure how to make it work without angular-izing, but I might be missing something.

That method gets called in core code:
https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/securityGroup/securityGroupReader.service.ts#L121-L123

And afaict the reference to the class in the provider module configuration must be either an angular module or a function:
https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/cloudProvider/providerService.delegate.ts#L19-L25

I have it configured as an angular module string here. Do you have examples I could look at from other providers where this is configured as a non-angular function?
https://github.com/clareliguori/deck/blob/ecs-security-group-view/app/scripts/modules/ecs/src/ecs.module.ts#L86

maggieneterval: > And afaict the reference to the class in the provider module configuration must be either an angular module or a function:

Unintuitively, I think in JS classes are actually functions! An example of a provider that has done away with most (all?) of its angular deps is cloudfoundry (thanks @Jammy-Louie!) :
https://github.com/spinnaker/deck/blob/master/app/scripts/modules/cloudfoundry/src/cf.module.ts

So for example, instead of cf's load balancer transformer being exported as an angular module, it's just a plain class:

https://github.com/spinnaker/deck/blob/master/app/scripts/modules/cloudfoundry/src/loadBalancer/loadBalancer.transformer.ts

That said, I just noticed that all the other provider's security group readers are in fact still angular modules, so please do feel free to just leave this as-is. There might even be something I'm totally missing that requires this reader in particular to remain a module. :woman_shrugging: 

P.S. Do you have write access to Deck yet or may I merge this for you?

clareliguori: I don't have write access yet, thanks!

PR #7366, opened 2019-08-28
 refactor(core): Reactify lock and unlock pipeline modal

Jammy-Louie:
* fixed the IPipeline model. `locked` should be an object and not of type boolean
 * Reactify Lock Pipeline modal
 * Reactify Unlock Pipeline modal


ezimanyi:
@Jammy-Louie @louisjimenez : This is causing the deck build to fail due to the `no-shadowed-variable` rule; looks like `LockPipelineModal.tsx` is shadowing the variable `props`.  Can you please fix ASAP?

PR #7370, opened 2019-08-30
feat(core): filter out providers that don't support creating security groups

clareliguori:
Allow some providers' security group support to be 'read only': viewing the existing security groups is supported, but new security groups cannot be created through Spinnaker.

Fixes spinnaker/spinnaker#3794

Today this only covers the ECS provider: security groups can be created through the associated AWS account with the AWS provider's security group creation flow, instead of duplicating that flow in the ECS provider.

The result of this change is that for a Spinnaker instance with AWS and ECS providers enabled, the 'Create Firewall' button goes straight to the AWS provider modal, instead of allowing a choice between the AWS and ECS provider.

Screenshot from 2019-08-30 12-06-02

But both AWS and ECS security groups are still shown in the list view:

Screenshot from 2019-08-30 12-07-49


app/scripts/modules/core/src/securityGroup/AllSecurityGroupsCtrl.js
```diff
@@ -67,8 +67,13 @@ module.exports = angular
         updateSecurityGroups();
       };
 
+      function createSecurityGroupProviderFilterFn(application, account, provider) {
+        let sgConfig = provider.securityGroup;
```
maggieneterval: can `const` this since it won't be reassigned
```suggestion
        const sgConfig = provider.securityGroup;
```

PR #7375, opened 2019-09-04
configure apache to log to stdout/stderr

paulczar:
Apache is configured by default to log to `/var/log...` which fills up real quick with logs from kubernetes health checks and regular traffic, eventually the pod runs out of space.  This modifies the config to log to stdout/stderr instead inline with docker/kubernetes best practices and should help protect the filesystem of the containers/pods.

Signed-off-by: Paul Czarkowski <username.taken@gmail.com>

### DIFF
--- a/docker/run-apache2.sh
+++ b/docker/run-apache2.sh
@@ -44,6 +44,10 @@ mv spinnaker.conf  /etc/apache2/sites-available
 
 a2ensite spinnaker
 
+# Disable default site
+
+a2dissite 000-default
+
 # Update ports.conf to reflect desired deck host
 
 cp docker/ports.conf.gen ports.conf
--- a/docker/setup-apache2.sh
+++ b/docker/setup-apache2.sh
@@ -1,5 +1,7 @@
 apt-get update
 apt-get install apache2 -y
+rm -rf /var/lib/apt/lists/* 
+
 service apache2 stop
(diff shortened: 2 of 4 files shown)


paulczar:
This is an example change of course, as it would be needed across all of the different spinnaker microservices, not just deck.

docker/run-apache2.sh
```diff
@@ -44,6 +44,10 @@ mv spinnaker.conf  /etc/apache2/sites-available
 
 a2ensite spinnaker
 
+# Disable default site
+
+a2dissite 000-default
```
paulczar: we're not using the default, so its good hygiene to disable it.

docker/setup-apache2.sh
```diff
@@ -1,5 +1,7 @@
 apt-get update
 apt-get install apache2 -y
+rm -rf /var/lib/apt/lists/*
```
paulczar: there's a bit of space to save by cleaning up the apt cache.

ezimanyi:
@Jammy-Louie : I'm not sure I'm the best person to look at this PR...is there someone more familiar with deck you could ask?

Jammy-Louie:
@ezimanyi Sure but this looks like a change we would want to adopt in all the other microservices and it is not just applicable to deck

PR #7376, opened 2019-09-04
feat(core/managed): Add resource dropdown with links to history + source JSON

erikmunson:
There's a couple bits of data about a resource we expect folks to care about early on: the event history of a resource, and the underlying config/definition for the resource that Keel has on file. This adds a dropdown to managed resources in the infrastructure views that lets folks see both, as new tabs with the raw JSON payload (opting for JSON instead of YAML to start, as we already do that across spinnaker and JSON is easy to use with the API ad-hoc). Design is quick and dirty for now, as we're not sure what we'll need in the coming months.

Screen Shot 2019-09-04 at 10 37 45 AM

Screen Shot 2019-09-04 at 10 38 23 AM

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedResourceDetailsIndicator.less
+++ b/app/scripts/modules/core/src/managed/ManagedResourceDetailsIndicator.less
@@ -1,4 +1,13 @@
-.ManagedResourceDetailsIndicator > .icon {
+// Selector specificity is the worst
+.band.band-info.ManagedResourceDetailsIndicator {
+  padding: 8px 10px 10px;
+}
+
+.ManagedResourceDetailsIndicator .rainbow-icon {
   text-shadow: 0px 1px 4px rgba(0, 0, 0, 0.5);
   padding-right: 5px;
+}
+
+.ManagedResourceDetailsIndicator .summary-message {
+  cursor: help;
 }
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/managed/ManagedResourceDetailsIndicator.tsx
```diff
@@ -12,35 +15,75 @@ export interface IManagedResourceDetailsIndicatorProps {
   entityTags: IEntityTags[];
 }
 
+const logClick = (label: string, resourceId: string) =>
```
erikmunson: also some usage tracking for fun.

app/scripts/modules/core/src/managed/ManagedResourceDetailsIndicator.less
```diff
@@ -1,4 +1,13 @@
...
+  padding: 8px 10px 10px;
+}
+
+.ManagedResourceDetailsIndicator .rainbow-icon {
   text-shadow: 0px 1px 4px rgba(0, 0, 0, 0.5);
   padding-right: 5px;
 }
+
+.ManagedResourceDetailsIndicator .summary-message {
+  cursor: help;
```
gcomstock: nice touch

erikmunson: I try (sometimes)

PR #7383, opened 2019-09-10
fix(google): add default gce instance type disk constants to front end

maggieneterval:
Band-aid for https://github.com/spinnaker/spinnaker/issues/4859.
Should be reverted pending investigation of root cause.


maggieneterval:

PR #7389, opened 2019-09-12
fix(kubernetes): remove former 24-char limit on services names

skannengiesser:
This PR removes the 24 character limitation on application name length for Kubernetes providers that got into deck with PR #2102 
According to various ressource (some of them mentioned in the code) that limitation is not in place since May 2016 or k8s > 1.4.0.

This change has been discussed in Slack prior to it being opened.

### DIFF
--- a/app/scripts/modules/kubernetes/src/loadBalancer/configure/wizard/basicSettings.html
+++ b/app/scripts/modules/kubernetes/src/loadBalancer/configure/wizard/basicSettings.html
@@ -14,7 +14,7 @@
         <input
           type="text"
           style="display: none"
-          ng-maxlength="24"
+          ng-maxlength="63"
           class="form-control input-sm"
           ng-model="loadBalancer.name"
           validate-unique="existingLoadBalancerNames"
@@ -83,7 +83,7 @@
       </div>
       <div class="form-group">
         <div class="col-md-9 col-md-offset-3" ng-if="form.loadBalancerName.$error.maxlength">
-          <validation-error message="Load Balancer name can only be 24 characters."></validation-error>
+          <validation-error message="Load Balancer name can only be 63 characters."></validation-error>
         </div>
       </div>
     </div>
--- a/app/scripts/modules/kubernetes/src/validation/applicationName.validator.js
+++ b/app/scripts/modules/kubernetes/src/validation/applicationName.validator.js
@@ -23,31 +23,16 @@ module.exports = angular
     }
 
     function validateLength(name, warnings, errors) {
-      // Kubernetes resource names must match [a-z]([-a-z0-9]*[a-z0-9])?
+      // general k8s resource restriction: 253 characters - https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+      // safe bet is 63 characters - (see also RFC 1035 or RFC 1123 - both setting DNS label maximum to 63 chars)
+      // for service names: - https://github.com/kubernetes/kubernetes/pull/29523 (until K8s 1.4.0 it was 24 characters https://github.com/kubernetes/kubernetes/issues/12463)
+      // or annotations:    - https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
+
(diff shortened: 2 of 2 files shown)


app/scripts/modules/kubernetes/src/validation/applicationName.validator.js
```diff
@@ -23,32 +23,17 @@ module.exports = angular
...
     function validateLength(name, warnings, errors) {
-      // Kubernetes resource names must match [a-z]([-a-z0-9]*[a-z0-9])?
-      let maxResourceNameLength = 63;
+      // general k8s resource restriction: 253 characters - https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+      // safe bet is 63 characters - (see also RFC 1035 or RFC 1123 - both setting DNS label maximum to 63 chars)
+      // for service names: - https://github.com/kubernetes/kubernetes/pull/29523 (until K8s 1.4.0 it was 24 characters https://github.com/kubernetes/kubernetes/issues/12463)
+      // or annotations:    - https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
 
-      let maxServiceNameLength = 24;
+      let maxResourceNameLength = 63;
```
maggieneterval:
```suggestion
      const maxResourceNameLength = 63;
```

ethanfrogers:
@maggieneterval @skannengiesser I don't see any issues with increasing the size. It's probably good to be aware that the Kubernetes documentation states this - 

```
By convention, the names of Kubernetes resources should be up to maximum length of 253 characters and consist of lower case alphanumeric characters, -, and ., but certain resources have more specific restrictions.
```

Also, why `63`? Is that some limitation somewhere else in the system? If so, we should comment why we chose `63` specifically.

app/scripts/modules/kubernetes/src/loadBalancer/configure/wizard/basicSettings.html
```diff
@@ -83,7 +83,7 @@
       </div>
       <div class="form-group">
         <div class="col-md-9 col-md-offset-3" ng-if="form.loadBalancerName.$error.maxlength">
-          <validation-error message="Load Balancer name can only be 24 characters."></validation-error>
+          <validation-error message="Load Balancer name can only be 63 characters."></validation-error>
```
ethanfrogers: Small nit, can you break `63` out into a variable so that we have to change it in fewer places if necessary?

ethanfrogers: @skannengiesser making sure you saw the above. Could you address?

skannengiesser: @ethanfrogers Yes, I saw your change request. Sorry, haven't had time so far and it might take me some time to figure out how to optimally do this. Just for clarification: You're asking me to have one shared variable for these 2 validation limits, which means across k8s v1 and k8s v2 providers? Maybe you can point me to where I should define it? Thanks a lot.

ethanfrogers: Yes, exactly. Honestly, I think it's fine if you just do it in this file. I just want to be sure that if we change the number to, say, `64` that I don't have to change it in more than one place.

skannengiesser: @ethanfrogers Sorry, as I'm having difficulties in setting up deck (and other necessary services) to be able to verify any changes, I don't actually dare changing anything like this. Putting a variable up for these two occurences shouldn't be too difficult, but I wouldn't feel good changing this without being able to verify.

maggieneterval: Completely agree that ideally this constant would be defined once in this file, going to merge this for now since hopefully we can get rid of this angular template altogether in the nearish future, and also since this change doesn't make it worse (the max # was already defined in multiple spots)

PR #7406, opened 2019-09-13
feat(pipeline): Custom alert for start manual execution dialog

alanmquach:
Allows pipelines to provide extra information/context to users manually triggering the pipeline.
Screenshot 2019-09-13 at 4 12 14 PM

Can also be injected when we're programatically opening the start manual execution modal.


christopherthielen:
assume this is configured via "edit pipeline as json"?

alanmquach:
Yes. Well I’m interested in using it programmatically, but yes for now it can be configured in JSON. We can add it to the pipeline config UI in the future.

PR #7409, opened 2019-09-16
feat(core/managed): add Managed Resources section to app config, allow opting out

erikmunson:
depends on https://github.com/spinnaker/gate/pull/898

This adds a new, feature flagged "Managed Resources" section to the app config, which initially supports one thing — 'pausing' management of resources through our new `veto` concept. The section only shows up if two things are true:
- you have `feature.managedResources` turned on
- your app specifically has resources managed by keel

This is slightly awkward because it means that if you _do_ have managed resources, that app config section 'pops' in async. In my experience it's not an issue and the plan is to do a more static configuration (where you always see the section if the feature flag is on) once this is less experimental.

**active:**
active

**paused:**
paused

cc @emjburns @robfletcher @asher @gcomstock


gcomstock:
Thanks for tagging me! Looks great and I wouldn't worry about the async loading. I would consider changing: "Spinnaker is configured to continuously manage some resources" to "Spinnaker is configured to continuously manage some _of this application's_ resources"

emjburns:
Love the rainbow. As I look at the "paused" vs "active" screens, I think the fact that the buttons looks so similar is a little confusing. I guess in addition to the rainbow I'd love a stop sign or something in front of "continuous management of resources is paused" to make it easier to quickly scan and tell the state of managment

erikmunson:
@emjburns @gcomstock another rev with some wording + visual tweaks:

new-paused-state

app/scripts/modules/core/src/application/config/managedResources/ManagedResourceConfig.tsx
```diff
@@ -0,0 +1,137 @@
...
+      <button
+        className="btn btn-primary"
+        disabled={pausePending}
+        onClick={paused ? resumeManagement : pauseManagement}
+        type="button"
+      >
+        {(!pausePending && <i className={classNames('fa sp-margin-xs-right', iconClass)} />) || <ButtonBusyIndicator />}{' '}
+        {paused ? 'Resume Management' : 'Pause Management'}
+      </button>
+      {pauseFailed && <div className="alert alert-danger sp-margin-l-top">Saving failed.</div>}
```
christopherthielen: can this use the `<ValidationMessage />` component? (which I totally want to rename to simply `<Message/>`)

app/scripts/modules/core/src/application/config/applicationConfig.controller.js
```diff
@@ -64,5 +65,14 @@ module.exports = angular
           this.notifications = notifications;
         });
       };
+
+      if (this.feature.managedResources) {
+        this.hasManagedResources = false;
+        ManagedReader.getApplicationSummary(this.application.name).then(({ hasManagedResources }) => {
+          $scope.$applyAsync(() => {
```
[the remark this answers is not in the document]
erikmunson: recording an in-person chat for posterity: we have no idea why. it's probably because of Some Subtle State Bug(tm) that popped up and we've be copy-pasting the same logic ever since.

app/scripts/modules/core/src/application/config/managedResources/ManagedResourceConfig.tsx
erikmunson: 44e5e02

PR #7413, opened 2019-09-19
fix(core/pipeline): standardize manual execution field layouts

erikmunson:
Most of the fields that end up inside a manual execution modal are currently either:
- A vanilla bootstrap layout + input with hand-rolled logic for everything
- A `FormField`/`FormikFormField` that sets up the input + state management, but doesn't have a label and uses vanilla bootstrap grids to add labels and layout.

When the modal itself got moved to react, one of the sub-components that got migrated was the dry run option, which got moved to a full-on `FormikFormField` with a `StandardFieldLayout`. Turns out that layout really does not play nicely with the existing standard bootstrap columns for the modal in a variety of ways — widths are off, flex layouts don't match, and snapping things to a stacked layout at small widths isn't consistent.

To get this fixed cosmetically for now, I'm moving the dry run field to use a pattern consistent with the other fields in this modal. I'd suggest that we should do two things, though I don't think they're super high priority:
1. Move all the vanilla inputs over to `FormField`/`FormikFormField` but keep the vanilla bootstrap layouts
2. Once ☝️is done, move all the bootstrap column layouts over to `StandardFieldLayout` so things change all at once rather than trying to interop between the two in weird ways

Just for funsies, I also made the help field on `Rebake` sit next to the label to be consistent with how we're trying to lay out help fields now.

**before:**
Screen Shot 2019-09-19 at 8 25 00 AM

**after:**
Screen Shot 2019-09-19 at 9 24 52 AM

### DIFF
--- a/app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
@@ -8,7 +8,14 @@ import { assign, clone, compact, extend, get, head, uniq, isArray, pickBy } from
 import { SubmitButton, ModalClose } from 'core/modal';
 import { Application } from 'core/application';
 import { AuthenticationService } from 'core/authentication';
-import { buildValidators, IModalComponentProps, ReactModal, SpinFormik, Markdown } from 'core/presentation';
+import {
+  buildValidators,
+  IModalComponentProps,
+  ReactModal,
+  SpinFormik,
+  Markdown,
+  LayoutProvider,
+} from 'core/presentation';
 import {
   IExecution,
   IExecutionTrigger,
@@ -25,6 +32,7 @@ import { Registry } from 'core/registry';
 import { SETTINGS } from 'core/config/settings';
 import { UrlParser } from 'core/navigation/urlParser';
 
+import { ManualExecutionFieldLayout } from './layout/ManualExecutionFieldLayout';
 import { PipelineOptions } from './PipelineOptions';
 import { CurrentlyRunningExecutions } from './CurrentlyRunningExecutions';
 import { StageManualComponents } from './StageManualComponents';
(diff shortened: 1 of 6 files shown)


christopherthielen:
Thanks for doing this!  This has been bugging me every day.

We already have a `ManualExecutionLayout` component which encapsulates the bootstrap layout markup.  I think I'd like to see that layout provided at the top of this dialog.  All the nested FormikFormFields will then use it for their layout.   In that way, we can continue switching fields over to `FormikFormField` without duplicating the layout markup everywhere.

erikmunson:
oh hey that's great. i wonder why this one never got the custom layout treatment.

app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
```diff
@@ -340,74 +348,76 @@ export class ManualExecutionModal extends React.Component<IManualExecutionModalP
...
+                      formatParameterConfig={this.formatParameterConfig}
+                      pipelineOptions={pipelineOptions}
+                      pipelineChanged={this.pipelineChanged}
+                      triggerChanged={this.triggerChanged}
+                      updateTriggerOptions={this.updateTriggerOptions}
+                    />
+                  )}
+                  {formik.values.pipeline && (
+                    <div className="form-group">
+                      <div className="col-md-6 col-md-offset-4">
```
erikmunson: nothing new, but this has looked really weird forever and I wanted to change it. It didn't respect the overall layout boundaries of the rest of the modal and always felt off.

app/scripts/modules/core/src/pipeline/manualExecution/PipelineOptions.tsx
```diff
@@ -106,14 +106,14 @@ export class PipelineOptions extends React.Component<IPipelineOptionsProps, IPip
       return (
         <div className="form-group row">
           <FormField
-            label={'Pipeline'}
+            label="Pipeline"
             onChange={this.pipelineSelected}
             value={formik.values.pipeline ? formik.values.pipeline.id : ''}
             input={props => (
-              <Select
+              <TetheredSelect
```
erikmunson: I switched this because when you're picking a pipeline, there's only one field and the modal is super short (which makes the modal + the dropdown itself scrollable and gets awkward).

erikmunson:
Alright, this PR got bigger but (I think) better. Mainly two things:
- I moved the `LayoutProvider` that lived down inside a notification component up to the entire modal so any fields that live here get the right layout
- I migrated the fields I found that were doing their own custom layouts (and/or using vanilla input fields) to use `FormField`/`FormikFormField` without custom layout code, so we're not snowflaking column layouts anymore

I'm sure there's other stuff (e.g. trigger-specific fields) that will need similar treatment, but we'll deal with those one-off.

cc @christopherthielen

PR #7415, opened 2019-09-20
fix(artifacts): Support custom artifacts with custom type

jervi:
Some recent changes to the Expected Artifacts feature has severely broken backwards compatibility with artifacts that have a custom type (for instance Igor can create artifacts with type `deb` or `rpm`). This commit will make Deck be more lenient when it encounters legacy artifacts that does not strictly adhere to the new standard. It will also add back the UI fields `type`, `name`, `version` and `location` for custom artifacts.

### DIFF
--- a/app/scripts/modules/core/src/artifact/react/ArtifactEditor.tsx
+++ b/app/scripts/modules/core/src/artifact/react/ArtifactEditor.tsx
@@ -3,6 +3,7 @@ import { cloneDeep, head } from 'lodash';
 import { IArtifactAccount } from 'core/account';
 import { ArtifactAccountSelector } from 'core/artifact';
 import { IArtifact, IPipeline } from 'core/domain';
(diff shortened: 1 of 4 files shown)


jervi:
Example:

Given this legacy expected artifacts definition:
```json
"expectedArtifacts": [
    {
      "defaultArtifact": {
        "customKind": true,
        "id": "47ae01ed-fca8-4513-972b-cf53f5dd9c1e"
      },
      "id": "d245c55c-84c5-46eb-82ba-d1c3ff3904ef",
      "matchArtifact": {
        "customKind": true,
        "id": "e3711912-8630-48e1-a209-69455cfa8f9f",
        "name": "my-artifact",
        "type": "deb"
      },
      "useDefaultArtifact": false,
      "usePriorArtifact": false
    }
  ],
```

Newer versions of Deck would display something like this when configuring the pipeline (the actual account would differ, but it would be the first retrieved from Clouddriver):
image

If you change to "custom-artifact", it will display this:
image

But if you save the pipeline, details from the original expected artifact definition are gone.

After applying this PR, Deck will display this instead:
image

PR #7418, opened 2019-09-23
feat(core/pipeline): Make custom artifacts more readable

jervi:
From
image

to
image

### DIFF
--- a/app/scripts/modules/core/src/pipeline/status/Artifact.tsx
+++ b/app/scripts/modules/core/src/pipeline/status/Artifact.tsx
@@ -44,7 +44,7 @@ export class Artifact extends React.Component<IArtifactProps> {
               {ArtifactIconService.getPath(type) ? (
                 <img className="artifact-icon" src={ArtifactIconService.getPath(type)} width="18" height="18" />
               ) : (
-                <span>{type}</span>
+                <span>[{type}]&nbsp;</span>
               )}
             </dt>
             <dd>


app/scripts/modules/core/src/pipeline/status/Artifact.tsx
```diff
@@ -44,7 +44,7 @@ export class Artifact extends React.Component<IArtifactProps> {
               {ArtifactIconService.getPath(type) ? (
                 <img className="artifact-icon" src={ArtifactIconService.getPath(type)} width="18" height="18" />
               ) : (
-                <span>{type}</span>
+                <span>[{type}] </span>
```
maggieneterval: tiny nit: I think the convention in Deck for adding a space after text in JSX is:
```suggestion
                <span>[{type}]{' '}</span>
```

jervi: Isn’t that only when it’s outside tags? Because React won’t render them otherwise?

erikmunson: We go by Prettier's rules on this one — their reason for adding `{' '}` is [primarily because JSX trims trailing whitespace](https://github.com/prettier/prettier/issues/4223#issuecomment-380093772). With that, the important part is whether or not the whitespace comes right before a line break, not whether spaces are inside or outside tags. In this case the whitespace is not being used across a line break, so no special syntax is needed.

The Prettier pre-commit hook will auto-format any whitespace that needs to use `{' '}` syntax, including when it makes the choice to split your one-liner across multiple lines. It's also smart enough to take `{' '}` **out** and replace them with spaces if they're not needed.

maggieneterval: Thank you for being awesome, @erikmunson and Prettier. :hugs:

PR #7419, opened 2019-09-23
feat(core/pipeline): Add timestamp for failed executions

jervi:
From
image

to
image


maggieneterval:
@jervi you're on a roll with these Deck fixes, thanks so much! LGTM!

PR #7421, opened 2019-09-23
fix(core/pipeline): Fix execution details chevron for grouped stages

jervi:
The Execution Details link in the pipeline view doesn't work if the first stage in the pipeline is a group. I think this was by design, but it has been a major cause of confusion for our users, so I think it makes sense to just expand the first stage with the first substage selected.
image


maggieneterval:
LGTM! Also adding the marvelous @kevinawoo who I think recently revamped this section of the UI :rocket:

PR #7425, opened 2019-09-24
fix(core/pipeline): migrate more manual execution field layouts

erikmunson:
so... in #7413 I made a pretty glaring exception in migrating over fields to `ManualExecutionFieldLayout`. Spoiler alert: it's parameters.

Screen Shot 2019-09-23 at 9 12 33 PM

Parameters ended up in a double-nested bootstrap grid, which is obviously great. This PR moves them over to the context-provided layout just like the other fields covered in #7413: 

Screen Shot 2019-09-23 at 9 15 41 PM

### DIFF
--- a/app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
@@ -364,7 +364,7 @@ export class ManualExecutionModal extends React.Component<IManualExecutionModalP
                   )}
                   {formik.values.pipeline && (
                     <div className="form-group">
-                      <div className="col-md-6 col-md-offset-4">
+                      <div className={pipelineOptions.length > 0 ? 'col-md-6 col-md-offset-4' : 'col-md-10'}>
                         <p>
                           This will start a new run of <strong>{formik.values.pipeline.name}</strong>.
                         </p>
--- a/app/scripts/modules/core/src/pipeline/manualExecution/Parameters.tsx
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/Parameters.tsx
@@ -47,51 +47,39 @@ export class Parameters extends React.Component<IParametersProps> {
         </p>
         {visibleParameters &&
           visibleParameters.map((parameter, i) => {
+            const fieldProps = {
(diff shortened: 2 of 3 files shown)


app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
```diff
@@ -364,7 +364,7 @@ export class ManualExecutionModal extends React.Component<IManualExecutionModalP
                   )}
                   {formik.values.pipeline && (
                     <div className="form-group">
-                      <div className="col-md-6 col-md-offset-4">
+                      <div className={pipelineOptions.length > 0 ? 'col-md-6 col-md-offset-4' : 'col-md-10'}>
```
erikmunson: after changing away from `col-md-10` in the last PR, I decided that each view (manual execution w/ pipeline picker + manual execution of a specific pipeline) needed it's own layout to look right.

app/scripts/modules/core/src/pipeline/manualExecution/layout/ManualExecutionFieldLayout.tsx
```diff
@@ -18,10 +18,11 @@ export class ManualExecutionFieldLayout extends React.Component<IFieldLayoutProp
...
     return (
       <div className="sp-margin-m-bottom">
-        <div className={'form-group'}>
+        <div className="form-group">
           {showLabel && (
-            <label className={'col-md-4 sm-label-right'}>
-              {label} {help}
+            <label className="col-md-4 sm-label-right break-word">
+              {label}
+              {required && <span>*</span>} {help}
```
erikmunson: the `required` treatment and `break-word` class were both on the hand-rolled parameter layouts, and it didn't seem problematic to make them a thing for all the fields on this modal.

app/scripts/modules/core/src/pipeline/manualExecution/Parameters.tsx
```diff
@@ -47,51 +47,39 @@ export class Parameters extends React.Component<IParametersProps> {
         </p>
         {visibleParameters &&
           visibleParameters.map((parameter, i) => {
+            const fieldProps = {
```
christopherthielen: ❤️

PR #7427, opened 2019-09-24
fix(core/pipeline): Fix revert button regression for templated pipelines

jervi:
When pressing the revert button for a templated pipeline, the original pipeline got merged with the rendered pipeline, resulting in a strange frankenstein pipeline that you couldn't save.


jervi:
Regression in https://github.com/spinnaker/deck/pull/7318
FYI @Jammy-Louie

Jammy-Louie:

PR #7428, opened 2019-09-24
fix(core/pipeline): Don't break templated pipelines when updating config

jervi:
When updatePipelineConfig is called on a templated pipeline, it breaks the UI rendering of the pipeline (since it is based on `$scope.plan`, not `$scope.pipeline`). This happens when you open the pipeline config screen for some pipelines (for instance if you have expected artifacts defined in the old format without artifact account).
Removing the line in this PR solves the problem, while it seems like the functionality still works (because angular watches `$scope.pipeline` and calls `updatePipeline()` when it changes).

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
+++ b/app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
@@ -537,7 +537,6 @@ module.exports = angular
         $scope.$applyAsync(() => {
           $scope.pipeline = _.cloneDeep($scope.pipeline);
           _.extend($scope.pipeline, changes);
-          $scope.renderablePipeline = $scope.pipeline;
           markDirty();
         });
       };


Jammy-Louie:

christopherthielen:
Looks like this commit has broken saving of pipelines (non-templated?).

Deck saves the pipeline successfully, but then the ORIGINAL pipeline JSON is rendered, causing confusion.  Refreshing the entire browser then gets the new pipeline JSON.

jervi:
Gah. Sorry about that. It’s midnight here now, but I’ll look at it first thing tomorrow morning. If you feel you need to revert it, just do so and I’ll create a new PR 👍

christopherthielen:
https://github.com/spinnaker/deck/pull/7437

I put the original code back in there, but conditionally applied it based on `!isTemplatedPipeline`

jervi:
I’ll test it tomorrow. Thanks!

PR #7435, opened 2019-09-24
feat(core/forms): Remove all async support from the spinnaker validation apis

christopherthielen:
After discussions with Alan Quach, we decided that async validation can
be externalized, and async support in this API reduced the ability for
developers to debug their form validations.  We also noted that we were
not using async validation anywhere in Deck.


alanmquach:
❤️ love it. Less is more!

PR #7440, opened 2019-09-26
fix(core/pipeline): Fix revert button for non-templated pipelines

jervi:
I accidentally broke this in #7427. Adding the removed line back, but only for non-templated pipelines.


louisjimenez:

PR #7441, opened 2019-09-26
fix(core/pipeline): "Depends On" doesn't always update when reverting

jervi:
Reverting these lines to pre #7318 fixes the issue

revert_button


jervi:
@Jammy-Louie PTAL

Jammy-Louie:

PR #7445, opened 2019-09-26
fix(pipeline): unset `locked` instead of `lock` when unlocking pipeline

Jammy-Louie:


Jammy-Louie:

PR #7448, opened 2019-09-27
feat(precondition): add custom message to precondition

gal-yardeni:
Adding a user description field when providing precondition expression. In case of failure, instead of showing a generic error message such as "'No reason provided", the user custom description will be shown.


gal-yardeni:
FYI @luispollo

app/scripts/modules/core/src/pipeline/config/preconditions/types/expression/additionalFields.html
```diff
@@ -16,4 +16,13 @@
       <input type="checkbox" ng-model="precondition.failPipeline" />
     </div>
   </div>
+
+  <div class="form-group row">
+    <div class="col-sm-3 sm-label-right">
+      Description <help-field key="pipeline.config.checkPreconditions.description"></help-field>
```
marchello2000: you probably are missing changes to `help.content.ts` to include the help text.

Might be useful in the helptext to say something like "This description will be shown to the user if the precondition evaluates to false"

marchello2000:
one note, otherwise looks good

app/scripts/modules/core/src/pipeline/config/preconditions/types/expression/additionalFields.html
marchello2000: Also... i wonder if it's better to call this `Error Description`..? I defer to @erikmunson

erikmunson: Yes, thanks Mark. Was going to pop in and comment on both these points...

- For the help field, you'll want to add something into `help.contents.ts` as Mark mentioned. Here's the definition for the 'failPipeline' field above this one so you have an example: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/help/help.contents.ts#L79

- I'd like to see this change from being called `description` to `failureMessage` — similar to the name of the thing we put into stages when they fail. Suspect that will connect/relate better with folks trying to figure out what to configure.

- To @gal-yardeni's point, I'd also recommend _not_ showing this when things work correctly. If it really is a description, let's show it in the list of fields and _not_ make it an error. If it's going to show up as an error, let's make it show up as an error and _not_ when things succeed.

gal-yardeni: Thanks @marchello2000 , will make the first change.
Not sure whether "error description" is better since the description field is shown in the details screen whether it was a success or failure.

app/scripts/modules/core/src/pipeline/config/stages/checkPreconditions/CheckPreconditionsExecutionDetails.tsx
```diff
@@ -1,12 +1,17 @@
...
+
 import { ExecutionDetailsSection, IExecutionDetailsSectionProps } from '../common';
 import { robotToHuman } from 'core/presentation/robotToHumanFilter/robotToHuman.filter';
 
 export function CheckPreconditionsExecutionDetails(props: IExecutionDetailsSectionProps) {
   const context = get(props.stage, 'context.context', {} as any);
+  const userDefinedMessage = context['description'];
+  const failureMessage = props.stage.failureMessage;
+  const stageFailureMessage =
+    failureMessage == null || failureMessage === ErrorMsgs.NO_REASON_PROVIDED ? userDefinedMessage : failureMessage;
```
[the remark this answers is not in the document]
gal-yardeni: Looks like this error message coming from the UI and not from ocra.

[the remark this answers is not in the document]
erikmunson: 👍 cool — per our in-person chat let's just check `failureMessage == null` here and omit the check for `No Reason Provided.`

gal-yardeni:
Addressed your comments - @erikmunson , @marchello2000

PR #7449, opened 2019-09-27
fix(core/pipeline): Revision history is vertically challenged in Safari

jervi:
Safari renders the viewport of the revision history with a height of 34px, which leaves a lot of whitespace unused, and also makes it hard to review the revision history as only two lines can be displayed at once.

image

The `padding-left` override also makes the revision history stop cropping the pluses and minuses (in all browsers).

image


louisjimenez:

louisjimenez:
Cherry picked failed because this is adding CSS to a class that didn't exist until after 1.1.6 was cut. Instead this can go in with the rest of the changes for 1.17.x

PR #7452, opened 2019-09-27
feat(core/utils): Add `traverseObject` which deeply walks object properties

christopherthielen:
A utility function I intend to use when adding some form validation enhancements.

```js
  it('walks nested objects inside array properties of an object', () => {
    const object = { foo: 1, bar: [{ name: 'abc' }, { name: 'def' }, { name: 'ghi' }] };
    traverseObject(object, defaultCallback);
    expect(keysWalked).toEqual(['foo', 'bar[0].name', 'bar[1].name', 'bar[2].name']);
    expect(valsWalked).toEqual([1, 'abc', 'def', 'ghi']);
  });
```

### DIFF
--- a/app/scripts/modules/core/src/utils/json/traverseObject.ts
+++ b/app/scripts/modules/core/src/utils/json/traverseObject.ts
@@ -1,26 +1,44 @@
 import { forIn, isPlainObject } from 'lodash';
 
 /**
- * Deeply walks an object tree and invokes `callback` for every simple property leaf node
- * (i.e., a property that is neither a nested object, nor an array)
+ * Deeply walks an object tree and invokes `callback` for each node
  *
  * @param object the object to walk
  * @param callback the callback to invoke on each simple leaf nodes.
  *        This callback receives the `path` and `value`.
  *        The `path` is a string representing the path into the object, which is compatible with lodash _.get(key).
  *        The `value` is the value of the simple leaf node.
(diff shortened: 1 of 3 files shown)


christopherthielen:
I realized we will sometimes want to walk all . nodes, not just leaf nodes, so I added a `traverseLeafNodesOnly` argument

app/scripts/modules/core/src/utils/json/traverseObject.ts
```diff
@@ -0,0 +1,44 @@
...
+  function maybeInvokeCallback(isLeafNode: boolean) {
+    if (contextPath !== null && (isLeafNode || !traverseLeafNodesOnly)) {
+      callback(contextPath, obj);
+    }
+  }
+
+  if (isPlainObject(obj)) {
+    maybeInvokeCallback(false);
+    forIn(obj, (val, key) => {
+      const path = contextPath ? contextPath + '.' : '';
```
erikmunson: should this be using `[]` instead of `.` to avoid weirdness with object keys that have periods in them? I remember Alan had to account for that case in formik field names at one point.

erikmunson: the thing i was referring to specifically: https://github.com/spinnaker/deck/commit/7f0fdb4da0af985e0f9312fd6b510fdc5e574df8

christopherthielen: Good call 👍

PR #7455, opened 2019-09-27
fix(kubernetes): Fix merge strategy field

ezimanyi:
* fix(kubernetes): Properly dirty patch stage on resource update 

  We're not currently not marking the stage dirty upon chaning the resource to be selected, which makes it impossible to save (without making some other change or waiting for a digest cycle to happen for another reason).

* fix(kubernetes): Fix manifest source defaulting in patch manifest 

  When creating a new patch manifest stage, we default the 'Manifest Source' selector to 'Text' but the 'Manifest' text box is not visible; the user needs to choose 'Artifact' then 'Text' again to make the text box visible.

* fix(kubernetes): Fix merge strategy field 

  We've been incorrectly setting the merge strategy on options.strategy instead of options.mergeStrategy since 1.15. Start writing to the correct field, and handle updating any stages affected by the bug by deleting the value of 'strategy' and copying it over to mergeStrategy.


ezimanyi:
@spinnakerbot :cherries: :pick: 1.16
@spinnakerbot :cherries: :pick: 1.15

PR #7456, opened 2019-09-27
fix(help text): update webhook help text

emjburns:
Does this look like better help text for the webhook page?


app/scripts/modules/core/src/help/help.contents.ts
```diff
@@ -423,11 +423,11 @@ const helpContents: { [key: string]: string } = {
   'pipeline.config.webhook.progressJsonPath':
     "JSON path to a descriptive message about the progress in the webhook's response JSON. (e.g. <samp>$.buildInfo.progress</samp>)",
   'pipeline.config.webhook.successStatuses':
-    'Comma-separated list of strings that will be considered as SUCCESS status.',
+    'Comma-separated list of strings (that will be returned in the response body) that will be considered as SUCCESS status.',
```
marchello2000: maybe also mention that this implies the user has specified:
`statusJsonPath` to parse the response

PR #7458, opened 2019-09-28
fix(core/task): properly cleanup TaskMonitor polling, fix digest thrashing

erikmunson:
In https://github.com/spinnaker/deck/pull/7433, I mentioned I had no idea why we didn't show the running time on canceled tasks. Now I kind of maybe know why. Along this wonderful journey I also seem to have found a fun cleanup bug in the code path for any React-based modal that uses a TaskMonitor.

*So*. Two things:
1. React-based modals that use a TaskMonitor rely on a `modalInstanceEmulation` method to generate a fake Angular modal. In theory, closing or dismissing the modal bound to a TaskMonitor [should trigger cleanup of the polling cycle for the task being watched](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/task/monitor/TaskMonitor.ts#L60). Turns out, that's [not actually what happens](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/task/monitor/TaskMonitor.ts#L60). This means that tasks which are still running will continue to be polled in the background every second until they eventually complete, and the polling handlers will mutate the appropriate task object with each cycle
2. Separately, when cancelling a task the confirmation modal uses the same `TaskReader.waitUntilTaskMatches()` poller under the hood as TaskMonitor. This will also mutate the task object every second

Turns out that when either 1, 2, or some horrible combination of both at the same time are happening we mutate task objects so rapidly and continuously that every digest cycle can get a new `runningTimeInMs` value. If Angular sees this happen more than 10 times in a row, it assumes we're doing something Very Bad(tm) like mutating scope fields inside rendering code and decides to throw a bunch of errors (along, perhaps, temporarily pausing that field from being watched?). When this gets bad enough (I tried launching a few manual tasks, closing the modals, then cancelling a task) my browser locked up and a few times I've/we've seen the page reset to the default route and lose page state. To fix:
1. The TaskMonitor modal emulator will now reject or resolve the deferred promise that cleanup is dependent on, and I've confirmed polling halts when closing/dismissing
2. Instead of trying to fundamentally change the way we render running time to alleviate the second issue with cancellation polling, the template will stop rendering the running time *while we are in the process of polling for cancellation* and resume showing running time when complete

Learning things is fun!


app/scripts/modules/core/src/task/task.read.service.ts
```diff
@@ -26,17 +26,13 @@ export class TaskReader {
     return API.one('tasks', taskId)
       .get()
       .then((task: ITask) => {
-        OrchestratedItemTransformer.defineProperties(task);
```
erikmunson: this seemed to be entirely unnecessary, but it's theoretically possible that doing this transform, then the potential `task.execution` transform, then a second identical transform is somehow necessary. I dunno.

PR #7460, opened 2019-09-29
refactor(core/validation): move FormValidator classes to separate files

christopherthielen:
Create a file for FormValidator and FormValidatorField.
Keep interfaces and helpers in validation.ts.


christopherthielen:
@alanmquach fyi

PR #7463, opened 2019-09-30
fix(bakeManifest): fix bake manifest UI rendering

ethanfrogers:
rendering was messed up when kustomize was disables via settings.
previously it would render both forms when no rendering option was
selected.


ethanfrogers:
@maggieneterval we'll want to cherry pick this into 1.16 as well. I'm not sure how we went so long without noticing this behavior.

ethanfrogers:

PR #7464, opened 2019-09-30
feat(displayName): Adding display name property for the bakery baseOs options

gal-yardeni:
When choosing a baseOS option in the bake stage, we can set a 'display name' which can we different than the actual 'id' value that the bakery expects for.


erikmunson:
can you post some quick screenshots of what it looks like when using a base image with a display name?

gal-yardeni:
Screen Shot 2019-09-27 at 2 10 55 PM


@erikmunson

PR #7468, opened 2019-10-01
fix(artifacts/bitbucket) Allow updates to the artifact text input even when the regex pattern does not match

battlecow:
Fixes https://github.com/spinnaker/spinnaker/issues/4958

Updated regex to include the bitbucket server regex pattern, allows updates to the text input box (value) even when the regex does not match to allow natural user input versus forcing copying and pasting the full URL.


erikmunson:
congrats on your first PR!

battlecow:
@spinnakerbot 🍒 ⛏ 1.16

battlecow:

PR #7473, opened 2019-10-01
fix(monitored deploy): properly initialize defaults

marchello2000:
* use `defaultsDeep` to correctly initlialize and not fail when `rollback` object is undefined
* also, show better UX when a stage hasn't run yet


app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
```diff
@@ -70,6 +73,11 @@ function getDeploymentMonitorAdditionalDetails(stage: IExecutionStage) {
   return null;
 }
 
+function isRunningOrComplete(stage: IExecutionStage) {
+  const runOrCompleteStatuses = ['RUNNING', 'SUCCEEDED', 'TERMINAL', 'FAILED_CONTINUE', 'PAUSED', 'CANCELED'];
```
erikmunson: the `IExecutionStage` object has a few helpers you can use to check things like this — perhaps using those fields will help here (not sure, haven't checked closely): https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/orchestratedItem/orchestratedItem.transformer.ts#L61-L90

christopherthielen: What statuses are you trying to guard against?  is it just `NOT_STARTED`, or does it include others?  If we inverted this status check, what would it test?

I am not very comfortable with assigning semantic meaning to lists (i.e., this list is "Running OR Complete") outside of a centralized area.  It worries me because we start to have different answers for "Is this stage complete?" in different areas of the app.  

Unfortunately, deck doesn't currently have centralized logic for testing these statuses, so there's no blaming here.  I'm just voicing my concern.

marchello2000: TOTALLY agree, and I was sort of sad writing this... I guess I can use `hasNotStarted` from the transformer @erikmunson mentioned, how do you feel about that?

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
```diff
@@ -108,17 +111,23 @@ export class DeploymentMonitorExecutionDetails extends React.Component<
...
-                <span>{stage.context.currentProgress}%</span>
-              </dd>
-              <dt>Details</dt>
-              <dd>{details || notProvidedFragment}</dd>
-              <dt className="sp-margin-l-bottom">Log</dt>
-              <dd className="sp-margin-l-bottom">{log ? <pre>{log}</pre> : notProvidedFragment}</dd>
-              <dt>More info</dt>
-              <dd>{additionalDetails ? '' : notProvidedFragment}</dd>
-              {additionalDetails}
+              {stage.hasNotStarted ? (
```
erikmunson: why the ternary with an empty string vs. `stage.hasNotStarted && (....)`?

erikmunson: or... maybe we want to add an exact inverse method (`stage.hasStarted`?)

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
```diff
@@ -108,17 +111,23 @@ export class DeploymentMonitorExecutionDetails extends React.Component<
...
+                  <dt className="sp-margin-l-bottom">Deploy %</dt>
+                  <dd className="sp-margin-l-bottom">
+                    <span>{stage.context.currentProgress}%</span>
+                  </dd>
+                  <dt>Details</dt>
+                  <dd>{details || notProvidedFragment}</dd>
+                  <dt className="sp-margin-l-bottom">Log</dt>
+                  <dd className="sp-margin-l-bottom">{log ? <pre>{log}</pre> : notProvidedFragment}</dd>
+                  <dt>More info</dt>
+                  <dd>{additionalDetails ? '' : notProvidedFragment}</dd>
```
erikmunson: not sure what's up with this markup... do you need an empty `<dd>` here if `additionalDetails` exists? Or could you consolidate this line and the one after it into:
```typescript
{additionalDetails || (<dd>notProvidedFragment</dd>)}
```

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
christopherthielen: yeah, I think you can simply call `stage.hasNotStarted()`, neat

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
marchello2000: yes, it's needed, because it's sort of a table:
if there is no additionalDetails:
```
more info | <not provided>
```

if we have the additionalDetails:
```
more info | 
detailKey | detailText/Link
detailKey | detailText/Link
```

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
marchello2000: i thought the ternary was easier to read, if not started, show nothing, otherwise ...

vs. if not not started show something. i am cool either way

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
erikmunson: ah i see, thanks. These `<dd>` / `<dt>` layouts get me confused.

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
erikmunson: 🤷‍♂ seems reasonable. only nitpicky suggestion would be to switch your usage of `''` here and on L127 to use `null` instead (wading into super nit territory, null is usually more idiomatic in JSX for 'I don't want to render anything here')

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
marchello2000: agree!

app/scripts/modules/core/src/pipeline/config/stages/monitoreddeploy/DeploymentMonitorExecutionDetails.tsx
marchello2000: gotcha, done

PR #7474, opened 2019-10-01
feat(plugins): able to load plugin stages into UI

bpowell:
# Motivation
This adds the ability to add stages as plugins in the UI. There is an interface for plugin developers to use to create a UI for their stages. The stage UI can be built using ReactJS.

# How It Works
When Deck is starting up, it will look to see if any plugins are defined in `spinnakerSettings`. If there are plugins defined, a script tag will be appended to the bottom of the page and the plugin will have its initialized method called. The plugin initialize method takes in the `Registry` so the stage can register itself as a stage. Then the plugin developer has to call `window.spinnakerSettings.onPluginLoaded` with their plugin object.

# Considerations
We removed the `ng-app` from `index.deck` because of using `modules` in a promise. The bootstrapping of deck now happens in a separate Javascript file `app/scripts/bootstrap.js`. This script is added to the page via webpack. The reason for this change was tests were failing when calling `bootstrap`. `bootstrap` only happened before when loading the page by hitting `index.deck`. The tests were expecting things to not be loaded, but with `bootstrap` the entire application was loaded.

When loading plugins we didn't use CommonJS or ES modules due to keeping the list of supported browsers larger.


ethanfrogers:
A little context on the test failures here - We iterated on this PR internally before submitting it upstream and experienced inconsistent failures between CI and running locally. The tests have consistently failed in Travis but @bpowell is the only person who has experienced them failing locally. I ran them myself and got a clean run. We thought that it _might be_ related to the tests running against our internal fork but, now that it's submitted, it appears to not be the case. Any insight would be greatly appreciated!

cc @maggieneterval, maybe you have an idea?

kevinawoo:
fyi, i've seen it break randomly on master too with the same error
<kbd></kbd>

```
....
Chromium 62.0.3202 (Ubuntu 0.0.0) ERROR
  {
    "message": "An error was thrown in afterAll\nScript error. thrown",
    "str": "An error was thrown in afterAll\nScript error. thrown"
  }
Chromium 62.0.3202 (Ubuntu 0.0.0): Executed 1354 of 1410 (skipped 13) ERROR (18.703 secs / 16.485 secs)
> Task :karma FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':karma'.
> Process 'command '/home/travis/build/spinnaker/deck/.gradle/nodejs/node-v10.15.1-linux-x64/bin/node'' finished with non-zero exit value 1

```

maggieneterval:
For anyone like me who was looking for additional context to help understand these changes, here are a few relevant PRs that Brandon sent me to look over:
- [Plugin creator docs PR](https://github.com/spinnaker/spinnaker.github.io/pull/1523)
- [Plugin consumer docs PR](https://github.com/spinnaker/spinnaker.github.io/pull/1524)
- [Halyard PR](https://github.com/spinnaker/halyard/pull/1437)

Looks like there are some pending conversations on the Halyard PR that might impact these changes, so we should probably wait until the Halyard changes are approved before merging this. In the meantime, adding some UI SIG folks as reviewers.

bpowell:
Closing this PR for now. https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1570203724049000

PR #7475, opened 2019-10-02
fix(artifacts/bitbucket): Update the help key to the correct reference

battlecow:
Fixes the help text to use the correct reference.


battlecow:

PR #7476, opened 2019-10-02
feat(core/presentation): Put margin between StandardFieldLayout's input and validation

christopherthielen:
Screen Shot 2019-10-02 at 10 50 37 AM

Adding margin in prep for switching to the newer form field validation styles


erikmunson:
nice 👍

PR #7480, opened 2019-10-02
fix(titus/pipeline): use proper git ssh URL for docker bakes

erikmunson:
Most Titus users at Netflix use the option to pull git repository information from the pipeline trigger, which uses expressions to build up a git URL which is passed to our internal bakery.

Turns out that when *not* using that option and specifying a branch/hash/etc. yourself in the stage, we weren't putting the appropriate query string `?` before the hash component of the URL.


erikmunson:
i don't make the rules, i just follow them 🤷‍♂

PR #7483, opened 2019-10-03
fix(artifacts/ui): Update speltext react element to maintain cursor position

battlecow:
This is my first foray into React so some more eyes would be good to have on this changeset.  Everytime an artifact element using react elements is updated the text input moves the cursor to the end of the input.  This PR maintains the state of the cursor across all elements subclassing the SpelText react element.

### DIFF
--- a/app/scripts/modules/core/src/widgets/spelText/SpelText.tsx
+++ b/app/scripts/modules/core/src/widgets/spelText/SpelText.tsx
@@ -33,9 +33,14 @@ export class SpelText extends React.Component<ISpelTextProps, ISpelTextState> {
   private readonly spelInputRef: any;
   private destroy$ = new Subject();
   private $input: JQLite;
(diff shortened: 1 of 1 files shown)


maggieneterval:
@battlecow you're on a roll with these Deck PRs, thank you so much!

I suspect that the root cause of the cursor jumping to the end of this input actually lives in one of its parent components causing a render cycle with some stale input value that is then corrected, but I'll dig into this a bit more and get back to you before addressing this PR.

maggieneterval:
@battlecow I started looking into this and ended up discovering that the root cause is how we're keeping React and Angular state in sync (I _think_)! PR is here if you want to take a look: https://github.com/spinnaker/deck/pull/7486

I'm going to close this PR for now, but I really really appreciate you bringing this issue to light! :pray: Keep the wonderful Deck contributions coming! :tada:

PR #7484, opened 2019-10-03
fix(triggers): fix a few minor issues with manual execution triggers

maggieneterval:
While investigating a bug with manual re-runs of pipelines triggered on Docker images (that turned out to be an [Echo issue](https://github.com/spinnaker/echo/pull/669)), I also found a few small regressions in Deck:

- Point "Artifact Constraints" field at correct help text

- When attempting to match a pipeline trigger to an execution trigger, use deep equality to compare properties (currently, there will never be a match because the pipeline trigger's `expectedArtifactIds`, a list, will never triple-equal the execution trigger's `expectedArtifactIds`). Also, only compare the pipeline trigger to the modified execution trigger, which if a manual trigger, has had its type changed from `manual` (which will never match the type of a pipeline trigger).
- When hydrating the matched pipeline trigger with properties from the execution trigger, the matched pipeline trigger should be the target and the execution trigger the source, not vice-versa. I think this logic was accidentally reversed in the conversion to React (see line 124 of `manualPipelineExecution.controller.js` before [this PR](https://github.com/spinnaker/deck/pull/7075/files)).

### DIFF
--- a/app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { Formik, Form } from 'formik';
 import { Modal } from 'react-bootstrap';
 import { Observable, Subject } from 'rxjs';
-import { assign, clone, compact, extend, get, head, uniq, isArray, isEmpty, pickBy } from 'lodash';
+import { assign, clone, compact, extend, get, head, uniq, isArray, isEmpty, isEqual, pickBy } from 'lodash';
 
 import { SubmitButton, ModalClose } from 'core/modal';
 import { Application } from 'core/application';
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx
```diff
@@ -289,7 +289,7 @@ export class ManualExecutionModal extends React.Component<IManualExecutionModalP
     const triggers = this.formatTriggers(pipeline && pipeline.triggers ? pipeline.triggers : []);
     let trigger: ITrigger;
     if (this.props.trigger) {
-      // Certain fields like correlationId will cause unexepcted behavior if used to trigger
+      // Certain fields like correlationId will cause unexepected behavior if used to trigger
```
erikmunson: ooo thanks for fixing my typos 😄

dbyron0: No good deed goes unpunished...sorry to be the bearer of bad news (s/unexepected/unexpected):
```suggestion
      // Certain fields like correlationId will cause unexpected behavior if used to trigger
```

maggieneterval: OMG I made it worse! Thanks @dbyron0, I'll fix this with my next Deck PR. So embarrassing :see_no_evil:

erikmunson: don't be emabrrasde magige, we ali maek ytpos msomeitmes!

PR #7485, opened 2019-10-03
fix(pipeline): triggers were not reverting in the ui

Jammy-Louie:
Triggers were not being reverted when revert was being triggered.

https://jaredpalmer.com/formik/docs/api/formik#enablereinitialize-boolean

set `enableReinitialize ` to true in formik to resolve the problem


maggieneterval:
@Jammy-Louie I flailed around in the UI for a while trying to figure out what you meant before asking this probably dumb question but...how do you revert a trigger?

And just to clarify my formik understanding, I'm assuming we should set `enableReinitialize` to true if and only if there are events that can occur outside the scope of the formik-managed form that should be able to reset the form values?

christopherthielen:
> we should set enableReinitialize to true if and only if there are events that can occur outside the scope of the formik-managed form that should be able to reset the form values?

Yes, exactly.  If something outside of formik is managing the form state then this flag will allow those changes to affect formik state.

christopherthielen:
> how do you revert a trigger?

When reverting edits to a pipeline with the Revert button on the bottom left

Jammy-Louie:

maggieneterval:
Thanks @christopherthielen, you learn something new about Spinnaker every day!

christopherthielen:
Unfortunately, this change has caused problems with the validation code in the react triggers.  I think the entire formik state gets reset, including `touched` and `errors`.


trigger bug

Jammy-Louie:
@christopherthielen I'll revisit this again and hopefully come up with something that satisfies everything

PR #7488, opened 2019-10-03
feat(core/presentation): Add helper functions for generating categorized validation messages

christopherthielen:
example: 

```js
const v = new FormValidator(values);
v.field('foo').withValidators(value => {
  return value === 'foo' && errorMessage('NO, NOT FOO!!!')
});
```

also moved all category stuff into categories.ts

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/validation/categories.spec.ts
+++ b/app/scripts/modules/core/src/presentation/forms/validation/categories.spec.ts
@@ -1,4 +1,4 @@
-import { categorizeErrorMessage, categorizeErrors } from './categorizedErrors';
+import { categorizeErrorMessage, categorizeErrors } from './categories';
 
 describe('categorizeErrorMessage', () => {
   it('returns an array of length 2', () => {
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/presentation/forms/validation/categories.ts
```diff
@@ -11,6 +24,14 @@ const inverseLabels: { [label: string]: IValidationCategory } = Object.keys(cate
   {},
 );
 
+export const asyncMessage = (message: string) => `Async: ${message}`;
+
```
christopherthielen: why is there a blank line here 🤦‍♂

app/scripts/modules/core/src/presentation/forms/validation/categories.spec.ts
```diff
@@ -1,29 +1,64 @@
...
+
+  it('errorMessage should prefix a message with Error:', () => {
+    expect(errorMessage('the quick brown fox')).toBe('Error: the quick brown fox');
+  });
+
+  it('infoMessage should prefix a message with Info:', () => {
+    expect(infoMessage('the quick brown fox')).toBe('Info: the quick brown fox');
+  });
+
+  it('messageMessage should prefix a message with Message:', () => {
```
christopherthielen: not sure how i feel about `messageMessage`

maggieneterval: giphy

PR #7491, opened 2019-10-04
fix(core): Separate how config and plans are updated, add tests

louisjimenez:
This adds some testing for configuring v2 templated pipelines. 

It also make sure that the pipeline config shown when editing as JSON does not display inherited items and only shows what is directly saved to the config. To see the pipeline plan, including its inherited template items, the user can go into the "Rendered Pipeline" view.

image

https://github.com/spinnaker/spinnaker/issues/4887


app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.html
```diff
@@ -126,6 +126,7 @@ <h3>
   <div class="pipeline-contents">
     <pipeline-config-view
       application="application"
+      is-v2-templated-pipeline="isV2TemplatedPipeline"
```
louisjimenez: this is the only change in this file, the rest is formatting

louisjimenez:

PR #7498, opened 2019-10-08
fix(core/pipeline): KLUDGE: use react 'key' to reinitialize formik when pipeline reverted

christopherthielen:
This is a workaround for angular/react interactions.

Previously, we had turned on `enableReinitialize` for trigger formik forms.
This allowed angular code to revert the pipeline and the changes would flow through through to formik via 'initialValues'.
However, turning on that prop had the unwanted side effect of reinitializing the form whenever _anything_ changed.
This caused problems with the validation logic.

This workaround causes the triggers component to be remounted whenever the pipeline is rolled back.

I think a longer term solution is to refactor the pipeline configuration screen to purely React, then adding adapter code for stages that are written in angularjs.  This would basically flip the angular/react adapter direction.


Jammy-Louie:

erikmunson:
wooo. this is fun. I approve.

PR #7501, opened 2019-10-08
refactor(application): reactify delete application section

Jammy-Louie:
Convert the Delete application section from Angular to React

OLD (Angular):
Screen Shot 2019-10-08 at 10 49 10 AM

NEW (React):
Screen Shot 2019-10-08 at 10 49 47 AM


app/scripts/modules/core/src/application/config/DeleteApplicationSection.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { FirewallLabel } from 'core/securityGroup/label';
+
+export interface IDeleteApplicationSection {
+  application: Application;
+}
+
+export function DeleteApplicationSection(props: IDeleteApplicationSection) {
+  const { application } = props;
+
+  function deleteApplication(): void {
```
caseyhebebrand: nit: You can use an arrow function here.

app/scripts/modules/core/src/application/config/DeleteApplicationSection.tsx
```diff
@@ -0,0 +1,59 @@
...
+      title: 'Deleting ' + application.name,
+      hasKatoTask: false,
+      onTaskComplete: () => {
+        ReactInjector.$state.go('home.infrastructure');
+      },
+    };
+
+    ReactInjector.confirmationModalService.confirm({
+      header: 'Really delete ' + application.name + '?',
+      buttonText: 'Delete ' + application.name,
```
caseyhebebrand: nit: For all the string interpolation in this function you can use ES6 template literals.

```js
`Really delete ${application.name}?`
```

Jammy-Louie: 👍

app/scripts/modules/core/src/application/config/DeleteApplicationSection.tsx
Jammy-Louie: 👍

PR #7505, opened 2019-10-09
fix(artifact/bitbucket): Bitbucket Use Default Artifact fix

jorgebee65:
# Bitbucket Default analysis
## Current behavior
When **Use Default Artifact** option is checked the **Object path** field is NOT editable.

The issue is due the method **onReferenceChanged** the variable **results** is always null
```
private onReferenceChanged = (reference: string) => {
	const pathRegex = new RegExp('/1.0/repositories/[^/]*/[^/]*/raw/[^/]*/(.*)$');
	const results = pathRegex.exec(reference);
	if (results !== null) {
		const clonedArtifact = cloneDeep(this.props.artifact);
		clonedArtifact.name = decodeURIComponent(results[1]);
		clonedArtifact.reference = reference;
		this.props.onChange(clonedArtifact);
	}
};
```
The correct way to do that is pulled out the **clonedAtifact** constant and then check if **results** exists.
```
    private  onReferenceChanged  = (reference:  string) => {
    	const  pathRegex  =  new  RegExp('/1.0/repositories/[^/]*/[^/]*/raw/[^/]*/(.*)$');
    	const  results  =  pathRegex.exec(reference);
    	const  clonedArtifact  =  cloneDeep(this.props.artifact);
    	clonedArtifact.reference  =  reference;
    	if (results  !==  null) {
    		clonedArtifact.name  =  results[1];
    	} else {
    		clonedArtifact.name  =  reference;
    	}
    	this.props.onChange(clonedArtifact);
    };
```
Doing this, the field is editable now.
  

## RegEx Issue
  The RegEx is not matching with the 1.0 api rest format.
**Current:**
`/1.0/repositories/[^/]*/[^/]*/raw/[^/]*/(.*)$`
  **Proposed**
  `.*/rest/api/1.0/[^/]*/[^/]*/repos/[^/]*/raw/(.*)$`

Replacing this regEx then the values for **name** and **reference** are well formed.

## Conclusion
Modify the method and change the RegEx in order to support 1.0 API rest.

  ```
private  onReferenceChanged  = (reference:  string) => {
	const  pathRegex  =  new  RegExp('.*/rest/api/1.0/[^/]*/[^/]*/repos/[^/]*/raw/(.*)$');
	const  results  =  pathRegex.exec(reference);
	const  clonedArtifact  =  cloneDeep(this.props.artifact);
	clonedArtifact.reference  =  reference;
	if (results  !==  null) {
		clonedArtifact.name  =  results[1];
	} else {
		clonedArtifact.name  =  reference;
	}
	this.props.onChange(clonedArtifact);
};
```

### DIFF
--- a/app/scripts/modules/core/src/help/help.contents.ts
+++ b/app/scripts/modules/core/src/help/help.contents.ts
@@ -148,7 +148,7 @@ const helpContents: { [key: string]: string } = {
       <p>The version of chart you want to trigger on changes to</p>`,
   'pipeline.config.expectedArtifact.defaultBitbucket.reference': `
       <p>The Bitbucket API file url the artifact lives under. The domain name may change if you're running your own Bitbucket server. The repository and path to files must be URL encoded.</p>
-      <p>An example is <code>https://api.bitbucket.org/1.0/repositories/$ORG/$REPO/raw/$VERSION/$FILEPATH</code>. See <a href="https://www.spinnaker.io/reference/artifacts/types/bitbucket-file/#fields">our docs</a> for more info.</p>`,
+      <p>An example is <code>https://api.bitbucket.com/rest/api/1.0/$PROJECTS/$PROJECTKEY/repos/$REPONAME/raw/$FILEPATH</code>. See <a href="https://www.spinnaker.io/reference/artifacts/types/bitbucket-file/#fields">our docs</a> for more info.</p>`,
   'pipeline.config.trigger.webhook.source': `
       <p>Determines the target URL required to trigger this pipeline, as well as how the payload can be transformed into artifacts.</p>
   `,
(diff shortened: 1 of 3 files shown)


ethanfrogers:
@jorgebee65 looks like there's a conflict on one of these files. Would you mind taking a look? 

Also, after taking a closer look at this, it'd be great if we had some API documentation to back up this change. I know the APIs vary wildly between Bitbucket Cloud and Server so it'd be good to clarify which version of Bitbucket this was tested against. If the original API was for Bitbucket Cloud and this is for server we'll probably be causing a backwards incompatible change here.

ethanfrogers:
Looking at [this documentation](https://developer.atlassian.com/bitbucket/api/2/reference/resource/) it does appear that the original regex was written against Bitbucket Cloud 1.0 APIs so I don't think this is the change we actually want to make. 

I feel like we may want to figure out a more complete solution to this issue. Perhaps we should look into making a Bitbucket Server artifact type to handle this?

ethanfrogers:
After talking with @jorgebee65 offline I think we've come up with a less backwards incompatible way to fix this. I'm going to close this PR for now.

PR #7510, opened 2019-10-10
feat(core/presentation): allow markdown in ValidationMessage

christopherthielen:
This renders markdown in the `ValidationMessage` component

```
const markdown = errorMessage(`A thing happened\n\n[Click here](http://internal.service.com/error?id=12345) for deets`);
return <ValidationMessage message={markdown} type={'error'} />
```

Screen Shot 2019-10-10 at 3 03 18 PM

### DIFF
--- a/app/scripts/modules/core/src/presentation/Markdown.less
+++ b/app/scripts/modules/core/src/presentation/Markdown.less
@@ -1,0 +1,9 @@
+.Markdown {
+  p {
+    margin: 0;
+  }
+
+  p + p {
+    margin-top: 4px;
+  }
+}
--- a/app/scripts/modules/core/src/presentation/Markdown.tsx
+++ b/app/scripts/modules/core/src/presentation/Markdown.tsx
@@ -1,40 +1,40 @@
 import * as React from 'react';
 import { HtmlRenderer, Parser } from 'commonmark';
 import * as DOMPurify from 'dompurify';
+import './Markdown.less';
 
 export interface IMarkdownProps {
   [key: string]: any;
 
   /** markdown */
   message: string;
+
   /** optional tag */
   tag?: string;
+
+  /** The className(s) to apply to the tag (.Markdown class is always applied) */
+  className?: string;
 }
 
 /**
  * Renders markdown into a div (or some other tag)
  * Extra props are passed through to the rendered tag
  */
-export class Markdown extends React.Component<IMarkdownProps> {
-  public static defaultProps: Partial<IMarkdownProps> = {
-    tag: 'div',
-  };
+export function Markdown(props: IMarkdownProps) {
+  const { message, tag: tagProp, className: classNameProp, tag = 'div', ...rest } = props;
 
-  private parser: Parser = new Parser();
-  private renderer: HtmlRenderer = new HtmlRenderer();
+  const parser = React.useMemo(() => new Parser(), []);
+  const renderer = React.useMemo(() => new HtmlRenderer(), []);
+  const className = `Markdown ${classNameProp || ''}`;
 
-  public render() {
-    const { message, tag, ...rest } = this.props;
-
-    if (message == null) {
-      return null;
-    }
-
-    const restProps = rest as React.DOMAttributes<any>;
-    const parsed = this.parser.parse(message.toString());
-    const rendered = this.renderer.render(parsed);
-    restProps.dangerouslySetInnerHTML = { __html: DOMPurify.sanitize(rendered) };
-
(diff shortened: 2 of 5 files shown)


app/scripts/modules/core/src/presentation/Markdown.tsx
```diff
@@ -1,40 +1,41 @@
...
  */
-export class Markdown extends React.Component<IMarkdownProps> {
-  public static defaultProps: Partial<IMarkdownProps> = {
-    tag: 'div',
-  };
-
-  private parser: Parser = new Parser();
-  private renderer: HtmlRenderer = new HtmlRenderer();
+export function Markdown(props: IMarkdownProps) {
+  const { message, tag: tagProp, className: classNameProp, ...rest } = props;
```
louisjimenez: what about destructuring with a default value here:
`tag = 'div'`
or is the concern that tag prop could be null?

app/scripts/modules/core/src/presentation/Markdown.less
```diff
@@ -0,0 +1,9 @@
+.Markdown {
```
louisjimenez: Curious if the capitalization here is a convention in deck?

erikmunson: Generally we apply capitalization like this when a few things are true:
1. The purpose of the class is to select all instances of a specific component type, and nothing else
2. The class sits on the root element of that component type
3. The class name is exactly identical to the name of the component type

The third point kind of goes hand-in-hand with the capitalization.

christopherthielen: This is a convention that has sort of organically appeared.  It's a simple way to scope CSS to a specific component.

See also: 

https://github.com/spinnaker/deck/blob/b5534656251c92180f8598cb22e8f50574e9cf1a/app/scripts/modules/core/src/presentation/forms/validation/ValidationMessage.less#L1

or in raw css:

https://github.com/spinnaker/deck/blob/23437377aace9cc6cd7fec8ecd0613b840cfe0d2/app/scripts/modules/core/src/projects/configure/FormikApplicationsPicker.css#L1

app/scripts/modules/core/src/presentation/Markdown.tsx
```diff
@@ -1,40 +1,41 @@
...
+  const tag = tagProp || 'div';
```
caseyhebebrand: Another option is to use a default prop for `tag` instead of aliasing then reassigning.

app/scripts/modules/core/src/presentation/Markdown.less
louisjimenez: Got it, thanks for the details.

app/scripts/modules/core/src/presentation/Markdown.tsx
```diff
@@ -1,40 +1,41 @@
...
-    }
+  if (message == null) {
+    return null;
+  }
 
-    const restProps = rest as React.DOMAttributes<any>;
-    const parsed = this.parser.parse(message.toString());
-    const rendered = this.renderer.render(parsed);
-    restProps.dangerouslySetInnerHTML = { __html: DOMPurify.sanitize(rendered) };
+  const restProps = rest as React.DOMAttributes<any>;
```
erikmunson: this makes me wish there was a nice way to describe `IMarkdownProps` as a union of `React.DOMAttributes<any>` and the other fields in a way where using `...rest` would automatically let the type system narrow back down to `React.DOMAttributes<any>`. Oh well.

app/scripts/modules/core/src/presentation/forms/validation/ValidationMessage.tsx
```diff
@@ -43,7 +46,7 @@ export const ValidationMessage = (props: IValidationMessageProps) => {
   return (
     <div className={`ValidationMessage ${containerClassName || containerClassNames[type] || ''}`}>
       {showIcon && <i className={iconClassName || iconClassNames[type] || ''} />}
-      <div className="message">{message}</div>
+      <div className="message">{isString(message) ? <Markdown message={message} /> : { message }}</div>
```
erikmunson: I think something about this syntax got me confused. can you help me figure out what `{ message }` is doing in this case? the way i'm reading it is that if `message` is anything other than a string, this will take the value of the `message` variable and make it the value of a `message` **field** inside an object, which will then get given to react to render. what am I missing?

christopherthielen: yes, nice catch!

PR #7511, opened 2019-10-10
fix(core/presentation): Make all DOMPurify'd links open in a new window

christopherthielen:
https://github.com/cure53/DOMPurify/tree/master/demos#hook-to-open-all-links-in-a-new-window-link

This makes it so `DOMPurify.purify(html)` will add `target="_blank"` to all links.  

The assumption here is that whenever we use DOMPurify, we're likely rendering user-provided content, and that any links provided should not blow away the spinnaker session.

### DIFF
--- a/app/scripts/modules/core/src/presentation/presentation.module.js
+++ b/app/scripts/modules/core/src/presentation/presentation.module.js
@@ -1,5 +1,7 @@
 'use strict';
 
+import * as DOMPurify from 'dompurify';
+
 const angular = require('angular');
 
 import { AUTO_SCROLL_DIRECTIVE } from 'core/presentation/autoScroll/autoScroll.directive';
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/presentation/presentation.module.js
```diff
@@ -29,20 +28,4 @@ module.exports = angular
...
-        node.setAttribute('rel', 'noopener noreferrer');
-      }
-      // set non-HTML/MathML links to xlink:show=new
-      if (!node.hasAttribute('target') && (node.hasAttribute('xlink:href') || node.hasAttribute('href'))) {
-        node.setAttribute('xlink:show', 'new');
-      }
-      return node;
-    });
-  });
+  .run(domPurifyOpenLinksInNewWindow);
```
erikmunson: 🙏

PR #7512, opened 2019-10-11
fix(webhook): Hide fail fast status code if they are pre-configured

afcastano:
This fixes spinnaker/spinnaker#5009

Check whether `failFastStatusCodes` field should be shown based on the preconfigured properties list.  This is needed because currently there is no way of pre-configuring this field.

Please refer to the above mentioned issue for more details.

This PR contains the front end changes. Backend changes are made in `orca` as part of https://github.com/spinnaker/orca/pull/3223


afcastano:
Thanks for reviewing this @marchello2000 
I've updated my fork with latest upstream changes. 
Fell free to merge
=)

PR #7514, opened 2019-10-11
fix(bakeManifest/helm): rawOverrides option

ethanfrogers:
adds an option for rawOverrides. this will trigger the use of `--set` in
the bake manifest stage.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/helm/BakeHelmConfigForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/helm/BakeHelmConfigForm.tsx
@@ -195,6 +195,17 @@ export class BakeHelmConfigForm extends React.Component<IBakeHelmConfigFormProps
         </StageConfigField>
         <StageConfigField
           fieldColumns={6}
+          helpKey={'pipeline.config.bake.manifest.helm.rawOverrides'}
+          label="Raw Overrides"
+        >
+          <CheckboxInput
+            value={stage.rawOverrides}
+            text={''}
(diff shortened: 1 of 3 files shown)


ethanfrogers:
related to spinnaker/rosco#441

app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/helm/BakeHelmConfigForm.tsx
```diff
@@ -193,6 +193,17 @@ export class BakeHelmConfigForm extends React.Component<IBakeHelmConfigFormProps
             />
           )}
         </StageConfigField>
+        <StageConfigField
+          fieldColumns={6}
+          helpKey={'pipeline.config.bake.manifest.helm.rawOverrides'}
```
maggieneterval:
```suggestion
          helpKey="pipeline.config.bake.manifest.helm.rawOverrides"
```

maggieneterval: We don't need to wrap string literals in JSX with brackets, but I see that the rest of this file doesn't follow this convention so feel free to leave and we can address in a future commit that we don't plan to cherry-pick

ethanfrogers: 👍 good call out!

ethanfrogers:
@spinnakerbot 🍒 ⛏ 1.16

ethanfrogers:

PR #7520, opened 2019-10-11
feat(ui): Show health check url beside target group

caseyhebebrand:
Introduce a health check link next to the Target Group in Status section of the instance's details. 

After: 
Screen Shot 2019-10-11 at 1 35 01 PM

### DIFF
--- a/app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
+++ b/app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
@@ -2,6 +2,7 @@
 
 const angular = require('angular');
 import _ from 'lodash';
+import { getAllTargetGroups, getTargetGroupHealthCheckInfo } from 'core/instance';
 
 import {
   CloudProviderRegistry,
@@ -70,6 +71,11 @@ module.exports = angular
         var displayableMetrics = instance.health.filter(function(metric) {
           return metric.type !== 'Amazon' || metric.state !== 'Unknown';
         });
+
+        // augment with target group healthcheck data
+        const targetGroups = getAllTargetGroups(app.loadBalancers.data);
+        getTargetGroupHealthCheckInfo(displayableMetrics, targetGroups);
+
         // backfill details where applicable
         if (latest.health) {
           displayableMetrics.forEach(function(metric) {
--- a/app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
+++ b/app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
@@ -2,7 +2,7 @@
 
 const angular = require('angular');
 import _ from 'lodash';
-import { getAllTargetGroups, getTargetGroupHealthCheckInfo } from 'core/instance';
+import { getAllTargetGroups, applyHealthCheckInfoToTargetGroups } from 'core/instance';
 
 import {
   CloudProviderRegistry,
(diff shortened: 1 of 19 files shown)


app/scripts/modules/core/src/instance/details/utils.js
```diff
@@ -0,0 +1,23 @@
+import { flatten } from 'lodash';
+
+export const getAllTargetGroups = loadBalancers => {
+  const allTargetGroups = loadBalancers.map(d => d.targetGroups);
+  const targetGroups = flatten(allTargetGroups).reduce((groups, tg) => {
```
christopherthielen: could use `keyBy` from lodash instead of a custom reducer here

app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
```diff
@@ -70,6 +71,11 @@ module.exports = angular
         var displayableMetrics = instance.health.filter(function(metric) {
           return metric.type !== 'Amazon' || metric.state !== 'Unknown';
         });
+
+        // augment with target group healthcheck data
+        const targetGroups = getAllTargetGroups(app.loadBalancers.data);
+        getTargetGroupHealthCheckInfo(displayableMetrics, targetGroups);
```
christopherthielen: if we are going to mutate data in a function, I'd like the function name to indicate this.  I was confused because there was a `get` method but I didn't see the return value used.  
Perhaps something like `applyHealthcheckInfoToTargetGroups`?

app/scripts/modules/titus/src/instance/details/instanceDetails.html
```diff
@@ -145,6 +145,15 @@ <h3 class="horizontal middle space-between flex-1" select-on-dbl-click>
             ng-repeat="targetGroup in metric.targetGroups"
           >
             <instance-load-balancer-health load-balancer="targetGroup"></instance-load-balancer-health>
+            <span class="pad-left small">
```
christopherthielen: Would it be possible to move this markup into the `instanceLoadBalancerHealth` component?

caseyhebebrand: Yes definitely possible. This component is also used for LoadBalancer type health metrics, so I will hide this segment if the url is not present.

app/scripts/modules/core/src/instance/loadBalancer/InstanceLoadBalancerHealth.tsx
```diff
@@ -43,17 +47,35 @@ export class InstanceLoadBalancerHealth extends React.Component<IInstanceLoadBal
...
         <OverlayTrigger placement="left" overlay={tooltip}>
           {healthDiv}
+          {ipAddress && healthCheckPath && healthCheckLinkSpan}
         </OverlayTrigger>
       );
     }
 
-    return healthDiv;
+    return (
+      <React.Fragment>
```
christopherthielen: I think we generally only use `<React.Fragment>` if we need to add a `key` prop.  We use the short version most other places: `<>`

app/scripts/modules/core/tsconfig.json
```diff
@@ -31,7 +31,8 @@
     "paths": {
       "@spinnaker/core": ["."],
       "core/*": ["*"],
-      "core": ["."]
+      "core": ["."],
+      "@spinnaker/amazon": ["../../amazon/lib"]
```
christopherthielen: this dependency should be removed

PR #7109, opened 2019-06-11
feat(k8s): Enable new artifacts workflow in Patch Manifest

jkschneider:
image

PR #7521, opened 2019-10-14
fix(kubernetes): add missing `app` config param for patch manifest st…

maggieneterval:
…ages

Closes https://github.com/spinnaker/spinnaker/issues/5022

Replaces missing `app` parameter from stage config (and removes `isNew` check, since we should properly default any `undefined` config values). Was previously defaulted in angular controller prior to [this commit](https://github.com/spinnaker/deck/pull/7109/files) which converted the Patch stage config to React to add support for artifacts rewrite.


maggieneterval:
@spinnakerbot :cherries: :pick: 1.16

PR #7523, opened 2019-10-14
fix(artifact/bitbucket): Bitbucket Use Default Artifact

jorgebee65:
This change allow to specify explicitly the attribute **name** for a Bitbucket Artifact instead of trying to parse it from the reference.  This decision is due the reference patterns for bitbucket vary widely between versions and would be difficult to maintain them per version.
We're just adding a name field so users can set that themselves. This is a backwards compatible change because existing artifacts will already have the name field populated.


ethanfrogers:
@spinnakerbot assign-issue maggieneterval

maggieneterval:
Hey @jorgebee65, thanks so much for the PR!

I don't know anything about bitbucket so just want to make sure I understand the workflows here: it looks like previously, users specified a reference, and we used regex to derive the name. Now, users can specify both a reference (which is no longer parsed using regex), and an explicit name. Do you now need to specify both the reference and the name? Or are there situations when users only need to specify one? If users now only need to specify one or the other, perhaps we can make that clear in the tooltip text?

ethanfrogers:
@maggieneterval @jorgebee65 is OOO this afternoon so I'll answer for him.

That's a good question. I'm not entirely sure if `name` is required anywhere other than the current (soon to be previous) implementation for Kustomize. Without tracking down everywhere artifacts are used it's probably _safe_ to assume that it's required. So, users will need to specify both, like how the first implementation of Github artifacts were defined. Before the regex was introduced you had to define both manually so this is kind of going back to the original implementation IIRC.

jorgebee65:
Hi @maggieneterval as @ethanfrogers mentioned now users will need to specify both fields because the regex was removed due the bitbucket API vary widely between versions.

PR #7529, opened 2019-10-15
fix(core/presentation): Fix null reference in FieldLayout components

christopherthielen:
The fix was adding a default value when destructuring the validation prop

```
const { label, help, input, actions, validation = {} } = props;
const { hidden, messageNode, category } = validation;
```

Also converted to functional components


christopherthielen:
> is the rest just prettier being aggressive?

switched from class component to functional component

erikmunson:
ah thanks, see that now. enough noise that my skimming didn't catch :)

caseyhebebrand:
> switched from class component to functional component

Nice! I am a huge fan of using functional components where possible. They help to reduce bundle size and custom hooks will help unlock abstraction, which I see being a big advantage for migrating controller logic that is now replicated across providers.

PR #7532, opened 2019-10-15
fix(monitored deploy): fix the rollback config to match what orca expects

marchello2000:
Note: `Manual` type is not yet implemented in orca, hence it's not used in deck, other then the enum definition

### DIFF
--- a/app/scripts/modules/core/src/deploymentStrategy/strategies/monitored/AdditionalFields.tsx
+++ b/app/scripts/modules/core/src/deploymentStrategy/strategies/monitored/AdditionalFields.tsx
@@ -13,8 +13,9 @@ import {
 
 export interface IMonitoredDeployCommand extends IServerGroupCommand {
   delayBeforeScaleDownSec: string;
-  rollback: {
-    onFailure: boolean;
+  failureActions: {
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/deploymentStrategy/strategies/monitored/AdditionalFields.tsx
```diff
@@ -119,6 +136,17 @@ export class AdditionalFields extends React.Component<
...
         </div>
+        <div className="col-md-12 checkbox" style={{ marginTop: 0 }}>
+          <label>
+            <input
+              type="checkbox"
+              checked={destroyFailedAsg}
+              disabled={!rollbackOnFailure}
+              onChange={this.destroyFailedAsgChange}
+            />
+            Destroy the failed server group after rollback <HelpField id="strategy.monitored.destroyFailedAsg" />
```
alanmquach: Didn't you want to grey this out?

marchello2000: yes... but it was too hard :)

PR #7533, opened 2019-10-16
Functions: List and Search functionality

sidmuls:
This PR provides following features:
1. Listing functions
2. Searching functions in the list

Note: core/LoadBalancer and amazon/loadBalancer were used as reference module components for these changes.

### DIFF
--- a/app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
+++ b/app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
@@ -1,0 +1,29 @@
+import { IFunction } from '@spinnaker/core';
+
+export interface IAmazonFunction extends IFunction {
+  credentials?: string;
+  role?: string;
+  runtime: string;
+  s3bucket: string;
+  s3key: string;
+  handler: string;
+  functionName: string;
+  publish: boolean;
+  description: string;
+  tags: [{}];
+  memorySize: number;
+  timeout: number;
+  envVariables: {};
+  tracingConfig: {
+    mode: string;
+  };
+  deadLetterConfig: {
+    targetArn: string;
+  };
+  KMSKeyArn: string;
+  vpcConfig: {
+    securityGroupIds: [];
+    subnetIds: [];
+    vpcId: string;
+  };
+}
--- a/app/scripts/modules/core/src/function/Function.tsx
+++ b/app/scripts/modules/core/src/function/Function.tsx
@@ -1,0 +1,51 @@
+import * as React from 'react';
+import { UISref, UISrefActive } from '@uirouter/react';
+
+import { Application } from 'core/application/application.model';
+import { IFunction } from 'core/domain';
+
+import { EntityNotifications } from 'core/entityTag/notifications/EntityNotifications';
+
+export interface IFunctionProps {
+  application: Application;
+  functionDef: IFunction;
+}
+
+export class Function extends React.Component<IFunctionProps> {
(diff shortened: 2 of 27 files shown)


app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
```diff
@@ -0,0 +1,29 @@
...
+  credentials?: string;
+  role?: string;
+  runtime: string;
+  s3bucket: string;
+  s3key: string;
+  handler: string;
+  functionName: string;
+  publish: boolean;
+  description: string;
+  tags: [{}];
```
caseyhebebrand: Is is possible to use a more explicit type or generics for the tags?

app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
```diff
@@ -0,0 +1,29 @@
...
+  memorySize: number;
+  timeout: number;
+  envVariables: {};
```
caseyhebebrand: If `envVariables` will not be empty it would be stronger to make the type `{ [key: string]: any }`.

app/scripts/modules/core/src/function/Function.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { IFunction } from 'core/domain';
+
+import { EntityNotifications } from 'core/entityTag/notifications/EntityNotifications';
+
+export interface IFunctionProps {
+  application: Application;
+  functionDef: IFunction;
+}
+
+export class Function extends React.Component<IFunctionProps> {
```
caseyhebebrand: Can we make this a functional component since it is stateless and presentational?

app/scripts/modules/core/src/function/FunctionDetails.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { Overridable, IOverridableProps } from 'core/overrideRegistry';
+
+export interface IFunctionDetailsProps extends IOverridableProps {}
+
+@Overridable('function.details')
+export class FunctionDetails extends React.Component<IFunctionDetailsProps> {
```
caseyhebebrand: We could also make this a functional component since it is presentational one-liner

app/scripts/modules/core/src/function/Functions.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export interface IFunctionsState {
+  initialized: boolean;
+  groups: IFunctionGroup[];
+  tags: IFilterTag[];
+}
+
+export class Functions extends React.Component<IFunctionsProps, IFunctionsState> {
+  private groupsUpdatedListener: Subscription;
+  private functionsRefreshUnsubscribe: () => any;
```
caseyhebebrand: It seems like this returns `void` rather than `any`, but maybe I am interpreting the function incorrectly

app/scripts/modules/core/src/function/Functions.tsx
```diff
@@ -0,0 +1,119 @@
...
+    FunctionState.filterService.clearFilters();
+    this.updateFunctionGroups();
+  };
+
+  private tagCleared = (): void => {
+    this.updateFunctionGroups();
+  };
+
+  public render(): React.ReactElement<Functions> {
+    const groupings = this.state.initialized ? (
```
caseyhebebrand: `groupings` should be its own component, especially since it has already been abstracted away from the rest of the presentational layer.

PR #7536, opened 2019-10-16
feat(provider/aws): Functions (listing and searching)

sidmuls:
Issue: https://github.com/spinnaker/spinnaker/issues/4921Issue
Summary:
Add functions support to Spinnaker UI. This is a net functionality that we've been working on for almost a year to this point. The initial backend work was done like three months back.
From then we've worked on creating the UI components with guidance from the Netflix UX team.

Cloud Provider(s): provider/AWS
Environment:
This worked on user's local environment and also through deploying docker images to kubernetes cluster.

Feature Area: @spinnaker/ui-ux-team
Description:
CRUD functionality for managing lambda functions, With this new we must be able to create, update and delete lambda lambda functions through Spinnaker UI
We must also be able to attach target groups to lambda functions using Spinnaker UI.
Additional Details:
Clouddriver backend details: https://github.com/spinnaker/clouddriver/tree/master/clouddriver-lambda


app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
```diff
@@ -0,0 +1,29 @@
...
+  credentials?: string;
+  role?: string;
+  runtime: string;
+  s3bucket: string;
+  s3key: string;
+  handler: string;
+  functionName: string;
+  publish: boolean;
+  description: string;
+  tags: [{}];
```
caseyhebebrand: Is it possible to use a more explicit type or generics for the tags?

app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
```diff
@@ -0,0 +1,29 @@
...
+  memorySize: number;
+  timeout: number;
+  envVariables: {};
```
caseyhebebrand: If `envVariables` will not be empty it would be stronger to make the type `{ [key: string]: any }`.

app/scripts/modules/core/src/function/Function.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { IFunction } from 'core/domain';
+
+import { EntityNotifications } from 'core/entityTag/notifications/EntityNotifications';
+
+export interface IFunctionProps {
+  application: Application;
+  functionDef: IFunction;
+}
+
+export class Function extends React.Component<IFunctionProps> {
```
caseyhebebrand: Can we make this a functional component since it is stateless and presentational only?

app/scripts/modules/core/src/function/FunctionDetails.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import { Overridable, IOverridableProps } from 'core/overrideRegistry';
+
+export interface IFunctionDetailsProps extends IOverridableProps {}
+
+@Overridable('function.details')
+export class FunctionDetails extends React.Component<IFunctionDetailsProps> {
```
caseyhebebrand: We could also make this functional since it is presentational one-liner

app/scripts/modules/core/src/function/Functions.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export interface IFunctionsState {
+  initialized: boolean;
+  groups: IFunctionGroup[];
+  tags: IFilterTag[];
+}
+
+export class Functions extends React.Component<IFunctionsProps, IFunctionsState> {
+  private groupsUpdatedListener: Subscription;
+  private functionsRefreshUnsubscribe: () => any;
```
caseyhebebrand: It seems like this returns void rather than any, but maybe I am interpreting the function incorrectly

app/scripts/modules/core/src/function/Functions.tsx
```diff
@@ -0,0 +1,119 @@
...
+    FunctionState.filterService.clearFilters();
+    this.updateFunctionGroups();
+  };
+
+  private tagCleared = (): void => {
+    this.updateFunctionGroups();
+  };
+
+  public render(): React.ReactElement<Functions> {
+    const groupings = this.state.initialized ? (
```
caseyhebebrand: `Groupings` should be its own component, especially since it has already been abstracted away from the rest of the presentational layer.

app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
sidmuls: it can be empty if no env vars are set on the function.

app/scripts/modules/core/src/function/Functions.tsx
sidmuls: the definition of that function maps to 
```
this.functionsRefreshUnsubscribe = app
      .getDataSource('functions')
      .onRefresh(null, () => this.updateFunctionGroups());
```
the `onRefresh` returns `any`

app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
caseyhebebrand: It looks like `{}` is also equiavalent to `{ [key: string]: any }`, so `{}` may be the best option unless there is a constant, explicit type when there is a tag present

app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
caseyhebebrand: got it, then feel free to leave it as is.

app/scripts/modules/core/src/function/FunctionDetails.tsx
alanmquach: Unfortunately, it looks like our `@Overridable` annotation can only decorate class components.

https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/overrideRegistry/Overridable.tsx#L36-L40

At least in its current form. 😞

caseyhebebrand: Thanks for pointing this out. A `PureComponent` could be a viable alternative in the current form, I would have to look at the differences between the two interfaces in react.

app/scripts/modules/core/src/function/FunctionGroupings.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+export interface IFunctionGroupingsState {
+  groups: IFunctionGroup[];
+}
+export class FunctionGroupings extends React.Component<IFunctionGroupingsProps, IFunctionGroupingsState> {
+  constructor(props: IFunctionGroupingsProps) {
+    super(props);
+
+    const { groups } = props;
+    this.state = {
```
caseyhebebrand: You can just directly use the props instead of assigning `groups` to the state. Especially since you are never mutating the groups within the component. It will re-render anyway if the groups in props are updated.

sidmuls: Pushed the change :)

caseyhebebrand:
LGTM, thanks for addressing the comments

app/scripts/modules/core/src/function/function.states.ts
```diff
@@ -0,0 +1,78 @@
...
+import { FunctionFilters } from './filter/FunctionFilters';
+
+export const FUNCTION_STATES = 'spinnaker.core.functions.states';
+module(FUNCTION_STATES, [APPLICATION_STATE_PROVIDER]).config([
+  'applicationStateProvider',
+  'stateConfigProvider',
+  (applicationStateProvider: ApplicationStateProvider, stateConfigProvider: StateConfigProvider) => {
+    const functionDetails: INestedState = {
+      name: 'functionDetails',
+      url: '/functionDetails/:provider/:account/:region/:vpcId/:name',
```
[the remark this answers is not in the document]
sidmuls: we can. removed it in my next push. Thank you for pointing this out :)

app/scripts/modules/core/src/function/function.read.service.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export class FunctionReader {
+  public static $inject = ['$q', 'functionTransformer'];
+  public constructor(private $q: IQService, private functionTransformer: any) {}
+
+  public loadFunctions(applicationName: string): IPromise<IFunctionSourceData[]> {
+    return API.one('applications', applicationName)
+      .all('functions')
+      .withParams({ region: 'us-west-2' })
```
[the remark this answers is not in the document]
sidmuls: removed in next push :)

app/scripts/modules/core/src/function/filter/FunctionFilterService.ts
```diff
@@ -0,0 +1,150 @@
...
+  public filterFunctionsForDisplay(functions: IFunction[]): IFunction[] {
+    return chain(functions)
+      .filter(fn => this.checkSearchTextFilter(fn))
+      .filter(fn => FilterModelService.checkAccountFilters(FunctionState.filterModel.asFilterModel)(fn))
+      .filter(fn => FilterModelService.checkRegionFilters(FunctionState.filterModel.asFilterModel)(fn))
+      .filter(fn => FilterModelService.checkProviderFilters(FunctionState.filterModel.asFilterModel)(fn))
+      .value();
+  }
+
+  private diffSubgroups(oldGroups: IFunctionGroup[], newGroups: IFunctionGroup[]): void {
```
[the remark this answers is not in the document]
sidmuls: added in next push.

app/scripts/modules/core/src/function/filter/FunctionFilterService.ts
```diff
@@ -0,0 +1,150 @@
...
+      if (!this.lastApplication) {
+        return null;
+      }
+    }
+
+    const groups: IFunctionGroup[] = [];
+    const functions = this.filterFunctionsForDisplay(application.functions.data);
+    const grouped = groupBy(functions, 'account');
+
+    forOwn(grouped, (group, account) => {
```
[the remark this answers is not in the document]
sidmuls: added in the next push.

app/scripts/modules/core/src/function/filter/FunctionFilterService.ts
```diff
@@ -0,0 +1,150 @@
...
+      }
+    });
+  }
+
+  public sortGroupsByHeading(groups: IFunctionGroup[]): void {
+    this.diffSubgroups(FunctionState.filterModel.asFilterModel.groups, groups);
+
+    // sort groups in place so Angular doesn't try to update the world
+    FunctionState.filterModel.asFilterModel.groups.sort((a, b) => {
+      if (a.heading < b.heading) {
```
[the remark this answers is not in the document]
sidmuls: done!

app/scripts/modules/core/src/function/filter/FunctionFilterService.ts
```diff
@@ -0,0 +1,150 @@
...
+import { FunctionState } from 'core/state';
+
+export class FunctionFilterService {
+  public groupsUpdatedStream: Subject<IFunctionGroup[]> = new Subject<IFunctionGroup[]>();
+
+  private lastApplication: Application;
+
+  constructor() {}
+
+  private addSearchFields(functionDef: IFunction): void {
```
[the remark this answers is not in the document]
sidmuls: I thought this fits here because it is specific to filtering.
I used loadbalancer module as reference.

app/scripts/modules/core/src/function/filter/FunctionFilterModel.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface IFunctionFilterModel extends IFilterModel {
+  groups: IFunctionGroup[];
+}
+
+export class FunctionFilterModel {
+  public asFilterModel: IFunctionFilterModel;
+
+  constructor() {
+    this.asFilterModel = FilterModelService.configureFilterModel(this as any, filterModelConfig);
+    FilterModelService.registerRouterHooks(this.asFilterModel, '**.application.insight.loadBalancers.**');
```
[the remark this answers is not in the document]
sidmuls: thank you for pointing this out :)
fixed in the next push.

app/scripts/modules/core/src/domain/IFunction.ts
```diff
@@ -0,0 +1,29 @@
...
+  type?: string;
+}
+
+export interface IFunction extends ITaggedEntity {
+  account?: string;
+  cloudProvider?: string;
+  description?: string;
+  name?: string;
+  functionName?: string;
+  provider?: string;
```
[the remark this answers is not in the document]
sidmuls: some providers may add the 'cloud provider' field with the key 'provider'
it was to accomodate for that discrepancy from the backend.
but we can probably remove it and enforce the filed key to be 'cloudProvider'
I have removed it in the next push.

app/scripts/modules/core/src/function/function.read.service.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  public listFunctions(cloudProvider: string): IPromise<IFunctionByAccount[]> {
+    return API.all('functions')
+      .withParams({ provider: cloudProvider })
+      .getList();
+  }
+
+  private normalizeFunction(functionDef: IFunctionSourceData): IPromise<IFunction> {
+    return this.functionTransformer.normalizeFunction(functionDef).then((fn: IFunction) => {
+      fn.cloudProvider = fn.cloudProvider || 'aws';
```
christopherthielen: Does the backend ever return functions without a cloudProvider field?  This seems a little odd since I assume the backend always returns 'aws' for Lambdas?

app/scripts/modules/core/src/function/function.states.ts
```diff
@@ -0,0 +1,72 @@
...
+      resolve: {
+        accountId: ['$stateParams', ($stateParams: StateParams) => $stateParams.account],
+        functionObj: [
+          '$stateParams',
+          ($stateParams: StateParams) => {
+            return {
+              name: $stateParams.name,
+              accountId: $stateParams.account,
+              region: $stateParams.region,
+              vpcId: $stateParams.vpcId,
```
christopherthielen: remove this line

PR #7540, opened 2019-10-18
feat(core/presentation): Add useDebouncedValue react hook

christopherthielen:
also renamed hooks files to '.hook.ts' and exported all from an index.ts

### DIFF
--- a/app/scripts/modules/core/src/presentation/hooks/useDebouncedValue.hook.ts
+++ b/app/scripts/modules/core/src/presentation/hooks/useDebouncedValue.hook.ts
@@ -1,0 +1,19 @@
+import { useEffect, useState } from 'react';
+
+/**
+ * A react hook which returns a debounced value
+ *
+ * @param value the immediate value
+ * @param debounceMs: the debounce time, in ms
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/presentation/hooks/useDebouncedValue.hook.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param value the immediate value
+ * @param debounceMs: the debounce time, in ms
+ */
+export function useDebouncedValue<T>(value: T, debounceMs: number): [T, boolean] {
+  const [debouncedValue, setDebouncedValue] = useState(value);
+  const [isDebouncing, setIsDebouncing] = useState(false);
+  useEffect(() => setIsDebouncing(value !== debouncedValue));
+  useEffect(() => {
+    const id = setTimeout(() => setDebouncedValue(value), debounceMs);
+    return () => clearTimeout(id);
```
louisjimenez: nice use of the cleanup func to remove the timeout, :+1:

PR #7541, opened 2019-10-18
fix: credentails typo

23doors:
Not sure if credentials/credentails field is actually used, seems a little obsolete as it is 1-1 with account and only present in job, not in manifest deployment. But still, this is definitely a typo here.


maggieneterval:
@ethanfrogers I just did a quick grep of Orca for `credentails` and couldn't find any uses, perhaps safe to remove?

ethanfrogers:
Thanks @erikmunson, this fell off my radar. Agree with your summary 👍 .

PR #7542, opened 2019-10-18
feat(provider/aws): Show load balancer warning based on settings

caseyhebebrand:
Enable a static AWS load balancer warning, upon load balancer creation. This is enabled by adding a warning in the settings file. 

Example usage: 
```
// settings.js
aws: {
  createLoadBalancerWarnings: {
     application: 'This is a warning message'
  }
}
```
Screen Shot 2019-10-18 at 1 12 35 PM


app/scripts/modules/amazon/src/loadBalancer/configure/AmazonLoadBalancerChoiceModal.tsx
```diff
@@ -87,6 +88,23 @@ export class AmazonLoadBalancerChoiceModal extends React.Component<
     return false;
   }
 
+  // This assumes that () and [] only exist in the link markup, and satisfies the use cases of this component.
+  private parseMessageForLinks(message: string) {
```
caseyhebebrand: Open to using a markdown parser, but most use cases in this component will only be simple strings. The first implementation happens to have a link with it.

christopherthielen: I would definitely prefer to use `<Markdown message={message} />` here

app/scripts/modules/amazon/src/loadBalancer/configure/AmazonLoadBalancerChoiceModal.tsx
```diff
@@ -133,6 +155,14 @@ export class AmazonLoadBalancerChoiceModal extends React.Component<
                   </div>
                 ))}
             </>
+            {Boolean(loadBalancerWarning) && (
```
christopherthielen: but we've typically used double bang for coercing to a boolean type in the rest of the codebase, i.e., `!!loadBalancerWarning`

this isn't actually a preference of ours, but it would be nice to be consistent code-style-wise

app/scripts/modules/amazon/src/loadBalancer/configure/LoadBalancerTypes.ts
```diff
@@ -11,7 +11,7 @@ export interface ICloseableLoadBalancerModal extends React.ComponentClass<ILoadB
 }
 
 export interface IAmazonLoadBalancerConfig {
-  type: string;
+  type: 'network' | 'application' | 'classic';
```
christopherthielen: 💯

app/scripts/modules/amazon/src/loadBalancer/configure/AmazonLoadBalancerChoiceModal.tsx
caseyhebebrand: Perfect, didn't see that this existed. I will swap it out.

app/scripts/modules/amazon/src/loadBalancer/configure/AmazonLoadBalancerChoiceModal.tsx
caseyhebebrand: Sure I can do that. I have no strong preference either, just a habit of using `Boolean(...)` from previous Flow rules.

erikmunson: There are days where I really miss Flow's magical `Boolean` inference powers...

PR #7543, opened 2019-10-18
feat(core/presentation): add revalidate api to IFormInputValidation prop

christopherthielen:
In the system of `FormField`, `FormikFormField`, validation, and `Input` components:

`FormField` and `FormikFormField` provide a `validation` prop to Input components (such as `TextInput`) with information about the current "touched" status, validation message, category, etc.  This enables the input to do things like It also provides an API to register _internal validators_ specific to the input type.  For example, a `SpelInput` may add spel validator.

This PR augments this interface  with a `revalidate()` prop which allows the input itself to trigger revalidation based on some external event (such as async data being loaded for autocompletion).  The parent component (`FormField`/`FormikFormField`) is then responsible for retriggering its validation logic.

example:

```js



function FancyInput(props) {
  const { validation, ...rest } = props.validation;

  // Use validation category to turn the input text RED when invalid
  const color = validation.category === 'ERROR' ? 'red' : 'black';

  // Fetching some async data in the input
  const fetchRepo = useData(() => asyncCheckRepoExists(), null, [props.value]);

  // Defining an internal validator
  const internalValidator = useCallback(() => 
    fetchRepo.status === 'RESOLVED' && fetchRepo.result === false && errorMessage('Repo doesn't exist')
  , []);

  // Registering an internal validator
  useEffect(() => {
    validation.addValidator(internalValidator);
    return () => validation.removeValidator(internalValidator);
  }, []);

  // Here's the new thing: revalidate whenever the async thing changes
  useEffect(() => validation.revalidate(), [fetchRepo.status, fetchRepo.result])
  
  return <input style={{ color }} {...rest} />
}
```

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
@@ -47,7 +47,6 @@ export class BakeManifestStageForm extends React.Component<
                 this.props.formik.setFieldValue('templateRenderer', o.target.value);
               }}
               value={stage.templateRenderer}
-              validation={{}}
               stringOptions={this.templateRenderers()}
             />
           </StageConfigField>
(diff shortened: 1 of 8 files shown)


christopherthielen:
I'll fix CI on monday

christopherthielen:
Tagging y’all because there’s a input example here thought might be useful to share. 

@caseyhebebrand

caseyhebebrand:
> Tagging y’all because there’s a input example here thought might be useful to share.
> 

Thanks for sharing!

PR #7547, opened 2019-10-21
fix(awslb): Preventing edits against orphaned load balancers

alanmquach:
When using load balancers from a different application and that other application does not exist, the previous behavior of "just using the current application" leads to unexpected behavior.

So we should prevent edits instead, if the application is not configured.

foobarbaz_·_Load_Balancers_·_Load_Balancer_Details__foobarbaz-alb__test_us-east-1_


caseyhebebrand:
LGTM

PR #7550, opened 2019-10-23
chore(settings): remove defaultCategory from settings.js

christopherthielen:
It seems this setting is no longer used by any code

Please review and let me know if this looks safe to remove.


settings.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+// Use environment variables when developing locqlly via 'yarn start', i.e.:
```
erikmunson:
```suggestion
// Use environment variables when developing locally via 'yarn start', i.e.:
```

PR #7559, opened 2019-10-24
feat(dataSources): widen + parameterize types, add 'managed' data source

erikmunson:
While working on Managed Delivery, we wanted to stop using entity tags for indicating Keel-managed things in the UI and add a new data source that could hold a 'summary' of Keel's view of an app — managed resources, some additional metadata that will be tacked on later like the status of various policies/vetos, and surely things we don't know we need yet. I quickly discovered that the existing data source implementation is highly coupled to the idea that _everything_ a data source holds is typed as `any[]`, and we have assumptions baked into the code that further narrow the typing down to "an array of objects that usually have certain properties available".

Thinking about it, I came to a couple conclusions:
- **It doesn't make sense anymore to require that data sources hold arrays of objects**, because the real value in data sources is providing a unified orchestrator for polling and data aggregation/joins/transformation.
- **We should do _something_ to unlock the ability to gradually add more static typing to data sources over time**, without boiling the ocean. As of right now the types for data sources are hard-coded into the class (`any[]`), and it's basically impossible to assert anything about types beyond that as a result.

So here we are, with this big PR. What I've chosen to try out/propose is conceptually not too complicated, but in practice touches a lot of code across providers and should be reviewed with a skeptical eye. I'm proposing we:
- **Widen out the _possible_ types a data source could wrap from `any[]` to `any`.** This means that if you get handed a data source you know nothing about, you'll need to confirm its type via runtime checks before doing anything to it like calling `.length` or built in array methods.
- **Add a single, required type parameter to the data source and its config object**, which represents the type of data it actually wraps. This parameter is used to express the relationship between what you'll get inside the `data` field, what you should return inside the `onLoad` callback, etc., so if you say that a data source wraps a particular type of data you'll get a small amount of oversight from TS.
- **Require that code which registers a new data source provides an explicit default value of the same type as the declared data type**, rather than implicitly defaulting it to `[]`. This one means that there are tons of different places defaults needed adding, but on the flip side it was easier to refactor because today the correct default is universally `[]`. While potentially controversial, I feel this is a good choice because when a data source can wrap any type of data, it makes some sense for the owner of the data source to decide what a reasonable default looks like and declare that for other consumers to see and understand.

If this goes well, I'd like to attempt to make it possible to get type information automatically when you dereference a data source off of an application model (e.g. `application.loadBalancers.data`). I ran into some trouble doing that this time, but setting up a real type system for data sources means we can leverage it later to accomplish more fun stuff.

I also decided to go ahead and add the new `managedResources` data source in the same PR. While this makes the size slightly larger, I felt it was very important to _actually use_ a data source with a new and different type as I was refactoring to make sure things really worked the way I wanted them to. The first commit contains all the data source refactoring and a very small amount of new data source code for Managed Delivery, which I think serves as a nice usage example of this new type scheme.

**Please give feedback if you have it**, this is potentially disruptive and the size/scope means things will probably break in unexpected places. I want to make sure folks like the direction before pulling the trigger.

### DIFF
--- a/app/scripts/modules/core/src/application/application.model.ts
+++ b/app/scripts/modules/core/src/application/application.model.ts
@@ -93,7 +93,11 @@ export class Application {
 
   private dataLoader: Subscription;
 
-  constructor(private applicationName: string, private scheduler: any, dataSourceConfigs: IDataSourceConfig[]) {
+  constructor(
+    private applicationName: string,
+    private scheduler: any,
+    dataSourceConfigs: Array<IDataSourceConfig<any>>,
+  ) {
     dataSourceConfigs.forEach(config => this.addDataSource(config));
     this.status$ = Observable.combineLatest(this.dataSources.map(ds => ds.status$)).map(statuses =>
       this.getDerivedApplicationStatus(statuses),
@@ -118,7 +122,7 @@ export class Application {
     };
   }
 
-  private addDataSource(config: IDataSourceConfig) {
+  private addDataSource<T>(config: IDataSourceConfig<T>) {
     const dataSource = new ApplicationDataSource(config, this);
     this.dataSources.push(dataSource);
     this[config.key] = dataSource;
@@ -222,21 +226,27 @@ export class Application {
   private setApplicationAccounts(): void {
     let accounts = this.accounts.concat(this.attributes.accounts || []);
     this.dataSources
-      .filter(ds => ds.credentialsField !== undefined)
-      .forEach(ds => (accounts = accounts.concat(ds.data.map(d => d[ds.credentialsField]))));
+      .filter(ds => Array.isArray(ds.data) && ds.credentialsField !== undefined)
+      .forEach(
+        (ds: ApplicationDataSource<any[]>) => (accounts = accounts.concat(ds.data.map(d => d[ds.credentialsField]))),
+      );
 
     this.accounts = uniq(accounts);
   }
 
   private extractProviderDefault(field: string): Map<string, string> {
     const results = new Map<string, string>();
-    const sources = this.dataSources.filter(d => d[field] !== undefined);
+    const sources: Array<ApplicationDataSource<any[]>> = this.dataSources.filter(
+      ds => ds[field] !== undefined && Array.isArray(ds.data) && ds.providerField !== undefined,
+    );
     const providers = sources.map(ds => ds.data.map(d => d[ds.providerField])).filter(p => p.length > 0);
     let allProviders: any; // typescript made me do it this way
     allProviders = union<string[]>(...providers);
     allProviders.forEach((provider: string) => {
       const vals = sources
-        .map(ds => map(ds.data.filter(d => d[ds.providerField] === provider), ds[field]))
+        .map(ds =>
+          map(ds.data.filter((d: any) => typeof d === 'object' && d[ds.providerField] === provider), ds[field]),
+        )
         .filter(v => v.length > 0);
       const allValues = union(...vals);
       if (allValues.length === 1) {
--- a/app/scripts/modules/core/src/managed/managed.dataSource.ts
+++ b/app/scripts/modules/core/src/managed/managed.dataSource.ts
@@ -1,0 +1,31 @@
+import { module, IQService } from 'angular';
+
+import { SETTINGS } from 'core/config/settings';
+import { ApplicationDataSourceRegistry } from 'core/application/service/ApplicationDataSourceRegistry';
+import { Application } from 'core/application';
+import { ManagedReader, IManagedApplicationSummary } from './ManagedReader';
+
+export const MANAGED_RESOURCES_DATA_SOURCE = 'spinnaker.core.managed.dataSource';
+module(MANAGED_RESOURCES_DATA_SOURCE, []).run([
+  '$q',
+  ($q: IQService) => {
+    if (!SETTINGS.feature.managedResources) {
+      return;
+    }
+    const loadManagedResources = (application: Application) => {
+      return ManagedReader.getApplicationSummary(application.name);
+    };
+
+    const addManagedResources = (_application: Application, data: IManagedApplicationSummary) => {
+      return $q.when(data);
+    };
+
+    ApplicationDataSourceRegistry.registerDataSource({
+      key: 'managedResources',
+      visible: false,
+      loader: loadManagedResources,
+      onLoad: addManagedResources,
+      defaultData: { hasManagedResources: false, resources: [] },
+    });
+  },
+]);
--- a/app/scripts/modules/core/src/managed/managed.dataSource.ts
+++ b/app/scripts/modules/core/src/managed/managed.dataSource.ts
@@ -1,9 +1,15 @@
 import { module, IQService } from 'angular';
 
+import { noop } from 'core/utils';
 import { SETTINGS } from 'core/config/settings';
 import { ApplicationDataSourceRegistry } from 'core/application/service/ApplicationDataSourceRegistry';
 import { Application } from 'core/application';
 import { ManagedReader, IManagedApplicationSummary } from './ManagedReader';
+import {
+  addManagedResourceMetadataToServerGroups,
+  addManagedResourceMetadataToLoadBalancers,
+  addManagedResourceMetadataToSecurityGroups,
+} from './managedResourceDecorators';
 
 export const MANAGED_RESOURCES_DATA_SOURCE = 'spinnaker.core.managed.dataSource';
 module(MANAGED_RESOURCES_DATA_SOURCE, []).run([
@@ -20,11 +26,18 @@ module(MANAGED_RESOURCES_DATA_SOURCE, []).run([
       return $q.when(data);
     };
 
+    const addManagedMetadataToResources = (application: Application) => {
+      application.serverGroups.ready().then(() => addManagedResourceMetadataToServerGroups(application), noop);
+      application.loadBalancers.ready().then(() => addManagedResourceMetadataToLoadBalancers(application), noop);
+      application.securityGroups.ready().then(() => addManagedResourceMetadataToSecurityGroups(application), noop);
+    };
+
     ApplicationDataSourceRegistry.registerDataSource({
       key: 'managedResources',
       visible: false,
       loader: loadManagedResources,
       onLoad: addManagedResources,
+      afterLoad: addManagedMetadataToResources,
       defaultData: { hasManagedResources: false, resources: [] },
     });
   },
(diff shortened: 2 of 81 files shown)


app/scripts/modules/core/src/securityGroup/securityGroupReader.service.ts
```diff
@@ -216,6 +216,7 @@ export class SecurityGroupReader {
     const nameParts: IComponentName = NameUtils.parseSecurityGroupName(securityGroup.name);
     securityGroup.stack = nameParts.stack;
     securityGroup.detail = nameParts.freeFormDetails;
+    securityGroup.moniker = NameUtils.getMoniker(nameParts.application, nameParts.stack, nameParts.freeFormDetails);
```
erikmunson: a fun bonus: now `ISecurityGroup` objects will always have a `moniker` on them, which is pretty nifty. they already have `stack` and `detail` fields that are empty strings when not present, which is kind of annoying and doesn't really look like anything else. I like the `moniker` object and it's contract around empty values being omitted from the object entirely.

app/scripts/modules/core/src/managed/managed.dataSource.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    const addManagedResources = (_application: Application, data: IManagedApplicationSummary) => {
+      return $q.when(data);
+    };
+
+    const addManagedMetadataToResources = (application: Application) => {
+      application.serverGroups.ready().then(() => addManagedResourceMetadataToServerGroups(application), noop);
+      application.loadBalancers.ready().then(() => addManagedResourceMetadataToLoadBalancers(application), noop);
+      application.securityGroups.ready().then(() => addManagedResourceMetadataToSecurityGroups(application), noop);
+    };
```
erikmunson: I wanted to do something fancy with observables here rather than stealing the way entity tags do it right now (where these functions are also called in the afterLoad handlers of every relevant data source), but it's weird to do today because you have to get an application model in a callback, then bind an observable chain to it, then cleanup that observable chain the next time a callback comes in because you don't have visibility into the application model lifecycle. 

I'm going to follow up with a proposed way of baking cross-data source joins into the data source API itself, since that has awareness of application model lifecycle outside the callbacks exposed here.

app/scripts/modules/core/src/managed/managedResourceDecorators.ts
```diff
@@ -0,0 +1,77 @@
...
+    serverGroup.managedResourceSummary = matchingResource;
+  });
+};
+
+export const addManagedResourceMetadataToLoadBalancers = (application: Application) => {
+  if (!SETTINGS.feature.managedResources) {
+    return;
+  }
+
+  const loadBalancerResources = getResourcesOfKind(application, ['classic-load-balancer', 'application-load-balancer']);
```
erikmunson: @emjburns @asher @robfletcher @gal-yardeni @luispollo something I started noticing while writing this — the whole thing of having very specific/discrete resource `kind`s is going to gradually build up a bit of an impedance mismatch when it comes to the generic, provider-agnostic modeling we do inside the rest of the API/UI. Here, for example, we have a bunch of load balancers in a list, and I say 'hey application, gimme your load balancers'. But to get the right list of resources that represent all the load balancers which are managed by keel, I need to encode a brittle whitelist of specific resource kinds that I as the programmer know will give me all the currently available load balancer kinds.

Right now LBs are the exception here, because we only have `cluster` and `security-group` otherwise. I'm wondering if we should strive for kinds that are more generic like those, and make load balancer types defined via the spec contents (e.g. `loadBalancerType: 'application'`). There's also a valid (I think) path where we say "no, this is the new world and things are different", but that causes a lot of pretty big and confusing questions to pop up about how we handle cases like this as they become more prevalent. Not proposing we take action on any of this immediately, just raising it as food for thought.

erikmunson:
Another point worth noting: because this adds a required field to the data source config object along with a new type parameter, anybody using data sources outside of the `deck` repo will require code changes with this change. At minimum, I know that `deck-kayenta` will need its data sources updated in OSS-land, and we have some internal stuff at Netflix that may need updating too.

erikmunson:
@maggieneterval funny enough, I didn't know until talking to @christopherthielen that the preferred access method is sort of secretly `application.dataSourceName.data`. As you saw, I somewhat casually moved to that notation where convenient but didn't go out of my way to move things if it was annoying or made the diff unnecessarily larger.

The current thinking is that by accessing via an object property, we could _theoretically_ infer types (e.g. we'd know that you were getting a `ApplicationDataSource<ILoadBalancer[]>` because that's the declared type of the `application.loadBalancer` field). I was amused to find that the _opposite_ recommendation is left in an old Chris Berry comment on the application model, for the same reason (typing). I suspect the typing he was talking about was being able to infer that we were going to get an `ApplicationDataSource` rather than defaulting to `any`. Either way, right now the two are very similar but I'd like to explore being able to have a data source also augment the Application model when it registers itself so we can get to the point of actually knowing what a data source's type is when you ask for it from an application.

Re: merging this on Monday after the 1.17 window closes, I _absolutely_ want to wait until then. This is going to require choreographing with deck-kayenta at a minimum, and some fun follow up bug fix PRs in all likelihood. Thanks for pointing out the window as I didn't realize we were that close.

app/scripts/modules/core/src/application/application.model.ts
```diff
@@ -79,9 +79,9 @@ export class Application {
    * Which data source is the active state
    * @type {ApplicationDataSource}
    */
-  public activeState: ApplicationDataSource = null;
+  public activeState: ApplicationDataSource<any> = null;
 
-  public activeDataSource$ = new ReplaySubject<ApplicationDataSource>(1);
+  public activeDataSource$ = new ReplaySubject<ApplicationDataSource<any>>(1);
```
christopherthielen: If you give the generic parameter a default value, you don't have to do this `<any>` thing all over the place.  

```
interface ApplicationDataSource<T = any> {```

Comment (erikmunson) [reply app/scripts/modules/core/src/application/application.model.ts]: yeah... I know and good call. I had originally left out a default as a convenient way of having the TS compiler give me a hit-list of spots the type was used so I could see whether any of them were capable of getting narrowed down beyond `any` (a few were). Then, Thursday Erik had some sort of half-baked idealogical thing going on where he thought forcing people to specify a type might encourage more specific typing in the future rather than throwing `any` around because it's the default, which led to keeping it in this state.

Monday Erik thinks that was maybe a little too ideological, and I'm going to just default it.

File: app/scripts/modules/core/src/managed/managedResourceDecorators.ts (hunk 3 above)
Comment (luispollo) [reply app/scripts/modules/core/src/managed/managedResourceDecorators.ts]: I think we should shoot for higher-level abstract "kinds" and add specific details as annotations in the specs. It might be a little trickier to resolve to concrete types under the covers, but to your point I think it's worth the effort because it will not only better align with the UI/API contracts we're exposing, but with the story we're trying to tell with MD. If there's any behavior we want to expose on the UI that is very specific to a concrete type, we could still do that based on the annotations, while keeping the API fairly agnostic.

We kinda talked about that idea here last week as well in the context of Titus vs. other types of container-based deployments: https://github.com/spinnaker/keel/pull/533#discussion_r337232907

Comment (luispollo) [reply app/scripts/modules/core/src/managed/managedResourceDecorators.ts]: P.S. On the other hand, I'm trying to be cognizant to avoid "premature optimization" while we don't have a whole lot of customers to provide actual user feedback, so I'm frequently unsure what feedback to give (or hold off) on changing some of these contracts..

**Comment by erikmunson** [comment]
@christopherthielen I updated this to default the new type parameter to `any`, along with rebasing on top of the new `functions` data source. I did go ahead and split out the meaningful chunks of changes to the best of my ability, so hopefully looking at this again will be a bit easier. That said, the first commit is still large and couples together both the type changes + adding `defaultData` attributes, because contrary to the wrong / accidentally outdated comments I left in the code (sorry, my bad) that field *is required when registering a data source*.

I thought about it a lot, and as you saw from my outdated comments I originally tried to make the `defaultData` field default to an empty array. IMO, going the required route is the right / best way to structure the API going forward if we truly want to think of data sources as 'wrappers for polled data' instead of 'wrappers for polled arrays of objects'. I saw a few possible approaches going into it:

1. `defaultData` is optional, default to array (non-breaking)
2. `defaultData` is optional, default to null (breaking)
3. `defaultData` is required (breaking)

I ended up on #3 because:
- #1 **used to be a sensible default**, but if data sources are actually going to be arbitrary data polling wrappers going forward a sensible default is not possible anymore, the code that registers a data source should decide what a sensible default is
- #2 is equally as breaking as #3, and makes the default experience more dangerous, confusing, and tedious (e.g. you might not realize you need to handle nulls, and if you do you have to do null-guarding everywhere or gate behind a `.ready()` call)
- That left me at #3, which IMO is a one-time breaking change cost (with a predictable and easy to communicate/understand required change) and after that is a good API to have going forward.

In addition to choreographing a `deck-kayenta` release, I'll make sure there's a note in the change log for the 1.18 release to anyone adding their own data sources in custom extensions explaining that they'll need to add `defaultData: []` to anywhere they call `registerDataSource`, along with covering the added ability to use different data types.

Do you agree/disagree with the thinking? Any strong objection to moving forward?

File: app/scripts/modules/core/src/application/application.model.ts [hunk 5]
```diff
@@ -118,7 +122,7 @@ export class Application {
     };
   }
 
-  private addDataSource(config: IDataSourceConfig) {
+  private addDataSource(config: IDataSourceConfig<any>) {
```
caseyhebebrand: If there are a finite number of data sources, my suggestion would be to either create an enum with those possible types to make this function more strict:
```
interface DataSources = IServerGroup | ILoadBalancer | IAnotherDataSource
private addDataSource(config: IDataSourceConfig<DataSources> { ... }
```
Or to introduce generics to this function as well. This one seems trickier based on when the function is invoked.
```
private addDataSource<T>(config: IDataSourceConfig<T>) { ... }
```

erikmunson: You're right that using generics here would be the most accurate typing, and I'll do a quick update to reflect that. Sadly, though, there really isn't any way to type this strictly enough that we get meaningful type info beyond `any` because:

- There is not a finite list of available data sources — some of them are added in provider modules outside core, others are added in repos outside `deck` like `deck-kayenta`, and still more are added in custom builds (like our internal build at Netflix)
- When the application model calls this method, it's doing it by looping over the list of all available data sources, which means the only type we can use to appropriately describe the contents is `any`

The 'big list of arbitrary data polling wrappers' architecture that currently exists doesn't lend itself to very useful static typing, but it does make it possible to coordinate polling across a lot of different endpoints beyond what was originally written into core. Not saying it's the perfect tradeoff today in 2019, but it has helped us do interesting things in the past.

caseyhebebrand: Got it, thanks for the additional context here.

christopherthielen:
> Do you agree/disagree with the thinking? Any strong objection to moving forward?

Go for it

erikmunson:
Thanks @christopherthielen! Now here's hoping that CI passes and I can do a rebase+merge on this before anybody merges to master again 🤣

PR #7561, opened 2019-10-24
fix(triggers): do not allow manual definition of docker image in trigger

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/4929

Disallows manual entry of a Docker image when configuring a trigger.

### DIFF
--- a/app/scripts/modules/docker/src/image/DockerImageAndTagSelector.tsx
+++ b/app/scripts/modules/docker/src/image/DockerImageAndTagSelector.tsx
@@ -32,6 +32,7 @@ export interface IDockerImageAndTagSelectorProps {
   onChange: (changes: IDockerImageAndTagChanges) => void;
   deferInitialization?: boolean;
   showDigest?: boolean;
+  allowManualDefinition?: boolean;
 }
 
 export interface IDockerImageAndTagSelectorState {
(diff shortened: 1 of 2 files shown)


app/scripts/modules/docker/src/image/DockerImageAndTagSelector.tsx
```diff
@@ -447,7 +450,7 @@ export class DockerImageAndTagSelector extends React.Component<
             <span className="field">
               <Select
                 value={defineManually}
-                disabled={imagesLoading}
+                disabled={imagesLoading || !allowManualDefinition}
```
erikmunson: thoughts on just removing this select field entirely when manual entry is disabled?

erikmunson: i guess we'd have to figure out what the visual treatment would look like since it lives in that grey header thing...

maggieneterval: Yeah, removing the select field entirely was my initial thought as well, followed by the realization that it lives in that header thing and that we'd probably want to convert this entire component to use the formik form fields at some point anyway so might not be worth handling the visual treatment now? But if you think it looks really weird with the select field disabled let me know!

erikmunson: I think you did the right thing — we just didn't account for wanting to omit this selection entirely when it was originally designed. You took this journey of discovery a little before me but now we've all ended up at the same place 🤣

PR #7563, opened 2019-10-24
fix(core/pipeline): make UX less bad when a pipeline stage never happened

erikmunson:
Today, when a pipeline stage never starts or otherwise doesn't kick off an execution, the experience is... pretty bad. We not only show you a broken execution link, we also use the _execution name_ to show which pipeline the stage is configured to run, which means you get a blank value there too. This adds an indicator that there is no execution to link out to, and also looks up the actual name of the appropriate pipeline so that we can show it instead of nothing.

 **Before:**
Screen Shot 2019-10-24 at 2 05 04 PM

**After:**
Screen Shot 2019-10-24 at 2 04 30 PM

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/pipeline/PipelineStageExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/pipeline/PipelineStageExecutionDetails.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import { UISref } from '@uirouter/react';
 
 import { IExecutionDetailsSectionProps, ExecutionDetailsSection, StageFailureMessage } from 'core/pipeline';
+import { IPipeline } from 'core/domain';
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/pipeline/PipelineStageExecutionDetails.tsx
```diff
@@ -13,6 +15,13 @@ export function PipelineStageExecutionDetails(props: IExecutionDetailsSectionPro
     current,
   } = props;
 
+  const { result: pipelineConfigs } = useLatestPromise<IPipeline[]>(() => {
+    application.pipelineConfigs.activate();
+    return application.pipelineConfigs.ready();
```
erikmunson: we have to activate and wait for readiness of the data source here, because on standalone views this data source isn't already activated when the page loads.

PR #7564, opened 2019-10-25
feat(rosco): Allow roscoMode per stage/execution

alanmquach:
This allows deck to be configured to compute `roscoMode` on a per stage/execution basis via an optional `roscoSelector` in settings:

```
roscoMode: false,
roscoSelector: (stage) => stage && stage.roscoMode,
snapshots: false,
```

### DIFF
--- a/app/scripts/modules/amazon/src/pipeline/stages/bake/awsBakeStage.js
+++ b/app/scripts/modules/amazon/src/pipeline/stages/bake/awsBakeStage.js
@@ -55,7 +55,9 @@ module.exports = angular
 
       $scope.viewState = {
         loading: true,
-        roscoMode: SETTINGS.feature.roscoMode,
+        roscoMode:
+          SETTINGS.feature.roscoMode ||
+          (typeof SETTINGS.feature.roscoSelector === 'function' && SETTINGS.feature.roscoSelector($scope.stage)),
         minRootVolumeSize: AWSProviderSettings.minRootVolumeSize,
         showVmTypeSelector: true,
       };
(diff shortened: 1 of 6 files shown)


app/scripts/modules/amazon/src/pipeline/stages/bake/awsBakeStage.js
```diff
@@ -85,6 +87,15 @@ module.exports = angular
...
           if (!$scope.stage.baseOs && $scope.baseOsOptions && $scope.baseOsOptions.length) {
             $scope.stage.baseOs = $scope.baseOsOptions[0].id;
+          } else if (
+            $scope.stage.baseOs &&
+            !($scope.baseOsOptions || []).find(baseOs => baseOs.id === $scope.stage.baseOs)
+          ) {
+            $scope.baseOsOptions.push({
+              id: $scope.stage.baseOs,
+              detailedDescription: 'Custom',
+              vmTypes: ['hvm', 'pv'],
```
erikmunson: oh hey this is kind of a cool idea. only downside is it means we lose the ability to validate OS options that we remove because they're deprecated. but maybe that's ok.

alanmquach: Well the current experience is terrible because it just JSEs.

If we wanted to aggressively validate against deprecated versions, I’d imagine we can return a “deprecated” blacklist or flag from the gate endpoint.

But not sure it’s worth that effort.

erikmunson: yeah, i went through basically the same thought process — i think this is good enough until we decide we need some kind of deprecated indicator

mybayern1974:
@alanmquach, the Azure provider breaks since Spinnaker version 1.17 and per our tier 1 investigation, this PR might be the root cause. More detail about Azure experience breaking see issue [5104](https://github.com/spinnaker/spinnaker/issues/5104). Could you please help take a look?
@duftler, @maggieneterval , given Azure is now broken, it's expected there will be a patch version release at the earliest date after the above issue got fixed.
@neil-yechenwei and @grayzu for awareness.

alanmquach:
@mybayern1974 I just took a look and it looks like I accidentally duplicated the string `SETTINGS.feature.roscoMode` (resulting in `SETTINGS.feature.roscoModeSETTINGS.feature.roscoMode`). Sorry about that. I've just created https://github.com/spinnaker/deck/pull/7631 which should fix it.

PR #7567, opened 2019-10-25
Revert "feat(provider/aws): Functions (listing and searching)"

plumpy:
Reverts spinnaker/deck#7536

This breaks things pretty dramatically for me. @maggieneterval reproduced the breakage.

When you click on a server group, the details pane doesn't open. In the console, I see 404 responses to this URL:
https://localhost:8084/applications/plumpstuff/functions

I don't know if this is breaking because there's some feature I don't have enabled or maybe there are unmerged changes to support this that go into gate and/or clouddriver? But it's pretty badly broken, so I'm going to roll it back.


plumpy:
cc @sidmuls

erikmunson:
👍 I logged on to do the same, only to see this PR. As far as I can tell there are a couple reasons:
1. There is no feature flag for functions, and as a result the function data source _always_ adds an item to the navigation menu under Infrastructure and starts trying to poll for data, whether you wanted that capability or not.
2. As plumpy noted, when there are no functions in an app the `/application/<app>/functions` endpoint returns a 404, not an empty array like the other infrastructure endpoints (server groups, lbs, etc.). Data sources _do not_ handle failed responses very well, and as plumpy observed one of the fun things about failed responses is that they can cascade out of a single data source and mess with other things like the ability to show the details pane for other kinds of infra.

I'm wondering if #2 is actually because (per plumpy's description) there is a flag for the functions controller in gate, and this is the behavior when that flag is off? If so, perhaps #2 is ok and expected behavior in this case. Either way, this should be feature flagged per #1, but if this is in fact he behavior of that endpoint when functions are enabled then that should change as well.

maggieneterval:
@erikmunson was about to add basically the exact same comment :smile: 
+1 to adding a feature flag and not registering the functions data source if disabled, similarly to how we handle [entity tags](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/entityTag/entityTags.dataSource.ts#L15).

PR #7569, opened 2019-10-26
fix(core/pipeline): make revision dropdown usable, layout tweaks

erikmunson:
When the pipeline revision history dropdown got moved to react, we started using `StandardFieldLayout` which has a pesky min width on its label. That caused some "fun" usability challenges with the width of the dropdown itself, which this fixes. I've also taken the liberty of improving some other layout nits, as the diff summary was both vertical misaligned and had a bunch of excess padding/margin that made it look a little funky.

**Before:**
Screen Shot 2019-10-25 at 6 34 02 PM
Screen Shot 2019-10-25 at 6 34 11 PM

**After:**
Screen Shot 2019-10-25 at 6 33 05 PM
Screen Shot 2019-10-25 at 6 34 19 PM

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/actions/history/ShowPipelineHistoryModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/actions/history/ShowPipelineHistoryModal.tsx
@@ -119,7 +119,7 @@ export function ShowPipelineHistoryModal(props: IShowHistoryModalProps) {
           {!loading && history.length > 1 && (
             <>
               <div className="history-header row horizontal">
-                <div className="revision-section col-md-4">
+                <div className="col-md-4">
                   <FormField
                     label="Revision"
                     layout={({ label, input }) => (
--- a/app/scripts/modules/core/src/pipeline/config/actions/history/ShowPipelineHistoryModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/actions/history/ShowPipelineHistoryModal.tsx
@@ -122,6 +122,12 @@ export function ShowPipelineHistoryModal(props: IShowHistoryModalProps) {
                 <div className="revision-section col-md-4">
                   <FormField
                     label="Revision"
+                    layout={({ label, input }) => (
+                      <div className="flex-container-h baseline margin-between-lg">
+                        <div className="bold">{label}</div>
+                        <div className="flex-grow">{input}</div>
+                      </div>
+                    )}
                     input={inputProps => (
                       <ReactSelectInput
                         clearable={false}
--- a/app/scripts/modules/core/src/pipeline/config/actions/history/ShowPipelineHistoryModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/actions/history/ShowPipelineHistoryModal.tsx
@@ -118,8 +118,8 @@ export function ShowPipelineHistoryModal(props: IShowHistoryModalProps) {
           )}
           {!loading && history.length > 1 && (
             <>
-              <div className="history-header row">
-                <div className="col-md-4">
+              <div className="history-header row horizontal">
+                <div className="revision-section col-md-4">
                   <FormField
                     label="Revision"
                     input={inputProps => (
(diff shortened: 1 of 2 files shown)


erikmunson:
forgot to add a screenshot of the vertical misalignment, which only happens when you select the current version and the 'restore this version' button is not present:
Screen Shot 2019-10-25 at 6 43 16 PM

app/scripts/modules/core/src/pipeline/config/actions/history/ShowPipelineHistoryModal.tsx
```diff
@@ -118,8 +118,8 @@ export function ShowPipelineHistoryModal(props: IShowHistoryModalProps) {
           )}
           {!loading && history.length > 1 && (
             <>
-              <div className="history-header row">
-                <div className="col-md-4">
+              <div className="history-header row horizontal">
+                <div className="revision-section col-md-4">
                   <FormField
                     label="Revision"
```
christopherthielen: this might be one of those cases where it would make sense to write an inline layout component instead of fight the css/markup in StandardFieldLayout (css which is likely going to change in the future). 

something like this might do the trick: 
```
layout={(label, input) => <div>{label}{input}</div>}
```

app/scripts/modules/core/src/pipeline/config/actions/history/showHistory.less
```diff
@@ -3,33 +3,40 @@
...
+
 .diff-summary {
   vertical-align: middle;
   display: inline-block;
-  margin: 0 20px;
+  margin-right: 20px;
   font-size: 120%;
   font-weight: 600;
   letter-spacing: -1px;
+
```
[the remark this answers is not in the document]
erikmunson: prettier was extremely unsatisfied with this file.

app/scripts/modules/core/src/pipeline/config/actions/history/ShowPipelineHistoryModal.tsx
erikmunson: you know, i originally thought 'aw man i'm gonna need my own layout', and then i had a spark of memory and said 'oh wait Chris overrode the min-width of this label in the Triggers component and it worked out great, i'm gonna steal that'. Guess I should've just tried out the "dumb" layout first.

erikmunson: Turns out to get this looking right you actually need a fair amount of layout code (side-by-side flex with the right margin, bold text, force the input to grow to size), which makes the markup look _slightly_ closer to writing a standard layout:

```jsx
layout={({ label, input }) => (
  <div className="flex-container-h baseline margin-between-lg">
    <div className="bold">{label}</div>
    <div className="flex-grow">{input}</div>
  </div>
)}
```

I'm cool with/prefer that, but do you personally prefer having this level of layout smarts hard-coded inline versus needing to refactor CSS overrides later?

christopherthielen: The idea behind the field layouts is that there are a lot of things that look similar, and we can change them all in lockstep by updating the layout component.  I think overusing the layouts for things which _dont look similar_ will make it harder to iterate on the layout itself down the road.  I'd lean towards adding the markup.

In fact, `FormikFormField` components make it really easy to: reuse a standard layout, apply validation, handle state management.  `FormField` components make it easy to reuse a layout, and apply validation.  In this UI, the select widget doesn't really benefit from any of those things.  Maybe `FormField` isn't the right choice here at all here, and it should just be markup with a select inside it.

```js
 <div className="flex-container-h baseline margin-between-lg">
    <div className="bold">Revision</div>
    <div className="flex-grow"><SelectInput ... /></div>
  </div>
```

christopherthielen: >  'oh wait Chris overrode the min-width of this label in the Triggers component and it worked out great, i'm gonna steal that'

It's not clear to me if this is abusing the layout or not.  I suspect it's borderline abuse.

erikmunson: yeah, it's not clear to me that the field components here are providing a ton of value. For now I'm gonna leave them as-is since this was mostly a quick dive in and out to fix this layout situation, but I'm certainly happier with the inline layout than coupling overrides to a field layout. Thanks for catching it 🙏

PR #7570, opened 2019-10-28
feat(google): add gce scale-down controls

maggieneterval:
Closes: https://github.com/spinnaker/spinnaker/issues/5072
Corresponding Clouddriver PR: https://github.com/spinnaker/clouddriver/pull/4129

- feat(google): add feature flag for gce scale down controls

  Adds a `gceScaleDownControlsEnabled` feature flag. GCE scale-down controls will be available to alpha-listed accounts only, so scale-down functionality in Spinnaker should be explicitly opted into. Once this feature set is generally available, we can remove the flag.

- feat(google): add gce scale-down controls to autoscaling policy ui

cK2XsSX4rAO

HORSQzASt3S


maggieneterval:
@plumpy this is ready for :eye: :eye_speech_bubble: :eyeglasses: :eyes: :roll_eyes:

PR #7579, opened 2019-10-29
fix(runJob): fix artifact output creation

ethanfrogers:
fixes artifact output creation by passing in the `updatePipeline` method


ethanfrogers:

PR #7584, opened 2019-10-30
fix(amazon): Update load balancer validations to match user expectations

caseyhebebrand:
Addressing validation issues on the create/edit load balancer user flow: 

1. Form validation for health check values was invoked on the user string, causing the validation to fail. ("Must be a number")
2. Inconsistent with [AWS](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html) rules. The summary is below, but more details can be found in the doc I linked. 

**HealthCheckTimeoutSeconds**
For target groups with TCP/TLS protocols - it must be 6 seconds for HTTP healthchecks and 10 seconds for TCP/HTTPS healthchecks.

For target groups with HTTP/HTTPS protocols, the range of values 2-120 should be acceptable.

**HealthCheckIntervalSeconds**
For TCP healthchecks, the interval must be 10 or 30 seconds.

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
@@ -13,8 +13,6 @@ import {
   spelNumberCheck,
   FormValidator,
 } from '@spinnaker/core';
-
-import { IAmazonApplicationLoadBalancer, IAmazonApplicationLoadBalancerUpsertCommand } from 'amazon/domain';
 import {
   isNameLong,
   isNameInUse,
@@ -24,6 +22,8 @@ import {
   isValidHealthCheckProtocol,
   spelNumber,
 } from '../common/targetGroupValidators';
+
+import { IAmazonApplicationLoadBalancer, IAmazonApplicationLoadBalancerUpsertCommand } from 'amazon/domain';
 
 export interface ITargetGroupsProps {
   app: Application;
@@ -93,13 +93,13 @@ export class TargetGroups extends React.Component<ITargetGroupsProps, ITargetGro
           .spelAware()
           .withValidators(spelNumber, checkBetween('unhealthyThreshold', 2, 10));
         builder.field('protocol', 'Protocol').required();
-        builder.field('healthCheckPath', 'HealthCheck Path').required();
+        builder.field('healthCheckPath', 'Health Check Path').required();
         builder
-          .field('healthCheckPort', 'HealthCheck Port')
+          .field('healthCheckPort', 'Health Check Port')
           .required()
           .spelAware()
           .withValidators(value => (value === 'traffic-port' ? null : spelNumberCheck(value)));
-        builder.field('healthCheckProtocol', 'HealthCheck Protocol').required();
+        builder.field('healthCheckProtocol', 'Health Check Protocol').required();
         builder
           .field('healthCheckTimeout', 'Timeout')
           .withValidators(isValidTimeout(item), checkBetween('healthCheckTimeout', 2, 120));
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
@@ -46,10 +46,12 @@ export class TargetGroups extends React.Component<ITargetGroupsProps, ITargetGro
 
   private checkBetween(errors: any, object: any, fieldName: string, min: number, max: number) {
     const field = object[fieldName];
-    if (!Number.isNaN(field)) {
+    const sanitizedField = Number.parseInt(field, 10);
+
+    if (!Number.isNaN(sanitizedField)) {
       const error =
-        Validators.minValue(min)(field, robotToHuman(fieldName)) ||
-        Validators.maxValue(max)(field, robotToHuman(fieldName));
+        Validators.minValue(min)(sanitizedField, robotToHuman(fieldName)) ||
+        Validators.maxValue(max)(sanitizedField, robotToHuman(fieldName));
       if (error) {
         errors[fieldName] = error;
       }
@@ -94,6 +96,20 @@ export class TargetGroups extends React.Component<ITargetGroupsProps, ITargetGro
           }
         });
       }
+      ['port', 'healthCheckInterval', 'healthyThreshold', 'unhealthyThreshold'].forEach(key => {
+        const err = spelNumberCheck(targetGroup[key]);
+        if (err) {
+          tgErrors[key] = err;
+        }
+      });
+
+      if (targetGroup.healthCheckProtocol === 'TCP') {
+        const field = Number.parseInt(targetGroup.healthCheckInterval, 10);
+        if (field !== 10 && field !== 30) {
+          tgErrors.healthCheckInterval = 'TCP health checks only support 10s and 30s intervals.';
+        }
+      }
+
       this.checkBetween(tgErrors, targetGroup, 'healthCheckTimeout', 2, 60);
       this.checkBetween(tgErrors, targetGroup, 'healthCheckInterval', 5, 300);
       this.checkBetween(tgErrors, targetGroup, 'healthyThreshold', 2, 10);
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
@@ -110,7 +110,20 @@ export class TargetGroups extends React.Component<ITargetGroupsProps, ITargetGro
         }
       }
 
-      this.checkBetween(tgErrors, targetGroup, 'healthCheckTimeout', 2, 60);
+      if (targetGroup.protocol === 'TCP' || targetGroup.protocol === 'TLS') {
+        if (targetGroup.healthCheckProtocol === 'HTTP' && Number.parseInt(targetGroup.healthCheckInterval, 10) !== 6) {
+          tgErrors.healthCheckTimeout = 'HTTP health check timeouts for TCP/TLS target groups must be 6s';
+        }
+
+        if (
+          (targetGroup.healthCheckProtocol === 'HTTPS' || targetGroup.healthCheckProtocol === 'TLS') &&
+          Number.parseInt(targetGroup.healthCheckInterval, 10) !== 10
+        ) {
+          tgErrors.healthCheckTimeout = 'HTTPS/TLS health check timeouts for TCP/TLS target groups must be 10s';
+        }
+      }
+
+      this.checkBetween(tgErrors, targetGroup, 'healthCheckTimeout', 2, 120);
       this.checkBetween(tgErrors, targetGroup, 'healthCheckInterval', 5, 300);
       this.checkBetween(tgErrors, targetGroup, 'healthyThreshold', 2, 10);
       this.checkBetween(tgErrors, targetGroup, 'unhealthyThreshold', 2, 10);
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
@@ -19,7 +19,7 @@ import {
   isDuplicateName,
   isValidTimeout,
   checkBetween,
-  isValidHealthCheckProtocol,
+  isValidHealthCheckInterval,
   spelNumber,
 } from '../common/targetGroupValidators';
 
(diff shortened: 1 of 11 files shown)


christopherthielen:
It might make sense to migrate these fields to `<SpelNumberInput>` instead.

This would also be a great opportunity to migrate this old `validate()` method to `FormValidation` and prove out the recent-ish `spelAware` validation methods from `validation.ts`

caseyhebebrand:
> For the cases where there are only a couple of specific values accepted, why not a radio input? For the cases where only one value is acceptable, why make people choose at all vs. just showing the value that will be used?

That is a great point. I will just show the chosen value for the two cases where only one value is acceptable. For the case where there are 2 options I will think a little more about the best user experience for this - either a radio or a tooltip and the same input as the other cases.

caseyhebebrand:
@erikmunson So I took a similar approach to what you were suggesting. For the strict use cases with only 1 valid option, I render a disabled input with the valid value and a `HelpField` explaining why they cannot select another value. For those with only a couple options I show a `HelpField` to explain which options are valid.

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.ts
```diff
@@ -0,0 +1,58 @@
+import { get } from 'lodash';
+import { Validators, robotToHuman, spelNumberCheck } from '@spinnaker/core';
+import { ITargetGroup } from 'amazon/domain';
+
+export const isNameInUse = (
+  existingNames: { [account: string]: { [region: string]: string[] } },
+  credentials: string,
+  region: string,
+) => (name: string) =>
+  get(existingNames, [credentials, region], []).includes(name.toLowerCase())
```
christopherthielen: Oh wow, TIL that the `path` argument to `get` can be an array!

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const isNameLong = (appNameLength: number) => (name: string) =>
+  name.length < 32 - appNameLength
+    ? null
+    : 'Target group name is automatically prefixed with the application name and cannot exceed 32 characters in length.';
+
+export const isDuplicateName = (duplicateGroups: string[]) => (name: string) =>
+  !duplicateGroups.includes(name) ? null : 'Duplicate target group name in this load balancer.';
+
+export const isValidTimeout = (targetGroup: ITargetGroup) => (value: string) => {
```
christopherthielen: should this be `(value: number)` now that you've migrated to NumberInput?

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+export const isValidHealthCheckInterval = (targetGroup: ITargetGroup) => (value: string) =>
+  targetGroup.healthCheckProtocol !== 'TCP' ||
+  (targetGroup.healthCheckProtocol === 'TCP' &&
+    (Number.parseInt(value, 10) === 10 || Number.parseInt(value, 10) === 30))
+    ? null
+    : 'TCP health checks only support 10s and 30s intervals';
+
+export const checkBetween = (fieldName: string, min: number, max: number) => (value: string) => {
```
christopherthielen: This one could probably be moved to `@spinnaker/core` validators as it's a common (non-amazon specific) case.  We could use it inside `NumberInput` as well.

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.ts
```diff
@@ -0,0 +1,58 @@
+import { get } from 'lodash';
```
christopherthielen: Awesome work with the validators.  The factory functions can be typed to return `IValidator`

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+          'us-east-1': ['targetgroup2'],
+        },
+      };
+      const actual = Validators.isNameInUse(existingGroups, 'test', 'us-east-1')(mockTargetGroup.name);
+      expect(actual).toEqual(null);
+    });
+  });
+
+  describe('of name length', () => {
+    it('returns error if the name is >32 additional characters', () => {
```
christopherthielen: might be worth explaining somehow in the `it` block that the application name is prepended

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+      const actual = Validators.isNameLong(12)(`${mockTargetGroup.name}areallylongname`);
```
christopherthielen: minor testing nit: In assertions like this, it's easier to understand the intention of the test if you use literal values instead of references, even if it means more "duplication".  For example, I think this assertion would be easier to grok if it was:

```suggestion
      const actual = Validators.isNameLong('application13'.length)('areallylongtargetgroupname');
```

This way, you don't have to wonder what `12` means, nor what is in `mockTargetGroup.name`

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.ts
```diff
@@ -0,0 +1,58 @@
...
+  get(existingNames, [credentials, region], []).includes(name.toLowerCase())
+    ? `There is already a target group in ${credentials}:${region} with that name.`
+    : null;
+
+export const isNameLong = (appNameLength: number) => (name: string) =>
+  name.length < 32 - appNameLength
+    ? null
+    : 'Target group name is automatically prefixed with the application name and cannot exceed 32 characters in length.';
+
+export const isDuplicateName = (duplicateGroups: string[]) => (name: string) =>
```
christopherthielen: This validator already exists in core as `valueUnique`

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.ts
```diff
@@ -0,0 +1,58 @@
...
+    const error =
+      Validators.minValue(min)(sanitizedField, robotToHuman(fieldName)) ||
+      Validators.maxValue(max)(sanitizedField, robotToHuman(fieldName));
+
+    return error;
+  }
+  return null;
+};
+
+export const spelNumber = (value: number | string) => spelNumberCheck(value);
```
christopherthielen: `spelNumberCheck` was created before the validation api was finalized.  I'd prefer to create a (non-spel aware) `isNumber` validator in core and use that, deprecating `spelNumberCheck` entirely.  `isNumber` could then be composed into the other number validators (`minValue`, `maxValue`, `between`)

app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
```diff
@@ -357,8 +341,7 @@ export class TargetGroups extends React.Component<ITargetGroupsProps, ITargetGro
                           <span className="wizard-pod-content">
                             <label>Path </label>
                             <SpInput
-                              className="form-control input-sm inline-text"
-                              error={tgErrors.healthCheckPath}
+                              // error={tgErrors.healthCheckPath}
```
christopherthielen: remove?

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+describe('Target Group validators', () => {
+  describe('of used names', () => {
+    it('returns an error if the name exists already', () => {
+      const existingGroups = {
+        test: {
+          'us-east-1': ['targetgroup'],
+        },
+      };
+      const actual = Validators.isNameInUse(existingGroups, 'test', 'us-east-1')(mockTargetGroup.name);
+      const expected = 'There is already a target group in test:us-east-1 with that name.';
```
erikmunson: is there value in asserting the exact error message in these tests (given that they'll likely just get copy-pasted across files when changed), as opposed to only asserting that the validator returns _some_ error message when it's expected to? e.g. `expect(isString(actual))`?

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+  region: 'us-east-1',
+  type: '',
+};
+
+describe('Target Group validators', () => {
+  describe('of used names', () => {
+    it('returns an error if the name exists already', () => {
+      const existingGroups = {
+        test: {
+          'us-east-1': ['targetgroup'],
```
erikmunson:
```suggestion
          'us-east-1': [mockTargetGroup.name],
```

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    it('returns null if the name is < 32', () => {
+      const actual = Validators.isNameLong(12)(mockTargetGroup.name);
+      expect(actual).toEqual(null);
+    });
+  });
+
+  describe('of duplicate names across the load balancer', () => {
+    it('returns an error if the name exists', () => {
+      const actual = Validators.isDuplicateName(['targetgroup'])(mockTargetGroup.name);
```
erikmunson:
```suggestion
      const actual = Validators.isDuplicateName([mockTargetGroup.name])(mockTargetGroup.name);
```

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+      expect(actual).toEqual(expected);
+    });
+
+    it('should be 6s and is valid', () => {
+      const actual = Validators.isValidTimeout(tg)('6');
+      expect(actual).toEqual(null);
+    });
+
+    it('should have a 10s timeout', () => {
+      const actual = Validators.isValidTimeout({ ...tg, healthCheckProtocol: 'HTTPS' })('9');
```
erikmunson: i found this double-override (first in the definition of `tg` and then inline in the `it` blocks) a bit hard to follow. Thoughts on just defining multiple variables at the top with friendlier names? e.g.
```ts
    const httpTargetGroup = {
      ...mockTargetGroup,
      protocol: 'TCP',
      healthCheckProtocol: 'HTTP',
    };

    const httpsTargetGroup = {
      ...mockTargetGroup,
      protocol: 'TCP',
      healthCheckProtocol: 'HTTPS',
    };

```

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+      const actual = Validators.isValidTimeout(mockTargetGroup)('10');
+      expect(actual).toEqual(null);
+    });
+  });
+
+  describe('of health check interval', () => {
+    const tg = {
+      ...mockTargetGroup,
+      healthCheckProtocol: 'TCP',
+    };
```
erikmunson: similar to above, `tcpTargetGroup` might be a bit friendlier.

app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
```diff
@@ -369,27 +352,42 @@ export class TargetGroups extends React.Component<ITargetGroupsProps, ITargetGro
...
+                                }
+                              />
+                            ) : (
+                              <SpelNumberInput
+                                error={tgErrors.healthCheckTimeout}
+                                required={true}
+                                value={targetGroup.healthCheckTimeout}
+                                min={2}
+                                max={120}
+                                onChange={(value: string) =>
```
erikmunson: I noticed that everywhere you're using `SpelNumberInput` (for both ALB/NLB) the type of this callback has been narrowed to `string` — what's the context behind that? Is there something downstream of these callbacks that gets angry if you let the `number | string` type flow through?

I only ask because in theory (not always in practice) it's preferable to let the types that are already defined flow, and in this case it seems like the narrowed `string` type also isn't entirely accurate re: what will actually happen at runtime?

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
caseyhebebrand: Its more of a nit-picky value to me. It brings value if we care about the specific language being used. I come from a world of very strict language requirements, therefore I have a habit of this, but I am happy to change this if you feel it is too strict.

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.spec.ts
caseyhebebrand: No strong opinions. I always lean toward repeating code as little as possible, but Chris expressed a similar concern as you so I will go make the references a bit more readable.

app/scripts/modules/amazon/src/loadBalancer/configure/common/targetGroupValidators.ts
caseyhebebrand: Great, will convert to that!

PR #7586, opened 2019-10-31
feat(provider/aws): Function create/update/delete feature.

sidmuls:
Issue: https://github.com/spinnaker/spinnaker/issues/4921Issue
Summary:
Add functions support to Spinnaker UI. This is a net functionality that we've been working on for almost a year to this point. The initial backend work was done like three months back.
From then we've worked on creating the UI components with guidance from the Netflix UX team.

Cloud Provider(s): provider/AWS
Environment:
This worked on user's local environment and also through deploying docker images to kubernetes cluster.

Feature Area: @spinnaker/ui-ux-team
Description:
CRUD functionality for managing lambda functions, With this new we must be able to create, update and delete lambda lambda functions through Spinnaker UI
We must also be able to attach target groups to lambda functions using Spinnaker UI.
Additional Details:
Clouddriver backend details: https://github.com/spinnaker/clouddriver/tree/master/clouddriver-lambda


app/scripts/modules/core/src/function/function.write.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    command: IFunctionUpsertCommand,
+    application: Application,
+    descriptor: string,
+    params: any = {},
+  ): IPromise<ITask> {
+    Object.assign(command, params);
+    return TaskExecutor.executeTask({
+      job: [command],
+      application,
+      description: `${descriptor} Function: ${command['functionName']}`,
```
caseyhebebrand: Can we use dot notation and add a fallback name so it isn't interpolated to "undefined"?
`${command.functionName || ''}`

app/scripts/modules/core/src/function/function.read.service.ts
```diff
@@ -36,7 +36,11 @@ export class FunctionReader {
   ): IPromise<IFunctionSourceData[]> {
     return API.all('functions')
       .withParams({ provider: cloudProvider, functionName: name, region: region, account: account })
-      .get();
+      .get()
+      .then((functions: IFunctionSourceData[]) => {
+        functions = this.functionTransformer.normalizeFunctionSet(functions);
+        return this.$q.all(functions.map(fn => this.normalizeFunction(fn)));
```
caseyhebebrand: Can you explain why it is necessary to normalize the function set and then go through and normalize each function? I don't know if this is possible, but can all the normalization happen in one step?

app/scripts/modules/core/src/function/CreateFunctionButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export interface IFunctionModalProps extends IModalComponentProps {
+  className?: string;
+  dialogClassName?: string;
+  app: Application;
+  forPipelineConfig?: boolean;
+  functionDef: IFunction;
+  command?: IFunctionUpsertCommand; // optional, when ejecting from a wizard
+  closeModal?(functionCommand: IFunctionUpsertCommand): void; // provided by ReactModal
+  dismissModal?(rejectReason?: any): void; // provided by ReactModal
```
caseyhebebrand: nit: 
```
closeModal?: (functionCommand: IFunctionUpsertCommand) => void; 
dismissModal?: (rejectReason?: any) => void; 
```

app/scripts/modules/core/src/function/CreateFunctionButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+        if (provider.CreateFunctionModal) {
+          provider.CreateFunctionModal.show({
+            app: app,
+            application: app,
+            forPipelineConfig: false,
+            function: null,
+            isNew: true,
+          });
+        } else {
+          // TODO
```
caseyhebebrand: Remove comment

app/scripts/modules/amazon/src/function/function.transformer.ts
```diff
@@ -1,7 +1,81 @@
...
 
 export class AwsFunctionTransformer {
   public normalizeFunction(functionDef: IAmazonFunction): IAmazonFunction {
-    return functionDef;
+    const normalizedFunctionDef: IAmazonFunction = functionDef;
+    normalizedFunctionDef.credentials = functionDef.account;
+    return normalizedFunctionDef;
+  }
+
+  public convertFunctionForEditing(functionDef: IAmazonFunction): IAmazonFunctionUpsertCommand {
```
caseyhebebrand: nit: Using object structuring here can get the job done in a more readable way. Then you can override/delete the attributes that are not consistent with the return you want. 

```
public convertFunctionForEditing = (functionDef: IAmazonFunction): IAmazonFunctionUpsertCommand => ({
  ...functionDef,
   envVariables: functionDef.environment ? functionDef.environment.variables : {},
   others,
})
```

app/scripts/modules/amazon/src/function/details/FunctionActions.tsx
```diff
@@ -0,0 +1,124 @@
...
+      // Load balancer is a part of a different application
+      ApplicationReader.getApplication(functionAppName)
+        .then(functionApp => {
+          this.setState({ application: functionApp });
+        })
+        .catch(() => {
+          // If the application can't be found, just use the old one
+          this.setState({ application: this.props.app });
+        });
+    }
```
caseyhebebrand: Typically the constructor is only used for assigning initial state, and not using `setState`. Since you have some async behavior, then I would recommend moving this logic to `componentDidMount` instead.

app/scripts/modules/amazon/src/function/configure/FunctionSettings.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { IAmazonFunctionUpsertCommand } from 'amazon/index';
+import { IAmazonFunction } from 'amazon/domain';
+
+export interface IFunctionSettingsProps {
+  formik: FormikProps<IAmazonFunctionUpsertCommand>;
+  isNew?: boolean;
+  functionDef: IAmazonFunction;
+}
+
+export class FunctionSettings extends React.Component<IFunctionSettingsProps>
```
caseyhebebrand: This can be a functional component.

app/scripts/modules/amazon/src/function/CreateLambdaFunction.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+    this.state = {
+      isNew: !props.functionDef,
+      functionCommand: funcCommand,
+      taskMonitor: null,
+    };
+  }
+
+  private _isUnmounted = false;
+  private refreshUnsubscribe: () => void;
```
caseyhebebrand: I feel like these variables should be part of the state. 

`_isUnmounted` is a tricky one.  If a task completes after a user closes the modal, does it matter if we truncate the logic in `onApplicationRefresh`? At that point it seems like the parent component should be keeping track of whether or not the component is open/closed, and that this flag would be better served as a prop since it offers insight about a higher level of the DOM.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+  Application,
+  ValidationMessage,
+} from '@spinnaker/core';
+
+import { FormikProps, FormikErrors, Field } from 'formik';
+import { IAmazonFunctionUpsertCommand } from 'amazon/index';
+import { IAmazonFunction } from 'amazon/domain';
+import { Subject, Observable } from 'rxjs';
+
+const availableRuntimes = [
```
caseyhebebrand: Is there a way to fetch this information easily so we do not have to manually update every time this list changes? If not no worries, I am just not sure how often this can/will change.

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
```diff
@@ -0,0 +1,80 @@
...
+    if (
+      values.deadLetterConfig.targetArn &&
+      !values.deadLetterConfig.targetArn.match(/arn:aws[a-zA-Z-]?:[a-zA-Z_0-9.-]+:./)
+    ) {
+      errors.deadLetterConfig = 'Invalid target ARN.';
+    }
+    return errors;
+  };
+
+  public componentDidMount() {
```
caseyhebebrand: This can be moved into the constructor, and you can just assign the initial state:
`this.state = { some: '' };`

app/scripts/modules/amazon/src/function/configure/FunctionTags.tsx
```diff
@@ -0,0 +1,52 @@
...
+  public validate = () => {
+    const errors = {} as any;
+
+    if (this.duplicateKeys) {
+      errors.vars = 'Tags have duplicate keys.';
+    }
+    return errors;
+  };
+
+  public componentDidMount() {
```
caseyhebebrand: Same comment about doing this by assigning state in the constructor.

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,247 @@
...
+  subnets: ISubnetOption[];
+  securityGroups: ISecurityGroupsByAccountSourceData;
+}
+
+export class Network extends React.Component<INetworkProps, INetworkState>
+  implements IWizardPageComponent<IAmazonFunctionUpsertCommand> {
+  constructor(props: INetworkProps) {
+    super(props);
+    this.getAllVpcs();
+    // this.getAllSubnets();
```
caseyhebebrand: erase

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,247 @@
...
+  private props$ = new Subject<INetworkProps>();
+  private destroy$ = new Subject<void>();
+
+  private getAllVpcs(): void {
+    VpcReader.listVpcs().then(Vpcs => {
+      this.state.vpcOptions = Vpcs;
+    });
+  }
+
+  public validate(): FormikErrors<IAmazonFunctionUpsertCommand> {
```
caseyhebebrand: Does Formik require a validate function even if no validation is required? It seems weird to have this function, but I haven't worked wit Formik enough to know.

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,247 @@
...
+  };
+
+  private toSubnetOption = (value: ISubnetOption): Option<string> => {
+    return { value: value.subnetId, label: value.subnetId };
+  };
+
+  private getSecurityGroupsByVpc = (sgs: ISecurityGroupsByAccountSourceData): Array<Option<string>> => {
+    const { values } = this.props.formik;
+    const sgOptions: Array<Option<string>> = [];
+    forOwn(sgs, function(sgByAccount, acc) {
```
caseyhebebrand: Would it make sense to just `filter` the `sgs` by all the attributes you listed, then map to `{ value, label }`?

app/scripts/modules/core/src/function/function.read.service.ts
sidmuls: The data returned by the `get()` call is of type [{..}].
I did it this way just to be consistent with how an array is handled for normalization. 
Should I use the normalization on the first element in the array ?
Would you like me to do it in some other way ?

app/scripts/modules/amazon/src/function/configure/FunctionSettings.tsx
sidmuls: It needs to implement the IWizardPageComponent interface. 
Can I still make it a functional component?

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
sidmuls: There is no API that serves this info currently.

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: I believe so. For example,  I was facing the following error: 
```
    Class 'Network' incorrectly implements interface 'IWizardPageComponent<IAmazonFunctionUpsertCommand>'.
  Property 'validate' is missing in type 'Network' but required in type 'IWizardPageComponent<IAmazonFunctionUpsertCommand>'.ts(2420)
```

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: I had to use `forOwn` because the `sgs` are of the form: 
```
export interface ISecurityGroupsByAccountSourceData {
    [account: string]: {
        [provider: string]: {
            [region: string]: ISecurityGroupSummary[];
        };
    };
}
```

I'm not how to use `forEach` on this sort of object. 
Would appreciate a suggestion :)

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: done. will appear in the next push.

app/scripts/modules/amazon/src/function/configure/FunctionTags.tsx
sidmuls: done. will appear in the next push.

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
sidmuls: done. will appear in the next push.

app/scripts/modules/amazon/src/function/function.transformer.ts
sidmuls: done. will appear in next push.

app/scripts/modules/core/src/function/CreateFunctionButton.tsx
sidmuls: I moved it to componentWillMount.

Will appear in the next push.

app/scripts/modules/core/src/function/CreateFunctionButton.tsx
sidmuls: done. will appear in next push.

app/scripts/modules/core/src/function/function.write.service.ts
sidmuls: done. Will be seen in the next push.

app/scripts/modules/amazon/src/function/details/FunctionActions.tsx
sidmuls: I moved it to componentWillMount.

Will appear in the next push.

app/scripts/modules/core/src/function/function.read.service.ts
caseyhebebrand: Got it, no I was just curious your rationale behind splitting up the nomalization.

app/scripts/modules/amazon/src/function/configure/FunctionSettings.tsx
caseyhebebrand: I am actually not sure

app/scripts/modules/amazon/src/function/configure/Network.tsx
caseyhebebrand: I see so you want to check the `key`, as opposed to the values. You could do a truthy check to see if the key exists and then the conditional would return if false preventing any errors. 

```
const sgOptions = sgs.map(sg => 
  if (sg[values.credentials] && sg[values.credentials] ['aws'] && 
  sg[values.credentials] ['aws'][values.region]) {
      const groups = sg[values.credentials] ['aws'][values.region];
      return groups.map({ value: group.id, label: group.name })
  }
 return []
);

return flatten(sgGroups)
```

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+    });
+    return (
+      <div className="container-fluid form-horizontal ">
+        <div className="sp-margin-m-bottom">
+          {isNew && (
+            <div className={className}>
+              <strong>Your load balancer will be named: </strong>
+              <HelpField id="aws.function.name" />
+              <span>{this.props.app.name.concat('-').concat(values.functionName)}</span>
+              <Field
```
christopherthielen: We're working towards consolidating forms and form error handling in Deck.  To that end, I'd like to try using a `FormikFormField` here.  If it doesn't work, then what you have already is fine.

```
<FormikFormField name="functionName" input={() => null} />
```

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+        </div>
+        <div className="sp-margin-m-bottom">
+          <FormikFormField
+            fastField={false}
+            name="credentials"
+            label="Account"
+            input={props => (
+              <ReactSelectInput
+                {...props}
+                inputClassName="cloudfoundry-react-select"
```
christopherthielen: You can't use css from `cloudfoundry`.  Why styles does this add?  I'd prefer we remove the custom class from this select completely if that's possible.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+                value={values.credentials}
```
christopherthielen: do not pass `value` prop to inputs being managed by a `FormikFormField`.  It comes automatically from `{...props}`

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
christopherthielen: Also, I notice that we have a FormikFormField on line 206 that should already be handling the error state.  Why are we displaying the same error here?  Can we delete this `<Field/>` and `<ValidationMessage/>`  completely?

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+        </div>
+        <div className="sp-margin-m-bottom">
+          <FormikFormField
+            name="runtime"
+            label="Runtime"
+            help={<HelpField id="aws.function.runtime" />}
+            input={props => (
+              <ReactSelectInput
+                {...props}
+                inputClassName="cloudfoundry-react-select"
```
christopherthielen: same comment as above about this classname.  Would prefer to remove the custom styling completely if possible.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+        </div>
+        <div className="sp-margin-m-bottom">
+          <FormikFormField
+            name="s3bucket"
+            label="S3 Bucket"
+            help={<HelpField id="aws.function.s3bucket" />}
+            input={props => <TextInput {...props} placeholder="S3 bucket name" />}
+          />
+        </div>
+        {errors.s3bucket && <ValidationMessage type="error" message={errors.s3bucket} />}
```
christopherthielen: `FormikFormField` should be rendering any error already, so probably delete this line.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+        <div className="sp-margin-m-bottom">
+          <FormikFormField
+            name="s3key"
+            label="S3 Key"
+            help={<HelpField id="aws.function.s3key" />}
+            input={props => <TextInput {...props} placeholder="object.zip" />}
+          />
+        </div>
+
+        <div className="sp-margin-m-bottom">
```
christopherthielen: Can we remove all these wrapper divs with `sp-margin-m-bottom`?  I think the form should look OK without them.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+          <FormikFormField
+            name="publish"
+            label="Publish"
+            input={props => (
+              <CheckboxInput
+                {...props}
+                value={values.publish}
```
christopherthielen: remove `value`

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+                onChange={() => this.props.formik.setFieldValue('publish', !values.publish)}
```
christopherthielen: I don't think this `onChange` is necessary either.

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { IAmazonFunction } from 'amazon/domain';
+
+export interface IFunctionDebugAndErrorHandlingProps {
+  formik: FormikProps<IAmazonFunctionUpsertCommand>;
+  isNew?: boolean;
+  functionDef: IAmazonFunction;
+}
+
+export interface IFunctionDebugAndErrorHandlingState {
+  some: string;
```
christopherthielen: remove unused state property

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+export class FunctionDebugAndErrorHandling
+  extends React.Component<IFunctionDebugAndErrorHandlingProps, IFunctionDebugAndErrorHandlingState>
+  implements IWizardPageComponent<IAmazonFunctionUpsertCommand> {
+  constructor(props: IFunctionDebugAndErrorHandlingProps) {
+    super(props);
+    this.state = { some: '' };
+  }
+  public validate = (values: IAmazonFunctionUpsertCommand): FormikErrors<IAmazonFunctionUpsertCommand> => {
+    const errors = {} as any;
```
christopherthielen: could also migrate this to `ValidateForm` api.
If we do migrate to ValidateForm, then all these AWS specific validators (for ARN, IAM Role, etc) could be placed in a file called `aws.validators.ts` and reused elsewhere.

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <div className="container-fluid form-horizontal ">
+        <div className="sp-margin-m-bottom">
+          Dead Letter Config
+          <FormikFormField
+            name="deadLetterConfig.targetArn"
+            label="Target ARN"
+            help={<HelpField id="aws.function.deadletterqueue" />}
+            input={props => <TextInput {...props} />}
+          />
+          {errors.deadLetterConfig && <ValidationMessage type="error" message={errors.deadLetterConfig} />}
```
christopherthielen: remove

app/scripts/modules/amazon/src/function/configure/Network.tsx
christopherthielen: this requirement is a wizard page thing

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,246 @@
...
+  public componentDidUpdate() {
+    this.props$.next(this.props);
+  }
+
+  public componentWillUnmount(): void {
+    this.destroy$.next();
+  }
+
+  public componentDidMount(): void {
+    const subnets$ = Promise.resolve(this.getAvailableSubnets())
```
christopherthielen: Since `subnets$` is not an observable, it shouldn't get the `$` suffix.  Alternatively, you can use `Observable.fromPromise()`

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,246 @@
...
+          <div className="sp-margin-m-bottom">
+            {values.credentials && (
+              <FormikFormField
+                name="vpcId"
+                label="VPC Id"
+                help={<HelpField id="aws.function.vpc.id" />}
+                fastField={false}
+                input={props => (
+                  <ReactSelectInput
+                    inputClassName="cloudfoundry-react-select"
```
christopherthielen: similar comment: remove `inputClassName` (unless absolutely necessary) and `value` props

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,246 @@
...
+          <div className="form-group">
+            <div className="col-md-4 sm-label-right">
+              <b>Security Groups </b>
+              <HelpField id="aws.function.subnet" />
+            </div>
+            <div className="col-md-7">
+              {sgOptions.length === 0 && (
+                <div className="form-control-static">No security groups found in the selected account/region/VPC</div>
+              )}
+              {values.credentials && values.credentials !== 'test' && values.vpcId ? (
```
christopherthielen: This check for `credentials !== 'test'` seems wrong.

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const { app, functionObj: functionFromProps } = this.props;
+    const functionDef: IFunction = app.functions.data.find((test: IFunction) => {
+      return (
+        test.functionName === functionFromProps.functionName &&
+        test.region === functionFromProps.region &&
+        test.account === functionFromProps.account
+      );
+    });
+
+    if (functionDef) {
```
christopherthielen: It looks like if `functionDef` is never found, the loading indicator will spin forever.

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
```diff
@@ -0,0 +1,127 @@
...
+  public render() {
+    const { app } = this.props;
+    const { loading, functionDef } = this.state;
+    if (loading) {
+      // Don't bother computing any children if we're loading
+      return <Details loading={loading} />;
+    }
+
+    const functionDetails = (
+      <>
```
christopherthielen: fragment tag not necessary here

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    const functionDetailsSection = (
+      <CollapsibleSection heading="Function Details">{functionDetails}</CollapsibleSection>
+    );
+
+    return (
+      <Details loading={this.state.loading}>
+        <Details.Header icon={<i className="fa icon-sitemap" />} name={this.state.functionDef.functionName}>
+          <div className="actions">
+            {
```
christopherthielen: braces not necessary

app/scripts/modules/amazon/src/function/details/FunctionActions.tsx
christopherthielen: `componentWillMount` is deprecated, I also recommend `componentDidMount`

app/scripts/modules/core/src/function/CreateFunctionButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+            app: app,
+            application: app,
+            forPipelineConfig: false,
+            function: null,
+            isNew: true,
+          });
+        } else {
+          ModalInjector.modalService
+            .open({
+              templateUrl: provider.createFunctionTemplateUrl,
```
christopherthielen: It looks like this code allows a provider to define CreateFunctionModal as AngularJS template/controller.  Going forward, we don't want new providers written using AngularJS, so I don't think this flexibility is needed.  I suggest removing this if/then and only supporting the React case.

app/scripts/modules/amazon/src/function/details/FunctionActions.tsx
sidmuls: `render()` is called before `componentDidMount()` and `application` variable is needed in `render()`.
That is why this `setState` was in the constructor.

app/scripts/modules/amazon/src/function/CreateLambdaFunction.tsx
sidmuls: > Do we know if `_isUnmounted` is necessary? I don't think it it because `refreshUnsubscribe()` is being called in `cWU`.
> 
> Also I think `refreshUnsubscribe` is fine as an instance variable, as it's unrelated to `render()`.

`_isUnmounted` is also used in `onApplicationRefresh`

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: Adding comments for making it easier to understand. Finding  it difficult to break it down as the structure of the object `sgs` is complex. :(

app/scripts/modules/amazon/src/aws.module.ts
```diff
@@ -130,6 +131,8 @@ module(AMAZON_MODULE, [
       LoadBalancersTag: AmazonLoadBalancersTag,
     },
     function: {
+      details: require('./function/details/AmazonFunctionDetails'),
```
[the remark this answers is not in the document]
sidmuls: you are right. Sorry about overlooking this. The change will appear in the next push. :P

app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
```diff
@@ -10,10 +10,13 @@ export interface IAmazonFunction extends IFunction {
   functionName: string;
   publish: boolean;
   description: string;
-  tags: [{}];
+  tags: string | { [key: string]: string };
   memorySize: number;
   timeout: number;
   envVariables: {};
+  environment: {
+    variables: {};
```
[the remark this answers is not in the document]
sidmuls: This is due to the back-end expectations.
The field name that clouddriver expects for environment variables during create/update operations is `envVariables`. However, when fetching the function details from AWS Lambda, it returns the field with name `environment.variables`.

app/scripts/modules/amazon/src/domain/IAmazonFunctionSourceData.ts
```diff
@@ -0,0 +1,48 @@
+import { IFunctionSourceData } from '@spinnaker/core';
+
+export interface IAmazonFunctionSourceData extends IFunctionSourceData {
```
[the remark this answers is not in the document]
sidmuls: Yep :)

app/scripts/modules/amazon/src/domain/IAmazonFunctionSourceData.ts
```diff
@@ -0,0 +1,48 @@
...
+  memorySize: number;
+  revisionId: string;
+  revisions: {};
+  role: string;
+  timeout: number;
+  tracingConfig: {
+    mode: string;
+  };
+  version: string;
+  // Some of the backend in clouddriver returns a vpcid (lowecase) only,
```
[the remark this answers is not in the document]
sidmuls: Removed this field as we now have vpcConfig instead of vpcid.

app/scripts/modules/amazon/src/function/CreateLambdaFunction.tsx
```diff
@@ -0,0 +1,244 @@
...
+import { ExecutionRole } from './configure/ExecutionRole';
+import { FunctionSettings } from './configure/FunctionSettings';
+import { FunctionEnvironmentVariables } from './configure/FunctionEnvironmentVariables';
+import { Network } from './configure/Network';
+import { AwsFunctionTransformer } from './function.transformer';
+import { FunctionTags } from './configure/FunctionTags';
+import { FunctionDebugAndErrorHandling } from './configure/FunctionDebugAndErrorHandling';
+
+export interface IAmazonCreateFunctionProps extends IFunctionModalProps {
+  app: Application;
```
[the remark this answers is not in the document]
sidmuls: Done. Change will appear in next push.

app/scripts/modules/amazon/src/function/CreateLambdaFunction.tsx
```diff
@@ -0,0 +1,244 @@
...
+  public static defaultProps: Partial<IAmazonCreateFunctionProps> = {
+    closeModal: noop,
+    dismissModal: noop,
+  };
+
+  constructor(props: IAmazonCreateFunctionProps) {
+    super(props);
+    const functionTransformer = new AwsFunctionTransformer();
+    const funcCommand = props.command
+      ? (props.command as IAmazonFunctionUpsertCommand) // ejecting from a wizard
```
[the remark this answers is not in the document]
sidmuls: On closer inspection I figured that there are only 2 possibilities for funcCommand. It will either be for creating new function or editing existing function. 
I wasn't sure what `ejecting from wizard` meant and just to be consistent with LoadBalancer
I added the casting. I used `CreateApplicationLoadBalancer` as reference.

I am removing the type casting and the change will appear in the next push.

app/scripts/modules/amazon/src/function/CreateLambdaFunction.tsx
```diff
@@ -0,0 +1,244 @@
...
+    const taskMonitor = new TaskMonitor({
+      application: app,
+      title: `${isNew ? 'Creating' : 'Updating'} your function`,
+      modalInstance: TaskMonitor.modalInstanceEmulation(() => this.props.dismissModal()),
+      onTaskComplete: () => this.checkForS3Update(functionCommandFormatted, descriptor),
+    });
+
+    taskMonitor.submit(() => {
+      functionCommandFormatted.type = 'lambdaFunction';
+      if (descriptor.includes('Update')) {
```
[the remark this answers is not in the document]
sidmuls: Done. Change will appear in he next push.

app/scripts/modules/amazon/src/function/CreateLambdaFunction.tsx
```diff
@@ -0,0 +1,244 @@
...
+    }
+  }
+
+  private onTaskComplete(values: IAmazonFunctionUpsertCommand): void {
+    this.props.app.functions.refresh();
+    this.refreshUnsubscribe = this.props.app.functions.onNextRefresh(null, () => this.onApplicationRefresh(values));
+  }
+
+  private checkForS3Update(functionCommandFormatted: IAmazonFunctionUpsertCommand, descriptor: string): void {
+    if (descriptor.includes('Update') && (functionCommandFormatted.s3bucket || functionCommandFormatted.s3key)) {
```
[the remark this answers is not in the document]
sidmuls: Done. change will appear in the next push.

app/scripts/modules/amazon/src/function/CreateLambdaFunction.tsx
```diff
@@ -0,0 +1,244 @@
...
+      functionCommandFormatted.operation = 'updateLambdaFunctionCode';
+      const { app } = this.props;
+      const taskMonitor = new TaskMonitor({
```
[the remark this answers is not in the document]
sidmuls: Oh you are right! Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/ExecutionRole.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { IAmazonFunction } from 'amazon/domain';
+
+export interface IExecutionRoleProps {
+  formik: FormikProps<IAmazonFunctionUpsertCommand>;
+  isNew?: boolean;
+  functionDef: IAmazonFunction;
+}
+
+export interface IExecutionRoleState {
+  some: '';
```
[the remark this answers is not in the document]
sidmuls: Right. I removed the state altogether.
Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/ExecutionRole.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <div className="col-md-11">
+          <div className="sp-margin-m-bottom">
+            <FormikFormField
+              name="role"
+              label="Role ARN"
+              fastField={false}
+              input={props => (
+                <TextInput
+                  {...props}
+                  type="text"
```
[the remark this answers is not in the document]
sidmuls: Done.

app/scripts/modules/amazon/src/function/configure/ExecutionRole.tsx
```diff
@@ -0,0 +1,64 @@
...
+                  placeholder="Enter role ARN"
+                  className="form-control input-sm no-spel"
```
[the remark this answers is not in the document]
sidmuls: Okay. Removed it. Change will appear in next push.

app/scripts/modules/amazon/src/function/configure/ExecutionRole.tsx
```diff
@@ -0,0 +1,64 @@
...
+                  name="role"
+                  value={values.role}
```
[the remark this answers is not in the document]
sidmuls: done. Change will appear in next push.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+    accounts: [],
+    existingFunctionNames: [],
+    regions: [],
+  };
+
+  constructor(props: IFunctionProps) {
+    super(props);
+
+    AccountService.listAccounts('aws').then(accounts => {
+      this.state = {
```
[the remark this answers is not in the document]
sidmuls: agreed.
Deleted the constructor.
the  `AccountService` observable is already being used in cDM. "I figured we can get rid of the constructor all together.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+      account$: formValues$.map(x => x.credentials).distinctUntilChanged(),
+      region$: formValues$.map(x => x.region).distinctUntilChanged(),
+      functionName$: formValues$.map(x => x.functionName).distinctUntilChanged(),
+      runtime$: formValues$.map(x => x.runtime).distinctUntilChanged(),
+      s3bucket$: formValues$.map(x => x.s3bucket).distinctUntilChanged(),
+      s3key$: formValues$.map(x => x.s3key).distinctUntilChanged(),
+      handler$: formValues$.map(x => x.handler).distinctUntilChanged(),
+    };
+
+    const allAccounts$ = Observable.fromPromise(AccountService.listAccounts('aws')).shareReplay(1);
```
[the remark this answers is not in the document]
sidmuls: `allAccounts$.takeUntil(this.destroy$).subscribe(accounts => this.setState({ accounts })`
is being done in cDM along with regions and functions like so: 
```
Observable.combineLatest(allAccounts$, accountRegions$, regionfunctions$)
      .takeUntil(this.destroy$)
      .subscribe(([accounts, regions, existingFunctionNames]) => {
        return this.setState({ accounts, regions, existingFunctionNames });
      });
```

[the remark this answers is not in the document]
sidmuls: I need to filter the accounts which have provider 'aws' 
I tried this:
```
const allAccounts$ = AccountService.accounts$.filter((accts, idx) => accts[idx].cloudProvider === 'aws' ? true: false).shareReplay(1);
```
but this returns all accounts and not filtered ones.
please advise on this. :)

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+      well: true,
+      'alert-danger': errors.functionName,
+      'alert-info': !errors.functionName,
+    });
+    return (
+      <div className="container-fluid form-horizontal ">
+        <div className="sp-margin-m-bottom">
+          {isNew && (
+            <div className={className}>
+              <strong>Your load balancer will be named: </strong>
```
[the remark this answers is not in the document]
sidmuls: Oops. :P corrected in next push

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
```diff
@@ -0,0 +1,271 @@
...
+              <HelpField id="aws.function.name" />
+              <span>{this.props.app.name.concat('-').concat(values.functionName)}</span>
```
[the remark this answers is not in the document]
sidmuls: Done.  will appear in next push.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
sidmuls: We can delete it. If we do so, it displays a red border indicating that there is an error. It does not display the error message.
I am deleting it and pushing it in the my next commit. :)

app/scripts/modules/amazon/src/function/CreateLambdaFunction.tsx
christopherthielen: oh I see it now

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
sidmuls: Done. Removed it. Change will appear in the next push. :)

app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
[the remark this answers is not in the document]
christopherthielen: 👍 Since this is greenfield project, I strongly suggest we fix this on the back end.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
sidmuls: done. change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
sidmuls: It does not display the error message. 
However, to be consistent, I am deleting it.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
sidmuls: you are right.
Removed :) Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
sidmuls: Removed :) Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/FunctionBasicInformation.tsx
sidmuls: Removed :) 
Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
sidmuls: done :)

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
sidmuls: done:)

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
```diff
@@ -0,0 +1,77 @@
...
+            name="tracingConfig.mode"
+            label="Mode"
+            help={<HelpField id="aws.function.tracingConfig.mode" />}
+            input={props => (
+              <ReactSelectInput
+                {...props}
+                inputClassName="cloudfoundry-react-select"
+                stringOptions={['Active', 'PassThrough']}
+                clearable={true}
+                value={values.tracingConfig.mode}
```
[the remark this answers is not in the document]
sidmuls: done :)

app/scripts/modules/amazon/src/function/configure/FunctionEnvironmentVariables.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { IAmazonFunctionUpsertCommand } from 'amazon/index';
+import { IAmazonFunction } from 'amazon/domain';
+
+export interface IFunctionEnvironmentVariablesProps {
+  formik: FormikProps<IAmazonFunctionUpsertCommand>;
+  isNew?: boolean;
+  functionDef: IAmazonFunction;
+}
+
+export interface IFunctionEnvironmentVariablesState {
```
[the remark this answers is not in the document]
sidmuls: Removed. Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/FunctionEnvironmentVariables.tsx
```diff
@@ -0,0 +1,71 @@
...
+  public render() {
+    const { values } = this.props.formik;
+    return (
+      <div className="form-group">
+        <div className="col-md-11">
+          <div className="sp-margin-m-bottom">
+            <b>Environment Variables (optional)</b>
+            <HelpField id="aws.function.env.vars" />
+          </div>
+          <MapEditor model={values.envVariables} allowEmpty={true} onChange={this.varsChanged} />
```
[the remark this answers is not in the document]
sidmuls: will fix this once `MapEditorInput` component is fixed.
@christopherthielen

app/scripts/modules/amazon/src/function/configure/FunctionEnvironmentVariables.tsx
```diff
@@ -0,0 +1,71 @@
...
+        </div>
+        <div className="col-md-11">
+          <div className="sp-margin-m-bottom">
```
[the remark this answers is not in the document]
sidmuls: Fixed. Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/FunctionTags.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { IAmazonFunctionUpsertCommand } from 'amazon/index';
+import { IAmazonFunction } from 'amazon/domain';
+
+export interface IFunctionTagsProps {
+  formik: FormikProps<IAmazonFunctionUpsertCommand>;
+  isNew?: boolean;
+  functionDef: IAmazonFunction;
+}
+
+export interface IFunctionTagsState {
```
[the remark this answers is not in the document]
sidmuls: Removed. Will appear in next push.

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,246 @@
...
+    regions: [],
+    subnets: [],
+    securityGroups: null,
+  };
+  private props$ = new Subject<INetworkProps>();
+  private destroy$ = new Subject<void>();
+
+  private getAllVpcs(): void {
+    VpcReader.listVpcs().then(Vpcs => {
+      this.state.vpcOptions = Vpcs;
```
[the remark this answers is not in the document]
sidmuls: Done.  Will appear in next push.

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: Renamed `subnets$` to `allSubnets`.
Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: Done. Will appear in next push.

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: Fixed. change will appear in the next push.

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
sidmuls: Removed.

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
sidmuls: Removed.

app/scripts/modules/core/src/function/CreateFunctionButton.tsx
sidmuls: Removed.

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
sidmuls: added `else` statement for setting `loading: false`

app/scripts/modules/amazon/src/function/configure/FunctionDebugAndErrorHandling.tsx
sidmuls: Done. change will appear in the next push

app/scripts/modules/amazon/src/function/details/FunctionActions.tsx
caseyhebebrand: You can set a default value in the constructor (`this.props.app` seems like a viable option), then use `componentDidMount` to immediately fetch the application and update the state if needed. 

If you need the values on render then you can also consider defaultng `application` to `null` and then render a loading spinner until the fetch has completed.

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
sidmuls: also added 'Function not found' message to be displayed in `else` case.

app/scripts/modules/amazon/src/function/configure/FunctionTags.tsx
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  public render() {
+    const { values } = this.props.formik;
+    return (
+      <div className="container-fluid form-horizontal ">
+        <FormikFormField
+          fastField={false}
+          name="tags"
+          input={props => <MapEditor {...props} model={values.tags} allowEmpty={false} onChange={this.varsChanged} />}
```
christopherthielen: I'm wondering if this one was missed, or if there is some reason you can't use `MapEditorInput`?

app/scripts/modules/amazon/src/function/configure/FunctionTags.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export interface IFunctionTagsProps {
+  formik: FormikProps<IAmazonFunctionUpsertCommand>;
+  isNew?: boolean;
+  functionDef: IAmazonFunction;
+}
+
+export class FunctionTags extends React.Component<IFunctionTagsProps>
+  implements IWizardPageComponent<IAmazonFunctionUpsertCommand> {
+  constructor(props: IFunctionTagsProps) {
```
christopherthielen: omit this empty constructor

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,248 @@
...
+    accounts: null,
+    regions: [],
+    subnets: [],
+    securityGroups: null,
+  };
+  private props$ = new Subject<INetworkProps>();
+  private destroy$ = new Subject<void>();
+
+  private getAllVpcs(): void {
+    VpcReader.listVpcs().then(Vpcs => {
```
christopherthielen: To avoid an unmount race condition: 

```js
Observable.fromPromise(VpcReader.listVpcs())
  .takeUntil(this.destroy$)
  .subscribe(vpcs => {
```

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,248 @@
...
+  private getAvailableSubnets(): IPromise<ISubnet[]> {
+    return SubnetReader.listSubnetsByProvider('aws');
+  }
+
+  private getAvailableSecurityGroups(): IPromise<ISecurityGroupsByAccountSourceData> {
+    return ReactInjector.securityGroupReader.getAllSecurityGroups();
+  }
+  private makeSubnetOptions(availableSubnets: ISubnet[]): ISubnetOption[] {
+    const subOptions: ISubnetOption[] = [];
+    availableSubnets.forEach(s => {
```
christopherthielen: simplified:
```
const subOptions: ISubnetOption[] = availableSubnets.map(s => ({ subnetId: s.id, vpcId: s.vpcId }));
```

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,248 @@
...
+      const opt: ISubnetOption = {
+        subnetId: s.id,
+        vpcId: s.vpcId,
+      };
+      subOptions.push(opt);
+    });
+    // we have to filter out any duplicate options
+    const uniqueSubOptions = Array.from(new Set(subOptions.map(a => a.subnetId))).map(subnetId => {
```
christopherthielen: simplified using `uniqBy` from lodash:
```js
    const uniqueSubOptions = uniqBy(subOptions, 'subnetId');
```

app/scripts/modules/amazon/src/function/configure/Network.tsx
```diff
@@ -0,0 +1,248 @@
...
+                name="vpcId"
+                label="VPC Id"
+                help={<HelpField id="aws.function.vpc.id" />}
+                fastField={false}
+                input={props => (
+                  <ReactSelectInput
+                    {...props}
+                    stringOptions={vpcOptions
+                      .filter((v: IVpc) => v.account === values.credentials)
+                      .map((v: IVpc) => v.id)}
```
christopherthielen: Not a task for today, but we generally don't show `vpc.id` in the UI, we try to use friendly names (like `vpc0` at netflix).  However, I don't know the details of this mapping off the top of my head.

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
```diff
@@ -0,0 +1,142 @@
...
+      </dl>
+    );
+
+    const functionDetailsSection = (
+      <CollapsibleSection heading="Function Details">{functionDetails}</CollapsibleSection>
+    );
+
+    return (
+      <Details loading={this.state.loading}>
+        {_.isEmpty(this.state.functionDef) ? (
```
christopherthielen:
```suggestion
        {isEmpty(this.state.functionDef) ? (
```

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
```diff
@@ -0,0 +1,142 @@
...
+                functionFromParams={{
+                  account: this.state.functionDef.account,
+                  region: this.state.functionDef.region,
+                  functionName: this.state.functionDef.functionName,
+                }}
+              />
+            </div>
+          </Details.Header>
+        )}
+        {!_.isEmpty(this.state.functionDef) ? functionDetailsSection : ''}
```
christopherthielen:
```suggestion
        {!isEmpty(this.state.functionDef) ? functionDetailsSection : ''}
```

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
```diff
@@ -0,0 +1,142 @@
...
+  IFunction,
+  Application,
+  AccountTag,
+  IOverridableProps,
+  Overrides,
+} from '@spinnaker/core';
+import { IAmazonFunctionSourceData, IAmazonFunction } from 'amazon/domain';
+import { FunctionActions } from './FunctionActions';
+import { AwsReactInjector } from 'amazon/reactShims';
+import * as _ from 'lodash';
```
christopherthielen:
```suggestion
import { isEmpty } from 'lodash';
```

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
```diff
@@ -0,0 +1,142 @@
...
+        functionDef: {},
+        loading: false,
+      });
+    }
+  }
+
+  public componentDidMount(): void {
+    const { app } = this.props;
+    const dataSource = app.functions;
+    dataSource.ready().then(() => {
```
christopherthielen: race condition here, let's use the `this.destroy$` observable pattern since this is a class component.  At some future date, we can refactor to a functional component and utilize `useData`

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const functionDef: IFunction = app.functions.data.find((test: IFunction) => {
+      return (
+        test.functionName === functionFromProps.functionName &&
+        test.region === functionFromProps.region &&
+        test.account === functionFromProps.account
+      );
+    });
+
+    if (functionDef) {
+      AwsReactInjector.functionReader
```
christopherthielen: race condition here, let's use the `this.destroy$` observable pattern since this is a class component.

app/scripts/modules/core/src/function/CreateFunctionButton.tsx
```diff
@@ -0,0 +1,61 @@
...
+  closeModal?: (functionCommand: IFunctionUpsertCommand) => void; // provided by ReactModal
+  dismissModal?: (rejectReason?: any) => void; // provided by ReactModal
+}
+
+export interface ICreateFunctionButtonProps {
+  app: Application;
+}
+
+export class CreateFunctionButton extends React.Component<ICreateFunctionButtonProps> {
+  constructor(props: ICreateFunctionButtonProps) {
```
christopherthielen: remove empty constructor

app/scripts/modules/amazon/src/aws.validators.ts
```diff
@@ -0,0 +1,27 @@
+import * as _ from 'lodash';
```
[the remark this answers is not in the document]
sidmuls: Done. Change will appear in the next push.

app/scripts/modules/amazon/src/aws.validators.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const awsArnValidator = (value: string, label: string) => {
+  const arn = value.match(/arn:aws[a-zA-Z-]?:[a-zA-Z_0-9.-]+:./);
+  return arn
+    ? undefined
+    : `Invalid ARN.  ${label} must match regular expression: /arn:aws[a-zA-Z-]?:[a-zA-Z_0-9.-]+:./`;
+};
+
+export const awsTagsValidator = (value: string | { [key: string]: string }, label: string) => {
+  return _.isEmpty(value) ? `Atleast one ${label} is required` : undefined;
```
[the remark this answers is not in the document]
sidmuls: Done. Change will appear in the next push

app/scripts/modules/amazon/src/function/configure/FunctionTags.tsx
sidmuls: Sorry I missed this one.
Changed it to MapEditorInput.

app/scripts/modules/amazon/src/function/configure/FunctionTags.tsx
sidmuls: Done. Change will appear in the next push

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: Done. Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: Done. Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: Done. Change will appear in the next push.

app/scripts/modules/amazon/src/function/configure/Network.tsx
sidmuls: Not all VPCs returned by the API `getAllVps()` have a name. All have the ID field.
I agree it is not user friendly. I will see to it that this is fixed in the future.

app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
sidmuls: Done. Change will appear in the next push.

app/scripts/modules/core/src/function/CreateFunctionButton.tsx
sidmuls: Done. change will appear in the next push.

PR #7588, opened 2019-10-31
chore(help): Update help text for GitHub artifact URL

justinrlee:


justinrlee:
I think commits get squashed so I shouldn't have to actually update the commit title.  Let me know if this isn't the case.

PR #7590, opened 2019-11-01
feat(managed): add resource status to infrastructure views

erikmunson:
We recently [added the ability to calculate a summarized "status"](https://github.com/spinnaker/keel/pull/433) for every declaratively managed resource. This status can indicate a variety of different states from the keel perspective, and provide valuable transparency into what's going on during normal operations and especially when something is wrong / abnormal. This PR adds iconography and explanatory text to Clusters, Load Balancers, and Firewalls with roughly similar visual style (screenshots below).

The PR itself is split into pretty atomic commits, I recommend going through them in order as a lot of the code is similar for each type of infrastructure and you can pattern match after the first couple of commits.

**Example of what this looks like when _all regions_ for a particular account + name grouping is managed:**
<details>
<summary>Cluster (all regions)</summary>
Screen Shot 2019-10-31 at 5 34 14 PM
</details>
<details>
<summary>Load Balancer (all regions)</summary>
Screen Shot 2019-10-31 at 5 35 16 PM
</details>
<details>
<summary>Firewall (all regions)</summary>
Screen Shot 2019-10-31 at 5 35 26 PM
</details>

**And examples of when _only some_ regions are managed:**
<details>
<summary>Cluster (some regions)</summary>
Screen Shot 2019-10-31 at 5 38 39 PM
</details>
<details>
<summary>Load Balancer (some regions)</summary>
Screen Shot 2019-10-31 at 5 39 42 PM
</details>
<details>
<summary>Firewall (some regions)</summary>
Screen Shot 2019-10-31 at 5 40 25 PM
</details>

**And finally, examples of the iconography and microcopy for each status a resource can be in:**
<details>
<summary>HAPPY</summary>
status-happy
</details>
<details>
<summary>CREATED</summary>
status-created
</details>
<details>
<summary>DIFF</summary>
status-diff
</details>
<details>
<summary>ACTUATING</summary>
status-actuating
</details>
<details>
<summary>PAUSED</summary>
status-paused
</details>
<details>
<summary>UNKNOWN</summary>
status-unknown
</details>
<details>
<summary>ERROR</summary>
status-error
</details>
<details>
<summary>UNHAPPY</summary>
status-unhappy
</details>

There's a lot of copy and whatnot in this PR, so your help and feedback is appreciated around things like typos, wording, clarity, etc. (cc @gcomstock)

### DIFF
--- a/app/scripts/modules/core/src/presentation/HoverablePopover.tsx
+++ b/app/scripts/modules/core/src/presentation/HoverablePopover.tsx
@@ -28,7 +28,10 @@ export interface IHoverablePopoverProps extends React.HTMLProps<any> {
   hOffsetPercent?: string;
   /** class to put on the popover content */
   className?: string;
-
+  /** class to put on the outermost element */
+  wrapperClassName?: string;
+  /** custom style attributes */
+  style?: React.CSSProperties;
   /** Rendered on the top of the popover content */
   title?: string;
   id?: string;
@@ -146,7 +149,18 @@ export class HoverablePopover extends React.Component<IHoverablePopoverProps, IH
   };
 
   public render() {
-    const { Component, template, placement, container, hOffsetPercent, id, title, className } = this.props;
+    const {
+      Component,
+      template,
+      placement,
+      container,
+      hOffsetPercent,
+      id,
+      title,
+      className,
+      wrapperClassName,
+      style,
+    } = this.props;
     const { popoverIsOpen, animation, placementOverride } = this.state;
     const { Wrapper } = this;
 
@@ -157,7 +171,12 @@ export class HoverablePopover extends React.Component<IHoverablePopoverProps, IH
     );
 
     return (
-      <Wrapper style={{ display: 'inline' }} onMouseEnter={this.handleMouseEvent} onMouseLeave={this.handleMouseEvent}>
+      <Wrapper
+        className={wrapperClassName}
+        style={{ display: 'inline', ...style }}
+        onMouseEnter={this.handleMouseEvent}
+        onMouseLeave={this.handleMouseEvent}
+      >
         {this.props.children}
         <Overlay
           show={popoverIsOpen}
--- a/app/scripts/modules/core/src/cluster/ClusterPod.tsx
+++ b/app/scripts/modules/core/src/cluster/ClusterPod.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as classNames from 'classnames';
 import { orderBy, partition, groupBy, map } from 'lodash';
 
 import { ClusterState } from 'core/state';
@@ -11,6 +12,7 @@ import { IClusterSubgroup, IServerGroupSubgroup } from './filter/ClusterFilterSe
 import { ISortFilter } from 'core/filterModel';
 import { ClusterPodTitleWrapper } from 'core/cluster/ClusterPodTitleWrapper';
 import { ServerGroupManager } from 'core/serverGroupManager/ServerGroupManager';
+import { ManagedResourceStatusIndicator } from 'core/managed';
 
 export interface IClusterPodProps {
   grouping: IClusterSubgroup;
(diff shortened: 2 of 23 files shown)


erikmunson:
ugh, forgot there were tests for those filter services. will add the appropriate test code shortly.

app/scripts/modules/core/src/presentation/HoverablePopover.tsx
```diff
@@ -28,7 +28,10 @@ export interface IHoverablePopoverProps extends React.HTMLProps<any> {
   hOffsetPercent?: string;
   /** class to put on the popover content */
   className?: string;
-
+  /** class to put on the outermost element */
+  wrapperClassName?: string;
+  /** custom style attributes */
+  style?: React.CSSProperties;
```
erikmunson: I feel some level of discomfort in adding more stuff to this component's API. That said, I ran into a bit of a wall because:
1) there is a mandatory wrapper div that doesn't get any classes right now, and in the flex layouts I'm working with adding layout classes to the outermost div is required to get things to fill space correctly
2) the very same wrapper div always gets `display: inline` slapped directly on it, which again is a no-go for the layouts I'm in which are flex all the way down and get totally screwed up by this

If people have other ideas, I'm happy to try to get these extra options out of here.

erikmunson:
alright test updates are squashed into their respective commits. really blew up the diff though.

gcomstock:
Thank you so much for all the screen shot examples

gcomstock:
I love all the descriptive text on hover. The only ambiguous thing to me is if the "Learn more" link takes me to the same place on all of them, or tells me about the exact status. If I saw the url I could probably infer that, so maybe it isn't worth changing. Otherwise, we could have the link text either say "Learn more about this" or "Learn more about statuses" Maybe not worth revising.

erikmunson:
Totally agree with the sentiment about that Learn More link — I feel the same way. My plan is actually to very shortly (like, maybe today) create a doc page on spinnaker.io that takes this help text in these bubbles, lays them all out in a list,  and rewords them to have more detail and be appropriate as a 'glossary' of all statuses. Then I'll change this Learn More link to point there, and deep-link to the specific status that's being shown in the UI. At the top of that doc I'll also link to the overall intro page that's being linked to from here right now.

gcomstock:
Perfect!

gal-yardeni:
This description is amazing! thanks for making this easy to understand!

app/scripts/modules/core/src/presentation/HoverablePopover.tsx
caseyhebebrand: Could the `style` prop be represented by another wrapper classname? Then only one prop is added.

erikmunson: Sadly since the `display: inline` property gets added as an inline style prop inside `HoverablePopover` we'd have to use `!important` on the the property in the class. I'm not entirely opposed to the idea, but I'm not sure which is better — an `!important` sprinkled in a stylesheet that's invisible until you check the `.less` file, or an extra prop on this component's API that's explicit when you're reading the code 🤔

christopherthielen: You can move the `display: inline` from inline CSS to `HoverablePopover.css`  then add an override for MD. Something like 
```css
.HoverablePopover {
  display: inline;
}

.HoverablePopover.ManagedResourceStatusIndicator {
  display: flex;
}
```

christopherthielen:
lgtm

app/scripts/modules/core/src/presentation/HoverablePopover.tsx
erikmunson: yeah, you're right it's worth trying this and seeing what happens — I was very anxious / risk-averse about unintended consequences from reducing the specificity, but maybe it's worth the risk to get that inline styling out of there

erikmunson: https://github.com/spinnaker/deck/pull/7597

PR #7591, opened 2019-11-01
feat(dataSources): log runtime error when defaultData isn't provided

erikmunson:
Since this is a new required field and might unexpectedly mess up someone's custom data source in a non-obvious way, let's log an error with some context to make the resolution more obvious.

I originally wanted to just throw a full-on `Error`, but those always get swallowed and end up making the application fail to load without showing up in the console.


christopherthielen:
> but those always get swallowed and end up making the application fail to load without showing up in the console.

What? Really?  Do we know why this happens?  I'd like this not to happen.

erikmunson:
@christopherthielen yeah it's been that way forever, and not just limited to data sources. Anything that throws during app setup/bootstrapping has its exception tossed out and the 'Application not found' thing comes up. Just took a few mins to build out a 'Something went wrong' state and log actual exceptions: https://github.com/spinnaker/deck/pull/7599

PR #7592, opened 2019-11-01
Hooks improvements

christopherthielen:
- Reduced unnecessary renders in `useLatestPromise` and `useDebouncedValue`
- Extracted new `useIsMountedRef` hook
- Added unit tests

### DIFF
--- a/app/scripts/modules/core/src/presentation/hooks/useDebouncedValue.hook.spec.tsx
+++ b/app/scripts/modules/core/src/presentation/hooks/useDebouncedValue.hook.spec.tsx
@@ -1,0 +1,125 @@
+import * as React from 'react';
+
+import { useDebouncedValue } from './useDebouncedValue.hook';
+import { mount } from 'enzyme';
+
+describe('useDebouncedValue hook', () => {
+  beforeEach(() => jasmine.clock().install());
+  afterEach(() => jasmine.clock().uninstall());
+
+  const timeoutMillis = 1000;
+  function Component(props: any) {
+    const { value, onChange, millis } = props;
+    const [debounced, isDebouncing] = useDebouncedValue(value, millis);
+
+    React.useEffect(() => onChange(value, debounced, isDebouncing), [value, debounced, isDebouncing]);
+    return <></>;
+  }
+
+  it('initially, debounced value is the same as the initial value', () => {
+    const spy = jasmine.createSpy();
+    mount(<Component value="a" onChange={spy} millis={timeoutMillis} />);
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalledWith('a', 'a', jasmine.anything());
+  });
+
+  it('initially, isDebouncing is false', () => {
+    const spy = jasmine.createSpy();
+    mount(<Component value="a" onChange={spy} millis={timeoutMillis} />);
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalledWith(jasmine.anything(), jasmine.anything(), false);
+  });
+
+  it('isDebounced is true during the time where the value is different than the debounced value', () => {
+    const spy = jasmine.createSpy();
+    const component = mount(<Component value="a" onChange={spy} millis={timeoutMillis} />);
+    component.setProps({ value: 'b' });
+    expect(spy).toHaveBeenCalledTimes(2);
+    const [value, debouncedValue, isDebouncing] = spy.calls.mostRecent().args;
+    expect([value, debouncedValue, isDebouncing]).toEqual(['b', 'a', true]);
+  });
+
+  it('after the timeout, debounced should equal value and isDebouncing is false', () => {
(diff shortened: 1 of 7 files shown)


app/scripts/modules/core/src/presentation/hooks/useLatestPromise.hook.spec.tsx
```diff
@@ -0,0 +1,161 @@
...
+  function Component(props: any) {
+    const { promiseFactory, deps, onChange } = props;
+    const useLatestPromiseResult: IUseLatestPromiseResult<any> = useLatestPromise(promiseFactory, deps);
+    const { status, result, error, requestId } = useLatestPromiseResult;
+
+    React.useEffect(() => onChange(useLatestPromiseResult), [status, result, error, requestId]);
+    return <></>;
+  }
+
+  function defer() {
```
erikmunson: old skool. love it

app/scripts/modules/core/src/presentation/hooks/useLatestPromise.hook.spec.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  it('has status RESOLVED if a promise resolved', async done => {
+    const spy = jasmine.createSpy('onChange');
+    const deferred = defer();
+    const component = mount(<Component promiseFactory={() => deferred.promise} deps={[]} onChange={spy} />);
+    expect(spy).toHaveBeenCalledTimes(2);
+
+    deferred.resolve('payload');
+    await deferred.promise;
+    component.setProps({});
```
erikmunson: for my own education: I would've expected the promise resolving to trigger a re-render on its own. Is this a thing about the way enzyme works? i.e. it never renders unless you explicitly tell it to?

app/scripts/modules/core/src/presentation/hooks/useDebouncedValue.hook.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  it('does not update debounced value until after the timeout', () => {
+    const spy = jasmine.createSpy();
+    const component = mount(<Component value="a" onChange={spy} millis={timeoutMillis} />);
+    component.setProps({ value: 'b' });
+
+    expect(spy).toHaveBeenCalledTimes(2);
+    expect(spy.calls.mostRecent().args).toEqual(['b', 'a', true]);
+
+    jasmine.clock().tick(timeoutMillis - 1);
```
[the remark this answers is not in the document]
christopherthielen: Previously, the spy was called twice and the last args were b, a, true.  After (timeout - 1)ms, I'm asserting that it has _still only been called twice_.  That is, it hasn't debounced anything yet.

app/scripts/modules/core/src/presentation/hooks/useDebouncedValue.hook.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+    expect(spy.calls.mostRecent().args).toEqual(['b', 'a', true]);
+
+    const halfTimeoutMillis = timeoutMillis / 2;
+    // Wait 500ms -- change the value to 'c' before 'b' is debounced
+    jasmine.clock().tick(halfTimeoutMillis); // clock is now 500ms
+    component.setProps({ value: 'c' });
+    expect(spy).toHaveBeenCalledTimes(3);
+    expect(spy.calls.mostRecent().args).toEqual(['c', 'a', true]);
+
+    // Wait 500ms more.  Debounced should still be 'a'
```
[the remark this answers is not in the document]
christopherthielen: (same as above, it's asserting that the spy didn't get called again)

app/scripts/modules/core/src/presentation/hooks/useLatestPromise.hook.spec.tsx
christopherthielen: Yeah, I think this is a thing about enzyme and I really don't understand it.  I also assumed resolving the promise should cause a re-render, but it doesn't seem to be the case (😢 ).  I also tried `component.update()` which is more explicit, but it doesn't seem to have any effect.  I learned that `setProps({})` will actually do the thing.

erikmunson: huh. Seems ok in cases like this to explicitly re-render, and it does make it easier to reason about what state things will be in when you call `expect`. At the same time it seems like it could get annoying if you're testing more complex renderings interactions. Oh well 🤷‍♂

app/scripts/modules/core/src/presentation/hooks/useDebouncedValue.hook.spec.tsx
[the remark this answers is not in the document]
erikmunson: ah yes i see — i wasn't piecing together the fact that in your dummy component the `useEffect` call has the value listed in its deps so you're basically asserting the value hasn't changed as part of this spy check.

PR #7598, opened 2019-11-06
feat(huaweicloud): first commit for huaweicloud

zengchen1024:
This pr is the first commit for huawei cloud and only includes the intialization works. The cloud provider proposal is spinnaker/community#17


zengchen1024:
@christopherthielen great! thanks for your comment!

PR #7600, opened 2019-11-06
fix(kubernetes): fix patchBody input in Patch (Manifest) stage

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5093

Corresponding Orca PR: https://github.com/spinnaker/orca/pull/3283

Clouddriver's patch operation expects Orca to send it a `patchBody` that is a list when the merge strategy is of type `json`, and a map otherwise. Prior to Spinnaker 1.15, Deck also saved the `patchBody` as either a list or map depending on merge strategy. In Spinnaker 1.15, the artifacts rewrite introduced a bug that caused the `patchBody` to be saved as a `Map` regardless of merge strategy, which broke the JSON merge strategy and Deck's YAML editor. This change:

1.  Saves the `patchBody` as a list regardless of merge strategy type (will always be a singleton list for `strategic` and `merge` strategy types, which is now validated by Orca), and
2. Re-formats any existing non-list `patchBody` as a list. Orca will handle sending either a list or map to Clouddriver, depending on the configured strategy.


maggieneterval:

PR #7602, opened 2019-11-07
feat(core/presentation): Made useData hook default result behavior more reasonable

christopherthielen:
Previously, the default result was only used if some of the deps were null or undefined.  This meant the default result was not used on the first render, nor during the time that the initial promise was resolving.  This also meant the default value _could be used after some value was resolved_, but while one of the deps returned to null or undefined.

Now, the default result is used immediately and while the first promise is PENDING.  It is used until the first promise turns to RESOLVED but then never used again.


app/scripts/modules/core/src/presentation/hooks/useData.hook.ts
```diff
@@ -1,23 +1,26 @@
...
 import { isNil } from 'lodash';
 
 import { useLatestPromise, IUseLatestPromiseResult } from './useLatestPromise.hook';
 
 /**
- * A react hook which invokes a promise returning callback whenever any of its dependencies changes.
- * Returns the provided default value whenever any dependency is null or undefined
+ * A react hook which invokes a promise factory callback whenever any of its dependencies changes.
+ *
+ * Returns the default result until one promise resolves.
```
erikmunson: To be consistent with L17:
```suggestion
 * Returns the default result until the first promise resolves.
```

PR #7604, opened 2019-11-07
feat(core/presentation): Mark all SpinFormik initialValues fields as 'touched'

christopherthielen:
Normally, FormikFormFields do not show validation errors until the user has tabbed away (the onBlur event).  This is intended to improve UX when a user first loads a blank form.  It would be horrible to see the entire form rendered in red because all the fields are required.

However, when a form is reloaded *with existing data*, we usually want to show validation errors immediately.

When the form is first rendered, mark all fields in initialValues as "touched".


christopherthielen:
This only sets `touched` for keys that exist in the `initialValues`, so it would have no effect if you passed `initialValues={{}}` for example.  If most of the fields are blank in this kayenta form, you can control that behavior by omitting those keys entirely in the object passed to `initialValues`.

PR #7605, opened 2019-11-07
fix(core/presentation): in FormikFormField use formik.validateForm() to revalidate

christopherthielen:
…instead of the non-existant formik.validate()


christopherthielen:
Apparently this was also fixed in a previous PR

PR #7608, opened 2019-11-07
feat(core/presentation): Create a useInternalValidator hook for FormInputs to use

christopherthielen:
This is a hook that should be used by Inputs that add their own internal validation code (think min/max number fields)

This hook does two things: 

- It automatically registers/unregisters the internal validator with the parent component.  It uses the `validation` prop which is currently provided only by `FormikFormField` and `FormField`.
- It stabilizes the `IValidator` callback reference so you can re-create the validate function on each render and get fresh closures.

See `NumberInput.tsx` for example usage


christopherthielen:
These commits were merged via https://github.com/spinnaker/deck/pull/7609

PR #7609, opened 2019-11-07
Spel input expression previews

christopherthielen:
feat(core/presentation): Add a SpelService to evaluate expressions on the Server based on a previous execution/stage id.
feat(core/presentation): Create a SpelInput component that evaluates the input text as an expression and generates a markdown preview

spel expression previews low


christopherthielen:
spel expression previews low framerate

app/scripts/modules/core/src/presentation/spel/SpelInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const fetchSpelPreview = useData(
+    () => SpelService.evaluateExpression(expression, previewExecutionId, previewStageId),
+    null,
+    [expression, previewExecutionId, previewStageId],
+  );
+
+  // re-validate whenever an async event occurs
+  React.useEffect(() => {
+    !isInitialRender && props.validation && props.validation.revalidate && props.validation.revalidate();
```
caseyhebebrand: An aside, but the new version of ts has optional chaining which would be great for expressions like these.

christopherthielen: yes!  I can't wait to start using that.  Feel free to take a stab at updating our TS dependency.  I know there were some typing errors that were surfaced with typescript 3.5 (which is why we're on 3.4).

caseyhebebrand: Same! Maybe I will take a stab at it after I am done with this set of PRs

PR #7610, opened 2019-11-07
feat(core/managed): add deep links to status reference doc

erikmunson:
Depends on https://github.com/spinnaker/spinnaker.github.io/pull/1585/ — don't merge until it's live.

There's now a very detailed doc (linked above) that covers what each status means, how they relate, and how you might want to react to a given status. With this change whenever someone sees the in-UI help text with context about the current status, they'll also be able to click the "Learn more about this" link to get deep linked directly to the entry for that status in the reference doc.

I also messed with the GA tracking a bit to log the status that was shown when a doc link got clicked, along with a slightly more future proof name.


erikmunson:
One additional change I just slipped in — the size of the link now that it's `Learn more about this` has made things visually weird, so I broke it out to its own line:

Screen Shot 2019-11-06 at 10 03 36 PM

(@gcomstock)

gcomstock:
Good call, thanks!

PR #7611, opened 2019-11-07
feat(core/managed): add RESUMED resource status

erikmunson:
Depends on https://github.com/spinnaker/spinnaker.github.io/pull/1585 and https://github.com/spinnaker/deck/pull/7610.

Support for the newly-added `RESUMED` status — feedback on wording / etc is welcome. Until #7610 gets merged there will be an extra commit at the beginning, this PR is only one commit.

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
+++ b/app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
@@ -116,7 +116,7 @@ const viewConfigurationByStatus = {
         </p>
         <p>
           Management was resumed after being temporarily paused. If Spinnaker detects that a drift from the declarative
-          configuration occured while paused, it will take automatic action to resolve the drift.
+          configuration occurred while paused, it will take automatic action to resolve the drift.
         </p>
         <p>
           You can pause and resume management on the{' '}
--- a/app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
+++ b/app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
@@ -106,6 +106,29 @@ const viewConfigurationByStatus = {
       </>
     ),
   },
+  RESUMED: {
+    iconClass: 'icon-md',
+    colorClass: 'info',
+    popoverContents: (id: string) => (
+      <>
+        <p>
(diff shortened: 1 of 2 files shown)


erikmunson:
Screen Shot 2019-11-06 at 10 10 07 PM

app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
```diff
@@ -102,8 +100,32 @@ const viewConfigurationByStatus = {
...
           </UISref>
-          . <LearnMoreLink id={id} />
+          .
+        </p>
+        <LearnMoreLink id={id} status={ManagedResourceStatus.PAUSED} />
+      </>
+    ),
+  },
+  RESUMED: {
+    iconClass: 'icon-md',
```
erikmunson: @gcomstock FYI, right now I'm just using the unadorned "M" icon because 1) I don't have an icon for this new "resumed" status 2) this status theoretically will only show up for a few seconds at a time, so unlikely to matter much.

If you have time, might be cool to get a dedicated icon (maybe like a play icon to compliment the pause one?)

gcomstock: Good idea! Here you are: https://github.com/spinnaker/deck/pull/7613

erikmunson: EPIC thanks for the fast follow up

app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
```diff
@@ -106,6 +106,29 @@ const viewConfigurationByStatus = {
...
+    iconClass: 'icon-md-resumed',
+    colorClass: 'info',
+    popoverContents: (id: string) => (
+      <>
+        <p>
+          <b>Continuous management was just resumed.</b>
+        </p>
+        <p>
+          Management was resumed after being temporarily paused. If Spinnaker detects that a drift from the declarative
+          configuration occured while paused, it will take automatic action to resolve the drift.
```
erikmunson:
```suggestion
          configuration occurred while paused, it will take automatic action to resolve the drift.
```

PR #7612, opened 2019-11-07
fix(artifacts): enable inline editing of base64 artifacts

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5102

The [artifacts rewrite commit](https://github.com/spinnaker/deck/pull/6634) added support for a React base64 artifact editor. In the [Angular implementation](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/base64/defaultBase64.artifact.ts#L83), the textarea value is bound to the value of the decoded base64, so does not need to be updated manually. In the React implementation, we are storing the value of the decoded base64 on state and need to manually update it in the textarea change handler.

143e2bc6-47bc-45eb-9d84-6f1346c306b0


maggieneterval:

PR #7617, opened 2019-11-08
feat(core/managed): visually refresh/rebrand infra details indicator

erikmunson:
We have a snazzy new set of branding and visual language, so let's use the crap out of it:

**Before:**
Screen Shot 2019-11-07 at 9 13 14 PM

**After:**
Screen Shot 2019-11-07 at 9 11 13 PM

**Dropdown open:**
Screen Shot 2019-11-07 at 9 11 19 PM

Note this is a modded version of [one of Greg's designs](https://gallery.io/projects/MCHbtQVoQ2HCZcM1NWkzJjWA/files/MCEJu8Y2hyDScb2TaOwWn_aqcRjXJwESiiI) that was originally done in the context of a new UI view that doesn't exist. I've adapted the sizing, spacing, and the dropdown to fit the current UI while trying to keep the same visual feel. 

(cc @gcomstock)


app/scripts/modules/core/src/managed/ManagedResourceDetailsIndicator.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.ManagedResourceDetailsIndicator .md-logo {
+  background-color: var(--color-black);
+  padding: 18px;
+}
+
+.ManagedResourceDetailsIndicator .summary-message {
+  font-size: 12px;
+  font-weight: 300;
```
erikmunson: @gcomstock I made the judgement call to drop this down to 300 because it felt like the text was throwing off the balance of the button/logo a bit — more than happy to go back on that if you prefer it at a normal weight or are worried about contrast (also only have a high-DPI display to judge with). Here's what normal weight looks like for comparison with the images in the PR description:

Screen Shot 2019-11-07 at 9 22 03 PM

gcomstock: Good call. This banner fits the space well.

gcomstock:
Looks good! Thanks

emjburns:
LGTM

PR #7618, opened 2019-11-08
fix(core/managed): properly diff/patch mutable infra groupings

erikmunson:
tl;dr I mistakenly thought that these filter services generated new grouping data and then pushed that out to consumers, but in reality the new data gets generated and then is used to diff + patch the original objects from the first time the grouping was created. Because I didn't implement this logic in previous PRs, changes to the `isManaged` and `managedResourceSummary` fields were not being picked up until a full reload.

Normally I would say this is a horrible kludge, but I suppose it's just the (gross) way of things while there's still some angular floating around since we need stable object references for those directives/components to behave correctly.

### DIFF
--- a/app/scripts/modules/core/src/cluster/filter/ClusterFilterService.ts
+++ b/app/scripts/modules/core/src/cluster/filter/ClusterFilterService.ts
@@ -473,6 +473,12 @@ export class ClusterFilterService {
         if (oldGroup.entityTags || newGroup.entityTags) {
           oldGroup.entityTags = newGroup.entityTags;
         }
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/cluster/filter/ClusterFilterService.ts
```diff
@@ -473,6 +473,12 @@ export class ClusterFilterService {
         if (oldGroup.entityTags || newGroup.entityTags) {
           oldGroup.entityTags = newGroup.entityTags;
         }
+        if (oldGroup.hasOwnProperty('isManaged') || newGroup.hasOwnProperty('isManaged')) {
```
erikmunson: checking for only the `isManaged` field is sufficient, because the two fields travel together.

PR #7622, opened 2019-11-09
Evaluate variables stage updates

christopherthielen:
Everyone, meet the new Evaluate Variables stage:

Nov-08-2019 16-12-32

Lots of changes here, this is a full rewrite:
- Swap out MapEditor in favor of FormikFormFields.  Note how the `label` prop is a nested `FormikFormField` 🤷‍♂ 
- Added Variable Preview and Validation via `SpelInput`
- Better field validation via Formik
- Added a Preview Execution/Stage picker.  This enables us to evaluate the previews against a previous execution and at the right spot in that execution.
- Prepopulate new stages with a blank row

Future work:

- Rewire autocomplete feature

- There are some backend changes coming from @marchello2000 (SpEL v4) that will cause these variables to get evaluated in order, enabling us to provide clear variable dependency rules.  For example, `host` variable is `myapp.${parameters.region}.${parameters.account}.netflix.net` and `delete_endpoint` is `https://${host}/delete`.  We will need to update this stage and SpelInput to preview all the variables in one shot instead of separate API calls.

Future thoughts around SpEL:
- I'd like push users to put all spel into Evaluate Variables stage
- We should only surface the fancy new spel preview UI in the Evaluate Variables stage, as a 🥕 
- We should have a UI for all stage config fields that allow you to pick a variable from a dropdown list
- We should have a clippy mascot that says "It looks like you have inline SpEL in your stage config! Would you like to move them to an Evaluate Variables stage?"
- FormikFormField and FormField components should be responsible for autodetecting SpEL expressions in values and rendering a SpelInput.  This would enable spel support globally.
- FormikFormField and FormField components should render a simple toggle that allows users to switch from (for example) a SelectInput to a VariablePicker input (which doesn't exist yet)

There are some mockups regarding these future thoughts done by @gcomstock available here: https://gallery.io/projects/MCHbtQVoQ2HCZcM1NWkzJjWA/files/MCEvKPS5RTlTG6LlhmLflRIS

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
@@ -2,21 +2,27 @@ import * as React from 'react';
 import { countBy } from 'lodash';
 import { FieldArray } from 'formik';
 
+import { IStage } from 'core/domain';
+import { FormikStageConfig, IFormikStageConfigInjectedProps, IStageConfigProps } from 'core/pipeline';
 import {
   FormValidator,
   FormikFormField,
   ILayoutProps,
+  IStageForSpelPreview,
   IValidator,
+  LayoutContext,
   LayoutProvider,
+  Markdown,
   SpelInput,
   StandardFieldLayout,
   TextInput,
   Tooltip,
   ValidationMessage,
   errorMessage,
+  useIsMountedRef,
 } from 'core/presentation';
-import { IStage } from 'core/domain';
-import { FormikStageConfig, IFormikStageConfigInjectedProps, IStageConfigProps } from 'core/pipeline';
+
+import { ExecutionAndStagePicker, IExecutionAndStagePickerProps } from './ExecutionAndStagePicker';
 
 import './EvaluateVariablesStageConfig.less';
 
@@ -26,16 +32,16 @@ export interface IEvaluatedVariable {
 }
 
 const variableNameValidator: IValidator = (val: string, label: string) =>
-  !val.match(/^[a-zA-Z0-9_]+$/) && errorMessage(`${label} should consist only of letters, numbers, or underscore`);
+  !val.match(/^[a-zA-Z_][a-zA-Z0-9_]+$/) && errorMessage(`${label} should consist only of letters, numbers, or underscore`);
 
-const getDuplicateKeyValidator = (variables: IEvaluatedVariable[]) => {
+const duplicateKeyValidatorFactory = (variables: IEvaluatedVariable[] = []) => {
   const keyCounts = countBy(variables.map(x => x.key), x => x);
   return (key: string) => keyCounts[key] > 1 && `Duplicate key '${key}'`;
 };
 
 export function validateEvaluateVariablesStage(stage: IStage) {
   const formValidator = new FormValidator(stage);
-  const duplicateKeyValidator = getDuplicateKeyValidator(stage.variables);
+  const duplicateKeyValidator = duplicateKeyValidatorFactory(stage.variables);
   formValidator.field('variables').withValidators(
     formValidator.arrayForEach(item => {
       item
@@ -48,8 +54,32 @@ export function validateEvaluateVariablesStage(stage: IStage) {
   return formValidator.validateForm();
 }
 
+function PreviewConfiguration(props: IExecutionAndStagePickerProps) {
+  return (
+    <div style={{ marginLeft: '16px', marginRight: '16px' }}>
+      <ValidationMessage
+        type={'info'}
+        message={
+          <>
+            <p className="bold">Variable Previews</p>
+
+            <ExecutionAndStagePicker {...props} />
+          </>
+        }
+      />
+    </div>
+  );
+}
+
 export function EvaluateVariablesStageConfig(props: IStageConfigProps) {
   const { application, stage, pipeline, updateStage } = props;
+  const [chosenStage, setChosenStage] = React.useState({} as IStageForSpelPreview);
+
+  const helpMessage =
+    'Define one or more variables by assigning a **name** _(string)_ and a **value** ' +
+    '_([SpEL Expression](https://www.spinnaker.io/guides/user/pipeline/expressions/))_. ' +
+    'The evaluated variables can be used in downstream stages, referencing them by name.';
+
   return (
     <FormikStageConfig
       application={application}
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -1,48 +1,232 @@
...
-      <div className="form-horizontal">
-        <StageConfigField label="Variables to evaluate">
-          <MapEditor
-            model={variablesObject}
-            allowEmpty={true}
-            onChange={(v: any) => this.mapChanged('variables', v)}
-            pipeline={pipeline}
-            valueCanContainSpel={true}
+const variableNameValidator: IValidator = (val: string, label: string) =>
+  !val.match(/^[a-zA-Z0-9_]+$/) && errorMessage(`${label} should consist only of letters, numbers, or underscore`);
```
marchello2000: there are 1 or 2 pipelines that `-` but prob makes sense to not allow it. Also, you should make that variable name doesn't start with a number - that will cause subsequent usages of that variable to fail
```suggestion
  !val.match(/^[a-zA-Z_][a-zA-Z0-9_]+$/) && errorMessage(`${label} should consist only of letters, numbers, or underscore`);
```

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -1,48 +1,232 @@
...
+        }
+      }}
+      touched={touchedOverride}
+      label={variableNameInputAsLabel}
+      actions={actions}
+      fastField={false}
+      input={inputProps => (
+        <SpelInput
+          {...inputProps}
+          placeholder="Variable value, e.g. ${trigger.properties.myproperty}"
```
marchello2000: nit: i would prefer a better/more real example, eg.
```suggestion
          placeholder="Variable value, e.g. ${trigger.buildInfo.number}"
```

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/ExecutionAndStagePicker.tsx
```diff
@@ -0,0 +1,144 @@
...
+    setStageId(exactStage || findCloseStageFromExecution(pipeline, execution, pipelineStage));
+  }, [pipeline, pipelineStage, execution]);
+
+  // When the chosen execution or stage changes, notify the parent
+  React.useEffect(() => {
+    onChange({ executionLabel: executionLabel(execution), executionId: execution && execution.id, stageId });
+  }, [execution, stageId]);
+
+  const stageOptions: Option[] = stages ? stages.map(x => ({ label: x.name, value: x.id })) : [];
+  const executionOptions: Array<Option<IExecution>> = executions ? executions.map(value => ({ value })) : [];
```
marchello2000: what if there are no prior executions? is there some helptext/indication that preview won't work

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -1,48 +1,155 @@
...
+function EvaluateVariablesStageForm(props: IFormikStageConfigInjectedProps) {
+  const { formik } = props;
+
+  React.useEffect(() => {
+    const { variables = [] } = props.formik.values;
+    const initialTouched = variables.map(() => ({ key: true, value: true }));
+    formik.setTouched({ variables: initialTouched } as any);
+  }, []);
+
+  const [deleteCount, setDeleteCount] = React.useState(0);
```
maggieneterval: Would it be possible to generate and persist some UUID for each variable so we don't need this? :pray:

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/ExecutionAndStagePicker.tsx
maggieneterval: +1 to some help text when there are no prior executions if possible!

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -1,48 +1,232 @@
...
-    return (
-      <div className="form-horizontal">
-        <StageConfigField label="Variables to evaluate">
-          <MapEditor
-            model={variablesObject}
-            allowEmpty={true}
-            onChange={(v: any) => this.mapChanged('variables', v)}
-            pipeline={pipeline}
-            valueCanContainSpel={true}
+const variableNameValidator: IValidator = (val: string, label: string) =>
```
caseyhebebrand: If there are other use cases for alphanumeric, snake case string validation then we should move this to the general validator file.

erikmunson: +1 honestly even if there aren't use cases elsewhere we can think of off the top of our heads right now, I'd like to see this in the core set of validators as it's a pretty generic requirement (might even be applicable to things like app/stack/detail on infra?)

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
```diff
@@ -1,48 +1,232 @@
...
+        .withValidators(variableNameValidator, duplicateKeyValidator);
+      item.field('value', 'Expression').required();
+    }),
+  );
+  return formValidator.validateForm();
+}
+
+function PreviewConfiguration(props: IExecutionAndStagePickerProps) {
+  return (
+    <div style={{ marginLeft: '16px', marginRight: '16px' }}>
```
erikmunson: I'd like to see stuff like this become `sp-margin-l-xaxis` so we can keep spacing a bit less loosey goosey. You happened to pick the value we use for those classes in this case, but still :smile:

app/scripts/modules/core/src/pipeline/config/stages/evaluateVariables/EvaluateVariablesStageConfig.tsx
erikmunson: theoretically we don't even need to persist it, as long as it's stable per-mount.

PR #7624, opened 2019-11-11

sanopsmx:
…stage #4792

Enhanced stage.html to

1) Get the roles of the application.
2) Display the roles of the application as a list only if the stage is Manual Judgment.

Enhanced stage.module.js to

1) Fetch the permissions of the application from the gate application url.
2) populate the list with only the roles of the application with no duplicates.

1) Fetch the permissions of the application from the gate application url.

1) Fetch the roles of the application, stage and the user during the execution.
2) Iterate through each of the user role to check if the role exists in the stage and application.
3) If yes/no, enable/disable the continue button.
4) Display the instruction that User does not have permissions to continue

### DIFF
--- a/app/scripts/modules/core/src/application/service/ApplicationReader.ts
+++ b/app/scripts/modules/core/src/application/service/ApplicationReader.ts
@@ -41,6 +41,17 @@ export class ApplicationReader {
       });
   }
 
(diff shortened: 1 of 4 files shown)


sanopsmx:
I am closing this Pull Request. I am raising a new one.

PR #7630, opened 2019-11-12
feat(provider/aws): Lambda function target support for AWS ALB

sidmuls:
This change enables creating Target group with lambda function as target type.
This change also includes fixes for consistency in `tags` and `targetGroups` fields. in Function definition classes.


app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
```diff
@@ -29,6 +29,7 @@ export interface IAmazonFunction extends IFunction {
     subnetIds: [];
     vpcId: string;
   };
+  targetGroups: string | string[];
```
christopherthielen: Why is this typed as`string | string[]`?  Why not just type these as `string[]`?

app/scripts/modules/amazon/src/loadBalancer/loadBalancer.transformer.ts
```diff
@@ -368,6 +368,7 @@ export class AwsLoadBalancerTransformer {
               stickinessEnabled: targetGroup.attributes['stickiness.enabled'] === 'true',
               stickinessType: targetGroup.attributes['stickiness.type'],
               stickinessDuration: Number(targetGroup.attributes['stickiness.lb_cookie.duration_seconds']),
+              multiValueHeadersEnabled: targetGroup.attributes['multiValueHeaders.enabled'] === 'true',
```
christopherthielen: should this be `'multiValueHeadersEnabled'`?

app/scripts/modules/amazon/src/function/configure/FunctionSettings.tsx
```diff
@@ -20,25 +20,25 @@ export class FunctionSettings extends React.Component<IFunctionSettingsProps>
...
+        <FormikFormField
+          name="timeout"
+          label="Timeout (seconds)"
+          help={<HelpField id="aws.functionBasicSettings.timeout" />}
+          input={props => <NumberInput {...props} min={1} max={900} />}
+        />
+        <FormikFormField
+          name="targetGroups"
+          label="Target Group Name"
+          help={<HelpField id="" />}
```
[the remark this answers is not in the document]
sidmuls: removed. Change will appear in the next push.

app/scripts/modules/amazon/src/domain/IAmazonFunction.ts
sidmuls: Currently we support adding only one target group instead of a list of target groups. Clouddriver expects it to be a string in Create/Update operation.
However, while fetching a function's details, AWS Lambda returns a list of string for the `targetGroups` field. 
Hence the type `string | string[]`

app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx
```diff
@@ -86,41 +86,42 @@ export class TargetGroups extends React.Component<ITargetGroupsProps, ITargetGro
...
 
-      ['port', 'healthCheckInterval', 'healthyThreshold', 'unhealthyThreshold'].forEach(key => {
-        const err = spelNumberCheck(targetGroup[key]);
-        if (err) {
-          tgErrors[key] = err;
-        }
-      });
-
+      if (targetGroup.targetType !== 'lambda') {
+        ['healthCheckInterval', 'healthyThreshold', 'unhealthyThreshold'].forEach(key => {
```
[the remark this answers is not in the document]
sidmuls: Ohh that was by mistake. Added it back.

app/scripts/modules/amazon/src/loadBalancer/loadBalancer.transformer.ts
sidmuls: apologies!
it should be lambda.multiValueHeaders.enabled
that is the name of the flag used by AWS Lambda
I have fixed it and pushed the change
thank you so much for pointing it out
that was a great catch!

caseyhebebrand:
@chris

> @caseyhebebrand note that this PR is modifying the `TargetGroups` component and your PR will have conflicts. I think we should merge this Lambda PR first, then redo the changes from this PR into the newer `TargetGroups.tsx` component.

Sounds good! I probably won't have a chance to get mine merged for the next couple days, so this works out.

PR #7631, opened 2019-11-12
fix(azure,gce): Fix typo breaking bake stage

alanmquach:
Somehow ended up with the string `SETTINGS.feature.roscoMode` duplicated, probably while copy/pasting #7564 to other providers resulting in 
```
SETTINGS.feature.roscoModeSETTINGS.feature.roscoMode
```
Which causes a javascript exception trying to access the second `feature` because `SETTINGS.feature.roscoModeSETTINGS` is likely never defined.


maggieneterval:
@spinnakerbot 🍒 ⛏ 1.17

maggieneterval:
@spinnakerbot cherry-pick 1.17

PR #7632, opened 2019-11-13
feat(gcb): Ability to invoke existing GCB triggers

afcastano:
Added ability to call existing GCB triggers.
This is part of: spinnaker/spinnaker#5076

Added a new `Build Definition Source` to the `GoogleCloudBuild` stage.

When this type is selected, new set of fields are shown:

- Trigger Name: A list of triggers populated when `GoogleCloudBuild` account changes.
- Trigger Type: It has to be one of `commitSHA`, `tagName` or `branchName`. Opted for a dropdown list because google api will fail if we send more than one trigger type.
- Value: The value for the trigger type.

<kbd></kbd>

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
@@ -16,8 +16,8 @@ import {
   yamlDocumentsToString,
   YamlEditor,
   TextInput,
+  IGcbTrigger,
 } from '@spinnaker/core';
-import { IGcbTrigger } from 'core/domain';
 
 export enum buildDefinitionSources {
   ARTIFACT = 'artifact',
@@ -25,11 +25,17 @@ export enum buildDefinitionSources {
   TRIGGER = 'trigger',
 }
 
+export enum triggerType {
+  BRANCH = 'branchName',
+  TAG = 'tagName',
+  COMMIT = 'commitSha',
+}
+
 interface IGoogleCloudBuildStageFormProps {
   googleCloudBuildAccounts: string[];
   gcbTriggers: IGcbTrigger[];
   updatePipeline: (pipeline: IPipeline) => void;
-  fetchGcbTriggers: (account: String) => void;
+  fetchGcbTriggers: (account: string) => void;
 }
 
 interface IGoogleCloudBuildStageFormState {
@@ -54,11 +60,10 @@ export class GoogleCloudBuildStageForm extends React.Component<
     this.props.formik.setFieldValue('buildDefinition', buildDefinition);
   };
 
-  private onAccountChange = (accountOption: Option) => {
+  private onAccountChange = (accountOption: Option<string>) => {
     const account = accountOption.value;
     this.props.formik.setFieldValue('account', account);
-    // OptionValues can also be number or boolean, but in this case, account is always a string.
-    this.props.fetchGcbTriggers(account as string);
+    this.props.fetchGcbTriggers(account);
   };
 
   private getAccountOptions = (): Array<Option<string>> => {
@@ -77,9 +82,9 @@ export class GoogleCloudBuildStageForm extends React.Component<
 
   private getTriggerTypes = (): Array<Option<string>> => {
     return [
-      { value: 'branchName', label: 'Branch name' },
-      { value: 'tagName', label: 'Tag Name' },
-      { value: 'commitSha', label: 'Commit SHA' },
+      { value: triggerType.BRANCH, label: 'Branch name' },
+      { value: triggerType.TAG, label: 'Tag Name' },
+      { value: triggerType.COMMIT, label: 'Commit SHA' },
     ];
   };
 
@@ -95,7 +100,7 @@ export class GoogleCloudBuildStageForm extends React.Component<
     this.props.formik.setFieldValue('buildDefinitionSource', e.target.value);
   };
 
-  private onTriggerTypeValueChange = (e: any): void => {
+  private onTriggerValueChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
     const stage = this.props.formik.values;
     if (stage.triggerType) {
       const path = `repoSource.${stage.triggerType}`;
@@ -109,17 +114,17 @@ export class GoogleCloudBuildStageForm extends React.Component<
 
   private onTriggerTypeChange = (selectedType: Option) => {
     this.props.formik.setFieldValue('triggerType', selectedType.value);
-    this.props.formik.setFieldValue('repoSource', undefined);
+    this.props.formik.setFieldValue('repoSource', null);
   };
 
-  private getTriggerTypeValue = () => {
+  private getTriggerValue = (): string => {
     const stage = this.props.formik.values;
     if (stage.triggerType) {
       const path = `repoSource.${stage.triggerType}`;
       return get(stage, path);
     }
 
-    return undefined;
+    return '';
   };
 
   private setArtifactId = (expectedArtifactId: string): void => {
@@ -196,7 +201,7 @@ export class GoogleCloudBuildStageForm extends React.Component<
           />
         )}
         {stage.buildDefinitionSource === buildDefinitionSources.TRIGGER && (
-          <React.Fragment>
+          <>
             <StageConfigField label="Trigger Name">
               <Select
                 clearable={false}
@@ -217,11 +222,11 @@ export class GoogleCloudBuildStageForm extends React.Component<
               <TextInput
                 type="text"
                 className="form-control"
-                onChange={this.onTriggerTypeValueChange}
-                value={this.getTriggerTypeValue()}
+                onChange={this.onTriggerValueChange}
+                value={this.getTriggerValue()}
               />
             </StageConfigField>
-          </React.Fragment>
+          </>
         )}
       </>
     );
--- a/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
@@ -15,16 +15,21 @@ import {
   StageConfigField,
   yamlDocumentsToString,
   YamlEditor,
+  TextInput,
 } from '@spinnaker/core';
+import { IGcbTrigger } from 'core/domain';
 
 export enum buildDefinitionSources {
   ARTIFACT = 'artifact',
   TEXT = 'text',
+  TRIGGER = 'trigger',
 }
 
 interface IGoogleCloudBuildStageFormProps {
   googleCloudBuildAccounts: string[];
+  gcbTriggers: IGcbTrigger[];
   updatePipeline: (pipeline: IPipeline) => void;
+  fetchGcbTriggers: (account: String) => void;
 }
 
 interface IGoogleCloudBuildStageFormState {
@@ -50,7 +55,10 @@ export class GoogleCloudBuildStageForm extends React.Component<
   };
 
   private onAccountChange = (accountOption: Option) => {
-    this.props.formik.setFieldValue('account', accountOption.value);
+    const account = accountOption.value;
+    this.props.formik.setFieldValue('account', account);
+    // OptionValues can also be number or boolean, but in this case, account is always a string.
+    this.props.fetchGcbTriggers(account as string);
   };
 
   private getAccountOptions = (): Array<Option<string>> => {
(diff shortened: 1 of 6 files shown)


afcastano:
Hi @ezimanyi , @maggieneterval 
This is the final part of https://github.com/spinnaker/spinnaker/issues/5076 😅.
As discussed in the issue, I opted for adding a new `Build Definition Source` instead of a new stage.

I have some questions though:
1. I don't know if I have to do anything extra for these new fields to accept `SpEL` expressions.
2. It's not clear to me how can I bubble up to the user errors that are thrown from the Google API. These could be quite common as we don't have a way to verify in the UI that the branch or tag exist for example.

Thank you for taking time to review these PR. Any help would be much appreciated.

maggieneterval:
@afcastano thanks so much for the PR and apologies for the slow followup, we've been traveling prior to Spinnaker Summit in San Diego. I'll take a closer look at this on Monday and address your (very good) questions, but overall looks great! 💯

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -15,16 +15,21 @@ import {
   StageConfigField,
   yamlDocumentsToString,
   YamlEditor,
+  TextInput,
 } from '@spinnaker/core';
+import { IGcbTrigger } from 'core/domain';
```
maggieneterval: Same comment as above -- let's add this import to the existing `@spinnaker/core` import.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -50,7 +55,10 @@ export class GoogleCloudBuildStageForm extends React.Component<
   };
 
   private onAccountChange = (accountOption: Option) => {
-    this.props.formik.setFieldValue('account', accountOption.value);
+    const account = accountOption.value;
+    this.props.formik.setFieldValue('account', account);
+    // OptionValues can also be number or boolean, but in this case, account is always a string.
+    this.props.fetchGcbTriggers(account as string);
```
maggieneterval: We won't need the `as` operator if we parameterize the `Option` type on line 57:

```
private onAccountChange = (accountOption: Option<string>) => {
```

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -60,17 +68,60 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
       { value: buildDefinitionSources.ARTIFACT, label: 'Artifact' },
+      { value: buildDefinitionSources.TRIGGER, label: 'Trigger' },
     ];
   };
 
   private onSourceChange = (e: any): void => {
     this.props.formik.setFieldValue('buildDefinitionSource', e.target.value);
   };
 
+  private onTriggerTypeValueChange = (e: any): void => {
```
maggieneterval:
```suggestion
  private onTriggerTypeValueChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
```

Looks like a bunch of the old code used `any` as the event type so totally understandable that you would copy this...but let's be as strict as possible with typing for new code :smile:

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -60,17 +68,60 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
+    const stage = this.props.formik.values;
+    if (stage.triggerType) {
+      const path = `repoSource.${stage.triggerType}`;
+      this.props.formik.setFieldValue(path, e.target.value);
+    }
+  };
+
+  private onTriggerChange = (selectedTrigger: Option) => {
```
maggieneterval:
```suggestion
  private onTriggerChange = (selectedTrigger: Option<string>) => {
```

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -60,17 +68,60 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
+    this.props.formik.setFieldValue('triggerId', selectedTrigger.value);
+  };
+
+  private onTriggerTypeChange = (selectedType: Option) => {
```
maggieneterval:
```suggestion
  private onTriggerTypeChange = (selectedType: Option<string>) => {
```

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -60,17 +68,60 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
+    this.props.formik.setFieldValue('triggerType', selectedType.value);
+    this.props.formik.setFieldValue('repoSource', undefined);
+  };
+
+  private getTriggerTypeValue = () => {
```
maggieneterval:
```suggestion
  private getTriggerTypeValue = (): string => {
```

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -60,17 +68,60 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
+    const stage = this.props.formik.values;
+    if (stage.triggerType) {
+      const path = `repoSource.${stage.triggerType}`;
+      return get(stage, path);
+    }
+
+    return undefined;
```
maggieneterval:
```suggestion
    return '';
```

Since this is setting the input value, let's default to an empty string.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -60,17 +68,60 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
+  private onTriggerTypeChange = (selectedType: Option) => {
+    this.props.formik.setFieldValue('triggerType', selectedType.value);
+    this.props.formik.setFieldValue('repoSource', undefined);
+  };
+
+  private getTriggerTypeValue = () => {
+    const stage = this.props.formik.values;
+    if (stage.triggerType) {
+      const path = `repoSource.${stage.triggerType}`;
+      return get(stage, path);
```
maggieneterval: Do we need lodash `get` here if we've already checked `stage.triggerType`?

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -144,6 +195,34 @@ export class GoogleCloudBuildStageForm extends React.Component<
             updatePipeline={this.props.updatePipeline}
           />
         )}
+        {stage.buildDefinitionSource === buildDefinitionSources.TRIGGER && (
+          <React.Fragment>
```
maggieneterval: Feel free to ignore this, but just a heads-up that you can just use the fragment shorthand here -- only time you should need explicit `React.Fragment` is if you need to key the element when it's spit out as a result of mapping some collection.
```suggestion
          <>
```

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -144,6 +195,34 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
+                onChange={this.onTriggerTypeChange}
+                options={this.getTriggerTypes()}
+                value={stage.triggerType}
+              />
+            </StageConfigField>
+            <StageConfigField label="Value">
+              <TextInput
+                type="text"
+                className="form-control"
+                onChange={this.onTriggerTypeValueChange}
```
maggieneterval:
```suggestion
                onChange={this.onTriggerValueChange}
```

Might be easier to read if we remove `type` from the name of the methods related to the value?

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -60,17 +68,60 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
     ];
   };
 
   private onSourceChange = (e: any): void => {
     this.props.formik.setFieldValue('buildDefinitionSource', e.target.value);
   };
 
+  private onTriggerTypeValueChange = (e: any): void => {
+    const stage = this.props.formik.values;
+    if (stage.triggerType) {
```
maggieneterval: If a user inputs a value before selecting a type, will this just fail to work without any feedback? Can we default to the first type and add `clearable=false` to the type `Select` to avoid this potentially confusing behavior?

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -60,17 +68,60 @@ export class GoogleCloudBuildStageForm extends React.Component<
...
+    }
+  };
+
+  private onTriggerChange = (selectedTrigger: Option) => {
+    this.props.formik.setFieldValue('triggerId', selectedTrigger.value);
+  };
+
+  private onTriggerTypeChange = (selectedType: Option) => {
+    this.props.formik.setFieldValue('triggerType', selectedType.value);
+    this.props.formik.setFieldValue('repoSource', undefined);
```
maggieneterval: Let's avoid assignment to `undefined` -- can we instead set to `triggerType: ''` or whatever a sane default would be here? Or if you don't think there should be one, let's use `null` to be explicit.

afcastano:
Hello @maggieneterval , 
Thank you very much for this detailed review. It looks good.
I'll address your comments and ping you back when ready.
Cheers,

afcastano:
@maggieneterval With regards to this:

> I think it's acceptable at this point for users to need to edit the stage JSON directly in order to use SpEL.

I don't completely understand how editing the JSON would enable users to define `SpEL` variables?
How is that different than writting the `SpEL` directly in one of the form fields, say for example the `Value` field for the `Branch trigger type`?

Also, Could you maybe point me to an example or any documentation explaining how to set up SpEL  on the JSON document for the stage?

Cheers

maggieneterval:
> I don't completely understand how editing the JSON would enable users to define SpEL variables?

Sorry, my comment was super confusing: for a normal text input, users can just type in SpEL inline by wrapping their expression: `${...}` (docs [here](https://www.spinnaker.io/guides/user/pipeline/expressions/)). In the case of a dropdown with predefined options (in this case, the Trigger Name and Trigger Type dropdowns), users would need to click "Edit Stage as JSON" and add the expression manually there.

> I'll address your comments and ping you back when ready.

Awesome, thanks! One other potential issue that came to mind is that since we are nesting the trigger value under a key that depends on the selected trigger type, we'll want to make sure we're handling cases where the trigger type changes after a trigger value has already been entered: should the value be cleared out, or persisted? Since we're nesting the value under a key that depends on the selected trigger type, we'll want to make sure to null out the previously-selected type key (or just not nest the value this way, although I'm guessing you've already committed the Orca implementation that depends on that structure?).

Have a great weekend!

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
afcastano: We do need it as `repoSource` might be empty if it is the first time the form is filled up.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
afcastano: Cool... has been a while since I wrote react code. Thanks for the suggestion.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
afcastano: If the user inputs a value without selecting a trigger, there should be an error displayed in the form. I implemented this here: https://github.com/spinnaker/deck/pull/7632/files#diff-c24618fef65c05c42e4c3a58f57f3b10R9
However, your suggestion makes sense. Will implement it.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
afcastano: I want to explicitly delete the repo source when a trigger type is selected since the GCB api will fail if there is more than one value in repo source. It could be one of `branchName, tagName, commitSha`.
I will set `null` on that value.
Thanks for the suggestion.

afcastano:
Hi @maggieneterval , with regards to:

> we'll want to make sure we're handling cases where the trigger type changes after a trigger value has already been entered: should the value be cleared out, or persisted? 

At the moment I am clearing out the value when the trigger type changes. It didn't make sense at the moment to persist that value because it is related to completely different subjects `Branch, Tag or Commit`.

I am doing this here: https://github.com/spinnaker/deck/pull/7632/files#diff-19e99516876a7c5272f898ac284ca493R112

afcastano:
Hello @maggieneterval , I have implemented your suggestions. Please let me know what do you think.

Cheers!
=)

afcastano:
With regards to this:

> (or just not nest the value this way, although I'm guessing you've already committed the Orca implementation that depends on that structure?).

The reason these values are nested, is because the GCB api expects it that way. I thought changing the structure somewhere in the middle could be more confusing. The structure stored in the stage is the same structure sent to the GCB api.
=)

maggieneterval:
@afcastano thanks so much for addressing all the feedback, looks great! Sorry for the slow follow-up on my end, I was OOO all last week. I'm just going to take one final pass through this tomorrow morning with fresh eyes before merging!

afcastano:
No worries at all. Let me know how it goes. :)

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -15,16 +15,27 @@ import {
...
+  BRANCH = 'branchName',
+  TAG = 'tagName',
+  COMMIT = 'commitSha',
 }
 
 interface IGoogleCloudBuildStageFormProps {
   googleCloudBuildAccounts: string[];
+  gcbTriggers: IGcbTrigger[];
   updatePipeline: (pipeline: IPipeline) => void;
+  fetchGcbTriggers: (account: String) => void;
```
maggieneterval:
```suggestion
  fetchGcbTriggers: (account: string) => void;
```

maggieneterval: Just one tiny change and this is ready to merge!

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
afcastano: oops,
done! =)

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
afcastano: Done... ! =)

afcastano:
@maggieneterval Done... =)
Thanks

PR #7633, opened 2019-11-13

sanopsmx:
…#4792

Enhanced stage.html to

Display the roles of the application as a list only if the stage is Manual Judgment.
Enhanced stage.module.js to

If yes/no, enable/disable the continue button.
Display the instruction that User does not have permissions to continue


sanopsmx:
Closing this PR. Will raise a new PR.

PR #7636, opened 2019-11-13

sanopsmx:
…#4792

Enhanced stage.html to

Display the roles of the application as a list only if the stage is Manual Judgment.
Enhanced stage.module.js to

If yes/no, enable/disable the continue button.
Display the instruction that User does not have permissions to continue.

### DIFF
--- a/app/scripts/modules/core/src/application/service/ApplicationReader.ts
+++ b/app/scripts/modules/core/src/application/service/ApplicationReader.ts
@@ -41,6 +41,17 @@ export class ApplicationReader {
       });
   }
 
+  public static getApplicationPermissions(applicationName: string): IPromise < any > {
+        return API.one('applications', applicationName)
+            .withParams({
+                expand: false
+            })
+            .get()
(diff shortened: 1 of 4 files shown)


sanopsmx:
Waiting for the feedback.

christopherthielen:
I assume this is waiting on https://github.com/spinnaker/governance/pull/80

maggieneterval:
> I assume this is waiting on spinnaker/governance#80

Yes, I believe so -- shall we close this for now? For changesets like this blocked by RFC approval (and coupled to a backend implementation that might change based on feedback on the RFC), would probably make sense to hold off on submitting the PR until the RFC has been merged, both to 1) encourage consensus on the RFC prior to implementation and 2) de-clutter Deck's open PRs. @christopherthielen what do you think?

maggieneterval:
@sanopsmx going to close this for now since this has been sitting for a few months while we wait for approval of your RFC; please feel free to re-open once the RFC is approved!

PR #7638, opened 2019-11-13
test(travis): Just testing a travis script DO NOT MERGE

alanmquach:
DO NOT MERGE.

Just testing that can perform a merge properly to minimize the diff it analyzes so that it can analyze ONLY the actual changes excluding the divergence from master (PR target branch).

Will also be checking that will fail with merge conflicts.


alanmquach:
Sweet. Looks like GitHub gives travis a commit that is already merged.

And travis will not run if there is a conflict.

PR #7644, opened 2019-11-19
fix(amazon): Copy capacity from current server group does not persist

caseyhebebrand:
When editing a Titus deploy stage and setting the capacity to "Copy the capacity form the current server group" in advanced mode, the state does not persist after saving + reloading. This PR fixes this so that the situation below does not happen. 

Example: 
Screen Shot 2019-10-28 at 12 34 50 PM
Screen Shot 2019-10-28 at 12 35 05 PM


caseyhebebrand:
> I don't have a ton of context about this area of the code, but AFAICT the diff seems reasonable 👍

TL;DR The conditional was checking an ephemeral UI toggle-state that reset when the modal was closed. It should look at toggle variable _and_ persisted value for the initial render.

PR #7658, opened 2019-11-26
feat(slack): Introduce slack support channel selector

caseyhebebrand:
A user can select a Slack support channel when creating an application, or editing an application. This will render in Application Attributes as a link to the channel. This is an optional field which is configured by (1) enabling the feature in deck settings and (2) providing a slack token to gate.

Screen Shot 2019-11-25 at 1 38 48 PM
Screen Shot 2019-11-25 at 1 39 13 PM


app/scripts/modules/core/src/application/modal/createApplication.modal.controller.js
```diff
@@ -129,5 +129,8 @@ module.exports = angular
         }
         this.createApplicationForTests();
       };
+
+      // Enables setting the attributes as a callback in react components
+      this.setAttribute = (name, value) => (this.application[name] = value);
```
christopherthielen: doesn't this need to be wrapped in `$applyAsync` for angular to pick it up?

app/scripts/modules/core/src/slack/SlackChannelSelector.tsx
```diff
@@ -0,0 +1,66 @@
...
+  callback: (name: string, value: any) => void;
+}
+
+export interface ISlackChannelSelectorState {
+  channels: ISlackChannel[];
+  selected: ISlackChannel;
+  loading: boolean;
+}
+
+export default class SlackChannelSelector extends React.Component<
```
christopherthielen: i'm curious why you chose to use a class component over a hooks component??

app/scripts/modules/core/src/slack/SlackChannelSelector.tsx
```diff
@@ -0,0 +1,66 @@
...
+  };
+
+  public render() {
+    const { channels, loading, selected } = this.state;
+
+    return (
+      <div className="form-group row">
+        <div className="col-sm-3 sm-label-right">Slack Channel</div>
+        <div className="col-sm-9">
+          <VirtualizedSelect
```
christopherthielen: Can this be migrated to `<ReactSelectInput mode="VIRTUALIZED" />`?  I'd like to move towards using a more opinionated set of  components across the board.

app/scripts/modules/core/src/slack/SlackReader.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    name_normalized: 'fakechannel',
+    num_members: 25,
+  },
+];
+
+describe('SlackReader', () => {
+  it('should fetch a list of public Slack channels', () => {
+    spyOn(SlackReader, 'getChannels').and.returnValue(Promise.resolve(mockSlackChannels));
+
+    SlackReader.getChannels().then((channels: ISlackChannel[]) => {
```
christopherthielen: this test isn't really asserting anything useful.  I can replace `SlackReader` reference with an empty stub and the test still passes:

```js
it('should fetch a list of public Slack channels', () => {
    const foo = { getChannels() {} };
    spyOn(foo, 'getChannels').and.returnValue(Promise.resolve(mockSlackChannels));

    foo.getChannels().then((channels: any[]) => {
      expect(foo.getChannels).toHaveBeenCalled();
      expect(channels.length).toEqual(2);
    });
  });
```

The only useful thing to assert that I can think of is if `API.one` got called with `slack/channels`.

yarn.lock
```diff
@@ -502,10 +502,10 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react-select@*", "@types/react-select@^1.2.3":
-  version "1.2.3"
-  resolved "https://registry.yarnpkg.com/@types/react-select/-/react-select-1.2.3.tgz#95b3af27e73d8fd4ba421c3dcfc2ce97f2603f92"
-  integrity sha512-D/A5W0b0SKxdFszwCp3AuVoMNS0+2w5YscFlUr8ue/Fycwm8mlS17H7MwMi37ByvjcgmCZeIK5FWSHIytHBfQg==
+"@types/react-select@^1", "@types/react-select@^1.2.4":
+  version "1.3.4"
```
christopherthielen: Thanks for updating this.  Was there a typing problem that this resolved?  If so, I'd prefer to commit this separately with a nice comment about what this fixes.

halconfig/settings.js
```diff
@@ -177,6 +177,7 @@ window.spinnakerSettings = {
     pipelines: true,
     pipelineTemplates: pipelineTemplatesEnabled,
     roscoMode: true,
+    slack: false,
```
christopherthielen: This would be useful for halyard users, I'm sure.  We can certainly address this in a future PR.

app/scripts/modules/core/src/application/modal/createApplication.modal.controller.js
caseyhebebrand: I haven't had any problem running locally without that.

app/scripts/modules/core/src/slack/SlackChannelSelector.tsx
caseyhebebrand: I think my main hesitation was ensuring that the API was only called one time, rather than after each render. I can refactor to a functional component with `useEffect` & a second parameter.

app/scripts/modules/core/src/slack/SlackChannelSelector.tsx
caseyhebebrand: Yes, did not see this component so I will swap it out.

yarn.lock
caseyhebebrand: Yes it was not recognizing the generics and causing several errors. I will update separately.

app/scripts/modules/core/src/slack/SlackReader.spec.ts
caseyhebebrand: Ok I will think about this more. I was trying to remove the dependency on the angular mocks. I assumed this would still call the API and on a successful response return the values, while on a failed response the test would fail.

halconfig/settings.js
caseyhebebrand: great! I wanted to disable in OSS to start since you also need to have a token.

app/scripts/modules/core/src/application/modal/createApplication.modal.controller.js
caseyhebebrand: Let me know if I am missing something, although I suppose it doesn't hurt to add as a safeguard.

christopherthielen: If it works locally, then 🤷‍♂

app/scripts/modules/core/src/slack/SlackChannelSelector.tsx
christopherthielen: If you refactor to functional component, use the `useData` or `useLatestPromise` hook!

app/scripts/modules/core/src/slack/SlackReader.spec.ts
christopherthielen: you should be able to do something like:
```
    const oneSpy = spyOn(API as any, 'one').and.callThrough();
    const getSpy = spyOn(API as any, 'getFn').and.returnValue(Promise.resolve(mockSlackChannels));
```
then you should be able to 
```
expect(oneSpy).toHaveBeenCalledWith('slack/channels);
```

... I think

app/scripts/modules/core/src/slack/SlackReader.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    num_members: 25,
+  },
+];
+
+describe('SlackReader', () => {
+  it('should fetch a list of public Slack channels', () => {
+    spyOn(SlackReader, 'getChannels').and.callThrough();
+    spyOn(API, 'one')
+      .and.callThrough()
+      .and.returnValue({ getList: () => Promise.resolve(mockSlackChannels) });
```
caseyhebebrand: @christopherthielen This was the best I could do using `callThrough` because I cannot spy on any of the private methods. I think it is valuable enough of a test to show that we have called `API` correctly, while still controlling the response. 

Another option is to utilize angular mocks like the other services, but I would prefer to reduce the dependency on angular as much as possible. 

Another

christopherthielen: Agreed that we should avoid using angular mocks in any new code.  this looks great to me, thanks for updating it!

PR #7660, opened 2019-11-26
chore(react-select): Update react-select type modules

caseyhebebrand:
Updating the react-select modules. Type errors were thrown because it was not recognizing generics.


app/scripts/modules/amazon/src/serverGroup/configure/AmazonImageSelectInput.tsx
```diff
@@ -343,7 +343,10 @@ export class AmazonImageSelectInput extends React.Component<IAmazonImageSelector
             filterOptions={false as any}
             noResultsText={searchNoResultsText}
             options={searchResults}
-            onInputChange={searchInput => this.searchInput$.next(searchInput)}
+            onInputChange={searchInput => {
+              this.searchInput$.next(searchInput);
+              return searchInput;
```
caseyhebebrand: This was a side effect of the update. This function should return a string, while enabling `serchInput$.next(...)` to continue being invoked properly.

PR #7661, opened 2019-11-27
Add modal components

erikmunson:
This introduces a new set of `Modal` components, which are intended to do a couple things:
- Add to our growing set of common components/design system building blocks optimized for Deck (and hopefully replace our usage of `react-bootstrap`'s `<Modal/>`)
- Serve as a testing ground for what it's like to implement and work with CSS Modules in hopes we can learn more about them and decide how to proceed going forward

The component API is somewhat simple (at least right now):
```tsx
const [ showModal, setShowModal ] = useState(false);

// some kind of handler code to decide when to show modal...

<Modal isOpen={showModal} onRequestClose={() => setShowModal(false)}>
  <ModalHeader>The title of this modal is "A Modal"</ModalHeader>
  <ModalBody>
    <div style={{ width: '500px', height: '500px' }}>This is a modal, pretty nifty if you ask me</div>
  </ModalBody>
  <ModalFooter
    primaryActions={
      <div>Wowowow</div> {/* optional, can be any element and appears on the right of the footer*/}
   }
    secondaryActions={
      <div>Coolio</div> {/* optional, can be any element and appears on the left of the footer*/}
   }
  />
</Modal>
```

There's some enter/leave animation, so much like `react-bootstrap` this component needs to be mounted _before_ you actually show it, and unmounted after you're done exiting it from view. Some examples at small + large screen sizes:

**Small screen:**
Screen Shot 2019-11-27 at 2 52 18 PM

**Large screen:**
Screen Shot 2019-11-27 at 2 36 21 PM

And a gif to give an idea of the enter/leave transitions:
modal-enter-leave

Apologies for the somewhat large diff, I didn't feel like it made much sense to pull any of these commits out into their own PRs as they're all pretty interdependent. I've done my best to chunk them into logical commits and recommend going through that way. A lot of the lines are also from stylesheet code which should be somewhat faster to read through.

cc @gcomstock


package.json
```diff
@@ -181,6 +183,7 @@
     "physical-cpu-count": "^2.0.0",
     "postcss-colorfix": "^0.0.4",
     "postcss-loader": "3.0.0",
+    "postcss-nested": "^4.2.1",
```
christopherthielen: is this addition necessary? Can we leverage less to do the same thing? i am not good with computer

app/scripts/modules/core/src/presentation/modal/Modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+  isOpen: boolean;
+  onRequestClose?: () => any;
+  children?: React.ReactNode;
+}
+
+export const Modal = ({ onRequestClose, isOpen, children }: IModalProps) => {
+  useContainerClassNames(isOpen ? [styles.backdropBlurEffect] : []);
+
+  const keydownCallback = ({ keyCode }: KeyboardEvent) => {
+    if (keyCode === 27) {
```
christopherthielen: add ` // esc`

app/scripts/modules/core/src/presentation/modal/Modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+      onRequestClose();
+    }
+  };
+  useEventListener(document, 'keydown', isOpen ? keydownCallback : null);
```
christopherthielen: really slick 👍

app/scripts/modules/core/src/presentation/modal/Modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+  onRequestClose?: () => any;
+  children?: React.ReactNode;
+}
+
+export const Modal = ({ onRequestClose, isOpen, children }: IModalProps) => {
+  useContainerClassNames(isOpen ? [styles.backdropBlurEffect] : []);
+
+  const keydownCallback = ({ keyCode }: KeyboardEvent) => {
+    if (keyCode === 27) {
+      onRequestClose();
```
christopherthielen: should this use `memoizedRequestClose`?

app/scripts/modules/core/src/presentation/modal/Modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+    }
+  };
+  useEventListener(document, 'keydown', isOpen ? keydownCallback : null);
+
+  const memoizedRequestClose = useLatestCallback(onRequestClose);
+  const modalContext = useMemo(() => ({ onRequestClose: memoizedRequestClose }), []);
```
christopherthielen: I believe this `useMemo` is unnecessary?

app/scripts/modules/core/src/presentation/modal/Modal.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <ModalContext.Provider value={modalContext}>
```
christopherthielen: Question.  Thought.  Whatever.  

I'm generally hesitant to add context because it can obfuscate the flow of data.
How would you feel about using a render prop instead?

```js
<Modal render={({ onRequestClose }) => (
  <ModalHeader onClose={onRequestClose} />
)} />

**Comment by christopherthielen** [comment]
One more thought/comment:

This is a component based API, but how do you feel about also supporting an imperative API like `ReactModal.show()`?

**Comment by christopherthielen** [comment]
What actually does the css module loading in webpack?

File: app/scripts/modules/core/src/presentation/colors.module.css [hunk 7]
```diff
@@ -0,0 +1,7 @@
+@value white: #fff;
```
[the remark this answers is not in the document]
erikmunson: I definitely should've left a note about this stuff (same goes for `breakpoints` and `layers`) — my intent here is to experiment with what it would be like to actually use CSS Modules for the full scope of things we do all the way from shared variables/building blocks up to actual components on the screen, so I am in some cases duplicating things or responsibilities that are already in the `styleguide` module. Once we align on a direction for this experimental stuff, the end-goal is to fold this set of variables into whatever becomes the 'paved path' way (whether that's anything like either of the two things we have now remains to be seen).

package.json
erikmunson: In theory we could use less to do pre-processing before we pass things along to get module-ified, but my impression was that we are generally not thrilled with the idea of continuing to use a heavy pre-processor "language" like less/sass going forward at all vs. just using tiny postcss transforms as needed. Perhaps that's something we need to address in a larger discussion.

erikmunson: So far I've taken the approach that we'll have 1) the 'legacy' set of non-module CSS that's usually written in `.less` files, and 2) the new set of module-ified CSS that only uses limited PostCSS transformations (nesting being probably the heaviest I'd want to drop in)

app/scripts/modules/core/src/presentation/modal/Modal.tsx
erikmunson: while it's not _super_ common to need this hook, it's amazing to use when you do need it. Totally different world from the class lifecycle way of doing it.

app/scripts/modules/core/src/presentation/modal/Modal.tsx
erikmunson: it's functionally equivalent to use either one here, because `useEventListener` also uses `useLatestCallback` under the hood so you can write event callbacks without worrying about closure freshness. Perhaps it's a bit pedantic, but I've generally been making an effort to only use memoized/ref'd versions of things in places where it's functionally necessary as a way to communicate meaning/significance to the reader.

app/scripts/modules/core/src/presentation/modal/Modal.tsx
erikmunson: Sadly it is necessary, because context providers re-render whenever the _reference_ of their `value` prop changes ([docs](https://reactjs.org/docs/context.html#caveats)). If you pass in an object literal that gets generated inside render, you end up with a chain reaction where every time the parent component renders you get a second render down the whole subtree underneath the context provider 😢

app/scripts/modules/core/src/presentation/modal/Modal.tsx
erikmunson: Funny enough, I went in circles a couple times re: whether to make the plumbing of the header close behavior something you glue together yourself vs. it being done for you under the hood.
After spending time with these components and thinking about other Modal components, how Formik uses render props vs. context, etc., I came around to thinking this behavior specifically is a good candidate for context. Specifically:

- Much like the wiring up of Formik's `Field` components to Formik state, it's really just "dumb" plumbing/noise that adds no value to the mental model of a consumer — it's going to work exactly the same way, every single time, and the most you as an app developer can really do to modify it is to mis-use it in some broken or sub-optimal way
- Letting the component system do the plumbing itself enforces a bit of separation between app code and the way modals use app code to present things on the screen/respond to inputs. I fully expect we will be layering higher level kinds of modals on top of this base modal component system, and it's entirely possible we'll want to do the same thing we're doing here with `ModalHeader` in other sub-components (dismiss buttons, links or other interactive chunks inside the body, etc.)
- One of the things I've observed about our use of `react-bootstrap` modals is that the interaction mechanics can be inconsistent in ways which are frustrating as an end user (and a little confusing as an engineer). Depending on how people choose to wire up the components and event handlers, sometimes you get `esc` key handling and sometimes not, sometimes you get a close button that does things a different way from the `esc` code path, we even have some modals that don't add close buttons at all (and I think there are some that are hand-built rather than using the `ModalClose` component). I'd love for the component/design systems we build to take the burden and cognitive load of maintaining consistent UX choices away from individuals and codify them into something that we can all confidently use the same way without needing to make the repetitive choices others have made hundreds of times already

After this little journey of self discovery, I think my personal opinion has become that if something:
  1) is intended and expected to be boilerplate plumbing that works the same way every time
  2) doesn't have a major impact on the mental model of the person consuming/reading it
  3) would benefit from being consistently applied
  4) isn't complicated enough in its data flow/state that hiding it is confusing 

then context is a good call. If any of the above aren't true or have sufficiently large asterisks, it's worth letting folks plumb them together manually using something like a render prop. I generally think that matches the approach Formik takes, but also always happy to eat my words if this proves to be annoying / not useful later on. I sincerely think this is a choice that has a low cost to change/walk back, because the refactor of consuming code would be pretty straightforward as long as the Modal components handle memoizing callbacks themselves.

erikmunson:
> This is a component based API, but how do you feel about also supporting an imperative API like ReactModal.show()?

100% yes, I intend to add a near-clone of `ReactModal` in a separate PR. Will likely try to move some of the things using `ReactModal` over to the new thing in a working branch then pull out the common code as a PR, as I was hesitant to bloat this diff and also want to make sure I've actually used the imperative replacement in a real thing before shipping it.

> What actually does the css module loading in webpack?

I was surprised to learn that the standard `css-loader` already does this natively via a couple config flags. I turned those on[ in a previous PR](https://github.com/spinnaker/deck/pull/7650/files#diff-d59466f48cdc8c626ac92e44aed3fc6dR97-R110). That PR also has evidence of some of the not-great tooling kludge I had to deal with to get CSS Modules working outside of the webpack loader story, which I will be writing about in the next day or so.

maggieneterval:
I am super late to the party, but just wanted to say :tada: :medal_sports: :100: :man: :computer: :smile:

maggieneterval:
@erikmunson no good deed goes unpunished -- I think this change is breaking the Jenkins build on master: `Cannot find module './ModalFooter.module.css'.` Are you seeing that locally as well?

erikmunson:
yesssss.... I am working on getting that stabilized right now. The stupid race condition I was hoping to avoid apparently decided to show up _after_ merging.

ttomsu:
Thoughts on reverting this until the race condition is fixed?

erikmunson:
AFAICT this is going to re-stabilize things: https://github.com/spinnaker/deck/pull/7674

It'll be merged momentarily.

PR #7662, opened 2019-11-29
Plugins shared libraries

christopherthielen:
This exposes a small set of shared libraries onto the spinnaker global object at `window.spinnaker.plugins.sharedLibraries`.  These libs can then be consumed by dynamically loaded plugins.


christopherthielen:
@bpowell @link108

christopherthielen:
Here's a proof of concept deck plugin that uses these shared libraries:
https://github.com/christopherthielen/rollup-starter-lib

Screen Shot 2019-11-29 at 1 41 26 PM

app/scripts/modules/core/src/plugins/pluginSupport.ts
```diff
@@ -0,0 +1,38 @@
+import * as lodash from 'lodash';
+import * as react from 'react';
+import * as react_dom from 'react-dom';
```
louisjimenez: I'm curious about the expected behavior for the bundler and bundle size here. Will each lib in full be put into the bundle so that it is accessible to dynamically loaded plugins via the global object?

christopherthielen: yes, these whitelisted libraries will be bundled in full into one of the initial deck bundles (which are loaded immediately on startup).  Then they're exposed on the global object for use by plugins.

app/scripts/modules/core/src/plugins/pluginSupport.ts
```diff
@@ -0,0 +1,38 @@
...
+import * as spinnaker_core from '@spinnaker/core';
+
+export const pluginSupport = {
+  sharedLibraryNames: [
+    'lodash',
+    'react',
+    'react-dom',
```
erikmunson: I see you have `@spinnaker/core` noted for removal/scoping down, but is it also fair to say `react-dom` will be unnecessary once we're past the experimental stage? Unless there's a great reason to have plugins calling `render()` and `createPortal()` on arbitrary DOM? I guess I assumed plugins would be calling explicit extension points that load them into the existing component tree.

app/scripts/modules/core/src/utils/consoleDebug.ts
```diff
@@ -18,6 +28,9 @@ export class ConsoleDebugWindow {
 }
 
 export const DebugWindow = new ConsoleDebugWindow();
+if (window) {
+  window.spinnaker = DebugWindow;
+}
```
erikmunson: what's the context behind this additional assignment vs. the one on L35? Is it mostly just to make use of the new interface augmentation on `Window`? If yes, why keep the other one?

erikmunson:
Oh also — as I was reading your plugin POC code in the other repo, I was thinking it might be nice for plugin devs to extract a rollup plugin that did all the custom externals configuration automatically (e.g. `rollup-plugin-spinnaker-globals`). Might make it easier to evolve the shared libs and other plugin build glue in that folks could just bump their dependency version rather than doing copy-pasta

app/scripts/modules/core/src/utils/consoleDebug.ts
christopherthielen: ack, that's a bug!

christopherthielen:
> I suspect this might be something you're already thinking about / have considered, but it came to mind that we _could_ theoretically scope down `@spinnaker/core` without actually creating a new NPM module

In order for this to work nicely (DX), we'd also need typings that are likewise scoped down.  This would allow plugin devs to import with auto-complete.  However, it seemed like a stopgap solution that might have gotchas, and would require a bunch of extra work that would be thrown away in the future.  Perhaps I'm not seeing a simpler approach though.  LMK

app/scripts/modules/core/src/plugins/pluginSupport.ts
christopherthielen: I was thinking `createPortal` is a legit use case

christopherthielen:
> Oh also — as I was reading your plugin POC code in the other repo, I was thinking it might be nice for plugin devs to extract a rollup plugin that did all the custom externals configuration automatically (e.g. `rollup-plugin-spinnaker-globals`). Might make it easier to evolve the shared libs and other plugin build glue in that folks could just bump their dependency version rather than doing copy-pasta

Yeah, exactly! I want to extract as much of the plugin development experience into an npm dependency.   Ideally this would contain all sorts of opinions including rollup config, eslint rules, etc.

app/scripts/modules/core/src/plugins/pluginSupport.ts
erikmunson: yeah, out of the two `createPortal` feels more legitimate. I'm still somewhat skeptical though — my philosophy/approach with these sorts of things is usually to lock it down and wait to hear what kinds of use cases come up to see if there's a less loosey-goosey way of solving it. Either way I doubt we'll get ourselves into a particularly bad/troublesome situation and I don't have a strong preference. I'm somewhat biased from spending a long time in the world of extensions/third party code.

erikmunson:
> In order for this to work nicely (DX), we'd also need typings that are likewise scoped down.

Already talked about this in person, and I agree the typing issue is a good justification for needing a separate NPM module.

bpowell:
@christopherthielen when we expose out the shared libraries like this, do the plugin projects need to have as `dependencies` like react and react-dom? I saw in the linked project that react and react-dom were listed, but my guess is not used based on the rollup config. I'm wondering if that will cause issues of version mismatch on things if people select a newer version of a library in the plugin code, but use the older library version from spinnaker.

christopherthielen:
>  I want to extract as much of the plugin development experience into an npm dependency.

@bpowell This npm dependency (perhaps a package named `@spinnaker/plugin`?) can transitively provide the common dependencies such as `lodash`, `react`, and `@spinnaker/ui` (or whatever we name the UI package).  This would be similar to kork for back end services.   When a plugin developer updates the `@spinnaker/plugin` package, they will get updated transitive dependencies such as `react`.

PR #7664, opened 2019-11-29
Fixes for lint errors

christopherthielen:
Fixes for eslint rules: `no-case-declarations`, `@spinnaker/strictdi`, `no-useless-escape`, `@typescript-eslint/ban-types` `@typescript-eslint/no-inferrable-types`

PR #7665, opened 2019-11-30
AngularJS StrictDI early warning system

christopherthielen:
This enables [angularjs strictdi mode](https://stackoverflow.com/questions/33494022/what-is-the-benefit-of-angularjs-strict-di-mode) and throws an error _before angularjs bootstraps_ if any injected code is not strictly annotated.

This differs from the current behavior where an error would be thrown at runtime when the injected code is attempted to be _used_.  This difference means non-compliant code can be detected much earlier.

Note: this will likely fail CI until https://github.com/spinnaker/deck/pull/7664 is merged first


christopherthielen:
Here's an example (I removed the explicit annotation from this controller):

Screen Shot 2019-11-30 at 2 28 27 PM

erikmunson:
From what I can tell, CI is failing because a provider in the `angular-mocks` code path is tripping the DI checks. I _think_ this was historically not a problem because strict DI isn't actually turned on for our karma tests, and now we're artificially erroring out about it. Tried to look into turning on `strictDI` via the injector in our `karma-shim` file, but I can't seem to really make it work and I'm somewhat suspicious that the `angular-mocks` library itself might have some incompatible module annotations.

Not totally sure what to do to resolve, one kind of terrible idea would be to just not load this new shim when we run things on karma 🤭

In the meantime I'm going to leave it as-is and might come back to it if I get bored enough in advance of Chris coming back from vacation.

PR #7670, opened 2019-12-03
Allow synthetic default imports

christopherthielen:
Enables `allowSyntheticDefaultImports` and `allowJs`.

Typescript 3.7 now supports `allowJs` and `declarations` at the same time. 
https://github.com/microsoft/TypeScript/issues/7546

`allowSyntheticDefaultImports` lets us do the normal javascript style import `import foo from 'foo'` instead of `import * as foo from 'foo'`


christopherthielen:
Depends on https://github.com/spinnaker/deck/pull/7668

PR #7672, opened 2019-12-03
fix(page): Do not auto-close modal on success

alanmquach:
Because this can complete so fast that users don't even see the task monitor, auto-closing causes users to not know whether their page even went through successfully.


app/scripts/modules/core/src/pagerDuty/PageModal.tsx
```diff
@@ -7,7 +7,7 @@ import { IPagerDutyService, PagerDutyWriter } from 'core/pagerDuty';
 import { NgReact, ReactInjector } from 'core/reactShims';
 import { SETTINGS } from 'core/config';
 import { SubmitButton } from 'core/modal';
-import { TaskMonitor } from 'core/task';
+import { TaskMonitor, TaskExecutor } from 'core/task';
```
erikmunson: doesn't look like `TaskExecutor` is used?

alanmquach: Whoops, thanks for catching that. It was residual from a bunch of crap I stuffed in to test this without actually paging people.

PR #7673, opened 2019-12-03
fix(core): fix NPE on first project cluster creation

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5174

Replaces yucky disabling (which wasn't even updating in certain scenarios because these FormikFormFields are not #formikFastFields) with #formikNative validation.


maggieneterval:

PR #7674, opened 2019-12-03
fix(webpack): use a less race-y way of generating CSS Module TS files

erikmunson:
The loader I had originally added to generate type declarations for CSS Module files seemed like it was working fine, but on Travis has been running into a race where the TS loader tries to lookup defs for the CSS Module files before their declarations are generated on the filesystem. That results in CI builds failing with errors re: not knowing about a module.

This switches to a [different webpack tool](https://github.com/dropbox/typed-css-modules-webpack-plugin) that runs during the plugin phase, before loaders kick in. Tested locally and it seems like it will be less race-y, but need to test on Travis to be sure.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,12 +1,19 @@
 'use strict';
 
 const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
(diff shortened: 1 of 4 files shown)


maggieneterval:
Thanks for the quick fix, @erikmunson, you're the best!

PR #7687, opened 2019-12-10
fix(tests): Use webpack-dev-middleware for functional tests

ezimanyi:
The new CSS loading plugin TypedCssModulesPlugin is causing webpack-dev-server to compile the application twice; the first generates the co-located .css.ts.d files and the second recompiles because of a change in sources.

This is breaking the functional tests, as the server reports healthy, causing the tests to start, but then immediately recompiles. This means that the first tests needs to wait for this second compilation pass, which is leading to intermittent (but frequent) errors.

The fix here is to turn off hot reloading of the app. There does not appear to be a way to do this using webpack-dev-server, so switch to using webpack-dev-middleware to set up the test server.

### DIFF
--- a/test/functional/tools/StaticServer.ts
+++ b/test/functional/tools/StaticServer.ts
@@ -1,8 +1,7 @@
 /// <reference path="wait-on.d.ts" />
 
-import * as waitOn from 'wait-on';
+const waitOn = require('wait-on');
 const configure = require('../../../webpack.config.js');
-
 const webpack = require('webpack');
 const middleware = require('webpack-dev-middleware');
 const express = require('express');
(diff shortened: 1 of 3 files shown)


ezimanyi:
I have confirmed locally that the first test now completes in ~5s instead of ~55s, so am fairly certain this should fix the failing functional tests.  (Also just pushed a minor second commit to fix a couple of things.)

test/functional/tools/StaticServer.ts
```diff
@@ -1,51 +1,48 @@
...
 
-import { spawn, ChildProcess } from 'child_process';
-import * as process from 'process';
 import * as waitOn from 'wait-on';
+const configure = require('../../../webpack.config.js');
+
+const webpack = require('webpack');
+const middleware = require('webpack-dev-middleware');
+const express = require('express');
+const app = express();
```
maggieneterval: Do we want to maybe either use `import` syntax for all the imports other than `configure`, or switch `waitOn` to use `require` just for consistency?

ezimanyi: Oh, great point...I'll switch them all to `require`.

PR #7688, opened 2019-12-10
fix(titus): Temporarily disabling previous image rollbacks for Titus

alanmquach:
`PREVIOUS_IMAGE` rollbacks for Titus were incompletely implemented making them dangerous to allow at this point.

Temporarily disabling them from being performed in the UI until they are addressed.

### DIFF
--- a/app/scripts/modules/titus/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
+++ b/app/scripts/modules/titus/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
@@ -89,7 +89,9 @@ module.exports = angular
 
         if (rollbackType === 'PREVIOUS_IMAGE') {
           // no need to validate when using an explicit image
-          return true;
+          // return true;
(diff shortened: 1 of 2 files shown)


app/scripts/modules/titus/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
```diff
@@ -89,7 +89,9 @@ module.exports = angular
 
         if (rollbackType === 'PREVIOUS_IMAGE') {
           // no need to validate when using an explicit image
-          return true;
+          // return true;
+          // KLUDGE: temporarily disabling previous image rollbacks for titus because they do not work properly
+          return false;
```
christopherthielen: I understand that this is a safety issue.  Will users be confused by this?  I suppose there's no feedback why the rollback is invalid.

alanmquach: Yeah.. agree. I was trying to come up with the right text/place. Think I got something that works.

PR #7690, opened 2019-12-11
chore(karma): Adding spec reporter to find problematic test

alanmquach:
Not sure if we want to merge this, but just trying to see if a travis run with this enabled gives us more insight into what is failing in master


alanmquach:
This turned out to not be useful at all in figuring out the offending test

PR #7692, opened 2019-12-11
feat(provider/ecs): add support for services with multiple target groups

pkandasamy91:
Fixes: spinnaker/spinnaker#4788

Should be merged in with: https://github.com/spinnaker/clouddriver/pull/4218

Task Definition Artifact:
Artifact Screenshot

Container Inputs:
Container Inputs Screenshot

Alert if No Target Groups are available:
No TG Available


app/scripts/modules/ecs/src/ecs.help.ts
```diff
@@ -60,6 +60,8 @@ const helpContents: { [key: string]: string } = {
   'ecs.containerMappingName':
     '<p>The name of the container. Name should match the <a href="https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html#ECS-Type-ContainerDefinition-name"><b>containerDefinition.name</b></a> field as it appears in the Task Definition.</p>',
   'ecs.containerMappingImage': '<p>The container image the named container should run.</p>',
+  'ecs.targetGroupMappings':
+    '<p>The list of expected target groups within the Task Definition and which port they should map to within the specified container.</p>',
```
clareliguori: I'm confused by this explanation. What do you mean by target groups within the task definition? Target groups are specified at the service level...

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/container.component.html
```diff
@@ -1,46 +0,0 @@
-<div class="container-fluid form-horizontal" ng-controller="ecsContainerImageController as containerImageController">
-  <div class="form-group">
-    <div class="col-md-4 sm-label-right">Container Image <help-field key="ecs.dockerimage"></help-field></div>
-    <div class="col-md-8">
-      <ui-select name="imageSelect" ng-model="$ctrl.command.imageDescription" class="form-control input-sm">
-        <ui-select-match placeholder="Please type your search...">{{ $select.selected.imageId }}</ui-select-match>
-        <ui-select-choices
-          repeat="image in $ctrl.command.backingData.filtered.images"
-          group-by="containerImageController.groupByRegistry"
```
clareliguori: Does the new react based ui group by registry, and allow searching and pagination? The UI should be usable when you have thousands of tags in your repository. I'm not sure if there are react based examples from other providers for scalable image selection

app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
```diff
@@ -94,62 +94,4 @@
       </div>
     </div>
   </div>
-
-  <div class="col-md-12" ng-if="$ctrl.command.viewState.dirty.targetGroup">
-    <div class="alert alert-warning">
-      <p>
-        <i class="fa fa-exclamation-triangle"></i>
-        The following target groups could not be found in the selected account/region/VPC and were removed:
```
clareliguori: Are we losing this logic? What happens when I select a different vpc higher up in the UI?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
```diff
@@ -94,62 +94,4 @@
...
-    </div>
-  </div>
-  <div class="form-group">
-    <div class="col-md-4 sm-label-right">
-      <b>Target Group</b>
-      <help-field key="ecs.loadBalancer.targetGroup"></help-field>
-    </div>
-    <div class="col-md-6">
-      <div class="form-control-static" ng-if="!$ctrl.command.backingData.filtered.targetGroups.length">
-        No target groups found in the selected account/region/VPC
```
clareliguori: Should we carry forward this logic and not show the target group mapping component if there are no target groups?

clareliguori:
@allisaurus would also be helpful to review this one

app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
pkandasamy91: The target group mapping component will always contain the Target Group, and Target Group Port. If one does not exist, it will be empty with an 'Add Target Group Mappings' button, but I think we should always show the section so that users can add new ones.

clareliguori: If targetGroupsAvailable is empty, won't the target group mappings row dropdown be empty?  That's the scenario where we might want to indicate that there are no target groups in selected VPC, and the user needs to select a different subnet to get some target groups (or go create some in that VPC)

pkandasamy91: It will, should we provide an error message if none are present, or rely on the user to figure it out since there are none in the drop down?

app/scripts/modules/ecs/src/ecs.help.ts
pkandasamy91: Updated

app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
clareliguori: I think users won't understand why there are none in the dropdown, or what action they can take to populate the dropdown (choose a diff VPC), so I think the error message we previously had was useful.

app/scripts/modules/ecs/src/ecs.help.ts
clareliguori: I still find this confusing.  Is this better?  

'The list of target groups through which the ECS service will receive load balancer traffic.  Each target group is mapped to a container name and port within the Task Definition to specify which container should be registered to the target group.'

allisaurus: yeah, the phrase "...target groups within the Task Definition" is the crux of the confusion. +1 to above reword

pkandasamy91: Addressed.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -0,0 +1,260 @@
...
+      });
+      cmd.targetGroup = '';
+    }
+
+    cmd.targetGroupMappings = uniqWith(defaultTargetGroup, isEqual);
+
+    this.state = {
+      imageDescription: cmd.imageDescription ? cmd.imageDescription : this.getEmptyImageDescription(),
+      computeUnits: cmd.computeUnits ? cmd.computeUnits : 256,
+      reservedMemory: cmd.reservedMemory ? cmd.reservedMemory : 256,
```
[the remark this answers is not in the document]
pkandasamy91: Good point, I think they'd be beneficial, but I'll send it in a later PR. Removed the defaults.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -0,0 +1,260 @@
...
+  private getIdToImageMap = (): Map<string, IEcsDockerImage> => {
+    const imageIdToDescription = new Map<string, IEcsDockerImage>();
+    this.props.command.backingData.filtered.images.forEach(e => {
+      imageIdToDescription.set(e.imageId, e);
+    });
+
+    return imageIdToDescription;
+  };
+
+  private getEmptyImageDescription = (): IEcsDockerImage => {
```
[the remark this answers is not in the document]
pkandasamy91: Added a TODO

app/scripts/modules/ecs/src/serverGroup/configure/wizard/taskDefinition/TaskDefinition.tsx
```diff
@@ -40,9 +43,27 @@ export class TaskDefinition extends React.Component<ITaskDefinitionProps, ITaskD
       defaultContainer = cmd.containerMappings[0].containerName;
     }
 
+    let defaultTargetGroup: IEcsTargetGroupMapping[] = [];
+    if (cmd.targetGroupMappings && cmd.targetGroupMappings.length > 0) {
+      defaultTargetGroup = cmd.targetGroupMappings;
+    }
+
+    if (cmd.targetGroup && cmd.targetGroup.length > 0) {
```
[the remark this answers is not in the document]
pkandasamy91: Wouldn't that cause string length to be 0?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/taskDefinition/TaskDefinition.tsx
```diff
@@ -40,9 +43,27 @@ export class TaskDefinition extends React.Component<ITaskDefinitionProps, ITaskD
...
+      defaultTargetGroup.push({
+        containerName: cmd.loadBalancedContainer || defaultContainer,
+        targetGroup: cmd.targetGroup,
+        containerPort: cmd.containerPort,
+      });
+      cmd.targetGroup = '';
```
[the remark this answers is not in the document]
pkandasamy91: Correct. When you click save, it saves it to Clouddriver, but otherwise, it's local

app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
pkandasamy91: Addressed. If target groups are not available, the Add New Target Group Mappings turns into an alert, and prevents the user from adding more mappings until some are available.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/networkingSelector.component.html
pkandasamy91: I don't think this logic actually works. Dirty.TargetGroups is never being set for the ECS provider (it is though for the Amazon provider)

pkandasamy91: @clareliguori would you have any objection if I implemented this in a separate PR?

clareliguori: No, provided that it doesn't actually work today, so we're not losing anything

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/container.component.html
clareliguori: How is this one being addressed?  Is this something we need to address in a different PR too?

pkandasamy91: This logic doesn't exist in the TaskDefinition.tsx as well. It's something I need to look into. If it's not a blocker, I'd suggest merging these changes in, and I'll address these 2 issues in the next PR.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/taskDefinition/TaskDefinition.tsx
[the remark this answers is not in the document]
allisaurus: actually, both of these conditions evaluate to [falsey](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) in JS so we could just do `!cmd.targetGroup` to be more concise.

if we also wanted to screen out whitespace-only strings, we could do `!_.trim(cmd.targetGroup)`, but I don't know what the current behavior is for this.

pkandasamy91:
> Per offline discussion, please add (successful) test output for deployments using EC2/bridge mode and an artifact (together or separate).
> 
> Other than that, LGMT save a couple nits re: falsey checks.

Confirmed. Bridge mode and Artifacts still work

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -0,0 +1,269 @@
...
+    if (cmd.containerMappings && cmd.containerMappings.length > 0) {
+      defaultContainer = cmd.containerMappings[0].containerName;
+    }
+
+    let defaultTargetGroup: IEcsTargetGroupMapping[] = [];
+    if (cmd.targetGroupMappings && cmd.targetGroupMappings.length > 0) {
+      defaultTargetGroup = cmd.targetGroupMappings;
+    }
+
+    if (cmd.targetGroup && cmd.targetGroup.length > 0) {
```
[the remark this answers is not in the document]
pkandasamy91: I actually want to check that target group is populated. Wouldn't this check the opposite?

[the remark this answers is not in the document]
allisaurus: ah! well in that case `if (cmd.targetGroup) {}` would be sufficient, b/c null or undefined or empty string are all falsey - but not going to block on this :P

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -0,0 +1,269 @@
...
+      account: '',
+      registry: '',
+      repository: '',
+      tag: '',
+    };
+  };
+
+  private pushTargetGroupMapping = () => {
+    const targetMaps = this.state.targetGroupMappings;
+    targetMaps.push({ containerName: '', targetGroup: '', containerPort: 80 });
```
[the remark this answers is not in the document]
allisaurus: per offline discussion: null or empty will be angry with number type and don't want to default to 0 b/c that's a bad port so keeping as 80 for now

PR #7685, opened 2019-12-10
Sweeping eslint changes

christopherthielen:
This is a set of commits that move us closer to pure ES6 module codebase.

### Always export the ng module name, not the module itself

This extracts the module name to a string constant (based on the name/path of the file it resides in). 
It exports the constant for use by dependers.  It also exports a `name` constant for backwards compatibility for use via `require('./module').name`.

```diff
-module.exports = angular.module('spinnaker.amazon.instanceType.service', []).factory('awsInstanceTypeService', [
+export const AMAZON_INSTANCE_AWSINSTANCETYPE_SERVICE = 'spinnaker.amazon.instanceType.service';
+export const name = AMAZON_INSTANCE_AWSINSTANCETYPE_SERVICE; // for backwards compatibility
+angular.module(AMAZON_INSTANCE_AWSINSTANCETYPE_SERVICE, []).factory('awsInstanceTypeService', [
```

### Import angularjs module dependencies by name.  

Migrate angularjs module dependencies to import the exported string identifier, and stop using `require('module').name`

```diff
+import { AMAZON_SECURITYGROUP_CONFIGURE_CONFIGSECURITYGROUP_MIXIN_CONTROLLER } from '../configure/configSecurityGroup.mixin.controller';
angular
  .module(AMAZON_SECURITYGROUP_CLONE_CLONESECURITYGROUP_CONTROLLER, [
-    require('../configure/configSecurityGroup.mixin.controller').name,
+    AMAZON_SECURITYGROUP_CONFIGURE_CONFIGSECURITYGROUP_MIXIN_CONTROLLER,
  ])
```

### use ES6 imports for angularjs module deps

for example:
- migrate from `require('@uirouter/angularjs').default` to import UIROUTER_ANGULARJS from '@uirouter/angularjs'
- migrate from `require('angular-ui-bootstrap')` to import ANGULAR_UI_BOOTSTRAP from 'angular-ui-bootstrap'

```patch
+import UIROUTER_ANGULARJS from '@uirouter/angularjs';
+import ANGULAR_UI_BOOTSTRAP from 'angular-ui-bootstrap';
angular
  .module(ECS_INSTANCE_DETAILS_INSTANCE_DETAILS_CONTROLLER, [
-    require('@uirouter/angularjs').default,
-    require('angular-ui-bootstrap'),
+    UIROUTER_ANGULARJS,
+    ANGULAR_UI_BOOTSTRAP,
    CONFIRMATION_MODAL_SERVICE,
  ])
```

[...]

PR #7697, opened 2019-12-12
Migrate from tslint to eslint and add custom deck eslint rules

christopherthielen:
_Let's see if this passes CI..._

TSLint is deprecated in favor of `eslint` + `@typescript-eslint` extensions.  

This PR:
- Removes `tslint` 
- Adds `@typescript-eslint` extensions
- Processes both `.js` *and* `.ts` with eslint
- Configures eslint rules for typescript code
- Adds custom eslint rules for AngularJS code

---

This PR creates an eslint plugin package in `/packages/eslint-plugin` with some standardized configuration for Deck along with some custom angularjs rules (to ease migration of legacy code to a standardized pattern).  The intention is that this package will be added as a dependency to deck, as well as to any plugin packages being developed.

Note: This package is currently installed in deck's package.json via `"@spinnaker/eslint-rules": "./packages/eslint-rules"` entry.  In the near future I'd like to publish the eslint plugin package to NPM and wire it up via standard npm dependencies (and then in the medium future use [lerna](https://github.com/lerna/lerna))

I created 5 eslint rules for angularjs to clean up some old patterns:

1) Do not use `new` syntax for controllers: `angular.module('foo', []).controller('fooController', new FooController())`
2) Do not export the angular module `module.exports = angular.module('foo', [])`
3) Do not use `require()` syntax to import angular `const angular = require('angular')`
4) Do not use `require()` syntax to import angular dependencies `angular.module('foo', [require('someDep').name])`
5) Always require strict DI annotations and do not allow unannotated angularjs code `.controller('myController', function($scope) {})`

I see these 5 AngularJS eslint rules as legacy code and possibly even temporary.  Once these "bad" patterns are no longer propagated and as long as new code is written in react, these rules will no longer needed.  I don't intend on improving them unless it would get us closer to some other goal (such as having pure ES6 modules). 

---

This PR is heavily related to https://github.com/spinnaker/deck/pull/7685, which represents the result of running these rules on the deck codebase.


christopherthielen:
@bpowell  @link108

packages/eslint-plugin/base.config.js
```diff
@@ -0,0 +1,55 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
```
louisjimenez: The custom rules in this PR are quite extensive :clap:. Is the ecmaVersion used for linting being determined by this parser? What version is being used?

christopherthielen: `extends: ['plugin:@typescript-eslint/recommended']` pulls in this file https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/base.json which enables ES6 modules.  I don't think the ecmaVersion is explicitly set anywhere 🤔

PR #7701, opened 2019-12-13
fix(projects): Fixing clusters error validation

alanmquach:
Falsy errors for `stack` and `detail` should actually not be placed on the `errors` object at all.

Otherwise, `errors` will be `{ stack: false, detail: false }` which will cause `Object.keys(errors).length` to be true and thus we will return an error object even though there aren't actually any errors.

This prevents the project from being saved due to validation errors.


maggieneterval:
@alanmquach :woman_facepalming: Sorry about this, thank you so much for fixing! Good to know about no falsy errors on the validation object :+1:

maggieneterval:

PR #7706, opened 2019-12-13
refactor(google): refactor GCB stage to use new FormikFormFields and …

maggieneterval:
…hooks APIs

The goal of this PR is to refactor the Google Cloud Build stage to use Deck's new best-practice patterns for wiring up a stage config form with FormikFormFields and handling data fetching and other side effects with hooks.

User experience improvements:
- Clear loading states for inputs with data dependent on pending requests
- Inline validation for missing required inputs
- Inputs that are dependent on missing prior inputs are disabled

Developer experience improvements:
- The declarative style of hooks makes it much easier to understand the implicit dependency graph in this component: we fetch all accounts once, we fetch available account triggers when the selected account changes, we clear no-longer-relevant fields when upstream form inputs change, etc.

Future work:
- Encapsulate pre- and post-artifact-rewrite stage config Formik setters and validators so we don't have to duplicate these across each Kubernetes stage that accepts artifacts.
- Encapsulate YAML editor state management and Formik setters so we don't have to duplicate these across each Kubernetes stage that accepts YAML.
- Add SpEL-aware FormikFormFields and enable experimentally in this stage only.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { cloneDeep, defaults } from 'lodash';
+import { cloneDeep } from 'lodash';
 
 import { FormikStageConfig, IFormikStageConfigInjectedProps, IStageConfigProps } from '@spinnaker/core';
 
@@ -15,10 +15,11 @@ export function GoogleCloudBuildStageConfig({
   updateStage,
 }: IStageConfigProps) {
   const stageWithDefaults: IGoogleCloudBuildStage = React.useMemo(() => {
-    return defaults(cloneDeep(stage), {
+    return {
       application: application.name,
       buildDefinitionSource: BuildDefinitionSource.TEXT,
-    });
+      ...cloneDeep(stage),
+    };
   }, []);
 
   return (
--- a/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
@@ -1,87 +1,36 @@
 import React from 'react';
-import { Observable, Subject } from 'rxjs';
-import { cloneDeep } from 'lodash';
+import { cloneDeep, defaults } from 'lodash';
 
-import { FormikStageConfig, IgorService, IStage, IStageConfigProps, IGcbTrigger } from '@spinnaker/core';
+import { FormikStageConfig, IFormikStageConfigInjectedProps, IStageConfigProps } from '@spinnaker/core';
 
-import { GoogleCloudBuildStageForm, buildDefinitionSources, triggerType } from './GoogleCloudBuildStageForm';
+import { GoogleCloudBuildStageForm } from './GoogleCloudBuildStageForm';
 import { validate } from './googleCloudBuildValidators';
+import { BuildDefinitionSource, IGoogleCloudBuildStage } from './IGoogleCloudBuildStage';
 
-interface IGoogleCloudBuildStageConfigState {
-  googleCloudBuildAccounts: string[];
-  gcbTriggers: IGcbTrigger[];
-}
+export function GoogleCloudBuildStageConfig({
+  application,
+  pipeline,
+  stage,
+  updatePipeline,
+  updateStage,
+}: IStageConfigProps) {
+  const stageWithDefaults: IGoogleCloudBuildStage = React.useMemo(() => {
+    return defaults(cloneDeep(stage), {
+      application: application.name,
+      buildDefinitionSource: BuildDefinitionSource.TEXT,
+    });
+  }, []);
 
-export class GoogleCloudBuildStageConfig extends React.Component<IStageConfigProps, IGoogleCloudBuildStageConfigState> {
-  private stage: IStage;
-  private destroy$ = new Subject();
-
-  public constructor(props: IStageConfigProps) {
-    super(props);
-    this.state = {
-      googleCloudBuildAccounts: [],
-      gcbTriggers: [],
-    };
-    const { stage: initialStageConfig } = props;
-    const stage = cloneDeep(initialStageConfig);
-    if (!stage.application) {
-      stage.application = props.application.name;
-    }
-    if (!stage.buildDefinitionSource) {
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
```diff
@@ -1,87 +1,36 @@
...
+import { BuildDefinitionSource, IGoogleCloudBuildStage } from './IGoogleCloudBuildStage';
+
+export function GoogleCloudBuildStageConfig({
+  application,
+  pipeline,
+  stage,
+  updatePipeline,
+  updateStage,
+}: IStageConfigProps) {
+  const stageWithDefaults: IGoogleCloudBuildStage = React.useMemo(() => {
```
maggieneterval: Any thoughts on a better way to initialize a stage config with defaults?

maggieneterval:
@christopherthielen @gcomstock @louisjimenez This is the stage on which I'd like to experimentally enable (coming soon) SpEL-aware FormikFormFields (smaller blast radius than the Kubernetes stages). I know we had initially discussed configuring SpEL-awareness on context, but as I actually began implementing it this way, I realized that there are several input types that it probably doesn't make sense to make SpEL-aware. Specifically, I don't think there is a user story for textarea inputs like the YAML editor or radio selects (that have only a handful of options that it wouldn't make sense to template because they determine the downstream state of the form) being SpEL-aware. With that in mind, how would we feel about having the SpEL-aware config be a property of the FormikFormField itself rather than a context that wraps an entire form? Otherwise, we will probably end up having to wrap individual FormikFormFields in context to opt them in and out of SpEL awareness, which at least in the case of this form would get pretty verbose. For clarity, I've outlined in red the only inputs it would make sense to make SpEL aware in this stage. Let me know what you think! :pray: :hugs: 

zE6LikEmFqn

0JGrTbBtEwH

0u1uW2iPZ8b

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
louisjimenez: This seems to work. An alternative, native way for doing this could be `{ application: application.name, buildDefinitionSource: buildDefinitionSource.TEXT, ...cloneDeep(stage)  }`

louisjimenez:
> With that in mind, how would we feel about having the SpEL-aware config be a property of the FormikFormField itself rather than a context that wraps an entire form? 

We took a look at this and I am leaning in this direction as well. @christopherthielen is there something significant we are forgoing by not lifting this state up to a context?

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageConfig.tsx
maggieneterval: :+1: :zap: updated, thanks! Was mostly curious if anyone had any thoughts on where the defaulting should happen (in the parent component that renders the `FormikStageConfig` vs. in the actual stage config component).

christopherthielen:
> how would we feel about having the SpEL-aware config be a property of the FormikFormField itself rather than a context that wraps an entire form

I think I'd like to have _both_ a spel-aware context setting _and_ a prop on a `FormikFormField` that opts in/out (where the prop overrides any context setting).  

If a stage config is mostly spel-aware, then it can default to spel (via the stage-level context provider) and opt out a few fields where it doesn't make sense (almost all of the amazon provider fields are spel aware).  If a form is mostly spel-unaware, then it can provide its own context at the top of the stage config and opt specific fields in.

maggieneterval:
> I think I'd like to have both a spel-aware context setting and a prop on a FormikFormField that opts in/out (where the prop overrides any context setting).

:+1: Sounds good to me!

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -1,234 +1,220 @@
...
+  ArtifactTypePatterns.GITLAB_FILE,
+  ArtifactTypePatterns.HTTP_FILE,
+  ArtifactTypePatterns.S3_OBJECT,
+);
+
+export function GoogleCloudBuildStageForm(props: IGoogleCloudBuildStageFormProps & IFormikStageConfigInjectedProps) {
+  const stage = props.formik.values;
+
+  const [rawBuildDefinitionYaml, setRawBuildDefinitionYaml] = React.useState(
+    stage.buildDefinition ? yamlDocumentsToString([stage.buildDefinition]) : '',
```
christopherthielen: Is `yamlDocumentsToString` expensive computationally? This will get run on every render.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -1,234 +1,220 @@
...
+        name="buildDefinitionSource"
+        input={(inputProps: IFormInputProps) => <RadioButtonInput {...inputProps} options={SOURCE_OPTIONS} />}
+      />
+      {stage.buildDefinitionSource === BuildDefinitionSource.TEXT && (
+        <FormikFormField
+          fastField={false}
+          label="Build Definition"
+          name="buildDefinition"
+          input={(inputProps: IFormInputProps) => (
+            <YamlEditor {...inputProps} value={rawBuildDefinitionYaml} onChange={onYamlChange} />
```
christopherthielen: it might be worth wrapping `YamlEditor` as a `FormInput` compatible component `YamlEditorInput` in some future PR.  It would be nice if each form that uses yaml didn't have to know how to manage single/multi-document yaml.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
maggieneterval: Great call, updated to [lazily initialize the state](https://reactjs.org/docs/hooks-reference.html#lazy-initial-state) with a function instead.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
maggieneterval: Yes, absolutely -- will do this as part of converting the Kubernetes stages that also use a YamlEditor.

PR #7708, opened 2019-12-13
Template version ui

louisjimenez:
https://github.com/spinnaker/spinnaker/issues/4670

This displays each saved version for a given pipeline template in the UI and allows specific versions to be viewed, deleted, or used to create new pipelines. It also updates tsconfigs to use es2019 so that funcs such as `flatMap` are linted properly.

image

Related PRs:
[Gate](https://github.com/spinnaker/gate/pull/989)
[Front50](https://github.com/spinnaker/front50/pull/658)
[Spin](https://github.com/spinnaker/spin/pull/235)


maggieneterval:
This is so exciting! Going to look in more detail on Monday with fresh eyes. :tada:

app/scripts/modules/cloudfoundry/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "jsx": "react",
-    "lib": ["es2017", "dom"],
+    "lib": ["es2019", "dom"],
```
christopherthielen: does typescript emit a polyfill for this, or does it emit es2019 features directly (requiring browser support)?

louisjimenez: @christopherthielen I believe that this will add support for type checking es2019, leave the compilation target at es6, and will not polyfill  for the 2019 spec. After looking at the webpack config, it seems we are also not polyfilling the existing 2017 spec or using babel for tsx files. Is there somewhere else where we are applying polyfills?

louisjimenez: Chatted with @christopherthielen and confirmed it is not getting polyfilled. I'm going to replace the es2019 usage with lodash.

PR #7714, opened 2019-12-17
chore(eslint): lint the entire app/ directory not just modules/

christopherthielen:
- include things like app.ts


maggieneterval:
:heart: :heart:

PR #7726, opened 2019-12-19
feat(aws/additionalIpRules): Add a component for aws security group a…

aravindmd:
- Adding a component for additional ip rules


app/scripts/modules/amazon/src/securityGroup/configure/additionalIpRules.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { ISecurityGroupDetail } from 'core/securityGroup';
+
+export interface IAdditionalIpRulesProps {
+  securityGroupDetails: ISecurityGroupDetail;
+}
+
+@Overridable('aws.securityGroup.additional.ipRules')
+export class AdditionalIpRules extends React.Component<IAdditionalIpRulesProps> {
+  public render(): any {
+    return null;
```
caseyhebebrand: Is this meant to always return `null`?  Will it return something different in `deck-nflx`?

aravindmd: yes that is correct, the purpose of this is to override the component for specific internal implementation.

caseyhebebrand: If there is a way to create this component  (or parts of it) in a generic way, then I would advocate for adding that code here and hiding it under a feature flag `RenderIfFeature` component.

Then you can add our version to deck-nflx. This way OSS can potentially use this in the future. If the component is _very_ custom, then this should be fine. Maybe add comments about what the intended purpose of this component could be.

aravindmd: I think at this point this component is custom , i will add some comments to make it clear and revisit later if we need to create the component in a generic way.

christopherthielen: I disagree.  I think we should not clutter OSS deck with feature flagged code that only makes sense inside of netflix.  If we do eventually create an open source version of this, then we can migrate the code to OSS Deck at that time.

caseyhebebrand: yes, I was missing some context here, but chatted offline about it. It sounds like it is too custom for this approach. I was under the impression that it could be abstracted to some level.

PR #7730, opened 2019-12-20
Add plugin support to Deck

christopherthielen:
This builds upon https://github.com/spinnaker/deck/pull/7474 which @bpowell submitted in October.  This version incorporates some of the ideas from the UI Plugin RFC: https://github.com/spinnaker/governance/pull/74

- adds a PluginRegistry
- switches from requirejs to native `import()`
- restores original bootstrap mechanism
- use the router to defer application start
- removes the `modules/plugins` for now (to be re-added later in `/packages`)

See https://github.com/christopherthielen/rollup-starter-lib for an example plugin that can be loaded via this mechanism.


christopherthielen:
@bpowell @link108 @claymccoy

halconfig/settings.js
```diff
@@ -232,6 +232,7 @@ window.spinnakerSettings = {
     },
   },
   pubsubProviders: ['google'], // TODO(joonlim): Add amazon once it is confirmed that amazon pub/sub works.
+  plugins: '{{%plugins%}}',
```
bpowell: @link108 didn't we have an issue with this when we were looking at it from the Halyard side of things?

app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -0,0 +1,71 @@
...
+      if (!module || !module.plugin) {
+        throw new Error(
+          `Successfully loaded plugin module from ${devUrl}, but it doesn't export an object called 'plugin'`,
+        );
+      }
+
+      const plugin: IDeckPlugin = module.plugin;
+
+      // Register extensions with deck.
+      plugin.stages?.forEach(stage => Registry.pipeline.registerStage(stage));
```
bpowell: does this mean currently this only works for stages and that's it?

christopherthielen: Yep, at the moment. More to be added later if this looks good.

christopherthielen: Yep, at the moment. More to be added later if this looks good.

halconfig/settings.js
christopherthielen: This is from the original pr. 
I’m going to lean on your expertise to make the halyard bits work since we don’t use halyard at Netflix

christopherthielen: This is from the original pr. 
I’m going to lean on your expertise to make the halyard bits work since we don’t use halyard at Netflix

christopherthielen:
Need to add tests.

ttomsu:
I think this PR is causing an issue in the 1.18.x release (cc @duftler)

Specifically, I'm getting this in the console:

```
plugin.module.ts:16 Uncaught (in promise) TypeError: core_src.SETTINGS.plugins.forEach is not a function
    at plugin.module.ts:16
    at Generator.next (<anonymous>)
    at plugin.registry.ts:78
    at new Promise (<anonymous>)
    at plugin_module_awaiter (plugin.registry.ts:78)
    at plugin.module.ts:9
    at Object.invoke (angular.js:5117)
    at runInvokeQueue (angular.js:5004)
    at angular.js:5014
    at forEach (angular.js:418)
```

I can't tell if this is merely a typo or completely broken - specifically in the `halconfig/settings.js` file, which has this line:

```
plugins: '{{%plugins%}}',
```

The extra set of { } seems to be different than the rest of the file (which I think is Jinja syntax). Even if the extra braces were correct, this seems to me to attempt to insert the entire `plugins` object from the Halconfig into the Deck config. Eric's NPE fix (https://github.com/spinnaker/deck/pull/7736) makes me think, though, that this object should be an array (confirmed by the above error which is doing a `forEach` over the value).

I'm going to open a PR that removes this line from `halconfig/settings.js` until we can figure out how to make this work.

PR #7734, opened 2019-12-22
fix(artifacts): set the reference attribute for helm artifact

Jammy-Louie:
spinnaker/spinnaker#4777

It appears the initial refactor of the helm artifact did not set the `reference` property.  This should rectify the issue and set the reference when the account changes as indicated by the previous implementation.

https://github.com/spinnaker/deck/blob/81386a3c55608dd1c31a30eff1d46f7c39627f30/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/helm/helm.artifact.ts#L27


maggieneterval:
@Jammy-Louie Thanks so much for looking into this! I actually just dug into this a bit and realized that we no longer require `reference` to be set since [this Rosco change](https://github.com/spinnaker/rosco/pull/442) (which should be in both 1.16 and 1.17). So I think we can close this for now?

PR #7737, opened 2020-01-02
fix(amazon): Fix additionalIpRules component imports

aravindmd:
- fix tsconfig for amazon and titus


app/scripts/modules/amazon/tsconfig.json
```diff
@@ -31,6 +31,7 @@
     "typeRoots": ["../../../../node_modules/@types"],
     "paths": {
       "@spinnaker/core": ["../../core/lib"],
+      "core/*": ["../../core/lib/*"],
```
aravindmd: Added this back to fix ci build.

PR #7740, opened 2020-01-06
feat(spel): add SpEL-awareness to FormikFormField

maggieneterval:
**- feat(spel): add SpEL-awareness to FormikFormField**

  Adds Formik SpEL configuration to context and FormikFormField props. Adds a single disabled-by-default mode that allows users to toggle between the supplied FormikFormField input and a freeform textarea input in which to enter SpEL, using new `SpelToggle` action button.
  
**- chore(spel): add tests for FormikFormField SpEL-awareness**

  Adds tests to confirm that the `SpelToggle` action button only renders when appropriate based on SpEL configuration on context and `FormikFormField` props, that a freeform SpEL input renders by default for form values that include SpEL when mode is enabled, and that toggling between the default and freeform SpEL input changes the input type as expected.
  
**- fix(style): make actions inline with input**

  Adds a few styles so that the `ResponsiveFieldLayout`'s `input` and `actions` render inline with the same in-between spacing as those of the `StandardFieldLayout` (without these style changes, the `actions` render below the `input`).

**- feat(gcb): enable Formik SpEL-awareness in GCB stage**

  Wraps `GoogleCloudBuildStageForm` in `FormikSpelContext` with default-to-freeform toggling enabled. Opts specific `FormikFormField`s out of SpEL-awareness using `spelConfig` prop. (Including this commit in this PR as an example of how to selectively enable new functionality).

aecc2771-b2cf-4108-8ddb-60950409db6a


app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
```diff
@@ -70,13 +82,56 @@ function FormikFormFieldImpl<T = any>(props: IFormikFormFieldImplProps<T>) {
...
 
+  const [inputMode, setInputMode] = React.useState(SpelAwareInputMode.DEFAULT);
+
+  const modeToInput: { [k in SpelAwareInputMode]: React.ComponentType<IFormInputProps> } = {
+    [SpelAwareInputMode.DEFAULT]: input,
+    [SpelAwareInputMode.FREEFORM]: SimpleSpelInput,
+  };
+
+  const freeformInputEnabled = firstDefined(
+    get(props, 'spelConfig.freeformInputEnabled', null),
```
christopherthielen: To get typings to flow through, we can use optional chaining.  Inspired by @anotherchrisberry  in #7742
```suggestion
    props.spelConfig?.freeformInputEnabled ?? null;
```

app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
```diff
@@ -70,13 +82,56 @@ function FormikFormFieldImpl<T = any>(props: IFormikFormFieldImplProps<T>) {
...
+    get(useContext(FormikSpelContext), 'freeformInputEnabled', null),
```
christopherthielen: `useContext(FormikSpelContext)` should never be nullish, so how about:
```suggestion
    useContext(FormikSpelContext).freeformInputEnabled,
```
I would also suggest extracting the useContext to a top level function variable.   my reasoning (opinion alert) is that it's easier to parse what hooks are in use if they're all at the top level.

app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
```diff
@@ -70,13 +82,56 @@ function FormikFormFieldImpl<T = any>(props: IFormikFormFieldImplProps<T>) {
...
+    }
+    const fieldValue = get(props.formik.values, name, '');
+    const isFieldValueSpel = SpelService.includesSpel(fieldValue);
+    if (isFieldValueSpel) {
+      setInputMode(SpelAwareInputMode.FREEFORM);
+    }
+  }, []);
+
+  const onSpelToggleClick = () => {
+    formik.setFieldValue(name, null);
```
christopherthielen: I wonder if clearing the field value is the right choice.  I suspect the answer varies depending on the input.

app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
```diff
@@ -70,13 +82,56 @@ function FormikFormFieldImpl<T = any>(props: IFormikFormFieldImplProps<T>) {
...
+    setInputMode(inputMode === SpelAwareInputMode.DEFAULT ? SpelAwareInputMode.FREEFORM : SpelAwareInputMode.DEFAULT);
+  };
+
   const renderField = ({ field }: FieldProps<any>) => {
-    const inputRenderPropOrNode = firstDefined(input, noop);
+    const inputRenderPropOrNode = firstDefined(modeToInput[inputMode], noop);
```
christopherthielen: I feel like this might be simpler if we co-located the default/freeform lookup with its usage and nuked the modeToInput mapping
```suggestion
    const inputOrSpel = inputMode === SpelAwareInputMode.DEFAULT ? input : SimpleSpelInput;
    const inputRenderPropOrNode = firstDefined(inputOrSpel, noop);
```
dis/agree?

If the intention is that the map can have additional modes added to it, can we use a function instead?

app/scripts/modules/core/src/presentation/forms/layouts/ResponsiveFieldLayout.tsx
```diff
@@ -27,8 +27,9 @@ export function ResponsiveFieldLayout(props: ILayoutProps) {
         </div>
         <div className="sp-formItem__right">
           <div className="sp-form">
-            <span className="field">
-              {input} {actions}
+            <span className="flex-container-h margin-between-lg field">
```
christopherthielen: I don't think this is quite right for mobile, but I also think it was broken before, so /shrug

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
```diff
@@ -127,94 +134,114 @@ export function GoogleCloudBuildStageForm(props: IGoogleCloudBuildStageFormProps
...
   }, [stage.triggerType]);
 
+  const freeformSpelEnabled = React.useMemo(
+    () => ({
+      freeformInputEnabled: true,
+    }),
+    [],
+  );
+
+  const freeformSpelDisabled = React.useMemo(
```
christopherthielen: If we stick with the object config, a ref is more succinct: 
``` const freeformSpelDisabled = React.useRef({ freeformInputEnabled: false }).current;```

File: app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx [hunk 7]
```diff
@@ -127,94 +134,114 @@ export function GoogleCloudBuildStageForm(props: IGoogleCloudBuildStageFormProps
     }
   }, [stage.triggerType]);
 
+  const freeformSpelEnabled = React.useMemo(
```
christopherthielen: What other configuration are you anticipating in `FormikSpelContext`?  Seeing this boilerplate and the usage down below (`spelConfig={freeformSpelDisabled}`) makes me wish for a simple `spelAware={false}`.

If the expectation is that we will add further things to the Spel context (i.e.: `{ freeformSpelDisabled: true, otherFieldToggle: false, thirdtoggle: true }`) then I have followup questions:  

- Would each component that provides this context object have opinions for _all the spel settings_ the context object contains?  For example, if we have three toggles but I only have an opinion about freeformSpelDisabled then I have to fetch the current context and merge in my settings.  
- How would you feel about an API where each setting was a separate context with a separate override prop?
- If we like the configuration-object approach, could these be pulled up into a `SpinFormikConfigContext` instead?

app/scripts/modules/core/src/presentation/spel/SpelToggle.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export function SpelToggle(props: ISpelToggleProps) {
+  const inputModeLabels: { [k in SpelAwareInputMode]: string } = {
+    [SpelAwareInputMode.DEFAULT]: 'Enter SpEL in freeform input',
+    [SpelAwareInputMode.FREEFORM]: 'Return to default input',
+  };
+
+  return (
+    <Tooltip value={inputModeLabels[props.inputMode]}>
+      <button className="btn btn-sm btn-default" onClick={props.onClick}>
```
christopherthielen: can we get an active state class on this button?  Greg used yellow in the mocks (which I don't personally love) but he's out until next monday.

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
maggieneterval: Thanks so much for the thorough review!

I was anticipating future configuration options for the following:
- Enabling a `VariablePicker` mode
- Enabling messages like the ones in Greg's mocks encouraging users to migrate SpEL to the Evaluate Variables stage

But I think you're right that each of these can just be a separate context with a separate override prop, and we can re-evaluate merging into a single config object if it becomes too many contexts/props.

app/scripts/modules/core/src/presentation/forms/layouts/ResponsiveFieldLayout.tsx
maggieneterval: To my untrained eye it looks...at least not worse than before?

e2ME35kYs25

app/scripts/modules/core/src/presentation/spel/SpelToggle.tsx
maggieneterval: Great call! Conditionally added the `active` class so it would get the default gray-ish `btn-default:active` styling, but please feel free to change when Greg returns if we feel strongly about the yellow!

app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
maggieneterval: Agree, thank you! Also nuked the mode to message mapping in `SpelToggle`.

app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
maggieneterval: Amaaaaazing thank you for the inspo @anotherchrisberry

app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/GoogleCloudBuildStageForm.tsx
maggieneterval: Updated so that the `FormikSpelContext` and `FormikFormField`'s `spelAware` prop are a bool rather than a config object.

app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
maggieneterval: At least in the stage fields I played around with, it might be confusing not to clear the value, since you could end up in a state where a valid string (say, `${parameters['account']}`) is persisted on the stage model but not reflected in the UI because the user has toggled back to the default input, which is some input that only accepts some list of predefined values.

Are we comfortable clearing the field value for now as a default? What input types were you thinking it would make more sense not to clear?

maggieneterval: @christopherthielen are you okay merging this as-is or do you have additional thoughts on _not_ clearing the field value under certain circumstances?

PR #7767, opened 2020-01-13
fix(plugins): Remove plugins entry that is causing runtime errors

ttomsu:
See explanation here: https://github.com/spinnaker/deck/pull/7730#issuecomment-573729170

Note that this will get cherry picked into release 1.18

cc @bpowell


ttomsu:

PR #7770, opened 2020-01-13
fix(appconfig): Fix error thrown when slack object is not defined

ttomsu:
Fixes https://github.com/spinnaker/spinnaker/issues/5306

Needs to be cherry picked for 1.18


ttomsu:

PR #7782, opened 2020-01-15
fix(plugins): updated halconfig to work with halyard string replacement

bpowell:


bpowell:
@christopherthielen @ttomsu here is a fix for halyard based deploys to properly replace plugin information via halyard. 

cc @ezimanyi

ttomsu:
Can you share a snippet of your halconfig that proves this works? Because from my understanding of how Halyard injects values into this file, this still won't work - Plugins still has a bunch of other non-array fields

bpowell:
Right now, Halyard doesn't do anything with plugins with respect to Deck. This change will allow Halyard to replace the value when Halyard gets updated to replace `{%plugins%}` based on the halconfig. That is what I am currently working on. Without this change, Halyard won't be able to add the plugin information needed.

ttomsu:
Yes, I understand that. What do you expect Halyard to inject as the value? Because as it looks to me, it's going to inject a [`Plugins` object](https://github.com/spinnaker/halyard/blob/a0aa2c435850787a9a59c17d059788a408833ac0/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/node/Plugins.java#L30), not an array of [`Plugin`s](https://github.com/spinnaker/halyard/blob/a0aa2c435850787a9a59c17d059788a408833ac0/halyard-config/src/main/java/com/netflix/spinnaker/halyard/config/model/v1/plugins/Plugin.java#L38)

bpowell:
Halyard will put in a list of plugin objects. Something like `[{'name':'myplugin1', 'version':'1.0.0', 'devUrl':'/plugins/index.js'},{...}]`. Essentially Halyard will look at all of the plugins and replace the value with a string that is a list of objects. Sorry if that wasn't clear

bpowell:
The plugin class will probably have to change a little bit to make this work btw. That is what I am currently looking at

christopherthielen:
I'm not at all familiar with halyard and settings.js. Will it correctly quote the contents of `{%plugins%}`?  

This approach using JSON.parse strikes me as potentially brittle.  What was the issue with dropping the contents of `{%plugins%}` directly into the file?  i.e.:

```js
    plugins: {%plugins%},
```

bpowell:
@christopherthielen isn't this file supposed to be valid javascript? I thought there was a step to validate that it was. Having `var plugins = {%plugins%}` would not be valid.

ttomsu:
There were two issues AFAICT:

1. `plugins` needs to be an array, but nothing was matching the Jinja replacement, so "{%plugins%}" was getting interpreted as a string literal, which caused downstream errors where a forEach() method was being called on a string. 

1. There were originally two sets of { }, which is an (unconfirmed) syntax error

bpowell:
@ttomsu that is correct. The extra set of `{}` was there because at first we were going to make a change in Halyard to do object replacement instead of string replacement. That work was never submitted to OSS. We wanted to make it more simple from a replacement side, so I thought doing a `Json.parse` on the string literal would be a better option. We have a thread in slack discussing this right now: https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1578510602005000

bpowell:
i'm going to close this PR and make loading plugin information based on `plugin-manifest.json` that Halyard will laydown for plugins

https://spinnakerteam.slack.com/archives/CK9FK4XDF/p1579137567020600?thread_ts=1578510602.005000&cid=CK9FK4XDF

PR #7788, opened 2020-01-16
feat(plugins): read plugin config from plugin-manifest.json

bpowell:


bpowell:
@christopherthielen based on todays SIG Platform meeting, a stop gap solution for plugins until the front50/gate work is compled

cc @robzienert @ttomsu @ezimanyi

bpowell:
@christopherthielen can I get some help with the failing travis build? I'm not sure why `import` works with a variable, but not a string

bpowell:
hey @christopherthielen I got it to work. Mind taking a look?

app/scripts/modules/core/src/plugins/plugin.module.ts
```diff
@@ -12,8 +11,24 @@ module(PLUGINS_MODULE, ['ui.router']).config([
...
+      const pluginModule = await import(/* webpackIgnore: true */ pluginManifestLocation);
+
+      if (!pluginModule || !pluginModule.PLUGINS) {
+        throw new Error(`Error loading plugins.`);
+      } else {
+        // @ts-ignore
+        pluginModule.PLUGINS.forEach(plugin => pluginRegistry.register(plugin));
+      }
+    } catch (error) {
+      console.error('No plugins found.');
```
christopherthielen: This error handling doesn't make much sense to me. 

1) It swallows the original error
2) It will log 'no plugins found' if `pluginRegistry.register()` fails, which is misleading

I'd prefer to see something like:
```suggestion
      console.error('Error registering plugin manifests');
      console.error(error);
```

app/scripts/modules/core/src/plugins/plugin.module.ts
```diff
@@ -12,8 +11,24 @@ module(PLUGINS_MODULE, ['ui.router']).config([
...
+    // Grab all plugins that are defined in plugin-manifest
+    // The format for plugin-manifest would be:
+    //    const PLUGINS = [{'name':'myPlugin', 'version':'1.2.3', 'devUrl':'/plugins/index.js'}]
+    //    export { PLUGINS }
+    try {
+      const pluginManifestLocation = '/plugin-manifest.js';
+      const pluginModule = await import(/* webpackIgnore: true */ pluginManifestLocation);
+
+      if (!pluginModule || !pluginModule.PLUGINS) {
+        throw new Error(`Error loading plugins.`);
```
christopherthielen:
```suggestion
        throw new Error(`Expected plugin-manifest.js to contain an export named 'plugins' but it did not.`);
```

app/scripts/modules/core/src/plugins/plugin.module.ts
```diff
@@ -12,8 +11,24 @@ module(PLUGINS_MODULE, ['ui.router']).config([
...
+      } else {
+        // @ts-ignore
+        pluginModule.PLUGINS.forEach(plugin => pluginRegistry.register(plugin));
```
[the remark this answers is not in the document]
bpowell: doing that still gives an error `Parameter 'plugin' implicitly has an 'any' type.`

PR #7795, opened 2020-01-22
fix(aws): Fixing cross-app security group lookup

alanmquach:
Given two applications `foo` and `bar` with `foo` having server groups and `bar` having load balancers, we "lose" security groups in the corner case where `foo` server groups are registered to `bar` load balancers. In this scenario

- Clicking into the server group details for a `foo` server group while looking at `bar`'s load balancers, not all security groups of the `foo` server group may be shown in the details
- More dangerously, if we edit the security groups, we'll actually submit a task with the missing groups absent which will be interpreted as a request to remove them. This means we could cause a removal of security groups simply because deck couldn't find them.

This is because we only try to find them in `app.securityGroups.data` and ids that are not found (`undefined`) are `compact`ed away.

I've added 2 fallbacks to address this:

1. Try to find the security group in the `securityGroupsIndex` which should contain ALL security groups
2. As a last resort, construct a slim, unhydrated object so that we can still list the missing group in the UI and allow it to remain present in any edit actions.


alanmquach:
Correct, the `app` prop is the app that owns the load balancer not the server group.

> Are there other issues in this scenario, where a user navigates to a server group in a different application via the load balancers view?

Upon closer inspection (of server group details), it doesn't look like there are any other issues in this scenario. Most of the information for the server group details is returned in the request for the specific server group, with security groups being the only exception, where it only contains a reference and we're expected to join it with the actual data (from cache or otherwise) to hydrate it.

PR #7798, opened 2020-01-22
feat(codebuild): Add AWS CodeBuild Stage

Kaixiang-AWS:
Add UI support for new AWS CodeBuild Stage (spinnaker/spinnaker#5292)


clareliguori:
Can you add a screenshot of the UI with these changes?

app/scripts/modules/core/src/pipeline/config/stages/awsCodeBuild/awsCodeBuildStage.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { AwsCodeBuildStageConfig } from './AwsCodeBuildStageConfig';
+import { AwsCodeBuildExecutionDetails } from './AwsCodeBuildExecutionDetails';
+import { validate } from './AwsCodeBuildValidator';
+
+Registry.pipeline.registerStage({
+  label: 'AWS CodeBuild',
+  description: 'Trigger a build in AWS CodeBuild',
+  key: 'awsCodeBuild',
+  producesArtifacts: true,
```
clareliguori: How does the CodeBuild artifact get tied to a Spinnaker pipeline artifact?

Kaixiang-AWS:
Configuration page
Screen Shot 2020-01-30 at 9 50 41 AM

Execution detail. Note that S3 logs URL or CloudWatch logs URL will only show if it's enabled.
Screen Shot 2020-01-30 at 9 51 33 AM
Screen Shot 2020-01-30 at 9 57 57 AM

app/scripts/modules/core/src/pipeline/config/stages/awsCodeBuild/awsCodeBuildStage.ts
Kaixiang-AWS: CodeBuild artifact cannot be tied to Spinnaker artifact for now. I will need to implement it in later versions. The basic idea I have now is configuring a Spinnaker artifact as an output artifact override so that CodeBuild can upload artifact to that location. Then that Spinnaker artifact could be used in the future stages.
A better solution would be creating a Spinnaker artifact after build finishes and bind the output artifact to that newly created Spinnaker artifact. Not sure if this is doable. I will have to further look into both options.

clareliguori: This 'producesArtifacts' attribute might need to be false until artifact support is plumbed through

clareliguori:
What do you think of plumbing through a ListProjects API through gate/igor, so that the project name is a dropdown?

Kaixiang-AWS:
That is a good idea. I can add a TODO item and add this functionality later.

PR #7801, opened 2020-01-23
fix(artifacts): Fix find artifacts from execution stage

ezimanyi:
The find artifacts from execution stage is not able to load the widget for each artifact type; creating the controller is broken because the constructors are not annotated with dependency injection information and we are running in strict-di mode.

Fix this by extracting the anonymous controller for each artifact type into a function and annotating it with injection information. (I'm not aware of a way to annotate an anonymous function, except
by doing Object.assign(function() {}, { $inject: 'abc' }) which
seemed less readable.)


maggieneterval:
I thought we merged this before the 1.18 release branches were cut, but we didn't, so I'm going to cherry-pick this now to close https://github.com/spinnaker/spinnaker/issues/5450

maggieneterval:
@spinnakerbot :cherries: :pick: 1.18

PR #7804, opened 2020-01-23
config(core): Remove MPTV2 UI feature flag

louisjimenez:
The MPTV2 UI has been stable for some time and with this previous [PR ](https://github.com/spinnaker/deck/pull/7708) gained support for managing template versions. This PR removes the feature flag and exposes UI components including the pipeline template view.

Linked [Halyard PR](https://github.com/spinnaker/halyard/pull/1510)


app/scripts/modules/core/src/pipeline/details/SingleExecutionDetails.tsx
```diff
@@ -189,7 +183,7 @@ export class SingleExecutionDetails extends React.Component<
                       <span> stage durations</span>
                     </label>
                   </div>
-                  {showConfigButton && (
+                  {
```
maggieneterval: Can probably get rid of these brackets?

app/scripts/modules/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -241,7 +240,7 @@ export class ExecutionGroup extends React.PureComponent<IExecutionGroupProps, IE
             execution={execution}
             pipelineConfig={pipelineConfig}
             application={this.props.application}
-            onRerun={pipelineConfig && isConfigurable ? this.rerunExecutionClicked : undefined}
```
maggieneterval: Just trying to understand why we can remove the `isConfigurable` check -- it looks like `isConfigurable` checked 1) that the feature flag is on, and 2) if the config has the `v2` schema. Why do we no longer need to check 2?

louisjimenez: Yeah this is tricky. Previously, the isConfigurable const was checking that if there was a pipeline config that it would not be a v2 kind if the flag was off. If the flag was on, then it would let the pipeline config pass the check regardless of type. Going forward all pipeline configs can be treated as configurable.

app/scripts/modules/core/src/pipeline/details/SingleExecutionDetails.tsx
louisjimenez: Yes, thanks!

louisjimenez:
fyi that this flag is being removed, @christopherthielen @erikmunson

chihchenghuang:
https://www.spinnaker.io/guides/user/pipeline/pipeline-templates/

I followed the instruction in the document but didn't work.
The error message says:
`Was passed main parameter '--managed-pipeline-templates-v2-ui' but no main parameter was defined in your arg class`

I traced back here, and then just figured out that the flag is removed in this PR.
Can you update the document as well?

louisjimenez:
@chihchenghuang Thanks for letting me know, could you share what version of Spinnaker and Halyard you were using?

louisjimenez:
@chihchenghuang The ability to control that flag through Halyard has been restored in Halyard 1.31.1. When Spinnaker 1.19 is released in March, the UI will no longer be behind a feature flag and Halyard will be updated accordingly.

PR #7810, opened 2020-01-24
Import aliases eslint rule

christopherthielen:
This PR adds four new ESLint rules:

### import-from-alias-not-npm

Source code in a package (i.e., `core`) should not import from its own npm package `@spinnaker/core`.
Instead, it should use an alias `core/foo`

### import-from-npm-not-alias

Source code in a package (i.e., `amazon`) should not import from a different package using an alias (i.e., `core/`)
Instead, it should import from `@spinnaker/core`

### import-from-npm-not-relative

Source code in a package (i.e., `amazon`) should not import from package (i.e., `core`) using a relative path.
Instead, it should import from `@spinnaker/core`

### import-relative-within-subpackage

Source code in a package (i.e., `core/presentation` should not import from the same subpackage using an alias `core/presentation`.
Instead, it should import relatively `../../path/file`

The other two commits updates the violating sources.


app/scripts/modules/core/src/function/FunctionGroupings.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Application } from '@spinnaker/core';
+import { Application } from 'core';
```
christopherthielen: I'd like to write a rule that bans imports from a raw alias (without a subpackage), but that's trickier and can come later

packages/eslint-plugin/test/import-relative-within-subpackage.spec.js
```diff
@@ -0,0 +1,78 @@
...
+    {
+      filename: '/root/spinnaker/deck/app/scripts/modules/amazon/src/package/amazon_source_file.ts',
+      code: `import { Anything } from '../subpackage/foo';`,
+    },
+    {
+      filename: '/root/spinnaker/deck/app/scripts/modules/core/subpackage/nest/core_source_file.ts',
+      code: `import { Anything } from 'core/subpackage';`,
+    },
+    {
+      // kubernetes subpackages are nested under v1 or v2 directory so treat them differently
```
maggieneterval: Not for long! https://github.com/spinnaker/governance/blob/master/rfc/eol_kubernetes_v1.md

PR #7812, opened 2020-01-26
fix(deck): use label for parameters in manual execution if supplied

sgarlick987:
@maggieneterval @ethanfrogers 

fixes the lost functionality mentioned in https://github.com/spinnaker/spinnaker/issues/4890 during angular -> react migration by displaying supplied labels on the manual execution gui.

@maggieneterval I was tracking down how to get the labels to display in the execution details and tracked it down to https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/ApplicationController.groovy#L124 . The problem is, it looks like this is returning a list of linkedhashmaps and the parameters themselves are just a map of name->values for the parameters. Do you have any thoughts on how we could easily get the labels returned as well? Would it be good to push this to a later PR?

Lastly, I updated ManualExecutionModalPage and related fixtures to include some test for the parameters with and without labels. I'm not entirely sure if how I handled the fixture update and if its the best place for those checks, but let me know and I can update if needed.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/manualExecution/Parameters.tsx
+++ b/app/scripts/modules/core/src/pipeline/manualExecution/Parameters.tsx
@@ -49,7 +49,7 @@ export class Parameters extends React.Component<IParametersProps> {
           visibleParameters.map((parameter, i) => {
             const fieldProps = {
               name: formikFieldNameForParam(parameter),
-              label: parameter.name,
+              label: parameter.label || parameter.name,
               help: parameter.description && <HelpField content={parameter.description} />,
               fastField: false,
               required: parameter.required,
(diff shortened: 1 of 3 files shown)


sgarlick987:
For adding them to the execution details, if I understand right gate is just passing stuff through? So I'm really looking at https://github.com/spinnaker/orca/blob/master/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy#L577 and that lead me to https://github.com/spinnaker/orca/blob/master/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/Trigger.kt and its impls in that package. From looking that, adding the labels there doesn't seem to be that great of an option?

ethanfrogers:
@sgarlick987 I'd have to do a little bit of digging but I'm pretty sure you should have access to the pipeline config from the component (or parents) that renders the parameters in the execution details (@maggieneterval can probably confirm). If so, it should be as simple as doing a cross-reference when rendering the values.

PR #7816, opened 2020-01-28
fix(core/confirmationModal): Supply the 'source' location when canceling

christopherthielen:
This fixes a bug in Enable Server Group modal.

Screen Shot 2020-01-28 at 10 02 14 AM

When the user selects the "No, I just want to enable the server group", it was closing the dialog and doing nothing else.
Now, it closes the first dialog and opens the next dialog (but not if they click the X in the top right corner)

----

Also, fixed the boolean logic that shows the "Reason" input.

# before
Screen Shot 2020-01-28 at 10 05 49 AM

# after

Screen Shot 2020-01-28 at 10 02 20 AM


maggieneterval:
@christopherthielen this seems like something we should cherry-pick into the current release; do you have any reservations about that?

christopherthielen:
@maggieneterval definitely cherrypick this one

maggieneterval:

PR #7817, opened 2020-01-28
feat(amazon): Expose addresses for IPv6 instances

caseyhebebrand:
Amazon supports instances with IPv6 addresses, so Spinnaker should expose those addresses in the instance details section. 

IP addresses from the v6 family are all public. Also, multiple interfaces can be attached to a v6 instance, so each may have N number of IPv6 addresses. Here are some possible scenarios for an aws instance:
1. One Private IPv4 with one public IPv4 (currently supported)
2. One Private IPv4 with one or more IPv6
3. Only IPv6 addresses, maximum one per network interface
4. One private IPv4, one public IPv4, and one or more IPv6 address (rare edge case)

This PR enables support for the last 3 scenarios.

### DIFF
--- a/app/scripts/modules/amazon/src/instance/details/instanceDetails.html
+++ b/app/scripts/modules/amazon/src/instance/details/instanceDetails.html
@@ -223,7 +223,9 @@
           <copy-to-clipboard class="copy-to-clipboard copy-to-clipboard-sm" text="ip" tool-tip="'Copy to clipboard'">
           </copy-to-clipboard>
         </dd>
-        <dt ng-if="instance.publicIpAddress">Public IP Address</dt>
+        <dt ng-if="instance.publicIpAddress || instance.ipv6Addresses.length">
+          Public IP Address<span ng-if="instance.ipv6Addresses.length > 1">es</span>
+        </dt>
         <dd ng-if="instance.publicIpAddress">
           <a href="http://{{instance.publicIpAddress}}:{{state.instancePort}}" target="_blank"
             >{{instance.publicIpAddress}}</a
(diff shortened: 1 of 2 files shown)


app/scripts/modules/amazon/src/instance/details/instanceDetails.html
```diff
@@ -235,6 +237,11 @@ <h3 class="horizontal middle space-between flex-1" select-on-dbl-click>
           >
           </copy-to-clipboard>
         </dd>
+        <dd ng-if="instance.ipv6Addresses.length" ng-repeat="ip in instance.ipv6Addresses">
+          <a href="http://{{ip}}:{{state.instancePort}}" target="_blank">{{ip}}</a>
+          <copy-to-clipboard class="copy-to-clipboard copy-to-clipboard-sm" text="ip" tool-tip="'Copy to clipboard'">
```
christopherthielen: Is this doing anything?  I don't see how the value of the IP is being passed to `copy-to-clipboard`

christopherthielen:
If you have time, it would be nice to get the entire ```    <collapsible-section heading="DNS">
``` refactored to React

app/scripts/modules/amazon/src/instance/details/instanceDetails.html
caseyhebebrand: It is passed using `text="ip"`, where `ip` comes from the `ng-repeat`.

caseyhebebrand:
> If you have time, it would be nice to get the entire ```

Can you explain? I think something wasn't copied here

christopherthielen:
@caseyhebebrand whoops, formatting error!  I was checking if you had time to convert the section to react.

caseyhebebrand:
> @caseyhebebrand whoops, formatting error! I was checking if you had time to convert the section to react.

@christopherthielen Yes! Do you mind if I create a subsequent PR? I was thinking of making a `LinkWithClipboard` component and then replacing all the occurrences in the details pane with that. This way I can isolate this feature from that enhancement.

christopherthielen:
sounds good.   FYI we already have a [CopyToClipboard](https://github.com/spinnaker/deck/blob/8e50c2e2c7d987a6c988f6f5d20657a2ca0a1b56/app/scripts/modules/core/src/utils/clipboard/CopyToClipboard.tsx#L23) component

PR #7818, opened 2020-01-28
feat(core): enable the "Override Helm chart artifact" checkbox in man…

dbyron0:
…ual execution by default

Since this is manual execution, it seems like a safe assumption that people want to
specify the helm chart to use.  Leaving it unchecked either leads to pipeline failure, or
to using an artifact from a prior execution (if enabled), both of which are surprising.

before:
helm_chart_override_unchecked

after:
helm_chart_override_checked


dbyron0:
Thanks @maggieneterval !

PR #7819, opened 2020-01-28
fix(amazon): Prevent error if target group is undefined

caseyhebebrand:
The instance details panel was showing "Instance not found" when an instance in application A was attached to a load balancer in application B. Throwing `error: group is undefined`. 

The bug causing the group to be `undefined` was fixed in a prior PR, so this PR adds an extra layer of safety if the `group` is not found.

Screen Shot 2020-01-28 at 12 40 57 PM

### DIFF
--- a/app/scripts/modules/amazon/src/instance/details/utils.ts
+++ b/app/scripts/modules/amazon/src/instance/details/utils.ts
@@ -16,7 +16,7 @@ export const applyHealthCheckInfoToTargetGroups = (
   healthMetrics.forEach(metric => {
     if (metric.type === 'TargetGroup') {
       metric.targetGroups.forEach((tg: ITargetGroup) => {
-        const group = targetGroups[tg.name];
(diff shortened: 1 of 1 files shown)


app/scripts/modules/amazon/src/instance/details/utils.ts
```diff
@@ -16,7 +16,7 @@ export const applyHealthCheckInfoToTargetGroups = (
   healthMetrics.forEach(metric => {
     if (metric.type === 'TargetGroup') {
       metric.targetGroups.forEach((tg: ITargetGroup) => {
-        const group = targetGroups[tg.name];
+        const group = targetGroups[tg.name] || ({} as ITargetGroup);
```
christopherthielen: might as well use the fancy new hotness

```suggestion
        const group = targetGroups[tg.name] ?? ({} as ITargetGroup);
```

PR #7822, opened 2020-01-29
feat(infrastructure/buttons): Add a new property configurable on sett…

jorgebee65:
Add a new property configurable on `settings-local.yml` to enable/disable the create **Kubernetes** infrastructure buttons (Create: Server Group, Load Balancer and Firewall).

The property:
```window.spinnakerSettings.kubernetesAdHocInfraWritesEnabled = false;```

Behavior:
createKubernetesInfrastructure = false
**Case 1**: The application contains just _Kubernetes_ provider then the button is disabled.
**Case 2**: The application contains more than one provider and one of them is _Kubernetes_, not disable the button but doesn't allow select kubernetes from the modal selector window.

createKubernetesInfrastructure = true
The button is enabled and allows select kubernetes in case the application uses it and has more than one provider.

### CODE REVIEW

**Comment by louisjimenez** [comment]
This brings up some good questions around how extensible this no infra edits feature should be. The practice of avoiding manual edits to infrastructure could be relevant to other providers outside of Kubernetes. So much so, that it may make sense to have the feature flag support turning this off across all providers. Another approach to throw out there, would be to have this tied at the application level with a default set to off.

As a general principal, changes I am making for new Kubernetes objects (e.g. Autoscalers) will not be including edits options in the Infra UI.

I agree with @maggieneterval that removing the buttons altogether would make for a clearer UX.

Will this feature flag be configurable via Halyard?

**Comment by maggieneterval** [comment]
> Another approach to throw out there, would be to have this tied at the application level with a default set to off.

:+1: Sounds like there's enough ambiguity around the ideal configuration level that this might be worth running by the Kubernetes SIG one more time, @jorgebee65 we'd love to have you join us! I'll add to the agenda for next meeting.

**Comment by jorgebee65** [comment]
**Hide** Buttons instead disable them. 
Add behavior for the **Edit** dropdown options in both Kubernetes Versions:

**V1**
- Clusters: (the Edit option does not exist so nothing changes)
- Load Balancer
- Firewall 

**V2**
- Cluster
- Load Balancer
- Firewall

@maggieneterval I'll be glad to join to that meeting, please let me know the details.

**Comment by maggieneterval** [comment]
Thanks for the update, @jorgebee65! The next meeting of the Kubernetes SIG is on Tuesday 2/25 at 1PM EST (meeting details [here](https://github.com/spinnaker/governance/blob/master/sig-kubernetes/README.md)). Would be great if you could quickly summarize these changes and poll the group on whether this behavior should be a global setting or configurable per-application (or at some other level that I'm not anticipating). Looking forward to chatting; thanks again for all your work here 🤗

**Comment by jorgebee65** [comment]
hi @maggieneterval as we discuss these are the changes: Global flag, configurable by **settings-local.js**, by default is disabled unless the property is true, apply to create, edit and delete. Discard changes for v1 just applicable for v2.

**Comment by jorgebee65** [comment]
The idea of this feature is to allow Spinnaker enable or disable: Clusters, LoadBalancers and Firewalls sections for each provider. (In this case I tested it with Kubernetes and AWS) 

So the configuration properties are: 
```
window.spinnakerSettings.adHocInfraWritesK8sCluster = true;
window.spinnakerSettings.adHocInfraWritesAWSCluster = true;

window.spinnakerSettings.adHocInfraWritesK8sLoadBalancer = true;
window.spinnakerSettings.adHocInfraWritesAWSLoadBalancer = false;

window.spinnakerSettings.adHocInfraWritesK8sFirewall = false;
window.spinnakerSettings.adHocInfraWritesAWSFirewall = false;
```
This way we can tell to spinnaker which buttons should been showed and also the selectable options, an example using Cluster and assuming we have enabled two providers: Kubernetes and AWS.

**Case 1: both enabled**
If we set both properties equals true then the Create Server Group button is going to be shown.
```
window.spinnakerSettings.adHocInfraWritesK8sCluster = true;
window.spinnakerSettings.adHocInfraWritesAWSCluster = true;
```
button enabled

And when clic over the button then shows both options Kubernetes and AWS:

selectable

**Case 2: Just one property is enabled**
If we set just one of the properties equals true the Create Server Group button is going to be shown but the difference is due just one provider is enabled then the button is going to open the modal panel according to the enabled Provider (in this example AWS modal).
```
    window.spinnakerSettings.adHocInfraWritesK8sCluster = false;
    window.spinnakerSettings.adHocInfraWritesAWSCluster = true;
```
aws default

**Case 3: both disabled**
If we set both properties equals false then the Create Server Group button is NOT going to be shown.
```
    window.spinnakerSettings.adHocInfraWritesK8sCluster = false;
    window.spinnakerSettings.adHocInfraWritesAWSCluster = false;
```
disabled

**Comment by jorgebee65** [comment]
Hi @maggieneterval , I just made some modifications related with the unit/functional tests in order to handle the new buttons behavior, also add the GCE provider so the properties should be as follows: 
 
```
window.spinnakerSettings.adHocInfraWritesGCECluster = true;
window.spinnakerSettings.adHocInfraWritesGCELoadBalancer = false;
window.spinnakerSettings.adHocInfraWritesGCEFirewall = false;
```

**Comment by maggieneterval** [comment]
Hi @jorgebee65, thanks for the update. I have to admit I'm a bit confused as to why we've expanded the scope of this change to include exposing feature flags for the AWS and GCE providers, and for each type of infrastructure -- I thought we had discussed at the Kubernetes SIG adding a single Kubernetes-specific feature flag that would gate all ad-hoc applies/edits/deletes?

**Comment by jorgebee65** [comment]
@maggieneterval you are absolutely right, I got a little confused because of a previous comment but now is fixed to: One single flag just for Kubernetes which affect the create/edit/delete behavior.  The other providers are not been affected and the behavior still be the same as usual.

**Comment by maggieneterval** [comment]
@jorgebee65 I pulled your branch down to test and I'm not sure this is working as expected: I have an application with two accounts, one Kubernetes V2 account and one GCE account, and when I click "Create Server Group" from the Infrastructure/Clusters tab, I am getting the Kubernetes V1 Create Server Group modal.

File: app/scripts/modules/core/src/config/settings.ts [hunk 1]
```diff
@@ -136,6 +136,7 @@ export interface ISpinnakerSettings {
   searchVersion: 1 | 2;
   triggerTypes: string[];
   useClassicFirewallLabels: boolean;
+  adHocInfraWrites: boolean;
```
maggieneterval, line 139: Do we want to make this flag name Kubernetes-specific, since it only affects the Kubernetes provider? Perhaps we could use something like `kubernetesAdHocInfraWritesEnabled`, which is perhaps a mouthful but at least descriptive?

app/scripts/modules/core/src/cloudProvider/providerSelection/ProviderSelectionService.ts
```diff
@@ -47,4 +58,26 @@ export class ProviderSelectionService {
       return provider;
     });
   }
+
+  public static isDisabled(app: Application, feature: string): boolean {
+    let isDisabled = true;
+    const BreakException = {};
```
maggieneterval, line 64: Hmm, what are we trying to do here? I don't love the use of a custom exception for the purpose of just breaking out of a loop...I'm also a bit confused as to why we are checking if the provider is `kubernetes` on line 68.

app/scripts/modules/kubernetes/src/v2/kubernetes.v2.module.ts
```diff
@@ -104,6 +109,7 @@ module(KUBERNETES_V2_MODULE, [
       detailsController: 'kubernetesV2ServerGroupDetailsCtrl',
       detailsTemplateUrl: require('./serverGroup/details/details.html'),
       transformer: 'kubernetesV2ServerGroupTransformer',
+      infra: SETTINGS.adHocInfraWrites,
```
maggieneterval, line 112: Why are we adding this `infra` property to each type of entity rather than just setting it as a top-level property of the provider? Also can we be a bit more descriptive than just `infra`?

app/scripts/modules/kubernetes/src/v1/kubernetes.v1.module.ts
```diff
@@ -112,6 +112,7 @@ module(KUBERNETES_V1_MODULE, [
       detailsController: 'kubernetesLoadBalancerDetailsController',
       createLoadBalancerTemplateUrl: require('./loadBalancer/configure/wizard/createWizard.html'),
       createLoadBalancerController: 'kubernetesUpsertLoadBalancerController',
+      infra: SETTINGS.adHocInfraWrites,
```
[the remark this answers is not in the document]
jorgebee65: Correct, the properties have been removed from here.

app/scripts/modules/core/src/config/settings.ts
jorgebee65: Changed

app/scripts/modules/core/src/cloudProvider/providerSelection/ProviderSelectionService.ts
jorgebee65: Refactored to use the methods **hasValue** and **getValue** instead.

app/scripts/modules/kubernetes/src/v2/kubernetes.v2.module.ts
jorgebee65: Fixed

jorgebee65:
@maggieneterval I modified the code according your comments, would you take a look?

maggieneterval:
Thanks @jorgebee65! I'll take a closer look next week but also looping in @danielpeach to review this as well.

maggieneterval:
Going to close this for now as @jorgebee65 is no longer working on Spinnaker, but if anyone would like to push this across the finish line, please feel free to re-open!

jorgebee65:
@maggieneterval I'm back, I reopened this PR and made the fixes removing the Kubernetes v1 provider.

kevinawoo (approved):
LGTM!

jorgebee65:
Hi @maggieneterval did you have the chance to review this changes?

kevinawoo (approved):
LGTM, going to merge this in.

PR #7824, opened 2020-01-29
fix(gce logo): Use a proper GCE logo

plumpy:
We were using a GCP logo which is ambiguous since AppEngine and GKE are
both also part of GCP.


duftler:
Can you please include a screen-grab of what it looks like in the dialog?

plumpy:
Before:
before

After:
after

plumpy:
> Looks like this is large increase in image size (7.19kb to 67.9kb). Is the higher res image needed? Also, new logo is snazzy

The new logo was 150x98, so was getting upscaled (at least on my monitor) and looking pretty crappy. I shrunk this one down to 512x512 (and fixed a few other things, like it was weirdly off-center) so now it's about 27k. Thanks for noticing.

PR #7826, opened 2020-01-31
feat(k8s): ManifestStatus can now be overridden

bpowell:


app/scripts/modules/kubernetes/src/v2/pipelines/stages/deployManifest/manifestStatus/DeployStatus.tsx
```diff
@@ -42,6 +43,7 @@ export interface IDeployStatusState {
   manifestIds: string[];
 }
 
+@Overridable('kubernetes.v2.pipeline.stages.deploymanifest.deploystatus')
```
ethanfrogers: Small nit - could you capitalize `deployManifest` and `deployStatus`?

maggieneterval:
@bpowell not super familiar with overridable components either; how does this work? Can you please update the PR description with 1) the problem this is trying to solve and 2) an example of using this including the React component you're overriding the default with and how it's wired up? Thanks!

bpowell:
@maggieneterval as for 2) There is an example in the code for overriding components. https://github.com/spinnaker/deck/blob/b6aabe18a2c71f194087c01fd15ec369460f5e70/app/scripts/modules/core/src/overrideRegistry/Overridable.tsx#L17

What this PR is trying to do is we would like to be able to change out that component with a different one (similar to how the header works or the [CreatePipelineModel](https://github.com/spinnaker/deck/blob/b6aabe18a2c71f194087c01fd15ec369460f5e70/app/scripts/modules/core/src/pipeline/create/CreatePipelineModal.tsx#L66) works)

maggieneterval:
Okay, thanks!

maggieneterval:
@bpowell Can you please let me know if you still need this? I am refactoring these components so whatever custom override you added will likely not work as expected after that refactor. I'll ping you on the refactor PR as well. Thanks!

bpowell:
@maggieneterval i'm not sure if this is still needed. might want to reach out to @claymccoy or @link108 to see

maggieneterval:
Thanks, @bpowell! @claymccoy @link108, please let me know if Armory is depending on any custom overrides for Kubernetes stage execution details components; I'm refactoring them as part of some [performance improvements](https://github.com/spinnaker/spinnaker/issues/5909) and don't want to break you!

PR #7827, opened 2020-02-03
chore(builds): Expand GHActions CI to master and release branches

ttomsu:


ttomsu:
I have no idea why this change would cause the build to fail. Help me @christopherthielen , you're my only hope.

ttomsu:
I don't know what happened here. Closing.

PR #7830, opened 2020-02-03
fix(amazon): parse healthCheckPort as integer before using it

vigneshm:
We [rely on `isNumber` check on `healthCheckPort`](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/amazon/src/instance/details/utils.ts#L20), but it is sent back as `string` by the back-end. Also, we [expect](https://github.com/spinnaker/deck/blob/0561762b9cd8131b75f5befb255637deeec3f400/app/scripts/modules/amazon/src/domain/IAmazonLoadBalancer.ts#L138) a mix of string and number as `healthCheckPort`.

In this PR, I'm parsing the value as integer before using it as the port number.


app/scripts/modules/amazon/src/instance/details/utils.ts
```diff
@@ -17,7 +17,7 @@ export const applyHealthCheckInfoToTargetGroups = (
     if (metric.type === 'TargetGroup') {
       metric.targetGroups.forEach((tg: ITargetGroup) => {
         const group = targetGroups[tg.name] ?? ({} as ITargetGroup);
-        const port = group.healthCheckPort && isNumber(group.healthCheckPort) ? group.healthCheckPort : group.port;
+        const port = parseInt(group.healthCheckPort as any, 10) ?? group.port;
```
caseyhebebrand: Can you be more explicit about the type of `group.healthCheckPort` instead of using `any`?

vigneshm: The type of `healthCheckPort` in its current form doesn't allow us to use `parseInt`. The core issue here is mismatch of the type definitions between front-end and back-end. In typescript, we are able to provide a more refined definition of `number | 'traffic-port'` whereas, *I guess*,  java doesn't allow such a type definition?

It makes a lot of sense to keep the type as `number | 'traffic-port'` in the front-end, but that makes it hard to use `parseInt`. I could do the following to make it work but I chose the other way since it is more compact.

```js
let port = group.port;

if (group.healthCheckPort != null && group.healthCheckPort !== "traffic-port") {
  port =
    typeof group.healthCheckPort === "string"
      ? parseInt(group.healthCheckPort, 10)
      : group.healthCheckPort;
}
```

I'll update it to the above to make it more explicit to avoid any confusion.

app/scripts/modules/amazon/src/instance/details/utils.ts
```diff
@@ -17,7 +17,12 @@ export const applyHealthCheckInfoToTargetGroups = (
     if (metric.type === 'TargetGroup') {
       metric.targetGroups.forEach((tg: ITargetGroup) => {
         const group = targetGroups[tg.name] ?? ({} as ITargetGroup);
-        const port = group.healthCheckPort && isNumber(group.healthCheckPort) ? group.healthCheckPort : group.port;
+        let port = group.port;
```
christopherthielen: I think this could be simpler and easier to understand the intent like this:
```
const useTrafficPort = group.healthCheckPort === 'traffic-port' || isNil(group.healthCheckPort);
const port = useTrafficPort ? group.port : group.healthCheckPort;
```

I think we don't really need to parseInt because we're only converting back to a string in a subsequent line
```
        tg.healthCheckPath = `:${port}${group.healthCheckPath}`;
```

PR #7834, opened 2020-02-04
chore(builds): Adds GHActions as CI job (PRs, master, release branches

ttomsu:
In the process of doing this, I added a dependency on [Puppeteer](https://github.com/puppeteer/puppeteer), which is the Chrome Dev Tools team-maintained version that pulls in everything needed to run Chrome in CI (see https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer).

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -40,8 +40,8 @@ module.exports = function(config) {
     },
 
     customLaunchers: {
-      Chrome_travis_ci: {
-        base: 'Chrome',
+      ChromeCI: {
+        base: 'ChromeHeadless',
         flags: ['--no-sandbox'],
       },
       ChromeActive: {
@@ -66,7 +66,7 @@ module.exports = function(config) {
     // web server port
     port: 8081,
 
-    browsers: [process.env.TRAVIS ? 'Chrome_travis_ci' : 'ChromeActive'],
+    browsers: [process.env.TRAVIS || process.env.GITHUB_ACTIONS ? 'ChromeCI' : 'ChromeActive'],
 
     colors: true,
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,75 @@
+name: Deck CI
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+      - release-*
+
+env:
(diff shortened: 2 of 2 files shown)


maggieneterval:
Oh, this is so exciting! :tada: 

@christopherthielen had a [WIP branch with similar changes](https://github.com/spinnaker/spinnaker/issues/5230#issuecomment-571265814), the most notable difference that I can see being that Production Build --> Build NPM Packages calls a custom script to build the NPM modules in the correct order.

ttomsu:
I reached out to him asking for the status but got this working before he responded. 

I'm willing to defer the guts of the build to him, but a.) this is working now (that WIP branch is ~30 days old), and b.) it conforms to the patterns set on the rest of the services (namely that it works on both `pull_request`s and `push`es to important branches.

@maggieneterval Would you be opposed to approving this now and modifying later as he finishes up work on the other branch?

karma.conf.js
```diff
@@ -3,6 +3,8 @@
 const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 const { TypedCssModulesPlugin } = require('typed-css-modules-webpack-plugin');
 
+process.env.CHROME_BIN = require('puppeteer').executablePath();
```
christopherthielen: are you sure this is necessary?  I didn't have any trouble with chrome headless on GHA.  I'd prefer to avoid adding another dependency if possible.

package.json
```diff
@@ -72,6 +72,7 @@
     "ngimport": "^0.6.0",
     "pretty-ms": "^5.0.0",
     "prop-types": "15.6.1",
+    "puppeteer": "^2.1.0",
```
christopherthielen: are you sure this is necessary? I didn't have any trouble with chrome headless on GHA. I'd prefer to avoid adding another dependency if possible.

ttomsu: I did not realize GHA has it's own, so you're right that this is not needed. Removed.

ttomsu:
I explicitly deleted that last step because that `build_modules.sh` file didn't exist (it obviously exists in your branch). You're welcome to append that last step (and all of the other 12 commits worth of changes) and we can be free of Metal Travis' tyranny.

PR #7836, opened 2020-02-05
fix(core): Use "Authorization" header instead of query parameter to get changelog (what's new) from gist

dragon3:
## WHAT

This pull request changes GItHub authentication method from query parameter to `Authorization` header for getting changelog from Gist.

## WHY

GitHub is deprecating authentication to the GitHub API using query parameters.
https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters

And it seems they just started sending email `[GitHub API] Deprecation notice for authentication via URL query parameters` with the following message to notify the deprecation when we use query parameter authentication 😅 

```
Hi <REDACTED>,

On February xx, 2020 at xx:xx (UTC) your personal access token (<REDACTED>) using Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.130 Safari/537.36 was used as part of a query parameter to access an endpoint through the GitHub API:

https://api.github.com/gists/0df1df27e7c4c8e2784a0a0eea750a99



Please use the Authorization HTTP header instead, as using the `access_token` query parameter is deprecated.

Depending on your API usage, we'll be sending you this email reminder once every 3 days for each token and User-Agent used in API calls made on your behalf.
Just one URL that was accessed with a token and User-Agent combination will be listed in the email reminder, not all.

Visit https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters for more information.

Thanks,
The GitHub Team
```

https://api.github.com/gists/0df1df27e7c4c8e2784a0a0eea750a99 is the gist for Spinnaker changelog (What's New).


dragon3:
I confirmed this change worked properly:

### Without access token
Could access the gist with low API rate limit:

without-access-token

### With access token
Could access the gist with high API rate limit because of the access token:

with-access-token

dragon3:
@maggieneterval 
Thank you for the review!

I'm looking forward to the new version with this change 😄

PR #7842, opened 2020-02-06
feat(plugins): Load plugins and plugin manifest from gate

christopherthielen:
Load manifest from gate URL: `/plugins/deck/plugin-manifest.json`
Continue to load pre-baked manifest from the local deck asset `/plugin-manifest.js`

*BREAKING CHANGE*: IPluginManifest renamed to IPluginMetaData and shape changed
from: `{ name, version, devUrl }` to: `{ id, version, url }`
*Migrate plugin-manifest.js to use `id` and `url`*

IPluginManifest now describes the shape of plugin-manifest.js

*BREAKING CHANGE*: Plugins are no longer loaded from settings.js or settings-local.js
*Migrate pre-baked plugin metadata to plugin-manifest.js*

### DIFF
--- a/app/scripts/modules/core/src/config/settings.ts
+++ b/app/scripts/modules/core/src/config/settings.ts
@@ -1,4 +1,3 @@
-import { IPluginManifest } from 'core/plugins/plugin.registry';
 import { cloneDeep, merge } from 'lodash';
 
 export interface IAdditionalHelpLinks {
@@ -130,7 +129,6 @@ export interface ISpinnakerSettings {
   providers?: {
     [key: string]: IProviderSettings; // allows custom providers not typed in here (good for testing too)
   };
-  plugins: IPluginManifest[];
   pubsubProviders: string[];
   quietPeriod: [string | number, string | number];
   resetProvider: (provider: string) => () => void;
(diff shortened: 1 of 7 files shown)


christopherthielen:
@bpowell

christopherthielen:
one oddity is the difference between gate's `plugin-manifest.json`:
```
[ { id: 'foo', version: '1.0' } ]
```
 and deck's `plugin-manifest.js`
```
{ 
  plugins: [ { id: 'foo', version: '1.0', url: '/plugins/foo/1.0/index.js' } ]
}
```

We may want to consider switching deck's baked-in manifest format so that it matches gate.  This would mean switching to a `.json` file and eliminating the top-level `plugins: []` (which only exists to avoid doing a default export from the javascript version)

jonsie:
@christopherthielen If we switch to gate's plugin-manifest.json schema would the URL path (i.e. `/plugins/foo/1.0/index.js`) no longer be necessary to source from config?  Are we essentially going on convention that the JS asset to load is always `pluginId/pluginVerison/index.js` ?

christopherthielen:
@jonsie The `url` in the manifest from a deck asset is still useful.  I think some operators may not use gate to proxy deck assets and want to use an artifact store.  I also think using a `url` from the manifest is important for local development.

bpowell:
code looks good. just have a question about gate. has gate been updated yet for the API endpoint or is that work still in progress?

christopherthielen:
> code looks good. just have a question about gate. has gate been updated yet for the API endpoint or is that work still in progress?

The gate endpoints are live on the master branch.  It's hidden behind a feature flag.

PR #7847, opened 2020-02-07
react(core): Introduce LinkWithClipboard component

caseyhebebrand:
A repeated code pattern in angular is the use of a link with a clipboard next to it. This component turns that code pattern into a react component and implements it on the instance details panel.

Screen Shot 2020-02-06 at 5 46 00 PM

### DIFF
--- a/app/scripts/modules/core/src/presentation/LinkWithClipboard.tsx
+++ b/app/scripts/modules/core/src/presentation/LinkWithClipboard.tsx
@@ -6,15 +6,11 @@ export interface ILinkWithClipboardProps {
   text: string;
 }
 
-export const LinkWithClipboard = ({ url, text }: ILinkWithClipboardProps) => {
-  // eslint-disable-next-line
-  console.log('here', url, text);
-  return (
-    <>
-      <a href={url} target="_blank">
-        {text}
-      </a>
-      <CopyToClipboard className="copy-to-clipboard copy-to-clipboard-sm" text={text} toolTip="'Copy to clipboard'" />
-    </>
-  );
-};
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/presentation/index.ts
```diff
@@ -7,6 +7,7 @@ export * from './hooks';
 export * from './HoverablePopover';
 export * from './LabelComponent';
 export * from './labeledValues';
+export * from './linkWithClipboard.component';
```
vigneshm: Should we make the react component available as well?

app/scripts/modules/core/src/presentation/LinkWithClipboard.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { CopyToClipboard } from 'core/utils';
+
+export interface ILinkWithClipboardProps {
+  url: string;
+  text: string;
+}
+
+export const LinkWithClipboard = ({ url, text }: ILinkWithClipboardProps) => (
+  <>
+    <a href={url} target="_blank">
```
vigneshm: Would `target="_blank"` be a reasonable expectation for users or should we make it configurable?

app/scripts/modules/core/src/presentation/index.ts
caseyhebebrand: Good call, I will add that for future use!

app/scripts/modules/core/src/presentation/LinkWithClipboard.tsx
caseyhebebrand: I think this is a reasonable expectation for users. This component will be used mainly for external links, in which case we would want a new page.

PR #7854, opened 2020-02-08
feat(package): auto-approve and merge pure package bumps

christopherthielen:
This needs a `GITHUB_TOKEN`  ~added to Deck's Settings Tab which can only be done by a member of the TOC.~
~https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets#about-encrypted-secrets~

apparently this token is provided by default.

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token

### DIFF
--- a/.github/workflows/approve_package_bumps.yml
+++ b/.github/workflows/approve_package_bumps.yml
@@ -1,0 +1,46 @@
+name: Deck CI
+
+on:
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


christopherthielen:
I merged this (and it didn't work because Github Actions bot doesn't have permission)
See https://github.com/spinnaker/deck/pull/7859#issuecomment-583877215

PR #7859, opened 2020-02-09
chore(purebump): use SPINNAKERBOT_TOKEN in github actions to auto-approve pure bumps

christopherthielen:


christopherthielen:
looks like github actions bot doesn't have permission to approve a PR.  We probably either need to either:
- add the GHA bot to an approvers list
- install an auth token for spinnakerbot and use that in the action (preferred?)

PR #7860, opened 2020-02-09
feat(deck): show expression in check precondition execution details

sgarlick987:
@maggieneterval @ethanfrogers this PR closes https://github.com/spinnaker/spinnaker/issues/5097 by showing the expression in the check precondition stage execution details .

When doing this I also noticed the expression itself was being escaped when displayed (eg parameters["not"] != "hello" -> parameters[\\"not\\"] != \\"hello\\"). This looks to be due to using JSON.stringify and the values being strings not objects. I didn't see any reason this would be needed otherwise, so I removed it. Let me know if you think it should stay and I can drop that commit.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/checkPreconditions/CheckPreconditionsExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/checkPreconditions/CheckPreconditionsExecutionDetails.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { get } from 'lodash';
+import { get, isBoolean, isString } from 'lodash';
 
 import { StageFailureMessage } from 'core/pipeline';
 
@@ -21,7 +21,9 @@ export function CheckPreconditionsExecutionDetails(props: IExecutionDetailsSecti
               .map(key => (
                 <div key={key}>
                   <dt>{robotToHuman(key)}</dt>
-                  <dd>{JSON.stringify(context[key])}</dd>
+                  <dd>
+                    {isString(context[key]) || isBoolean(context[key]) ? context[key] : JSON.stringify(context[key])}
+                  </dd>
                 </div>
               ))}
             <div>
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/checkPreconditions/CheckPreconditionsExecutionDetails.tsx
```diff
@@ -17,11 +17,11 @@ export function CheckPreconditionsExecutionDetails(props: IExecutionDetailsSecti
         <div className="col-md-12">
           <dl className="dl-horizontal">
             {Object.keys(context)
-              .filter(key => !['expression', 'failureMessage'].includes(key) && context[key] !== null)
+              .filter(key => !['failureMessage'].includes(key) && context[key] !== null)
               .map(key => (
                 <div key={key}>
                   <dt>{robotToHuman(key)}</dt>
-                  <dd>{JSON.stringify(context[key])}</dd>
+                  <dd>{context[key]}</dd>
```
maggieneterval: I think the reason we are stringifying the value here is that not all context values are guaranteed to be strings. For example, I think `context.regions` will be a list when the precondition you are checking in this stage is the size of a cluster, rather than an expression. Ideally we would strongly type each potential context value so we would know how to display it here, but that's probably outside the scope of what you're trying to do here. Perhaps you could simply check if `context.expression` is defined and print it if so, outside this logic that is simply mapping over everything on context. Let me know if that makes sense.

sgarlick987: Got it, I thought that might be the case.  

Since both the expression and expression result are strings here, and maybe others, would it be fine to do something like

`['expression', 'expressionResult', 'someother'].includes(key) ? context[key] : JSON.stringify(context[key])`

or 

`typeof context[key] === 'string' ? context[key] : JSON.stringify(context[key])`

The first isn't greatest since you'd need to update the list if adding new string kvs, but would at least keep the text clean until more refactoring is done. The second I would think would work too, but my js is pretty rusty so not sure if that would have other issues.

Let me know what you think and I can test/update PR later tonight.

sgarlick987: Had a free moment to test this, do you see any problems with this?

`isString(context[key]) || isBoolean(context[key]) ? context[key] : JSON.stringify(context[key])`

That should cover at least the expression precondition text, and would get picked up by any of the others that might have string/bool values?

maggieneterval: Was hoping some more elegant solution would occur to me but this should at least both solve the problem you set out to solve (rendering the original expression in the execution details) and not break rendering of any non-primitive context values in this component. :+1:

PR #7861, opened 2020-02-10
feat(codebuild): Add more options to AWS CodeBuild stage

Kaixiang-AWS:
Add 3 optional fields for AWS CodeBuild stage. First is source field. Users can
select a pre-existing Spinnaker artifact as the source or create an artifact from
execution context. Second is source version field. Users can input the version of
the source. This version could be the commit ID, branch name or tag name. Third is
environment variables. The environment variables provided will be propagated into
the build.


Kaixiang-AWS:
Not override source
Screen Shot 2020-02-10 at 11 12 43 AM
Override source with artifact
Screen Shot 2020-02-10 at 11 14 15 AM
Override source with artifact from execution context
Screen Shot 2020-02-10 at 11 12 55 AM

clareliguori:
This can be in a separate PR, but it would be helpful to have those little help text widgets next to the input fields (like you see for "URL" under Source Artifact Override, so you can more clearly explain what each field is for.

Kaixiang-AWS:
> This can be in a separate PR, but it would be helpful to have those little help text widgets next to the input fields (like you see for "URL" under Source Artifact Override, so you can more clearly explain what each field is for.

I will have a PR for help text after I finish all UI changes.

PR #7862, opened 2020-02-10
fix(kubernetes): Shorten displayed digest to the first 7 characters i…

ttomsu:
…n clusters tab

This has annoyed me for a long time - displaying the entire digest hash causes wrapping and general ugliness. Clipping it to just 7 characters enables the user to see the important bits quicker.

This only clips it in the Clusters tab. The details section keeps the whole digest. Preview:

VhZUtzoDE68

I'm new to Typescript, so I have no idea if this is the "proper" way to do things in TSX, so please correct me where needed.


christopherthielen:
This might be better handled using CSS

app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
```diff
@@ -72,7 +72,16 @@ export class ImageList extends React.Component<IServerGroupHeaderProps, IImageLi
   }
 
   public render() {
-    const images = this.props.images.sort();
+    const shorten = function(val: string): string {
+      const sha = val.lastIndexOf('sha256:');
+      if (sha > 0) {
+        // 14 = "sha256:" + 7 characters of hash
+        return val.substr(0, sha + 14) + '...';
```
maggieneterval: I think `String.prototype.substr` is [considered legacy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) so let's use `String.prototype.substring` if possible.

app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
```diff
@@ -72,7 +72,16 @@ export class ImageList extends React.Component<IServerGroupHeaderProps, IImageLi
   }
 
   public render() {
-    const images = this.props.images.sort();
+    const shorten = function(val: string): string {
+      const sha = val.lastIndexOf('sha256:');
+      if (sha > 0) {
```
maggieneterval: Since `lastIndexOf` will return -1 if no match is found, we should probably use `> -1` as the condition here, right?

ttomsu:
@christopherthielen If there is some CSS way to do this, I'm more than happy to defer to you. I could stumble around trying to figure out this "is it modern?" TypeScript + less + webpack setup, but I bet you could fix it in a fraction of the time. Would you mind taking a stab at it?

robzienert:
@anotherchrisberry is on parental leave. cc @christopherthielen @alanmquach - any hopes of taking a look at this?

ethanfrogers:
I'm not sure if this would work as is or take some reworking of the elements but `text-overflow: ellipsis` will automatically truncate the any overflow and add `...`.

app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
christopherthielen:
```suggestion
        return val.substring(0, sha + 14) + '...';
```

app/scripts/modules/core/src/serverGroup/ServerGroupHeader.tsx
christopherthielen:
```suggestion
      if (sha > -1) {
```

christopherthielen:
After revisiting this and discussing with Travis, I think his idea is the better experience for users.

ttomsu:
Moved to `ServerGroup`

PR #7867, opened 2020-02-11
fix(amazon/availabilityZone): Simplify state management for availability zone selection

vigneshm:
The primary purpose of this PR is to remove the use of dropdown selector to select between manual and automatic selection of availability zones. We do not have a flag to determine manual vs automatic selection, instead we use the selected availability zones and the pre-configured preferred zones to determine that. This confuses users when they see automatic selected even though they chose manual (but ended up selecting preferred zones).

The new UI treatment looks the following way.

availbilityzones

I also used this opportunity to fix a few bad patterns that might have crept in due to angular to react migration. The state management for availability zone selection is spread across `AvailabilityZoneSelector.tsx`, `ServerGroupZones.tsx`, `serverGroupConfiguration.service.ts` and `serverGroupCommandBuilder.service.js` and I've made the following changes.

1. Removed local component state (required by `AvailabilityZoneSelector.tsx`)  from formik.
2. Removed preferred zones loading and computation from `AvailabilityZoneSelector.tsx` and made it a stateless component.
3. Removed unused code from `LoadBalancerLocation.tsx`. This component only allows selection of availability zones computed from subnets and the removed code wasn't actually affecting anything in the UI.
4. Retained form level state management at `serverGroupConfiguration.service.ts`

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/common/LoadBalancerLocation.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/common/LoadBalancerLocation.tsx
@@ -156,14 +156,10 @@ export class LoadBalancerLocation extends React.Component<ILoadBalancerLocationP
       ([allSubnets, subnetPurpose]) => allSubnets && allSubnets.find(subnet => subnet.purpose === subnetPurpose),
     );
 
-    // I don't understand why we use subnet.availabilityZones here, but region.availabilityZones below.
     const availabilityZones$ = subnet$.map(subnet => (subnet ? uniq(subnet.availabilityZones).sort() : []));
-
-    // Update selected zones when the selected region changes
-    const regionZones$ = form.region$
-      .withLatestFrom(accountRegions$)
-      .map(([currentRegion, accountRegions]) => accountRegions.find(region => region.name === currentRegion))
-      .map(region => (region ? region.availabilityZones : []));
+    availabilityZones$.takeUntil(this.destroy$).subscribe(availabilityZones => {
+      this.props.formik.setFieldValue('regionZones', availabilityZones);
+    });
 
     const moniker$ = Observable.combineLatest(appName$, form.stack$, form.detail$).map(([app, stack, detail]) => {
       return { app, stack, detail, cluster: NameUtils.getClusterName(app, stack, detail) } as IMoniker;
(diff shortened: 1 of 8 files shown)


vigneshm:
Oops, didn't notice the failing tests! Will fix and update the PR.

gcomstock:
Thanks Vignesh!

vigneshm:
I messed up this PR! I'll see if I can fix them here, otherwise I'll close this one and create a new one!

vigneshm:
This PR is ready for review now :)

app/scripts/modules/amazon/src/loadBalancer/configure/common/LoadBalancerLocation.tsx
```diff
@@ -156,14 +156,10 @@ export class LoadBalancerLocation extends React.Component<ILoadBalancerLocationP
       ([allSubnets, subnetPurpose]) => allSubnets && allSubnets.find(subnet => subnet.purpose === subnetPurpose),
     );
 
-    // I don't understand why we use subnet.availabilityZones here, but region.availabilityZones below.
```
christopherthielen: @vigneshm do you understand why we used subnet.availabilityZones here, but region.availabilityZones below?

app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/ServerGroupZones.tsx
```diff
@@ -35,15 +40,18 @@ export class ServerGroupZones extends React.Component<IServerGroupZonesProps>
 
   public render() {
     const { values } = this.props.formik;
+    const currentAvailabilityZones = values.availabilityZones.sort();
+    const preferredZones = values.backingData.preferredZones[values.credentials]?.[values.region] || [];
+    const isUsingPreferredZones = currentAvailabilityZones.join() === preferredZones.join();
```
christopherthielen: do we need to `.sort()` these?

vigneshm: Good catch. Will update it!

app/scripts/modules/amazon/src/loadBalancer/configure/common/LoadBalancerLocation.tsx
vigneshm: I found two things related to these changes.

1. `availabilityZones` from `account.region` do not get used at all here. `AvailabilityZoneSelector.tsx` overrides the selected zones when mounted and only makes the availability zones (computed from subnets) up for selection.
2. @robzienert did confirm in a slack thread that this (i.e computing availability zones from subnets) is the expected behavior here.

robzienert: +1 -- I would say calculating AZs from subnets over regions is always the correct decision, especially if dealing with networking.

vigneshm:
Can't push this in its current state. Need to get more information about how AZs are set up before we can continue with this PR. Will update this PR with more info.

vigneshm:
Going to drop this PR since we want to keep the current UI of selecting 'automatic' & 'manual'. When 'automatic' is selected, deck shouldn't send any `availabilityZones` to the back-end. Instead, the back-end must use a configured set of preferred zones and pass that information to aws' launch configuration.

PR #7868, opened 2020-02-11
feat(plugins): Prefer plugins from deck manifest (over gate) if they have the same ID

christopherthielen:
If deck has a plugin-manifest which includes a plugin ID that is also served by gate, prefer the one in deck.

This can happen for two reasons:
- Deck has baked in manifest and plugins
- Developer is updating a plugin and is running in dev mode

### DIFF
--- a/app/scripts/modules/core/src/plugins/plugin.registry.spec.ts
+++ b/app/scripts/modules/core/src/plugins/plugin.registry.spec.ts
@@ -20,14 +20,14 @@ describe('PluginRegistry', () => {
 
   describe('.register()', () => {
     it('should validate plugin manifests', () => {
-      expect(() => pluginRegistry.registerPluginMetaData({} as any)).toThrowError(/Invalid plugin manifest/);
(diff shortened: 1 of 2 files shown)


christopherthielen:
@bpowell ptal

plugin-manifest.js
```diff
@@ -1,2 +1,2 @@
-const plugins = [];
+const plugins = [{ id: 'com.robzienert.helloworld', version: 'snapshot', url: '/plugins/helloworld/index.js' }];
```
bpowell: Will adding this is without the corresponding `index.js` file cause any issues?

christopherthielen: oh wow, nice catch!  That's not supposed to be there.

bpowell:
@christopherthielen is there something blocking this that I can help unblock?

christopherthielen:
@bpowell no, I was just waiting for an oss-approver to thumbs up this, but I'll just self merge

PR #7873, opened 2020-02-12
chore: test auto merge package bumps

christopherthielen:
maybe will it work if it's a in-repo branch and the workflow hasn't been modified?


christopherthielen:
@alanmquach it works!

PR #7883, opened 2020-02-12
fix(amazon): Add a warning when no oidc conifgs are found

caseyhebebrand:
When authenticating an ALB listener rule, the select field is empty if there are no configs found. This is due to configuration being nonexistent or being configured improperly for the application. This PR adds a warning in the dropdown (in addition to the help field), so it does not look like an error rendering the options.
Screen Shot 2020-02-12 at 11 05 23 AM

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
@@ -772,12 +772,11 @@ const Action = (props: {
         props.oidcConfigs.length > 0 &&
         (!clientId || props.oidcConfigs.find(c => c.clientId === clientId)));
 
-    const oidcOptions =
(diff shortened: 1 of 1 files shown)


app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
```diff
@@ -771,6 +771,14 @@ const Action = (props: {
       (props.oidcConfigs &&
         props.oidcConfigs.length > 0 &&
         (!clientId || props.oidcConfigs.find(c => c.clientId === clientId)));
+
+    const oidcOptions =
+      props.oidcConfigs && props.oidcConfigs.length ? (
```
christopherthielen: might as well use some es2020 sugar, (and remove the || [] guard):
```suggestion
      props.oidcConfigs?.length ? (
              props.oidcConfigs.map(config => <option key={config.clientId}>{config.clientId}</option>)
```

PR #7884, opened 2020-02-13
fix(aws): Fix to set right state while cloning security group

aravindmd:
- Setting `state.isNew` to `true` during clone which is used in the overridden component


aravindmd:
merging the changes

PR #7887, opened 2020-02-13
fix(npm): Fixing assert script slightly

alanmquach:
Trying a quick fix

### DIFF
--- a/app/scripts/modules/assert_package_bumps_standalone.sh
+++ b/app/scripts/modules/assert_package_bumps_standalone.sh
@@ -12,16 +12,8 @@ fi
 if [[ -n $TRAVIS || -n $GITHUB_ACTIONS ]] ; then
   echo "git fetch -q..."
   git fetch -q
-  CI_TARGET_BRANCH=origin/master
-  pwd
-  ls -al
-  cd app
-  ls -al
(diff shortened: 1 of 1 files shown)


app/scripts/modules/assert_package_bumps_standalone.sh
```diff
@@ -12,8 +12,8 @@ fi
 if [[ -n $TRAVIS || -n $GITHUB_ACTIONS ]] ; then
   echo "git fetch -q..."
   git fetch -q
-  CI_TARGET_BRANCH=origin/master
-  cd app/scripts/modules || exit 1;
+  CI_TARGET_BRANCH=origin/master # dirname should maybe work in both cases?
+  cd "$(dirname "$0")" || exit 1;
```
christopherthielen: move ```
  cd "$(dirname "$0")" || exit 1;
```
up to the top and remove the else?

alanmquach: Would like to see if this fixes @aravindmd's publish first then button it up in a following PR.

I think what is happening is when run in CI we are at the root but when run by the publish, the github script has already CD'd us into modules?

PR #7892, opened 2020-02-13
feat(codebuild): Add support for overriding image and buildspec

Kaixiang-AWS:
Add support for image overriding, which allows user to build with an image from
upstream stages or triggers.

Add support for buildspec overriding, which allows user to parameterize his/her
buildspec.

Add heading for each section to make the UI more clear.

Screen Shot 2020-02-13 at 11 43 28 AM

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/awsCodeBuild/AwsCodeBuildStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/awsCodeBuild/AwsCodeBuildStageForm.tsx
@@ -16,6 +16,7 @@ import {
   StageArtifactSelector,
   TextInput,
   useData,
+  YamlEditor,
 } from 'core';
 import { CheckboxInput } from 'core/presentation';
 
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/awsCodeBuild/AwsCodeBuildStageForm.tsx
```diff
@@ -114,6 +121,22 @@ export function AwsCodeBuildStageForm(props: IAwsCodeBuildStageFormProps & IForm
...
+          <YamlEditor {...inputProps} value={get(stage, 'buildspec')} onChange={onYamlChange} />
+        )}
+      />
+      <h4>Environment Configuration</h4>
+      <FormikFormField
+        fastField={false}
+        label="Image"
+        name="image"
+        input={(inputProps: IFormInputProps) => <TextInput {...inputProps} />}
+      />
```
clareliguori: As a todo, it'd be neat to use the Docker image selector, so you can get a dropdown of images to select from

Kaixiang-AWS: I will take a look at the Docker image selector to see if it could be used here

Kaixiang-AWS: I looked at the `DockerImageAndTagSelector`. This is used for selecting an image but not matching against the incoming docker images. The image selected will be fixed. For example, if I configured 2 triggers for 2 different images, this selector could only select one of them. No matter which image triggers the pipeline, CodeBuild will always build using the image selected, which is definitely not what we want. For this `TextInput` approach, user has the ability by specifying `${trigger['repository']}/${trigger['tag']}` in this field.

PR #7896, opened 2020-02-14
fix(docker): Fix handling Docker Image name with digest

dragon3:
## WHAT

This pull request fixes handling Docker Image name (imageId) with digest.

## WHY

I found that manual execution with specifying Docker image `Digest` doesn't work even if the digest is correct.

Screen_Shot_2020-02-14_at_15_38_52

The reason is that currently `DockerImageUtils` doesn't parse/generate Docker image name with digest properly.

When we use digest, Docker image name (imageId) should be this (`@` is separator):
```
gcr.io/project/my-image@sha256:28f82eba
```

But now it's generated like this (`:` is separator):
```
gcr.io/project/my-image:sha256:28f82eba
```

## REF

https://github.com/docker/distribution/blob/a8371794149d1d95f1e846744b05c87f2f825e5a/reference/reference.go#L6


maggieneterval:
@dragon3 thanks so much for this PR! I especially appreciate your detailed PR description and the unit tests you added. Were you able to manually test these changes against a dev Spinnaker setup?
 I'm asking because I'm not 100% sure that these changes actually resolve the problem you reported around manual executions not parsing digests correctly. I think that logic lives in the [DockerTriggerTemplate](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/docker/src/pipeline/trigger/DockerTriggerTemplate.tsx#L73) component: specifically, it looks like the `updateArtifact` method assumes that the input will always be a `tag` string rather than a `digest`. Let me know what you think!

dragon3:
Hi @maggieneterval 
Thank you for the comment.

> @dragon3 thanks so much for this PR! I especially appreciate your detailed PR description and the unit tests you added. Were you able to manually test these changes against a dev Spinnaker setup?
> I'm asking because I'm not 100% sure that these changes actually resolve the problem you reported around manual executions not parsing digests correctly. I think that logic lives in the [DockerTriggerTemplate](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/docker/src/pipeline/trigger/DockerTriggerTemplate.tsx#L73) component: specifically, it looks like the `updateArtifact` method assumes that the input will always be a `tag` string rather than a `digest`. Let me know what you think!

Oops... You are right. 
Actually I haven't tested these changes against a dev Spinnaker setup yet... 
It seems that I found and fixed a different bug from the original one which I'd like to fix...

Let me check `DockerTriggerTemplate` to fix the original issue 🙇

dragon3:
Hi @maggieneterval 
I've updated the `DockerTriggerTemplate` and push it (ec5c720)
I also confirmed it fixed the original issue with my dev Spinnaker setup.

Screen Shot 2020-02-17 at 9 45 02

Screen Shot 2020-02-17 at 9 53 55

Screen Shot 2020-02-17 at 9 54 19

---

Could you please review this pull request again when you have a chance? 🙏 
Let me know if you have any questions or feedback.

Thanks!

rbadillo:
Are we expecting this fix in the 1.18 release? 

Thanks in advance for fixing this. I just ran into this problem this morning.

dragon3:
> Are we expecting this fix in the 1.18 release?

I hope, but I'm not sure.

@maggieneterval 
What do you think?

> Thanks in advance for fixing this. I just ran into this problem this morning.

Sure 😁

app/scripts/modules/docker/src/image/DockerImageUtils.ts
```diff
@@ -8,7 +8,12 @@ export interface IDockerImageParts {
 export class DockerImageUtils {
   // Split the image id up into the selectable parts to feed the UI
   public static splitImageId(imageId = ''): IDockerImageParts {
-    const imageParts = imageId.split(':');
+    let imageParts: string[];
+    if (imageId.indexOf('@') > 0) {
```
maggieneterval: [`indexOf`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf) returns`-1` if the value is not found, so technically we should switch `>0` to `>-1`. Alternately, [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) can make this a lot cleaner:

```suggestion
    if (imageId.includes('@')) {
```

app/scripts/modules/docker/src/pipeline/trigger/DockerTriggerTemplate.tsx
```diff
@@ -70,21 +70,29 @@ export class DockerTriggerTemplate extends React.Component<
     this.setState({ lookupType: newType });
   };
 
-  private updateArtifact(command: IPipelineCommand, tag: string) {
-    this.props.updateCommand('extraFields.tag', tag);
+  private updateArtifact(command: IPipelineCommand, tagOrDigest: string) {
```
maggieneterval: One thing I noticed while testing this is that we don't call `updateArtifact` when we toggle between `tag` and `digest` lookup types. This seems problematic because it's possible to get into a state where `extraFields.tag` points to your tag as expected, but `extraFields.artifacts[0]` still references a digest you had entered previously. It seems like what we actually want to do is replace this line from `lookupTypeChanged`:

`this.props.updateCommand('extraFields.tag', newType === 'tag' ? this.state.selectedTag : this.state.digest);`

with something like this:

`this.updateArtifact(this.props.command, newType === 'tag' ? this.state.selectedTag : this.state.digest)`

What do you think?

app/scripts/modules/docker/src/image/DockerImageUtils.ts
dragon3: Thanks! I fixed it (15a3e10)

app/scripts/modules/docker/src/pipeline/trigger/DockerTriggerTemplate.tsx
dragon3: Thank you for the comment.

Yes, you are right.
It's better to update the `lookupTypeChanged` to replace `this.props.updateCommand(...` with `this.updateArtifact(...`.

I'm on it 👍

dragon3: Fixed in https://github.com/spinnaker/deck/pull/7896/commits/1642c2a7afcb82cc49b9f1b5abefa649e5f35782

maggieneterval: Thanks for the fix; looks good to me!

maggieneterval:
@rbadillo @dragon3 We can likely cherry-pick this fix into the next patch release of 1.18; however, I defer to the current release manager @duftler for the final call.

duftler:

maggieneterval:

dragon3:
@maggieneterval 
Thank you for the review and merge!

PR #7902, opened 2020-02-15
feat(plugins): copies over plugin-manifest from a halyard deploy

bpowell:


christopherthielen:
@bpowell does this imply that plugin manifest js isn’t being used yet? I’d like to switch it to `json` And if nobody is using it yet, we can switch it without consequence.

bpowell:
@christopherthielen sure, we can change it to `plugin-manifest.json`. I don't think anyone right now is using it.

bpowell:
closing in favor of https://github.com/spinnaker/deck/pull/7941

PR #7903, opened 2020-02-16
feat(core/presentation): add Table components, multiple table layouts

erikmunson:
Don't merge, depends on #7899 and #7901. CI will likely fail until those are merged.

This is another chapter of adding some re-usable components to match @gcomstock's new design work, in support of a Managed Delivery view. My goal was to provide a high-level component API for creating tables which — similar to previous work on form fields — could serve as a generic scaffold with pluggable points for different kinds of table layouts. I'm starting with two (very different) layout implementations because that's what my current work calls for, but the idea is that adding or iterating on layouts themselves should be pretty isolated from actual usage (vs. requiring refactors of all consuming code).

To start, here's some example screenshots of the end result:

<details>
<summary>"Mobile" display size</summary>
Screen Shot 2020-02-14 at 7 56 50 PM
</details>

<details>
<summary>Larger width display size</summary>
Screen Shot 2020-02-14 at 7 56 43 PM
</details>

As shown, I've started with two layout examples: one is a 'minimal' visual style that leverages native tables to get an auto-arranged layout without much fuss or complexity. The other is what I'm calling a 'standard' visual style (this is the style we're expecting will be most commonly used) laid out using CSS grids, with some fancy features added in (expanding rows, responsive + mobile layout, sorting, etc.). Both these layouts plug into the same layout contract used by some higher level components: `Table`, `TableRow`, and `TableCell`.

The responsibilities of those externally-facing components are primarily:
- Give a common, consistent component structure and API to all tables regardless of layout
- Handle frequently used, often stateful table patterns like row expandability or row indenting that consumers and individual layouts can leverage
- Do runtime validation for things like required/dependent prop combinations so layouts can focus on their own needs (and any layout-specific validation)

For a simple example, here's what the `minimalNativeTableLayout` looks like in use:
<details>
<summary>Minimal layout</summary>

```jsx
<Table layout={minimalNativeTableLayout} columns={['configuration', 'change', 'current', 'desired']}>
  <TableRow>
    // indent is useful for tables where rows have hierarchical relationships
    <TableCell indent={2}>
      <SingleLineString>keeldemo</SingleLineString>
    </TableCell>
    <TableCell>
      <SingleLineString className="text-italic">removed</SingleLineString>
    </TableCell>
    <TableCell>
      <BreakString>keeldemo</BreakString>
    </TableCell>
    <TableCell>
      <BreakString>null</BreakString>
    </TableCell>
  </TableRow>
</Table>
```

</details>

And as a more involved example, here's what it looks like to use a `standardGridTableLayout` with sorting and expandable rows:
<details>
<summary>Standard layout</summary>

```jsx
// First we setup some state for sorting and create column definitions.
// While the simple example used an array of strings for the column names,
// this one shows how column definitions can add additional behavior.
const [sortColumn, setSortColumn] = React.useState('When');
const [sortDirection, setSortDirection] = React.useState<'ascending' | 'descending'>('descending');
const mainTableColumns = React.useMemo(
  () =>
    ['Where', 'What', 'When'].map(name => ({
      name,
      // works like a controlled input — add a `sortDirection` to indicate a sort,
      // react to changes via `onSort` 
      sortDirection: name === sortColumn ? sortDirection : undefined,
      onSort: () => {
        const newDirection = sortDirection === 'ascending' ? 'descending' : 'ascending';
        console.log(`sorting ${name} column to direction: ${newDirection}`);

        setSortColumn(name);
        setSortDirection(newDirection);
      },
    })),
  [sortColumn, sortDirection],
);

// This layout accepts some of its own configuration parameters
// to control how large each column is relative to the others (among some other options)
<Table layout={standardGridTableLayout([4, 2, 2.6])} columns={mainTableColumns} expandable={true}>
  <TableRow
    // Works like an uncontrolled input — if `defaultExpanded` is set
    // that will be the initial expanded state after mount until a user interacts with it.
    defaultExpanded={true}
    renderExpandedContent={() => (
      // Render whatever you want,
      // in the example case we stick a minimal native table here
    )}
  >
    <TableCell>TEST > keeldemo-prestaging</TableCell>
    <TableCell>
      Drift Detected
    </TableCell>
    <TableCell>
      2018-10-17 15:51:00 PDT 30 minutes ago
    </TableCell>
  </TableRow>
</Table>
```

</details>

While the diff is a little larger, each commit is pretty tightly scoped and I recommend going one-by-one. Because this is still in the process of getting its first real usage, I'm deferring investing more time in documenting the available configuration options for the Table components and specific layouts until this PR is reviewed/iterated and the API settles slightly more. If you have feedback about API design or overall approach, now is the time!

### DIFF
--- a/app/scripts/modules/core/src/presentation/tables/Table.tsx
+++ b/app/scripts/modules/core/src/presentation/tables/Table.tsx
@@ -1,0 +1,71 @@
+import * as React from 'react';
+
+import { useIsMobile, useDeepObjectDiff } from 'core/presentation';
+
+import { ITableRowLayoutProps } from './TableRow';
+import { ITableCellLayoutProps } from './TableCell';
+import { TableContext } from './TableContext';
+
+const { useState, useMemo } = React;
+
+export interface ITableColumn {
+  name: string;
+  sortDirection?: 'ascending' | 'descending';
+  onSort?: () => any;
+}
+
+export type ITableColumns = Readonly<Array<string | ITableColumn>>;
+
+export interface ITableLayoutProps {
+  columns: ITableColumn[];
+  isMobile: boolean;
+  expandable?: boolean;
+  expandAll: () => any;
+  children?: React.ReactNode;
+}
+
+export interface ITableLayout {
+  TableLayout: React.ComponentType<ITableLayoutProps>;
+  TableRowLayout: React.ComponentType<ITableRowLayoutProps>;
+  TableCellLayout: React.ComponentType<ITableCellLayoutProps>;
+}
+
+export interface ITableProps {
+  layout: ITableLayout;
+  columns: ITableColumns;
+  expandable?: boolean;
+  children?: React.ReactNode;
+}
+
+export const Table = (props: ITableProps) => {
+  const isMobile = useIsMobile();
+  const [allExpanded, setAllExpanded] = useState(false);
+
+  const { layout, ...propsWithoutLayout } = props;
+
+  const normalizedColumns = props.columns.map(column => (typeof column === 'string' ? { name: column } : column));
+  const tableContext = useMemo(
+    () => ({
+      layout,
+      isMobile,
+      allExpanded,
+      setAllExpanded: () => setAllExpanded(false),
+      expandable: props.expandable,
+      columns: normalizedColumns,
+    }),
+    [layout, isMobile, props.expandable, allExpanded, useDeepObjectDiff(props.columns)],
+  );
+
+  const { TableLayout } = layout;
+
+  return (
+    <TableContext.Provider value={tableContext}>
+      <TableLayout
+        {...propsWithoutLayout}
+        isMobile={isMobile}
+        columns={normalizedColumns}
+        expandAll={() => setAllExpanded(true)}
+      />
+    </TableContext.Provider>
+  );
+};
--- a/app/scripts/modules/core/src/presentation/tables/TableContext.tsx
+++ b/app/scripts/modules/core/src/presentation/tables/TableContext.tsx
@@ -1,0 +1,14 @@
+import * as React from 'react';
+
+import { ITableLayout, ITableColumn } from './Table';
+
+export interface ITableContext {
+  layout: ITableLayout;
+  isMobile: boolean;
+  expandable: boolean;
+  allExpanded: boolean;
+  setAllExpanded: (allExpanded: boolean) => any;
+  columns: ITableColumn[];
+}
+
+export const TableContext = React.createContext<ITableContext>(null);
--- a/app/scripts/modules/core/src/presentation/tables/TableRow.tsx
+++ b/app/scripts/modules/core/src/presentation/tables/TableRow.tsx
@@ -1,0 +1,81 @@
+import * as React from 'react';
+
+import { ITableCellPropsWithInternalFields } from './TableCell';
+import { ITableContext, TableContext } from './TableContext';
+
+const { useState, useContext, useEffect } = React;
+
+export interface ITableRowProps {
+  defaultExpanded?: boolean;
+  renderExpandedContent?: () => React.ReactNode;
+  children?: React.ReactNode;
+}
+export type ITableRowLayoutProps = ITableRowProps &
+  Pick<ITableContext, 'isMobile' | 'columns'> &
+  Pick<ITableRowProps, 'renderExpandedContent'> & {
+    tableExpandable: boolean;
+    rowExpandable: boolean;
+    expanded: boolean;
+    setExpanded: (expanded: boolean) => any;
+  };
+
+const logDefaultExpandedError = () => {
+  console.error(
+    new Error(
+      `You used the defaultExpanded prop on a row inside a <Table/> that's not expandable.
+       Either remove the defaultExpanded prop or make the <Table/> expandable.`,
+    ),
+  );
+};
+
+const throwCellColumnError = () => {
+  throw new Error(
+    `A <TableCell/> was added that didn't match any of the columns defined on its parent <Table/>.
+     Check if there are an equal number of <TableCell/> elements to the number of columns provided to the <Table/>.
+    `,
+  );
+};
+
+export const TableRow = ({ defaultExpanded, renderExpandedContent, children }: ITableRowProps) => {
+  const { layout, expandable, allExpanded, setAllExpanded, isMobile, columns } = useContext(TableContext);
+  const [expanded, setExpanded] = useState(defaultExpanded || false);
+
+  // Expand individual rows when the table expands all its rows at once
+  useEffect(() => {
+    allExpanded && renderExpandedContent && setExpanded(true);
+  }, [allExpanded]);
+
+  if (!expandable && defaultExpanded != null) {
+    logDefaultExpandedError();
+  }
+
+  const childrenWithIndex = React.Children.map(children, (cell, index) => {
+    return React.cloneElement(cell as React.ReactElement<ITableCellPropsWithInternalFields>, { index });
+  });
+
+  if (childrenWithIndex.length > columns.length) {
+    throwCellColumnError();
+  }
+
+  const { TableRowLayout } = layout;
+
+  return (
+    <TableRowLayout
+      tableExpandable={expandable}
+      rowExpandable={expandable && !!renderExpandedContent}
+      expanded={expanded}
+      setExpanded={expand => {
+        setExpanded(expand);
+        // When all rows were previously expanded, clear away that state
+        // the first time an individual row is collapsed
+        if (allExpanded && !expand) {
+          setAllExpanded(false);
+        }
+      }}
+      renderExpandedContent={renderExpandedContent}
+      isMobile={isMobile}
+      columns={columns}
+      children={childrenWithIndex}
+    />
+  );
+};
--- a/app/scripts/modules/core/src/presentation/tables/standardGridTableLayout.less
+++ b/app/scripts/modules/core/src/presentation/tables/standardGridTableLayout.less
@@ -1,0 +1,196 @@
+.StandardGridTableLayout {
+  display: grid;
+
+  .standard-grid-table-header {
+    display: grid;
+    position: sticky;
+    top: 0;
+    font-weight: 600;
+    -webkit-font-smoothing: antialiased;
+    -moz-osx-font-smoothing: grayscale;
+    color: var(--color-regal);
+    font-size: 13px;
+    z-index: 100;
+    align-items: flex-end;
+    min-height: 48px;
+    grid-column: 1 / -1;
+    background-color: var(--color-smoke);
+    border-bottom: 2px solid var(--color-blumine);
+
+    .standard-grid-table-cell {
+      align-items: flex-end;
+      padding-bottom: 6px;
+
+      &.sortable {
+        border-radius: 16px 16px 0 0;
+        cursor: pointer;
+        height: 32px;
+
+        &:hover {
+          background-color: rgba(160, 180, 220, 0.15);
+        }
+
+        & > .sort-indicator {
+          border-radius: 16px;
+          background-color: var(--color-blumine);
+          width: 16px;
+          height: 16px;
+          margin-left: 6px;
+          margin-bottom: 1px;
+          position: relative;
+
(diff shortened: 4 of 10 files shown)


app/scripts/modules/core/src/presentation/tables/TableContext.tsx
```diff
@@ -0,0 +1,14 @@
...
+export interface ITableContext {
+  layout: ITableLayout;
+  isMobile: boolean;
+  expandable: boolean;
+  allExpanded: boolean;
+  setAllExpanded: (allExpanded: boolean) => any;
+  columns: ITableColumn[];
+}
+
+export const TableContext = React.createContext<ITableContext>(null);
```
erikmunson: While this context-based approach is _slightly_ roundabout when passing things around, I prefer it over using render props because basically **every table related component** in the subtree of a `<Table/>` needs a slice of this context in some way or another, and you'd end up with `{...props}` in a gazillion places without getting much of any noticeable benefit (at least as far as I can tell). Happy to talk about other approaches if folks have suggestions, but would prefer to avoid too much error-prone manual wiring.

app/scripts/modules/core/src/presentation/tables/standardGridTableLayout.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <div className="cell-content">{children}</div>
+    </div>
+  );
+};
+
+export const standardGridTableLayout = (sizes: ITableColumnSizes) => ({
+  getTableLayout: (props: ITableLayoutProps) => <TableLayout {...props} sizes={sizes} />,
+  getTableRowLayout: (props: ITableRowLayoutProps) => <TableRowLayout {...props} sizes={sizes} />,
+  getTableCellLayout: (props: ITableCellLayoutProps) => <TableCellLayout {...props} />,
+});
```
erikmunson: this is an example of how a layout can accept some of its own layout-specific configuration without jumping through major hoops.

app/scripts/modules/core/src/presentation/tables/TableRow.tsx
```diff
@@ -0,0 +1,83 @@
...
+    logDefaultExpandedError();
+  }
+
+  if (expandable && !renderExpandedContent) {
+    logExpandedContentError();
+  }
+
+  const childrenWithIndex = React.Children.map(children, (cell, index) => {
+    return React.cloneElement(cell as React.ReactElement<ITableCellPropsWithInternalFields>, { index });
+  });
```
erikmunson: not thrilled with having to do a children map+clone here, but it's in service of being able to give an individual `TableCell` the name of the column it represents so folks don't have to manually wire up names across many different components. I thought that was a pretty valuable thing to add, so I took the tradeoff and added some validation below to alert folks to potential misconfiguration (when there are the wrong number of cells compared to columns).

app/scripts/modules/core/src/presentation/tables/Table.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export type ITableColumns = Readonly<Array<string | ITableColumn>>;
+
+export type ITableLayoutProps = Omit<ITableProps, 'layout' | 'columns'> & {
+  columns: ITableColumn[];
+  isMobile: boolean;
+  expandAll: () => any;
+};
+
+export interface ITableLayout {
```
caseyhebebrand: I think exporting these as components would make the most sense, given the use of React context and for composing custom layouts. 

In the case where you may need only part of the `minimalNativeTableLayout`, it feels cleaner to write:
```
<Table 
    layout={{ table: MinimumNativeTableLayout, row: CustomRow, cell:  StandardCellLayout }}
    { ...otherProps }
/>
```

app/scripts/modules/core/src/presentation/tables/Table.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export interface ITableProps {
+  layout: ITableLayout;
+  columns: ITableColumns;
+  expandable?: boolean;
+  children?: React.ReactNode;
+}
+
+export const Table = (props: ITableProps) => {
+  const isMobile = useIsMobile();
```
caseyhebebrand: This is ok for now, but I am still of the mindset that `isMobile` is a global state of the page. The table, as a subcomponent of the page, should be viewing this boolean, not setting it.

app/scripts/modules/core/src/presentation/tables/Table.tsx
```diff
@@ -0,0 +1,67 @@
...
+      setAllExpanded: () => setAllExpanded(false),
+      expandable: props.expandable,
+      columns: normalizedColumns,
+    }),
+    [layout, isMobile, props.expandable, allExpanded, props.columns],
+  );
+
+  return (
+    <TableContext.Provider value={tableContext}>
+      {layout.getTableLayout({
```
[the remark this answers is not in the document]
erikmunson: 👍 yeah, I went back and forth on this a bit and in hindsight I think I agree. One nice freebie you get out of this is the ability to use hooks in layout code without having to render a component inside a component the way I did in `standardGridTableLayout` (in the current impl trying to use a hook in layout code would be weird because it's technically not at the top of the render function for the component it lives inside).

app/scripts/modules/core/src/presentation/tables/Table.tsx
erikmunson: Would your preference be that individual layout implementations ask for the value from `useIsMobile` separately? e.g. each `getXYZLayout` implementation would call `useIsMobile`?

After discussing it I'm not actually against that approach, but it does mean I'd probably have to do some performance optimization on `useIsMobile` sooner than anticipated.

app/scripts/modules/core/src/presentation/tables/Table.tsx
erikmunson: After hearing your feedback + Chris's, I am inclined to switch over to thinking about the individual layout bits as components instead of `getXYZLayout` functions. That said, I'm not 100% sure how much I want to encourage folks to do mix-and-match layout overriding. I suspect in many cases table implementations will have some implicit dependencies between — for example — the row implementation and the cell implementation.

Either way moving to think about layout implementations as components instead of functions that return elements _should_ enable the kind of overriding in your example.

app/scripts/modules/core/src/presentation/tables/standardGridTableLayout.less
```diff
@@ -0,0 +1,196 @@
...
+
+        & > i {
+          font-size: 24px;
+          margin-bottom: -4px;
+        }
+      }
+    }
+  }
+
+  .standard-grid-table-row {
```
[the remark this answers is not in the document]
erikmunson: Definitely a valid thing to ask! With the caveat that this is all early stages and will probably change a lot, @gcomstock and I have put some time and research into exactly this question and we feel relatively confident that it will be able to serve the majority of the tabular display use cases we're currently aware of. A couple nuances though:

- **Greg and I do not currently see things like infrastructure views, functions, etc. as belonging in the "Table" family.** Greg's perspective is that those views are not really tables, they don't have columns, and it's not the right fit to try to force them into a tabular data API (e.g. something focused on columns and rows). He and I have been calling the table-esque bits of those views "Object lists" or "Object rows", and I suspect that when it comes time to implement views using that visual pattern we will come up with a standalone API whose whole job is to be **really great** at modeling that pattern, since it's one we know will be an important and central building block of our most critical views.

- **The list of places we actually use tabular data is not that long.** I don't think we have a huge diversity of tables today, though Greg is certainly leveraging them more than we have historically. We audited our current and planned table usage and right now it's seeming like this 'standard' layout style will serve:
  - MD history (this will be the first usage to prove it out)
  - Tasks view
  - Kayenta reports table
  - FP rollouts (Netflix internal only)
  - Astrid artifact/server group usage breakdowns (Netflix internal only)

Happy to dig into this a bit more tactically if you're interested, I'm trying to lean into experimenting with this vs. having everything lined up in the first pass.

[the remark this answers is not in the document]
gcomstock: I approve of this message

erikmunson:
Thanks for all the great feedback folks!

@christopherthielen just wanted to address some of your critical feedback to make sure whatever I do next is in a positive direction:

> The sorting API seems pretty awkward and could be rethought.

100% agree, and in part this is a result of not actually needing to use the sorting stuff in the first view this was built for (MD history view). I actually went through 3 different iterations of sorting before settling on this one, and it took some self control to stop iterating and show it as-is for feedback. Because it won't be used immediately I may not block merging this on figuring out sorting (and I bet we won't fully figure it out until we have to use it for something). That said, I'd really like to get more detail on what turned you off about it. I have my own criticisms of it, but also had frustrations with the other approaches I tried. More feedback would be super helpful there.

> It seems like these things are more like renderers than layout. Compare to the FormField Layout code

I'd like to understand what you see as the main difference between what you're calling a layout vs. a renderer — I have generally considered them different names for the same thing and chose layout here to be consistent with the form APIs. From my perspective the main attributes of both the form layouts and these layouts/renderers is that they take in some data and JSX elements (inputs/labels in forms, rows/cells/cell content in tables), and arrange the JSX elements in some specific way (optionally using the other data they received). Perhaps I'm just coming at the definition/structure of those things from a different angle than you are? I don't have much of a preference on naming as long as we feel like end users can grok how to use the respective components.

> The use of Context seems fine to me. I wonder though if the implementation could be simplified if layouts just fetched the contexty bits from a react hook?

I thought about this, and originally didn't do it because I was imagining using `useContext` directly and felt the stuff that's currently stored in context had too many internal-API-only bits (e.g. the state flags for expanding/collapsing, etc.). Your comment got me thinking about a custom hook that'd do the things we're doing in the table components right now to massage context into something ready for layouts to consume, which seems like it could be cool. I'll play around with it, but have some worries about the fact that each layout (table, row, cell) has different slices/transformations of the context it cares about. I don't want to push that slicing burden onto layout code directly.

app/scripts/modules/core/src/presentation/tables/Table.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const tableContext = useMemo(
+    () => ({
+      layout,
+      isMobile,
+      allExpanded,
+      setAllExpanded: () => setAllExpanded(false),
+      expandable: props.expandable,
+      columns: normalizedColumns,
+    }),
+    [layout, isMobile, props.expandable, allExpanded, props.columns],
```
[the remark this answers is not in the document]
erikmunson: yes. I was admittedly kind of lazy here. I think in a theoretical world it'd be more flexible/more performant in some subset of cases to let consumers of these components decide when to invalidate memoization for `columns`, but clearly even the person that wrote it can't be trusted to do that consistently. Given that `columns` isn't a super large/deeply nested object I'm inclined to take over the burden and do what you suggest.

app/scripts/modules/core/src/presentation/tables/Table.tsx
erikmunson: With the updates I pushed to render these layout implementation bits as components instead of functions, I believe the use case you described is now possible (though still not something I'd like to encourage often).

app/scripts/modules/core/src/presentation/tables/Table.tsx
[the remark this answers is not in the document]
erikmunson: Added, thanks! ✅

erikmunson:
Alright. I'm incredibly grateful for all the detailed feedback, and I think I've addressed the majority of it now. I'm going to go ahead and get this first iteration merged, but here are some things I _didn't_ directly address yet which I'll be following up on:

- I'm going to try to get more feedback on the sorting API, but for now I'm not directly using it so I'd like to move forward and keep working on that as a background task
- I agree with the previously stated opinions that individual components should be referencing `useIsMobile()` directly rather than it being baked into the table layout API, but I want to do one more pass at making that hook more performant and capable of handling multiple subscriptions. Once I do that I'll remove `isMobile` from the table components + layouts, and instead use the hook directly in layout implementations
- I'd like to chat a bit more with Chris about the layout vs. renderers topic before actually changing the name, will follow up with whatever the result of that is
- I'll be looking into using hooks instead of props for the layout/context bits, but not totally sure I'll go in that direction. Pending experimentation

PR #7905, opened 2020-02-17
feat(plugins): Switch from /plugin-manifest.js to /plugin-manifest.json

christopherthielen:
BREAKING CHANGE: Migrate the deck plugin-manifest asset from `.js` to `.json` to match the json output from gate's [/plugins/deck/plugin-manifest.json](https://github.com/spinnaker/gate/blob/master/gate-deck-plugins/src/main/kotlin/com/netflix/spinnaker/gate/plugins/DeckPluginsController.kt#L38-L42)


christopherthielen:
@bpowell

christopherthielen:
@bpowell @link108 ptal

PR #7912, opened 2020-02-18
feat(plugins): imperative initialization

bpowell:


app/scripts/modules/core/src/plugins/plugin.registry.spec.ts
```diff
@@ -53,6 +53,7 @@ describe('PluginRegistry', () => {
       pluginModule = {
         plugin: {
           stages: [({ id: 'mystage' } as any) as IStageTypeConfig],
+          undefinedExtensions: () => {},
```
christopherthielen: I don't know what `undefinedExtensions` means.  Let's call this `initialize()` ?

bpowell: sure thing! that works for me

app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -168,6 +169,9 @@ export class PluginRegistry {
       // Register extensions with deck.
       plugin.stages?.forEach(stage => Registry.pipeline.registerStage(stage));
 
+      // Run code that currently does not have an extension point
+      plugin.initialize && plugin.initialize();
```
erikmunson: Some fancy new TypeScript sugar:
```suggestion
      plugin.initialize?.();
```

PR #7915, opened 2020-02-18
feat(codebuild): Add support for multiple sources

Kaixiang-AWS:
Refactor the stage definition so that all source related fields go into one single structure.
This is helpful when it comes to multiple sources because of the reusability that the single
structure provides. For primary source, artifact is not required. Users can override source
version without touching source artifact. However, in secondary sources, users must specify
an artifact. Otherwise, there's no way to link a secondary source version to a secondary source
defined in the project configuration.

Screen Shot 2020-02-18 at 2 28 00 PM

Screen Shot 2020-02-18 at 2 28 35 PM

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/awsCodeBuild/AwsCodeBuildStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/awsCodeBuild/AwsCodeBuildStageForm.tsx
@@ -2,8 +2,6 @@ import React from 'react';
 import { get } from 'lodash';
 
 import {
-  ArtifactTypePatterns,
-  excludeAllTypesExcept,
   FormikFormField,
   IArtifact,
   IExpectedArtifact,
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/pipeline/config/stages/awsCodeBuild/AwsCodeBuildStageForm.tsx
```diff
@@ -41,17 +34,21 @@ export function AwsCodeBuildStageForm(props: IAwsCodeBuildStageFormProps & IForm
   );
 
   const onYamlChange = (buildspec: string, _: any): void => {
-    props.formik.setFieldValue('buildspec', buildspec);
+    props.formik.setFieldValue('source.buildspec', buildspec);
```
caseyhebebrand: All of these functions (`onYamlChange`, `setArtifactId`, `setArtifact`, `updateSources`) are just updating a form field value. Can we create a generic function that accepts `fieldName` and `fieldValue` as parameters and updates the field?

Kaixiang-AWS: I tried to come up with something like
```
interface Field {
  fieldName: string;
  fieldValue: any;
}
const updateFormField = (...keys: Field[]): void => {
  keys.forEach((key: Field) => {
    return props.formik.setFieldValue(key.fieldName, key.fieldValue);
  });
};
<YamlEditor 
  {...inputProps}
  value={get(stage, 'source.buildspec')}
  onChange={(buildspec: string, _: any) => 
    updateFormField({fieldName: 'source.buildspec', fieldValue: buildspec})
  }
/>
```
it turns out that it makes the code more complex and unreadable when the number of fields increases. For example
```
const setArtifact = (artifact: IArtifact): void => {
  props.formik.setFieldValue('source.sourceArtifact.artifact', artifact);
  props.formik.setFieldValue('source.sourceArtifact.artifactId', null);
};
```
will become
```
(artifact: IArtifact) =>
  updateFormField(
    {fieldName: 'source.sourceArtifact.artifact', fieldValue: artifact},
    {fieldName: 'source.sourceArtifact.artifactId', fieldValue: null}
  )
```
I think the `setFieldValue` method provided by `formik` is generic enough and there's no way to write a more generic method outside of it. Let me know what you think.

caseyhebebrand: I was just thinking of:
```
const onFieldChange = (fieldName: string, value: T) => 
  props.formik.setFieldValue(fieldName, value);
```
And then:
```
onChange={ (value: boolean) => onFieldChange<boolean>('field.name', value) }
``` 
For `sourceArtifact` you can update the entire object this way, or provide a conditional for this edge case.  Then you can use this one function for any field change.

PR #7917, opened 2020-02-19
feat(spel): default new pipelines to spel v4

marchello2000:
now that spel v4 is live and glorious in orca let's make deck create new pipelines with that as default

### DIFF
--- a/app/scripts/modules/core/src/pipeline/create/CreatePipelineModal.tsx
+++ b/app/scripts/modules/core/src/pipeline/create/CreatePipelineModal.tsx
@@ -89,6 +89,7 @@ export class CreatePipelineModal extends React.Component<ICreatePipelineModalPro
       application: this.props.application.name,
       limitConcurrent: true,
       keepWaitingPipelines: false,
+      spelEvaluator: 'v4',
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/create/CreatePipelineModal.tsx
```diff
@@ -389,10 +390,7 @@ export class CreatePipelineModal extends React.Component<ICreatePipelineModalPro
                     </div>
                     <div className="col-md-7">
                       <Select
-                        options={[
```
marchello2000: i guess the linter autofixed this?

marchello2000:
> LGTM. As an aside, do you intend to pull this back out at some point in the future when v4 has been thoroughly tested and older pipelines get migrated, or do you expect this will be necessary permanently (i.e. v3 won't get EOL'd for the foreseeable future)?

Good question. There are breaking changes between versions. I believe we can migrate everyone at NFLX but not sure about OSS (likely also true). I, unfortunately, don't have time to do that now, but it's def on my todo list

PR #7923, opened 2020-02-20
feat(managed): Scaffolding environments route

alanmquach:
Bare minimum for the new route so we can iterate. Not yet exposed via nav.

Up next: probably data sources

### DIFF
--- a/app/scripts/modules/core/src/managed/Environments.tsx
+++ b/app/scripts/modules/core/src/managed/Environments.tsx
@@ -1,0 +1,5 @@
+import React from 'react';
+
+export default function Environments() {
+  return <div>For there shall be no greater pursuit than that towards desired state.</div>;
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+export default function Environments() {
+  return <div>For there shall be no greater pursuit than that towards desired state.</div>;
```
erikmunson: beautiful

alanmquach: > we should figure out a strategy for when to surface this in the nav
Yes, my thoughts exactly. I imagine it may also evolve as MD matures.

PR #7926, opened 2020-02-20
feat(codebuild): List project names

Kaixiang-AWS:
Change text input field to select input field for project name
so that user could select project name from a dropdown list
instead of manually type in the name.

Fix a null pointer issue when creating a new CodeBuild stage.

Screen Shot 2020-02-20 at 2 05 44 PM


Kaixiang-AWS:
depend on https://github.com/spinnaker/gate/pull/1070

PR #7932, opened 2020-02-21
fix(plugins): if gate doesn't have plugin-manifest.json, ignore the error

bpowell:


app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -115,6 +115,10 @@ export class PluginRegistry {
       .get()
       .catch((error: any) => {
         console.error(`Failed to load ${uri} from ${source}`);
+        // If we cannot hit the Gate URL, ignore it
+        if (error.data.status === 404) {
```
christopherthielen: How could this happen?  This endpoint should always be in gate, as far as I can tell

bpowell: Not if you are playing with master of Deck and Gate that hasn't been updated in a while. Both @claymccoy and I found this while doing some testing

christopherthielen: ok, but let's delete this in a month or two.

christopherthielen:
@bpowell if you mark your pull requests as "allow edits from maintainers" then mergify will keep your PRs up to date with master automatically.

PR #7933, opened 2020-02-21
fix(core/pipeline): handle missing trigger fields in importDeliveryConfig stage

erikmunson:
Currently when there's an issue with the trigger used for the `importDeliveryConfig` stage and some fields are missing, we throw trying to call `substring()` on the `hash` field and the entire executions view goes blank (along with permanently busting the entire UI until a full page reload).

This change uses a `Not found` fallback for all the trigger-related fields we surface, so things render and the page continues to work. Looks like this:

Screen Shot 2020-02-21 at 1 21 00 PM

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
@@ -4,6 +4,8 @@ import { ExecutionDetailsSection, IExecutionDetailsSectionProps, StageFailureMes
 import { IGitTrigger } from 'core/domain';
 import { SETTINGS } from 'core/config';
 
+const NOT_FOUND = 'Not found';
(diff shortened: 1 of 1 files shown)


erikmunson:
my 2 cents: if something about the trigger is _misconfigured_, I'd expect it to show up as a validation error on the trigger. In the open source world the only way to run into this problem is if you misconfigure your Git trigger (i.e. don't add the required fields), so in that world you'd already be seeing a validation error while configuring. 

The reason we never saw any config-time validation internally at Netflix is because our internal source code trigger is actually a source code trigger _plus_ and artifact trigger _plus_ a build trigger, so when you leave out critical bits of config for a git/source code trigger we look at it and go 'well, you configured this _artifact_ trigger correctly, so I guess you're good to go'.

I'd rather look at that trigger validation problem as the one to solve (probably by splitting apart our internal trigger) since that problem applies not just to this stage, but to anything regarding validation (for all 3 kinds of triggers).

PR #7939, opened 2020-02-24
feat(core/managed): add managed resource history modal

erikmunson:
**Depends on #7936 and #7937, don't merge yet.**

We've had a nifty API for a while that lets you see an event history of all the things that have happened to a managed resource over time — changes in desired/actual state, tasks being launched, etc. Now it's time to make all that available to folks directly in deck while they're looking at a resource. As a fun bonus, this is also one of the first views in deck that has full (for some definition of "full") mobile/small screen support. Here's what it looks like:

<details>
<summary>Smaller screen</summary>
 
Screen Shot 2020-02-24 at 9 28 46 AM
Screen Shot 2020-02-24 at 9 29 04 AM
</details>

<details>
<summary>Larger screen</summary>
 
Screen Shot 2020-02-24 at 9 26 51 AM
Screen Shot 2020-02-24 at 9 27 01 AM
</details>

<details>
<summary>Interactions on small screens</summary>
 
history-modal-mobile
</details>

<details>
<summary>Interactions on larger screens</summary>
 
history-modal-desktop
</details>

<details>
<summary>Expanding all events at once (larger screens only)</summary>
 
history-modal-desktop-expandall
</details>

<details>
<summary>When we can't get events from the API</summary>
 
Screen Shot 2020-02-23 at 1 30 09 PM
</details>

Some things I didn't do in this PR, but will probably add shortly:
 - Polling every N seconds for updated events (and displaying them at the top)
 - Help text for each event type (similar to how we have help text for each resource status
 - Maybe a banner that appears at the top of the modal when the resource is paused, with a link to unpause?

Feedback is, as always, very welcome on all aspects.

cc @gcomstock @emjburns @luispollo @gal-yardeni @asher @robfletcher @lorin

### DIFF
--- a/app/scripts/modules/core/src/domain/IManagedEntity.ts
+++ b/app/scripts/modules/core/src/domain/IManagedEntity.ts
@@ -34,3 +34,56 @@ export interface IManagedResource {
   managedResourceSummary?: IManagedResourceSummary;
   isManaged?: boolean;
 }
+
+export enum ManagedResourceEventType {
+  ResourceCreated = 'ResourceCreated',
+  ResourceUpdated = 'ResourceUpdated',
+  ResourceDeleted = 'ResourceDeleted',
+  ResourceMissing = 'ResourceMissing',
+  ResourceValid = 'ResourceValid',
+  ResourceDeltaDetected = 'ResourceDeltaDetected',
+  ResourceDeltaResolved = 'ResourceDeltaResolved',
+  ResourceActuationLaunched = 'ResourceActuationLaunched',
+  ResourceCheckError = 'ResourceCheckError',
+  ResourceCheckUnresolvable = 'ResourceCheckUnresolvable',
+  ResourceActuationPaused = 'ResourceActuationPaused',
+  ResourceActuationResumed = 'ResourceActuationResumed',
+}
+
+export interface IManagedResourceDiff {
+  [fieldName: string]: {
+    key: string;
+    diffType: 'CHANGED' | 'ADDED' | 'REMOVED';
+    desired?: string;
+    actual?: string;
+    fields: IManagedResourceDiff;
+  };
+}
+
+export interface IManagedResourceEvent {
+  type: ManagedResourceEventType;
+  apiVersion: string;
+  kind: string;
+  id: string;
+  application: string;
+  timestamp: string;
+  plugin?: string;
+  tasks?: Array<{ id: string; name: string }>;
+  delta?: IManagedResourceDiff;
+  message?: string;
+  reason?: string;
+}
+
+export type IManagedResourceEventHistory = IManagedResourceEvent[];
+
+export type IManagedResourceEventHistoryResponse = Array<
+  Omit<IManagedResourceEvent, 'delta'> & {
+    delta?: {
+      [key: string]: {
+        state: 'CHANGED' | 'ADDED' | 'REMOVED';
+        desired: string;
+        current: string;
+      };
+    };
+  }
+>;
--- a/app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
+++ b/app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
@@ -1,0 +1,227 @@
+import React from 'react';
+import classNames from 'classnames';
+import { DateTime } from 'luxon';
+import { UISref } from '@uirouter/react';
+
+import {
+  ModalHeader,
+  ModalBody,
+  Table,
+  TableRow,
+  TableCell,
+  standardGridTableLayout,
+  useData,
+} from 'core/presentation';
+
+import { relativeTime, timestamp } from 'core/utils';
+import { IManagedResourceSummary, IManagedResourceDiff } from 'core/domain';
+import { AccountTag } from 'core/account';
+import { ManagedReader } from 'core/managed';
+import { Spinner } from 'core/widgets';
+
+import { ManagedResourceDiffTable } from './ManagedResourceDiffTable';
+
+import './ManagedResourceHistoryModal.less';
+
+export interface IManagedResourceHistoryModalProps {
+  resourceSummary: IManagedResourceSummary;
+}
+
+const { useMemo } = React;
+
+const viewConfigurationByEventType = {
+  ResourceCreated: {
+    displayName: 'Created',
+    iconClass: 'icon-md-created',
+    level: 'info',
+  },
+  ResourceUpdated: {
+    displayName: 'Config updated',
+    // Needs it's own icon
+    iconClass: 'icon-md-created',
+    level: 'info',
+  },
+  ResourceDeleted: {
+    displayName: 'Deleted',
+    // Needs it's own icon
+    iconClass: 'icon-md-missing-resource',
+    level: 'info',
+  },
+  ResourceMissing: {
+    displayName: 'Missing',
+    iconClass: 'icon-md-missing-resource',
+    level: 'info',
+  },
+  ResourceValid: {
+    displayName: 'Valid',
+    // Needs it's own icon
+    iconClass: 'icon-md-delta-resolved',
+    level: 'info',
+  },
+  ResourceDeltaDetected: {
+    displayName: 'Drift detected',
+    iconClass: 'icon-md-delta-detected',
+    level: 'info',
+  },
+  ResourceDeltaResolved: {
+    displayName: 'Drift resolved',
+    iconClass: 'icon-md-delta-resolved',
+    level: 'info',
+  },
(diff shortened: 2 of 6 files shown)


emjburns:
(1) A dumb question: for the un-expanded events, the carrot is a backwards carrot `<` when closed. I thought people usually used `>`? Like for example your attached images and gifs are under a filled in carrot like `>` (not that this matters, I'm just wondering, clearly it is bothering me a little).
(2) Should the row you're looking at or just clicked on be highlighted in the large view? If you're debugging something in that view I think it would be nice to have some sort of visual anchoring - the events look incredibly similar.
(3) Are you planning on doing filtering by time or event type from this view, or is it going to be in another view? 

(4) I love everything and can't wait to use this!!!

app/scripts/modules/core/src/domain/IManagedEntity.ts
```diff
@@ -34,3 +34,56 @@ export interface IManagedResource {
   managedResourceSummary?: IManagedResourceSummary;
   isManaged?: boolean;
 }
+
+export enum ManagedResourceEventType {
+  ResourceCreated = 'ResourceCreated',
```
gal-yardeni: We have 2 more resource event types that I think we should include here - `ResourceTaskFailed` and `ResourceTaskSucceeded` - which actually indicates the orca task status of the resource.

erikmunson: I originally wanted to include these, but none of them appear to have any sort of data attached to them about _which specific task they are in reference to_. I found it a bit confusing / weird to look at because there would be a `ResourceActuationLaunched` event with a specific task, then a bunch of other events (e.g. `ResourceDeltaResolved`, `ResourceValid`), then a `ResourceTaskFailed` or `ResourceTaskSucceeded` without any task info. Made it tough to follow exactly what was happening.

If we can get those events to include a `tasks` list the same way `ResourceActuationLaunched` has, I'll make sure to surface them here.

app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
```diff
@@ -0,0 +1,227 @@
...
+      )}
+      {diff && <ManagedResourceDiffTable diff={diff} />}
+    </div>
+  );
+};
+
+export const ManagedResourceHistoryModal = ({ resourceSummary }: IManagedResourceHistoryModalProps) => {
+  const {
+    id,
+    moniker: { app, stack, detail },
```
luispollo: Since you're joining the moniker parts into a single name just below, and considering that @robfletcher is in the process of removing monikers, would it make sense to parse the API result directly into a single name property?

erikmunson: We generally prefer using `moniker` objects as the canonical representation in UI code — 9 times out of 10 the non-trivial things we want to do with names/monikers are super inconvenient if you have to constantly parse out the name into an object (and then give it a name and pass it around, etc. etc.). Joining on `-` for actually plopping the name on the screen at the end of all the complicated data massaging is the easy part 😉

luispollo:
@erikmunson This looks great! The one thing I missed seeing that I think would make a big difference is some visual indication as to what type of resource each event refers to. Even in the expanded view, you have to look at the details and try to infer what the resource is based on the properties displayed. 

For example, take this event:
image

Looking at the details of the delta, I can see there was a change in the `launchConfiguration`, so it must be a cluster. And since the `imageId` is for an AMI, it must be an EC2 cluster. It'd be nice if that was apparent in the list and/or details. Maybe we could use the same icons we use in the infrastructure view?

erikmunson:
> Even in the expanded view, you have to look at the details and try to infer what the resource is based on the properties displayed.

@luispollo fwiw I do agree in the context of us expanding this view out to surface events from multiple resources at the same time, and even as-is it'd be a nice visual cue. I'm just avoiding that rabbit hole for now in favor of the 3 improvements I listed in the PR description, as I currently think those are the biggest value-adds to what's implemented here.

erikmunson:
@emjburns I'm gonna try to answer your questions in order:

> A dumb question: for the un-expanded events, the carrot is a backwards carrot < when closed. I thought people usually used >? Like for example your attached images and gifs are under a filled in carrot like > (not that this matters, I'm just wondering, clearly it is bothering me a little).

Not a dumb question, it's a new design with a new language to learn. Those expand-y bits are part of [our new family of Table components](https://github.com/spinnaker/deck/pull/7903), and I'll let @gcomstock chime in if he disagrees with my explanation:

Generally the reason you see arrows in that `>` orientation (like in the GitHub example) is because they are pointing _toward the content_. Visually that means that as you open the expandable thing, the arrow flips to sort of 'unfurl' the content, and 'retracts' back up when you close it (almost like a window shade or a door). In the case of this design, the expand-y thing is on the right hand side, so we mirror the icon to give it the exact same effect/quality. It's still pointing toward the content, just from the opposite side. I believe the reason we've moved to keeping the expansion indicators on the right instead of the left is because it's often pretty distracting/noisy to have a bunch of identical arrows as the first thing you see when you're trying to quickly scan a bunch of list items — moving it to the right means that as you scan, you can focus on the content.

Separately, if we kept it on the right but flipped the icon we'd actually be using a _different_ visual pattern which has come to be broadly accepted as hinting that by selecting a row you'll get launched into a whole different detail view — a good example is rows / lists in iOS (like in the Settings app) where tapping on a row with the little `>` means you're going to dive into a different screen entirely.

> Should the row you're looking at or just clicked on be highlighted in the large view? If you're debugging something in that view I think it would be nice to have some sort of visual anchoring - the events look incredibly similar.

You mean like when you click a button and it gets a glowing outline around it until you click on something else? We already add a white background to anything expanded to help it stand out, but it's not a bad idea to have a focus state (which is what those glow-y/outline-y things on buttons are called) so you can see what you most recently clicked on. Wiring that up would be a bit too finicky for me to want to do it in this PR but I'll stick it in my notes.

> Are you planning on doing filtering by time or event type from this view, or is it going to be in another view?

Everything is always changing, but at the moment the plan is to (later) add some filtering options above the table that let you slice and dice which kinds of evens you want to see, the timeframe, types of resources, etc. Then when you click on a thing in the UI to open a history modal, we'd automatically choose some filters to match what you want to see — e.g. if I open a history modal from a specific cluster automatically filter to just that cluster's events, if I click on something for a whole environment/app/etc. then filter to that scope. There's a _ton_ of additional complexity hiding in there (making the right API calls at the right times, paginating within a giving filtered query, etc.) but we'll get there when the time is right.

app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+import './ManagedResourceHistoryModal.less';
+
+export interface IManagedResourceHistoryModalProps {
+  resourceSummary: IManagedResourceSummary;
+}
+
+const { useMemo } = React;
+
+const viewConfigurationByEventType = {
```
erikmunson: @gcomstock fyi, I commented in here some event types that we don't currently have bespoke icons for. In the absence of specific icons, I've temporarily chosen some existing MD icons that I think are close enough to work for now (you can see some of them in the screenshots)

app/scripts/modules/core/src/domain/IManagedEntity.ts
gal-yardeni: Following our conversation, I doubled check and we currently do not log the task name. Adding an example payload for the `ResourceTaskFailed` event : `{"type":"ResourceTaskFailed","apiVersion":"titus.spinnaker.netflix.com/v1","kind":"cluster","id":"titus:cluster:titustestvpc:gyardeniapp-env4","application":"gyardeniapp","reason":"Too many retries.  10 attempts have been made to bake http://bakery.test.netflix.net:7001/api/v1/us-east-1/bake/b-4065khb8bt1frwg1jxjkw5b1y4 ... its probably not going to happen.","timestamp":"2020-02-11T23:27:02.989Z"}`. I'll be happy to add the task name if you think it's important info!

erikmunson: Ah ok got it — thanks for checking! I'd love it if we could add the exact same `tasks` list that's [present on `ResourceActuationLaunched`](https://github.com/spinnaker/keel/blob/master/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/events/ResourceEvent.kt#L200) to both `ResourceTaskFailed` _and_ `ResourceTaskSucceeded`. 

I made us a GitHub issue so we can keep track: https://github.com/spinnaker/keel/issues/829

PR #7940, opened 2020-02-24
fix(amazon/iamRole): Do not default iamRole in deploy edit scenario

vigneshm:
While editing a deploy stage without an `iamRole` specified, deck shows a default value. To the end-user, it appears that this field is set but it is empty. Fixing that in this PR so that the UI reflects the exact configuration.


christopherthielen:
@vigneshm can you rebase this and fix your commit comment?

vigneshm:
Rebased and updated the commit message

christopherthielen:
@vigneshm  fwiw, the comment should start with `fix(amazon/serverGroup)`, to fit the pattern `fix(module/submodule)`

vigneshm:
> @vigneshm fwiw, the comment should start with `fix(amazon/serverGroup)`, to fit the pattern `fix(module/submodule)`

Ah, my bad.

PR #7949, opened 2020-02-24
fix(ecs): use preferSourceCapacity instead of useSourceCapacity when configuring a serverGroup

pkandasamy91:
Use preferSourceCapacity instead of useSourceCapacity when configuring a serverGroup to use the previous server groups capacity.

Fixes: https://github.com/spinnaker/spinnaker/issues/4690

Screen Shot 2020-02-24 at 3 37 45 PM

Screen Shot 2020-02-24 at 3 38 42 PM

### DIFF
--- a/app/scripts/modules/ecs/src/serverGroup/configure/serverGroupCommandBuilder.service.js
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/serverGroupCommandBuilder.service.js
@@ -149,6 +149,7 @@ angular
               ecsClusterName: '',
               targetGroup: '',
               copySourceScalingPoliciesAndActions: true,
+              preferSourceCapacity: true,
               useSourceCapacity: true,
               viewState: {
                 useAllImageSelection: false,
--- a/app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
@@ -52,7 +52,11 @@
   <div class="form-group">
     <div class="col-md-12 checkbox">
       <label>
-        <input type="checkbox" ng-model="$ctrl.command.useSourceCapacity" />
+        <input
+          type="checkbox"
+          ng-model="$ctrl.command.useSourceCapacity"
+          ng-change="$ctrl.command.preferSourceCapacity=$ctrl.command.useSourceCapacity"
+        />
         <b>If available, use the previous server group's capacity</b>
       </label>
       <help-field key="ecs.capacity.overwrite"></help-field>


allisaurus:
can you confirm after making this change that you can set this option and:
1. successfully deploy a brand new server group (error case defined in https://github.com/spinnaker/spinnaker/issues/4690 ) 
2. successfully deploy to an existing server group

if you've already done these tests, a couple summary bullet points in the description are fine(or you can add screen shots of the successful deployments for bonus points)

clareliguori:
I can't find any other classes that actually use the preferSourceCapacity field other than this aws provider-specific one, and it looks to me like BOTH useSourceCapacity and preferSourceCapacity would have to be set:
https://github.com/spinnaker/orca/blob/fdc5f6a53affddc56051171accd97493fdc7b660/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/CaptureSourceServerGroupCapacityTask.groovy#L42-L44

pkandasamy91:
@spinnakerbot add-label do-not-merge

pkandasamy91:
> can you confirm after making this change that you can set this option and:
> 
>     1. successfully deploy a brand new server group (error case defined in [spinnaker/spinnaker#4690](https://github.com/spinnaker/spinnaker/issues/4690) )
> 
>     2. successfully deploy to an existing server group
> 
> 
> if you've already done these tests, a couple summary bullet points in the description are fine(or you can add screen shots of the successful deployments for bonus points)

Successfully deployed in both cases!

pkandasamy91:
> I can't find any other classes that actually use the preferSourceCapacity field other than this aws provider-specific one, and it looks to me like BOTH useSourceCapacity and preferSourceCapacity would have to be set:
> https://github.com/spinnaker/orca/blob/fdc5f6a53affddc56051171accd97493fdc7b660/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/CaptureSourceServerGroupCapacityTask.groovy#L42-L44

@clareliguori @allisaurus PR ready for review. 

Currently, setting preferSourceCapacity to the value of useSourceCapacity, not sure what if we should continue to do that or always set preferSourceCapacity to true (both cases successfully deploy and allow users to enable/disable useSourceCapacity successfully)

Setting preferSourceCapacity true without setting useSourceCapacity as true does not seem to have any impact to the deployment, and does not use the current server group configurations.

pkandasamy91:
> > I can't find any other classes that actually use the preferSourceCapacity field other than this aws provider-specific one, and it looks to me like BOTH useSourceCapacity and preferSourceCapacity would have to be set:
> > https://github.com/spinnaker/orca/blob/fdc5f6a53affddc56051171accd97493fdc7b660/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/pipeline/providers/aws/CaptureSourceServerGroupCapacityTask.groovy#L42-L44
> 
> @clareliguori @allisaurus PR ready for review.
> 
> Currently, setting preferSourceCapacity to the value of useSourceCapacity, not sure if we should continue to do that or always set preferSourceCapacity to true (both cases successfully deploy and allow users to enable/disable useSourceCapacity successfully)
> 
> Setting preferSourceCapacity true without setting useSourceCapacity as true does not seem to have any impact to the deployment, and does not use the current server group configurations.

Spoke to Allie offline, and don't think we should worry about this unless we decide to change the Horizontal Scaling section of the create server group modal to include fail or fallback options

PR #7952, opened 2020-02-25
refactor(core/account): In AccountSelectInput, delegate to SelectInput and ReactSelectInput

christopherthielen:
3 commits:
- Fix an async test
- Support disabled options in `<SelectInput />`
- Delegate to `SelectInput` and `ReactSelectInput` in `AccountSelectInput`

### DIFF
--- a/app/scripts/modules/core/src/account/AccountSelectInput.tsx
+++ b/app/scripts/modules/core/src/account/AccountSelectInput.tsx
@@ -1,10 +1,12 @@
 import React from 'react';
 import { $q } from 'ngimport';
 import { flatten, isEqual, map, uniq, xor } from 'lodash';
-import Select, { Option } from 'react-select';
+import { Option } from 'react-select';
 
 import { createFakeReactSyntheticEvent } from 'core/presentation/forms/inputs/utils';
-import { IFormInputProps } from '../presentation';
+import { IFormInputProps } from 'core/presentation/forms/inputs';
+import { ReactSelectInput } from 'core/presentation/forms/inputs/ReactSelectInput';
+import { SelectInput } from 'core/presentation/forms/inputs/SelectInput';
 
 import { AccountService, IAccount } from './AccountService';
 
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/account/AccountSelectInput.tsx
```diff
@@ -94,73 +96,17 @@ export class AccountSelectInput extends React.Component<IAccountSelectInputProps
     }
   }
 
-  private SimpleSelect = () => {
```
christopherthielen: replaced this with `<SelectInput/>`

app/scripts/modules/core/src/account/AccountSelectInput.tsx
```diff
@@ -1,10 +1,12 @@
 import React from 'react';
 import { $q } from 'ngimport';
 import { flatten, isEqual, map, uniq, xor } from 'lodash';
-import Select, { Option } from 'react-select';
+import { Option } from 'react-select';
 
 import { createFakeReactSyntheticEvent } from 'core/presentation/forms/inputs/utils';
-import { IFormInputProps } from '../presentation';
+import { IFormInputProps } from 'core/presentation/forms/inputs';
+import { ReactSelectInput } from 'core/presentation/forms/inputs/ReactSelectInput';
```
christopherthielen: I had to import deeply to avoid a circular module dependency 😢 
Some day we'll build these packages with rollup and find all these circular dependencies (there will be a LOT)

app/scripts/modules/core/src/account/AccountSelectInput.tsx
```diff
@@ -94,73 +96,17 @@ export class AccountSelectInput extends React.Component<IAccountSelectInputProps
...
-          ))}
-        </select>
-      </div>
-    );
-  };
+  public render() {
+    const { value, readOnly } = this.props;
+    const { mergedAccounts } = this.state;
 
-  private FilterableSelect = () => {
```
christopherthielen: replaced this with `<ReactSelectInput>`

PR #7959, opened 2020-02-26
feat(managed-delivery): Improve errors displayed from unparseable delivery configs

luispollo:
Adds a bit more logic to the Import Delivery Config stage's execution details component so it displays more user-friendly error messages when the error is in parsing a delivery config.

Closes https://github.com/spinnaker/keel/issues/818

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
@@ -43,14 +43,16 @@ function buildCustomErrorMessage(error: IDeliveryConfigImportError) {
     return CustomErrorMessage(error.message);
   }
 
+  const pathExpression = error.details.pathExpression.replace('.', '/');
+  /* eslint-disable */
   const errorMessage =
     {
-      missing_property: `The following property is missing: \`${error.details.pathExpression}\``,
-      invalid_type: `The type of property \`${error.details.pathExpression}\` is invalid.`,
-      invalid_format: `The format of property \`${error.details.pathExpression}\` is invalid.`,
-      invalid_value: `The value of property \`${error.details.pathExpression}\` is invalid.`,
+      missing_property: `The following property is missing: \`${pathExpression}\``,
+      invalid_type: `The type of property \`${pathExpression}\` is invalid.`,
+      invalid_format: `The format of property \`${pathExpression}\` is invalid.`,
+      invalid_value: `The value of property \`${pathExpression}\` is invalid.`,
     }[error.details.error] || 'Unknown error';
-
+  /* eslint-enable */
   return CustomErrorMessage(errorMessage, error.details.message);
 }
 
--- a/app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
@@ -24,11 +24,11 @@ const CustomErrorMessage = (message: string, debugDetails?: string) => {
       <div className="alert alert-danger">
         <b>There was an error parsing your delivery config file.</b>
         <br />
-        <Markdown message={message} />
+        <Markdown message={message} style={{ wordBreak: 'break-all' }} />
         <br />
         {debugDetails && (
           <CollapsibleSection heading={({ chevron }) => <span>{chevron} Debug Details</span>}>
-            <pre>{debugDetails}</pre>
+            <pre style={{ whiteSpace: 'pre-wrap' }}>{debugDetails}</pre>
           </CollapsibleSection>
         )}
       </div>
@@ -45,14 +45,16 @@ function buildCustomErrorMessage(error: IDeliveryConfigImportError) {
     return CustomErrorMessage(error.message);
   }
 
-  const pathExpression = error.details.pathExpression.replace('.', '/');
-  /* eslint-disable */
+  // Replace dots with slashes for paths because it feels more familiar. Also ditch the very first slash/dot as it just adds noise.
+  const pathExpression = error.details.pathExpression.substring(1).replace(/\./g, '/');
+  // Turn off linter for the block below because it wants CamelCase property names.
+  /* eslint-disable @typescript-eslint/camelcase */
   const errorMessage =
     {
-      missing_property: `The following property is missing: \`${pathExpression}\``,
-      invalid_type: `The type of property \`${pathExpression}\` is invalid.`,
-      invalid_format: `The format of property \`${pathExpression}\` is invalid.`,
-      invalid_value: `The value of property \`${pathExpression}\` is invalid.`,
+      missing_property: `The following property is missing: <br/> \`${pathExpression}\``,
+      invalid_type: `The type of the following property is invalid: <br/> \`${pathExpression}\``,
+      invalid_format: `The format of the following property is invalid: <br/> \`${pathExpression}\``,
+      invalid_value: `The value of the following property is invalid: <br/> \`${pathExpression}\``,
     }[error.details.error] || 'Unknown error';
   /* eslint-enable */
   return CustomErrorMessage(errorMessage, error.details.message);
--- a/app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
@@ -6,41 +6,41 @@ import { SETTINGS } from 'core/config';
 
 const NOT_FOUND = 'Not found';
 
-export class DeliveryConfigImportErrorDetails {
+interface IDeliveryConfigImportErrorDetails {
   error: string;
   message: string;
   pathExpression: string;
-
-  getErrorMessage() {
-    let errorMessage = 'Error parsing delivery config: <br/>';
-    switch (this.error) {
-      case 'missing_property':
-        errorMessage += 'The following property is missing: ' + this.pathExpression;
-        break;
-
-      case 'invalid_type':
-        errorMessage += 'The type of property `' + this.pathExpression + '` is invalid.';
-        break;
-
-      case 'invalid_format':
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
```diff
@@ -6,6 +6,43 @@ import { SETTINGS } from 'core/config';
 
 const NOT_FOUND = 'Not found';
 
+export class DeliveryConfigImportErrorDetails {
```
erikmunson: @luispollo let's meet up in person and chat through how this works/how you want it to work — I see what you're trying to do and I think you're starting down the right general direction, but I also get the sense you _might_ be applying some of the implicit behavior you'd expect to see in Java-land around automatically turning data structures into classes with methods on them. 

I think it's easier to explain how this would work in JS/TS if we do it verbally vs. trying to write out the behavior in a PR comment.

luispollo: Sure. I'll be in the office on Thursday, let's meet then.

luispollo:
@erikmunson Hopefully the code looks better now. Here's what the UI looks like:
image
image

If you have any ideas on how to make either better, please let me know!

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
```diff
@@ -2,10 +2,58 @@ import React from 'react';
...
+  if (!error.details) {
+    return CustomErrorMessage(error.message);
+  }
+
+  const errorMessage =
+    {
+      missing_property: `The following property is missing: \`${error.details.pathExpression}\``,
+      invalid_type: `The type of property \`${error.details.pathExpression}\` is invalid.`,
+      invalid_format: `The format of property \`${error.details.pathExpression}\` is invalid.`,
+      invalid_value: `The value of property \`${error.details.pathExpression}\` is invalid.`,
```
luispollo: I have the keys as strings (e.g. `'missing_property'`) but when I commit the file it looks like there's a pre-commit hook that runs and changes that into symbols, except the linter complains they have to be camel-case. 😝 

Not sure how to fix that...

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
```diff
@@ -2,10 +2,62 @@ import React from 'react';
...
+
+  const pathExpression = error.details.pathExpression.replace('.', '/');
+  /* eslint-disable */
+  const errorMessage =
+    {
+      missing_property: `The following property is missing: \`${pathExpression}\``,
+      invalid_type: `The type of property \`${pathExpression}\` is invalid.`,
+      invalid_format: `The format of property \`${pathExpression}\` is invalid.`,
+      invalid_value: `The value of property \`${pathExpression}\` is invalid.`,
+    }[error.details.error] || 'Unknown error';
```
luispollo: The code formatter we run in the git pre-commit hook is changing the properties in this object from strings to symbols as you can see. This is what's causing the build to break:
```
  src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx:51:5 - error TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ missing_property: string; invalid_type: string; invalid_format: string; invalid_value: string; }'.
    No index signature with a parameter of type 'string' was found on type '{ missing_property: string; invalid_type: string; invalid_format: string; invalid_value: string; }'.
```
How do I stop the formatter from messing with my code? 🤔

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
```diff
@@ -2,10 +2,69 @@ import React from 'react';
...
+          <CollapsibleSection heading={({ chevron }) => <span>{chevron} Debug Details</span>}>
+            <pre style={{ whiteSpace: 'pre-wrap' }}>{debugDetails}</pre>
+          </CollapsibleSection>
+        )}
+      </div>
+    </div>
+  );
+};
+
+function buildCustomErrorMessage(error: IDeliveryConfigImportError) {
```
erikmunson: Let's stay consistent with function syntax within the same file:

```suggestion
const buildCustomErrorMessage = (error: IDeliveryConfigImportError) => {
```

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
```diff
@@ -2,10 +2,69 @@ import React from 'react';
...
+  message: string;
+  pathExpression: string;
+}
+
+interface IDeliveryConfigImportError {
+  message: string;
+  details?: IDeliveryConfigImportErrorDetails;
+}
+
+const CustomErrorMessage = (message: string, debugDetails?: string) => {
```
erikmunson: Let's turn this into a React component — you'll end up with something that looks like this:

```tsx
const CustomErrorMessage = ({ message, debugDetails }: { message: string, debugDetails?: string }) => {
```
And then rather than calling it as a function, you'll render it like this:

```tsx
<CustomErrorMessage message={errorMessage} debugDetails={error.details.message} />
```

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
```diff
@@ -2,10 +2,69 @@ import React from 'react';
...
+  if (!error.details) {
+    return CustomErrorMessage(error.message);
+  }
+
+  // Replace dots with slashes for paths because it feels more familiar. Also ditch the very first slash/dot as it just adds noise.
+  const pathExpression = error.details.pathExpression.substring(1).replace(/\./g, '/');
+
+  const errorMessage = ERROR_MESSAGE_MAP[error.details.error]
+    ? ERROR_MESSAGE_MAP[error.details.error] + `<br/>\`${pathExpression}\``
+    : 'Unknown error';
```
erikmunson:
```suggestion
  const errorMessage = ERROR_MESSAGE_MAP[error.details.error]
    ? `${ERROR_MESSAGE_MAP[error.details.error]} <br/>\`${pathExpression}\``
    : 'Unknown error';
```

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
```diff
@@ -2,10 +2,69 @@ import React from 'react';
...
+
+  return CustomErrorMessage(errorMessage, error.details.message);
```
erikmunson: In addition to turning `CustomErrorMessage` into a React component, I recommend decoupling this `buildCustomErrorMessage` function from `CustomErrorMessage`. Instead, consider returning the error message/details as data and then passing that data to the component. For example instead of the code currently on line 99 you might write:

```tsx
// maybe buildErrorMessage returns an object like { summary: string, debugDetails: string }?
const errorMessage = buildErrorMessage(stage.context.error as IDeliveryConfigImportError);

{errorMessage ? <CustomErrorMessage message={errorMessage.summary} debugDetails={errorMessage.debugDetails} /> : <StageFailureMessage stage={stage} />}
```

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
```diff
@@ -2,10 +2,69 @@ import React from 'react';
...
 import { ExecutionDetailsSection, IExecutionDetailsSectionProps, StageFailureMessage } from 'core/pipeline';
 import { IGitTrigger } from 'core/domain';
+import { CollapsibleSection, Markdown } from 'core/presentation';
 import { SETTINGS } from 'core/config';
 
 const NOT_FOUND = 'Not found';
 
+// Turn off linter for the block below because it wants CamelCase property names.
+/* eslint-disable @typescript-eslint/camelcase */
+const ERROR_MESSAGE_MAP: any = {
```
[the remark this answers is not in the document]
luispollo: This is what ultimately fixed the TypeScript error about trying to index into the object with a string.

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
luispollo: I wrote this in both formats but weren't sure one was more idiomatic than the other. What's the advantage of the `const` form?

erikmunson: there's not really a huge difference in practice most of the time, we just tend to stick to one style when we're in the same file/module.

the biggest different between the two listed here is that the `const` version is using an arrow function (`() => {}`), which binds the function to the "lexical" `this` rather than giving each function invocation its own fresh `this` binding. If that doesn't make any sense to you right now, it's because old school JS's `this` semantics were super confusing and basically made no sense 😆. Even though I'm using an arrow function in this example, you could also assign a regular `function` to a `const` binding like `const buildErrorMessage = function buildErrorMessage() {};`. That has it's own bizarre differences ('hoisting' behavior is different), but I recommend you just worry about that right now! 😉

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
luispollo: Done!

app/scripts/modules/core/src/pipeline/config/stages/managed/ImportDeliveryConfigExecutionDetails.tsx
luispollo: Done!

PR #7960, opened 2020-02-26
fix(core/managed): add even more message fields on history events

erikmunson:
Turns out we have even more kinds of fields for messages on these history events than I had previously thought. This just adds one more. Looks like the other error message treatments:

Screen Shot 2020-02-25 at 6 45 29 PM


erikmunson:
i super pinky promise.

PR #7961, opened 2020-02-26
fix(core/managed): remove arrow visual treatment on history modal

erikmunson:
We've consistently gotten feedback that the grey arrow on each resource in the `Where` column is confusing because it _looks_ like an expand/collapse arrow, but it isn't. That combined with folks needing to re-learn that expand/collapse controls are on the right has been leading to a bit of awkwardness. Since for now there's really only two sections to the hierarchy in the `Where` column (account + resource name) I'm just removing it. We can figure out what visual treatment works better to communicate hierarchy better separately.

Before: 

Screen Shot 2020-02-26 at 1 58 50 PM

After:

Screen Shot 2020-02-26 at 1 55 06 PM


erikmunson:
fyi @gcomstock

gcomstock:
Good call! Yeah we'll solve the problem as needed. Thanks!

PR #7964, opened 2020-02-26
fix(test/functional): fix functional tests by increasing the timeout to 60 seconds

christopherthielen:
Looks like we can "fix" the functional tests by bumping up the timeout.  I'd like to merge this before the migration to cypress lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,7 +89,7 @@ jobs:
 
       - name: Functional Tests
         if: steps.purebump.outputs.ispurebump != 'true'
-        run: node test/functional/run.js --replay-fixtures --serve-deck --imposter-port=8084 --browser=chrome --headless
+        run: yarn functional
 
   build:
     name: Production Build
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "int": "protractor protractor.conf.js",
     "modules": "./gradle/buildModules.sh",
     "prettier": "prettier --write 'app/**/*.{js,jsx,ts,tsx,html,css,less}'",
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -89,7 +89,7 @@ jobs:
 
       - name: Functional Tests
         if: steps.purebump.outputs.ispurebump != 'true'
-        run: node test/functional/run.js --replay-fixtures --serve-deck --imposter-port=8084 --browser=chrome --headless
+        run: yarn functional
```
christopherthielen: instead of duplicating this, delegate to the yarn task.

package.json
```diff
@@ -25,7 +25,7 @@
     "int": "protractor protractor.conf.js",
     "modules": "./gradle/buildModules.sh",
     "prettier": "prettier --write 'app/**/*.{js,jsx,ts,tsx,html,css,less}'",
-    "functional": "test/functional/run.js --serve-deck --replay-fixtures --browser chrome --headless"
+    "functional": "test/functional/run.js --replay-fixtures --serve-deck --imposter-port=8084 --browser=chrome --headless"
```
christopherthielen: make the yarn task match what was in the workflow

test/functional/tests/core/pages/Page.ts
```diff
@@ -2,7 +2,7 @@ import '@wdio/sync';
 import { defaultsDeep } from 'lodash';
 
 const DEFAULT_OPTIONS = {
-  awaitExistsTime: 10000,
+  awaitExistsTime: 60000,
```
christopherthielen: 640k should be enough for anyone

PR #7967, opened 2020-02-27
fix(core): Fix missing maxRemainingAsgs

alanmquach:
There were 2 places where we failed to hydrate the `redblack` strategy with it's standard parameters (one of which is maxRemaningAsgs):

1. When adding a new cluster configuration but opting to copy the configuration from an existing server group, we were simply setting `command.strategy = 'redblack'` without running the `initializationMethod`
2. When changing from a strategy other than `none` to `redblack`, the `initializationMethod` actually gets run, however we set `maxRemaningAsgs` to `2` if `!command.strategy`, but that does not guarantee that `command.maxRemainingAsgs` has a value.


app/scripts/modules/core/src/deploymentStrategy/strategies/redblack/redblack.strategy.ts
```diff
@@ -28,7 +28,7 @@ DeploymentStrategyRegistry.registerStrategy({
       rollback: {
         onFailure: false,
       },
-      maxRemainingAsgs: !command.strategy ? 2 : command.maxRemainingAsgs,
```
christopherthielen: Can you verify that this Boolean logic is correct?

alanmquach: The old logic? I imagine it was there to preserve the value if someone were to switch strategies then switch back

alanmquach: Oh but I suppose `!command.strategy` was probably never very correct to start. `command.maxRenainingAsgs === undefined` is probably more appropriate

marchello2000:
❤️

app/scripts/modules/core/src/deploymentStrategy/strategies/redblack/redblack.strategy.ts
```diff
@@ -28,7 +28,7 @@ DeploymentStrategyRegistry.registerStrategy({
...
+      maxRemainingAsgs: command.maxRemainingAsgs !== undefined ? command.maxRemainingAsgs : 2,
```
christopherthielen: wondering if you're intentionally excluding `null` check, or if we can do:

```suggestion
      maxRemainingAsgs: command.maxRemainingAsgs ?? 2,
```

alanmquach: Oh right! That's a thing we can do now! Nope, not intentional.

PR #7968, opened 2020-02-27
refactor(core/presentation): Switch the default value of formik fastField from true to false

christopherthielen:
The behavior caused by this default has surprised me many times and wasted much time.  I think we should default to false and allow form creators to opt in if the form is slow.


maggieneterval:
:tada: :+1:

PR #7969, opened 2020-02-27
Issues with UI #5484

vvtelu:
fix(deck):Issues with UI [#5484](https://github.com/spinnaker/spinnaker/issues/5484)

Alignment of text is fixed, by default showing one account and rest gets displayed (n More)


maggieneterval:
@vvtelu thanks for the PR. I'm not sure this change completely resolves the issue you linked to -- if the first account's name is very long, won't we run into the same issue with the pipeline name overflowing? Also, the issue you linked to says `clicking on configure on the pipeline name results in No pipeline with that name, whereas selecting the same pipeline from the configure dropdown works correctly.` -- were you able to reproduce that issue?

maggieneterval:
Hi @vvtelu, hope all is well! I'm going to close this PR for now, but if you'd like to continue discussing solutions to this problem, let's keep the conversation going on the [GitHub issue](https://github.com/spinnaker/spinnaker/issues/5484). Thanks!

PR #7980, opened 2020-02-28
fix(amazon): Add brackets to IPv6 url

caseyhebebrand:
Add brackets to IPv6 urls


caseyhebebrand:
> I have no idea why brackets are needed because I don't understand IPv6, but I trust your judgment 😄

@erikmunson Thanks! It is because the colons in v6 conflict with the port and the protocol declaration, so the bracket help delineate the address.

PR #7981, opened 2020-02-28
Update eslint plugin to include react hooks rules

vigneshm:
Adding `react-hooks` rules.


vigneshm:
https://github.com/spinnaker/deck/pull/7982 is the right one.

PR #7984, opened 2020-02-29
feat(core/presentation): add usePollingData hook for scheduled fetching

erikmunson:
[`useData`](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/hooks/useData.hook.ts) is great. Sometimes in places where the most appropriate choice for data fetching would be `useData`, it's also desirable to poll for data on an interval to keep pulling in fresh state. This PR adds a new hook called `usePollingData` which does two things:

- Delegate data fetching behavior directly to `useData`, effectively mirroring its behavior
- Subscribe to a [`SchedulerFactory`](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/scheduler/SchedulerFactory.ts) with the chosen polling interval, and refresh the result from `useData` whenever the scheduler decides to fire

Why `SchedulerFactory`? Because it's a bit fancier than just setting up a `setInterval`/`setTimeout` on its own — it handles things like going online/offline and suspending polling when switching to other tabs/windows. It also makes the polling behavior relatively consistent with application data sources (though the polling interval isn't synced up with data sources).

Feedback welcome!


app/scripts/modules/core/src/presentation/hooks/usePollingData.hook.spec.tsx
```diff
@@ -0,0 +1,126 @@
+import React, { useEffect } from 'react';
+import { mount } from 'enzyme';
+import { IUseLatestPromiseResult } from './useLatestPromise.hook';
+import { usePollingData } from './usePollingData.hook';
+
+describe('usePollingData hook', () => {
```
erikmunson: Because this hook really just delegates down to `useData` and `SchedulerFactory` directly with just a teeny bit of `useEffect` logic thrown in, I didn't feel it was valuable to go deep on tests. These verify the basic assumptions that we'll return a proper result and refresh on an interval.

app/scripts/modules/core/src/presentation/hooks/usePollingData.hook.ts
```diff
@@ -0,0 +1,44 @@
...
+ * The promise factory is not called if any of the deps are null or undefined.
+ *
+ * This can be useful when fetching data based on a users keyboard input, for example.
+ * This behavior is similar to RxJS switchMap.
+ *
+ * @param callback the callback to be invoked whenever dependencies change and on a polling schedule
+ * @param defaultResult the default result returned before the first promise resolves
+ * @param pollingInterval the time interval (in milliseconds) for how often to invoke the callback
+ * @param deps array of dependencies, which (when changed) cause the callback to be invoked again
+ * @returns an object with the result and current status of the promise
```
erikmunson: I'm not sure I love making consumers use 4 variadic args like this, but it also seemed potentially nice to keep the API familiar since `useData` is common in newer components. Would it be better as an object?

PR #7985, opened 2020-02-29
feat(core/managed): poll for new history events in modal

erikmunson:
**Depends on ~~https://github.com/spinnaker/deck/pull/7984~~ and https://github.com/spinnaker/keel/pull/842**

When folks are looking at the resource history modal and trying to get a sense of what's been happening with their resource, it's valuable to update the view in real time. This change uses the [`usePollingData`](https://github.com/spinnaker/deck/pull/7984) hook to poll for new events every 10 seconds. The bulk of the changes are actually for optimizing rendering, not for wiring up the polling itself.

Rendering the event table itself isn't that much work, but when event rows are expanded the diff tables inside them can have complex component trees which in aggregate consume an unacceptable amount of rendering time. That problem is pretty simple to work around with memoization, it just means we have to do some mildly annoying array slicing and dicing when we receive new requests. In the future we're hoping to add `before` and `after` query parameters to the API this modal calls so that we can effectively have the API do this work for us — i.e. fetch the latest N events, then poll for anything after the newest timestamp in the initial request so we can prepend onto the existing data without even looking at it.

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
+++ b/app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import classNames from 'classnames';
 import { DateTime } from 'luxon';
 import { UISref } from '@uirouter/react';
@@ -11,11 +11,12 @@ import {
   TableRow,
   TableCell,
   standardGridTableLayout,
-  useData,
+  usePollingData,
+  usePrevious,
 } from '../presentation';
 
 import { relativeTime, timestamp } from 'core/utils';
-import { IManagedResourceSummary, IManagedResourceDiff } from 'core/domain';
+import { IManagedResourceSummary, IManagedResourceDiff, IManagedResourceEvent } from 'core/domain';
 import { AccountTag } from 'core/account';
 import { ManagedReader } from 'core/managed';
 import { Spinner } from 'core/widgets';
@@ -28,7 +29,7 @@ export interface IManagedResourceHistoryModalProps extends IModalComponentProps 
   resourceSummary: IManagedResourceSummary;
 }
 
-const { useMemo } = React;
+const EVENT_POLLING_INTERVAL = 10 * 1000;
 
 const viewConfigurationByEventType = {
   ResourceCreated: {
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
```diff
@@ -166,7 +205,7 @@ export const ManagedResourceHistoryModal = ({ resourceSummary }: IManagedResourc
               </button>
             </div>
           )}
-          {historyEventStatus === 'RESOLVED' && (
+          {!isLoading && historyEventStatus !== 'REJECTED' && (
```
erikmunson: all this loading related logic is primarily to ensure that when a polling request happens and the result from `usePollingData` switches to `PENDING`, we don't a loading indicator, but do if we're making a request after one failed.

alanmquach: Might be worth just making that another variable so the intent is a bit more clear?

app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
```diff
@@ -99,6 +100,41 @@ const viewConfigurationByEventType = {
   },
 } as const;
 
+const mergeNewEvents = (next: IManagedResourceEvent[], previous: IManagedResourceEvent[]) => {
```
alanmquach: This seems like an appropriate solve for this particular case because this is an event stream where the events are basically immutable.

We might need to start thinking about what we would do if that wasn't the case, i.e. a state of the world type of thing where the data we're looking at is more deeply nested and not immutable.

erikmunson: yes strongly agree (as we've chatted about offline). I'd like to find a more appropriate and generalized solution than what we currently have in existing routes (e.g. executions) since those areas had angular compatibility as constraint and rely on all kinds of `.forceUpdate()` tomfoolery.

app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
erikmunson: 👍 very yes. will do.

christopherthielen:
Should this be a new data source?

erikmunson:
@christopherthielen I thought about it, and decided not. This data / view is never going to be tied to the application model/lifecycle as far as I know, and I'm skeptical of the value in stuffing things into a data source when they're effectively not using any of the actual data source features / behavior. The only data-source-y bit of functionality this view requires is a polling cycle, which is why I just wrapped `SchedulerFactory` in a hook instead.

Separately, there's also going to be a bunch of API-driven filtering and pagination stuff happening in this view over the coming quarters and I'm not entirely sure how ergonomic doing that in data sources is — do we do anything like that right now in other views?

Do you see other aspects of leveraging the data source API that would be useful/beneficial here?

PR #7987, opened 2020-03-01
feat(core/presentation): add imperative showModal API (née ReactModal)

erikmunson:
After [adding a family of Modal components](https://github.com/spinnaker/deck/pull/7661), one of the next major steps toward shifting existing modals over to new ones is support for an imperative API that can be called from places like event handlers, etc. This job has been done for a while by [`ReactModal`](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/ReactModal.tsx), which in my opinion has served us quite well. While it's possible to swap out `ReactModal`'s internals to use new components, we make use of some `react-bootstrap` options in a variety of modals and I'd like to do a proper migration for those rather than forcing a big-bang switchover where we try to solve every modal's problems all at once. This led me to want to add a second API in `core/presentation` alongside `ReactModal` which we can roll out in phases until it replaces `ReactModal` entirely.

I had a few observations about the current `ReactModal` API:
- The `ReactModal` class/'service' is really just a function, and could probably just be unwrapped from its class
- The types for `ReactModal.show` are useful and nice, but IMO could be stronger and provide more coverage (e.g. all results for close/dismiss are always `any`)
- Because of the choice to use promise rejection for the 'dismiss' event, `ReactModal` has to add its own empty `.catch(() => {})` to make the behavior work as desired. As a result we're swallowing real exceptions in addition to misusing promise rejection a bit. Using rejection also makes things awkward with `async`/`await` because your `try`/`catch` is really behaving as an `if`/`else`
- The user-provided component receives no router context, which we've usually gotten around in the past by just slapping `@UIRouterContext` on class components. It'd be convenient to have context provided by default

The API I ended up with returns a promise which _always_ resolves, and supplies information about the result via an `IModalResult` object. The result has a discriminator (`status`) which can be checked in consuming code to get precise type coverage. When no type information is provided by the modal component, it falls back to the existing behavior of typing all results as `any`. This means we get (optional) increased type coverage and can go back to letting real exceptions flow through without swallowing. The user-supplied component is also wrapped in a `<UIRouterContextComponent/>` so children get router context for free.

Usage example below, very open to feedback (including "don't do it"). I suspect as we touch more complicated modals (especially ones that use multiple `react-bootstrap` options) we'll need to add some kind of extra configuration object in to do things with the `<Modal />` itself, but I'm ok adding things into the API incrementally as we discover what the needs of migrated modals are. If this gets merged one of the first test cases will be confirmation modals, then wizards after that.

<details>
<summary>Usage example</summary>

```tsx
// Leaving either of the two type parameters out of IModalComponentProps
// defaults them to `any`
interface IMyCoolModalProps extends IModalComponentProps<boolean, string> {
  color: 'red' | 'blue';
}

const MyCoolModal = ({ color, closeModal, dismissModal }: IMyCoolModalProps) => {
  const onClick = () => {
    // TS enforces argument types for both callbacks
    color === 'red' ? closeModal(true) : dismissModal('a string'); 
  };

  return <span onClick={onClick}>{color}</span>;
};


showModal(
  MyCoolModal,
  { color: Math.random() > .5 ? 'red' : 'blue' }
).then(({ status, closeResult, dismissResult }) => {
  if (status === 'CLOSED') {
    // TS sees `closeResult` as a boolean here
  } else {
    // TS sees `dismissResult` as an optional string here
  }
}).catch((error) => {
  // Now we're free to catch actual exceptions in the catch block
});
```
</details>


erikmunson:
Really trivial usage example: https://github.com/spinnaker/deck/pull/7988

app/scripts/modules/core/src/presentation/ReactModal.tsx
```diff
@@ -2,14 +2,7 @@ import React from 'react';
...
 
-/** The Modal content Component will be passed these two props */
-export interface IModalComponentProps {
-  // Close modal with a result value (i.e., OK button)
-  closeModal?(result?: any): void;
-
-  // Dismiss/reject modal (i.e., Cancel button)
-  dismissModal?(rejectReason?: any): void;
-}
+import { IModalComponentProps } from './modal';
```
erikmunson: I think that besides optionally parameterizing the result types (which is done in this PR) the shape of `IModalComponentProps` is remains very useful and has no reason to change. Since it's already got the right shape and name, there's no reason to duplicate it, instead let's just move it to a new home and continue using it.

app/scripts/modules/core/src/presentation/modal/showModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+    render();
+
+    setTimeout(() => {
+      if (!mountNode) {
+        return;
+      }
+
+      show = true;
+      render();
+    });
```
erikmunson: I'm somewhat confident we didn't need any of this enter transition stuff with `react-bootstrap` because they were doing it themselves inside their component lifecycle hooks.

app/scripts/modules/core/src/presentation/modal/showModal.tsx
```diff
@@ -0,0 +1,107 @@
...
+      show = false;
+      render();
+    };
+
+    const handleClose = handleResultWith((result: C) => ({ status: 'CLOSED', closeResult: result }));
+    const handleDismiss = handleResultWith((result: D) => ({ status: 'DISMISSED', dismissResult: result }));
+    const handleRequestClose = () => handleDismiss(null);
+
+    function render() {
+      ReactDOM.render(
```
vigneshm: Have you considered using [react portal](https://reactjs.org/docs/portals.html#usage) here instead?

erikmunson: Good Q, the `Modal` component itself [actually renders in a portal](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/modal/Modal.tsx#L33-L45) — the reason this imperative API requires a real mount point is because it's intentionally _not_ rendering the `Modal` component into an existing component tree. Portals are great, but they inherently have to be mounted into an existing component tree.

vigneshm: Ah, right. You had mentioned earlier that you didn't explicitly want these kinds of modals to be in the same react tree.

erikmunson:
> ❤️ that is some bad-ass typescript right there

Slowly but surely I'm picking up some of your tricks! 👼

PR #7988, opened 2020-03-01
feat(core/managed): use showModal API for history modal

erikmunson:
Depends on https://github.com/spinnaker/deck/pull/7987

This modal really isn't tied to its parent component tree at all, and should be rendered imperatively. Implementation may change depending on the outcome of https://github.com/spinnaker/deck/pull/7987


app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
```diff
@@ -4,14 +4,15 @@ import { DateTime } from 'luxon';
...
+  IModalComponentProps,
   ModalHeader,
   ModalBody,
   Table,
   TableRow,
   TableCell,
   standardGridTableLayout,
   useData,
-} from 'core/presentation';
+} from '../presentation';
```
erikmunson: this is so that extending `IModalComponentProps` doesn't explode when TS can't transform the `core` alias in custom builds.

PR #7989, opened 2020-03-02
fix(artifacts): fix editing expected artifacts in artifact rewrite tr…

maggieneterval:
…igger flow

Closes https://github.com/spinnaker/spinnaker/issues/5513

Updates to expected artifacts configured against triggers were not saving correctly, because we were updating the pipeline config with whatever value we had saved on the `TriggerPageContent`'s state rather than the updated value being passed to the callback as a result of editing. I'm not convinced that we need to duplicate the expected artifacts data to state to ["to avoid race condition"](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/triggers/TriggersPageContent.tsx#L32) (as the comment on the change that introduced this bug states), but I'll save that investigation for a more holistic audit of the artifacts rewrite logic.


maggieneterval:
@spinnakerbot :cherries: :pick: 1.18

PR #7997, opened 2020-03-03
feat(core/managed): add history modal button to status popovers, revise language

erikmunson:
**Depends on #7987 and #7988**

Now that we have a way of viewing historical resource events in the UI, it makes sense to offer a quick way to jump into that view when learning about a resource's current status in the infrastructure view. This PR adds a "History" button to all status popovers that looks like this:

Screen Shot 2020-03-03 at 8 25 43 AM

In addition, I've tweaked the language on a few of the statuses to offer guidance around when it makes sense to check the History (e.g. `DIFF`, `ACTUATING`, `ERROR`, etc.).

As a bonus, I fixed the deep-linking to the docs which broke when they moved from Reference to Guides. I'll be opening up a separate PR to modify the spinnaker.io doc page with updated screenshots for each status as well.

(cc @gcomstock)

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
+++ b/app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
@@ -28,11 +28,8 @@ const viewConfigurationByStatus: { [status in ManagedResourceStatus]: IViewConfi
           <b>Action is being taken to resolve a drift from the declarative configuration.</b>
         </p>
         <p>
-          Check the{' '}
-          <UISref to="home.applications.application.tasks">
-            <a>tasks view</a>
-          </UISref>{' '}
-          to see work that's in progress. <LearnMoreLink resourceSummary={resourceSummary} />
+          Check this resource's History to see details and track the work currently in progress.{' '}
+          <LearnMoreLink resourceSummary={resourceSummary} />
         </p>
       </>
     ),
@@ -62,8 +59,8 @@ const viewConfigurationByStatus: { [status in ManagedResourceStatus]: IViewConfi
           <b>A drift from the declarative configuration was detected.</b>
         </p>
         <p>
-          Spinnaker will automatically take action to bring this resource back to its desired state.{' '}
-          <LearnMoreLink resourceSummary={resourceSummary} />
+          Spinnaker will automatically take action to bring this resource back to its desired state. Check the History
+          to see details and track progress. <LearnMoreLink resourceSummary={resourceSummary} />
         </p>
       </>
     ),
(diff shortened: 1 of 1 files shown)


gcomstock:
Awesome idea! Thanks Erik.

app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
```diff
@@ -27,11 +28,8 @@ const viewConfigurationByStatus: { [status in ManagedResourceStatus]: IViewConfi
           <b>Action is being taken to resolve a drift from the declarative configuration.</b>
         </p>
         <p>
-          Check the{' '}
-          <UISref to="home.applications.application.tasks">
-            <a>tasks view</a>
-          </UISref>{' '}
-          to see work that's in progress. <LearnMoreLink resourceSummary={resourceSummary} />
+          Check this resource's History to see details and track the work currently in progress.{' '}
```
luispollo: Newbie question: is the `{' '}` the "right" way of forcing a space character?

app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
```diff
@@ -27,11 +28,8 @@ const viewConfigurationByStatus: { [status in ManagedResourceStatus]: IViewConfi
...
+          <LearnMoreLink resourceSummary={resourceSummary} />
```
luispollo: Love a good self-explanatory name for a component. ❤️

app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
```diff
@@ -180,45 +179,61 @@ const LearnMoreLink = ({ resourceSummary }: { resourceSummary: IManagedResourceS
...
+          <button
+            className="passive flex-none"
+            onClick={() => toggleResourcePause(resourceSummary, application, hidePopover)}
+          >
+            <i className="fa fa-play" /> Resume management of this resource
+          </button>
+        )}
+        {application.isManagementPaused && (
+          <UISref to="home.applications.application.config" params={{ section: 'managed-resources' }}>
+            <a>Resume application management</a>
```
luispollo: Since this is displayed in the context of a specific resource, shouldn't there be a warning here that this will affect all resources under the application?

app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
erikmunson: yep! it's a sort of weird pattern, but this is how you make sure you end up with a space between the end of a text block and the next JSX element _if the point where they meet is broken across a line break_. If it's all on one line a normal space is enough (e.g. `this is some text <MyComponent />`. If you write a single-line thing like my example and then run it through Prettier, it will automatically add that syntax if it decides to break your JSX component onto the next line (that's what happened here).

app/scripts/modules/core/src/managed/ManagedResourceStatusIndicator.tsx
erikmunson: this is actually a link to the app config page, where you get deep-linked into the "Managed Resources" section with set of copy/explanation about resuming the application — making that relationship/impact super clear is actually why Chris Berry originally added this as a link, instead of having it do the pause in a modal like with resource pausing.

erikmunson:
Related PR for updated documentation + screenshots: https://github.com/spinnaker/spinnaker.github.io/pull/1732

erikmunson:
@Mergify rebase

PR #8010, opened 2020-03-05
fix(core): Order executions by startTime not endTime

alanmquach:
While there could be an argument to be made that it's nice to see more recently running things, overall sorting by anything other than `startTime` is mostly unexpected and confusing to the user.

Because the executions endpoint returns a limited number of things ordered by `startTime` not `endTime`, ordering by`endTime` is only more confusing because requesting newer items could get sandwiched between existing items, which is just plain weird.


erikmunson:
okwiththis

PR #8011, opened 2020-03-06
Preconfigured job: use react keys and extract preconfiguredjob reader

christopherthielen:
feat(core/pipeline): Create a preconfiguredJob.reader
fix(core/pipeline): use react keys in preconfigured job config component loop


app/scripts/modules/core/src/pipeline/config/stages/preconfiguredJob/preconfiguredJob.reader.ts
```diff
@@ -0,0 +1,30 @@
...
+  waitForCompletion?: boolean;
+  parameters?: IPreconfiguredJobParameter[];
+  producesArtifacts: boolean;
+}
+
+export const PreconfiguredJobReader = {
+  list(): IPromise<IPreconfiguredJob[]> {
+    return API.one('jobs')
+      .all('preconfigured')
+      .useCache()
```
erikmunson: nice.

PR #8013, opened 2020-03-09
fix(core): Fixing AccountSelectInput sync issue

alanmquach:
I believe what happened here is that the event got removed from the pool but `event.target` is pointing to the DOM node whose value still got reset by the rerender, therefore reading `event.target.value` asynchronously still gave us the reset value.

I think we're stuck with just capturing the new value instead?


danielpeach:
@Mergifyio backport release-1.19.x

PR #8016, opened 2020-03-09
feat(amazon/sharding): Enable server group sharding in deck

vigneshm:
### What is the sharding feature?

Spinnaker currently provides a way to create clusters via `stack` and `details` fields. The sharding feature allows us to potentially create more dimensions for clusters.

### How is it implemented?

This feature re-uses the `freeformDetails` field and introduces a new format to specify the sharding identifiers. The format is `x1shard1-x2shard2-other`. In this example `shard1` and `shard2` are the new shard identifiers and `other` will continue to be used as how `details` was used earlier. At the moment we are limiting the additional number of shards to 2. However this could be easily changed in future to include additional ones. This format is completely hidden away from the user as deck will show two new shard fields in pipeline configuration flow and server group clone flow.

The PR is split into 2 commits and it would be easier to review them separately. The first commit includes a new utils module that helps for parsing and forming the `freeformDetails` string. The second commit includes new form fields that accept shard identifiers.

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/configure/ServerGroupShardingUtils.ts
+++ b/app/scripts/modules/amazon/src/serverGroup/configure/ServerGroupShardingUtils.ts
@@ -1,0 +1,110 @@
+import { chain, isEmpty, partition, sortBy, times, toPairs } from 'lodash';
+
+export interface ParsedShards {
+  shards: { [shardIndex: string]: string };
+  other: string;
+}
+
+function createEmpty(dimensions: number, other = ''): ParsedShards {
+  const result: ParsedShards = { shards: {}, other };
+  times(dimensions).forEach(index => (result.shards[`x${index + 1}`] = ''));
+
+  return result;
+}
+
+/**
+ * Given a free form details string, this function parses, validates and returns the shards along with remaining
+ * detail string. This function also accepts a `dimension` param that restricts the number of accepted shard
+ * identifiers.
+ *
+ * freeFormDetails = 'x1Shard1-x2Shard2-x3Shard3-Other', dimensions = 3
+ * Returns: {shards: {x1: 'Shard1', x2: 'Shard2', x3: 'Shard3'}, other: 'Other'}
+ *
+ * freeFormDetails = 'x2Shard2-x1Shard1-x3Shard3-Other', dimensions = 2
+ * Returns: {shards: {x1: 'Shard1', x2: 'Shard2'}, other: 'x3Shard3-Other'}
+ */
+export function parse(freeFormDetails: string, dimensions = 2): ParsedShards {
+  // Valid shard strings must begin with x\d or -x\d.
+  if (!freeFormDetails.match(/^-?x\d/)) {
+    return createEmpty(dimensions, freeFormDetails);
+  }
+
+  // Prepend with a '-' and split it so that potentialShards look like the following
+  // [ "" or "-", "1", "Shard1", "2", "Shard2", "3", "Shard3-Other"].
+  const potentialShards = `-${freeFormDetails}`.split(/-x(\d)/);
+
+  // Remove the empty first entry.
+  potentialShards.shift();
+
+  let shards: Array<[number, string]> = [];
+  let other = '';
+  let i = 0;
+
+  // By the end of this loop, we're trying to group the pairs of shard index and shard identifier so that `shards`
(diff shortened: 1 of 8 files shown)


vigneshm:
Unable to add reviewers here for some reason, so explicitly tagging @alanmquach @christopherthielen @erikmunson @caseyhebebrand .

cc @cfieber

cfieber:
FWIW the server side implementation on this is an internal only extension at the moment

app/scripts/modules/amazon/src/serverGroup/configure/ServerGroupShardingUtils.ts
```diff
@@ -0,0 +1,110 @@
...
+ * detail string. This function also accepts a `dimension` param that restricts the number of accepted shard
+ * identifiers.
+ *
+ * freeFormDetails = 'x1Shard1-x2Shard2-x3Shard3-Other', dimensions = 3
+ * Returns: {shards: {x1: 'Shard1', x2: 'Shard2', x3: 'Shard3'}, other: 'Other'}
+ *
+ * freeFormDetails = 'x2Shard2-x1Shard1-x3Shard3-Other', dimensions = 2
+ * Returns: {shards: {x1: 'Shard1', x2: 'Shard2'}, other: 'x3Shard3-Other'}
+ */
+export function parse(freeFormDetails: string, dimensions = 2): ParsedShards {
```
caseyhebebrand: A bit of a nit-pick, but naming this `parseShard` or `getParsedShard` would be helpful since this is specific to shard rules and parsing could also happen to JSON objects, other strings, etc.

app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
```diff
@@ -0,0 +1,229 @@
...
+ * When the user is editing the value of this field, it is possible to introduce incorrect values such as hyphens,
+ * spaces, dots etc. These incorrect values will get combined to form the consolidated value for `freeformDetails`
+ * field. This consolidated value will be parsed again when `ServerGroupDetailsField` is re-rendered, but the parsing
+ * will fail this time since the consolidated value does not conform to the expected format. This will result in a
+ * state where the error cannot be shown to the user and won't be provided a way to fix it.
+ *
+ * Keeping a local error state allows us to bail out from informing the parent about the change and provide the user
+ * a way to fix the error.
+ */
+function useLocalError(
```
caseyhebebrand: This hook would be valuable as a reusable hook. If you move it to the `hooks` folder in `core/presentation` other fields with this same subtlety can utilize this hook.

app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
```diff
@@ -0,0 +1,229 @@
...
+    </>
+  );
+}
+
+export interface OtherDetailsFieldProps {
+  value: string;
+  onChange: (value: string) => void;
+  onError: (error: string) => void;
+}
+function OtherDetailsField({ value, onChange, onError }: OtherDetailsFieldProps) {
```
caseyhebebrand: The `ShardField` and the `OtherDetailsField` have the same behavior and similar return values. We could make this more reusable by extracting both components to a generic `InputWithLocalError` that accepts the validator function, label, and help text as props (in addition to the existing props).

app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
vigneshm: Thanks, I wasn't sure of the usefulness of this hook outside of this use case and was hesitant to move it out of the file. But if you feel this is useful, I will definitely move it out after making a few changes to make it a bit generic.

app/scripts/modules/amazon/src/serverGroup/configure/ServerGroupShardingUtils.ts
christopherthielen: agreed, especially since we tend to re-export all symbols up to the top level.  I'm starting to reconsider this practice because it makes for a very crowded namespace, and potential for collisions.

vigneshm:
Closing this one. We'll be experimenting with this feature for a while internally to find rough edges and push it to OSS later.

PR #8020, opened 2020-03-10
feat(md): Types and new data source for environments

alanmquach:
Depends on https://github.com/spinnaker/keel/pull/848

The types are probably the most interesting thing in this PR. There's a fair amount of duplication in different views of an entity, for example:

 * `environments` vs `artifact.version.environments`
    * `environments`: a light summary of all resources and all artifacts related to those resources
    * `artifact.version.environments`: a view of each environment as it relates to the artifact's specific version (i.e. time that version was deployed in that specific environment)
* `artifacts` vs `environment.artifacts` vs `resource.artifact`
    * `artifacts`: summary of all artifacts and all versions also including a join between a specific version and it's environment
    * `environment.artifacts`: a view of each artifact that is in a given environment (i.e. approved, vetoed)
    * `resource.artifact`: (not yet present) a view of a specific artifact as it relates to the resource (i.e. deploying, pending)

The shapes or mostly different, so I'm not sure if it's even useful to have common attributes between the types. I also don't know if there's a sane way to name them.

`optional` and `optIn` mean this datasource is only enabled if explicitly enabled.

`hidden` means this (for now) is not yet exposed in the application config UI.


app/scripts/modules/core/src/domain/IManagedEntity.ts
```diff
@@ -24,6 +24,49 @@ export interface IManagedResourceSummary {
...
+  name: string;
+  state: string;
+  deployedAt?: string;
+  replacedAt?: string;
+  replacedBy?: string;
+}
+
+export interface IManagedArtifactVersion {
+  version: string;
+  environments: IManagedArtifactVersionEnvironmentSummary[];
```
erikmunson: crazy idea: what if... we just didn't give this thing a name and inlined it? Seems somewhat futile to try to name a thing that's really just a particular join on a couple other discrete things.

Can always extract it like `IManagedArtifactVersion['environments']` if / when you need it on it's own.

app/scripts/modules/core/src/domain/IManagedEntity.ts
```diff
@@ -24,6 +24,49 @@ export interface IManagedResourceSummary {
...
+    current: string;
+    pending: string[];
+    approved: string[];
+    previous: string[];
+    vetoed: string[];
+  };
+}
+
+export interface IManagedEnviromentSummary {
+  artifacts: IManagedEnvironmentArtifact[];
```
erikmunson: Similarly to `IManagedArtifactVersionEnvironmentSummary`, what if we just inlined this?

app/scripts/modules/core/src/managed/managed.dataSource.ts
```diff
@@ -34,6 +34,14 @@ module(MANAGED_RESOURCES_DATA_SOURCE, []).run([
       application.securityGroups.ready().then(() => addManagedResourceMetadataToSecurityGroups(application), noop);
     };
 
+    const loadEnvironments = (application: Application) => {
+      return ManagedReader.getEnvironmentsSummary(application.name);
+    };
+
+    const addEnvironments = (_application: Application, data: IManagedApplicationSummary) => {
```
erikmunson: Should this be `IManagedApplicationEnvironmentsSummary`? (or `IManagedApplicationSummary<'resources' | 'artifacts' | 'environments'>` if you do that thing I commented about)

app/scripts/modules/core/src/managed/managed.dataSource.ts
```diff
@@ -42,5 +50,24 @@ module(MANAGED_RESOURCES_DATA_SOURCE, []).run([
...
+      hidden: true,
+      icon: 'fa fa-fw fa-xs fa-code-branch',
+      loader: loadEnvironments,
+      onLoad: addEnvironments,
+      defaultData: {
+        applicationPaused: false,
+        hasManagedResources: false,
+        environments: [],
+        artifacts: [],
+        resources: [],
```
erikmunson: is duplicating the `resources` data into this data source a temporary/convenience thing, or do you intend on it being permanent?

app/scripts/modules/core/src/domain/IManagedEntity.ts
```diff
@@ -24,6 +24,49 @@ export interface IManagedResourceSummary {
...
+  environments: IManagedArtifactVersionEnvironmentSummary[];
+}
+
+export interface IManagedArtifactSummary {
+  name: string;
+  type: string;
+  versions: IManagedArtifactVersion[];
+}
+
+export interface IManagedApplicationEnvironmentsSummary extends IManagedApplicationSummary {
```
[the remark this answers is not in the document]
alanmquach: giphy-1

app/scripts/modules/core/src/domain/IManagedEntity.ts
alanmquach: Yeah, I was thinking that too. Will go with that, at least for now.

app/scripts/modules/core/src/managed/managed.dataSource.ts
alanmquach: Well originally temporary/convenience, but thinking about it some more I think it might be more permanent.

I don't like the duplication either but because the `managedResource` data source is `visible: false`, it doesn't contribute to the application ready state, so I don't have a guarantee that it will be available when my route loads.

alanmquach:
Whoops, that merge was unintended... I was trying to update my branch with the latest master.

alanmquach:
Reverted and reopeened as #8022

PR #8025, opened 2020-03-10
Preconfigured job stage

christopherthielen:
This allows registration of a preconfigured job stage UI. The parameter configuration for the stage is fetched from gate.  

This PR also enables plugins to export preconfiguredJob stages

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/PipelineRegistry.spec.ts
+++ b/app/scripts/modules/core/src/pipeline/config/PipelineRegistry.spec.ts
@@ -7,7 +7,7 @@ import { IRegion } from 'core/account/AccountService';
 import { Registry } from 'core/registry';
 import { ITriggerTemplateComponentProps } from '../manualExecution/TriggerTemplate';
 import { PipelineRegistry } from './PipelineRegistry';
-import { IPreconfiguredJob, PreconfiguredJobReader } from './stages/preconfiguredJob';
+import { IPreconfiguredJob, makePreconfiguredJobStage, PreconfiguredJobReader } from './stages/preconfiguredJob';
 
 const mockProviderAccount = {
   accountId: 'abc',
@@ -186,8 +186,7 @@ describe('PipelineRegistry: API', function() {
   describe('preconfigured stage', function() {
     beforeEach(mock.inject());
 
-    const makeStage = () => ({ key: 'job', alias: 'preconfiguredJob' } as IStageTypeConfig);
-
+    // Gate response
     const makeJobMetadata = () => {
       return {
         type: 'job',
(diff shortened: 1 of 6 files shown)


app/scripts/modules/core/src/pipeline/config/stages/preconfiguredJob/preconfiguredJobStage.ts
```diff
@@ -1,39 +1,43 @@
...
+  const { label, description, type, waitForCompletion, parameters, producesArtifacts } = job;
+
+  return {
+    label,
+    description,
+    key: type,
+    alias: 'preconfiguredJob',
+    addAliasToConfig: true,
+    restartable: true,
+    // Overriden when the
```
erikmunson: some missing text here?

app/scripts/modules/core/src/pipeline/config/PipelineRegistry.spec.ts
```diff
@@ -182,6 +183,51 @@ describe('PipelineRegistry: API', function() {
     );
   });
 
+  describe('preconfigured stage', function() {
+    beforeEach(mock.inject());
+
+    const makeStage = () => ({ key: 'job', alias: 'preconfiguredJob' } as IStageTypeConfig);
```
erikmunson: am I reading correctly that folks writing stage plugins will still have to add `alias: 'preconfiguredJob'` to their plugin stage config? if so, any idea how ok/terrible it would be to handle that for them in `registerPreconfiguredJobStage()`?

christopherthielen: There's a helper `buildPreconfiguredJobStage` that builds a preconfigured job stage from the PreconfigureJob metadata (the shape that Gate returns).  So in actual code (not tests) it would look more like this:

```
export const treasureDeployStage: IStageTypeConfig = {
  ...buildPreconfiguredJobStage({
    uiType: 'CUSTOM',
    type: 'treasureDeploy',
    label: `Treasure Deploy with UI (EXPERIMENTAL)`,
    noUserConfigurableFields: false,
    description: 'Publishes static assets to treasure,
    producesArtifacts: false,
  }),
  component: TreasureStageConfig, // stage config
  // executionDetailsSections: [PreconfiguredJobExecutionDetails, ExecutionDetailsTasks],
  executionDetailsSections: [ExecutionDetailsTasks],
  validateFn: validate,
};
```
As I'm looking at this code now, I see that a lot of it is duplicated from the backend orca registration.  I think I'll refactor this a bit more in hopes of simplifying it further for devs.

christopherthielen:
@erikmunson I updated the registration logic to apply more preconfigured job metadata from gate.  Now, callers will do something like this:

```js
export const treasureDeployStage: IStageTypeConfig = {
  ...makePreconfiguredJobStage('treasureDeploy'),
  component: TreasureStageConfig,
  executionDetailsSections: [ExecutionDetailsTasks],
  validateFn: validate,
};
```

PR #8026, opened 2020-03-10
Update eslint plugin

christopherthielen:
chore(package): Update to eslint-plugin 1.0.5 (adds react hooks linter rules)

    fix(reacthooks): Fix React Hooks linter errors

    Left the warnings to be dealt with/ignored later


christopherthielen:
There are many many react-hooks linter warnings that are now surfaced.  They are all due to not explicitly declaring dependencies that are used inside an effect.  

I'm not convinced the deps rule should be followed blindly.  For example, I updated the MapEditorInput according to its naive recommendations and got into an infinite effect loop due to unstable callbacks (solvable).

app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/ManualExecutionBakeManifest.tsx
```diff
@@ -26,6 +26,14 @@ export function ManualExecutionBakeManifest(props: ITriggerTemplateComponentProp
     props.updateCommand('extraFields.artifacts', updatedArtifacts);
   };
 
+  React.useEffect(() => {
+    if (overrideArtifact === false) {
+      removeHelmArtifact();
+    } else {
+      updateHelmArtifact(defaultArtifact);
+    }
+  }, [overrideArtifact]);
```
caseyhebebrand:
```
React.useEffect(() => {
  overrideArtifact === false ? removeHelmArtifact() : updateHelmArtifact(defaultArtifact); 
}, [overrrideArtifact]);
```

christopherthielen: I disagree with this refactor.  

_Opinion warning_: I prefer that ternaries are used only as expressions, i.e. something that evaluates to a value and is returned or captured in a variable.  In this case, these are pure side effects, and I actually prefer an if/else block.

christopherthielen:
There are almost 55 or so unfixed warnings from the react-hooks rule.  When we run `yarn lib` to build the packages, the warnings are being printed to the console.  Github Actions is somehow surfacing the warnings into the PR review (which is SUPER COOL, but not something I want to look at in every PR).

Until we either a) fix all the linter warnings or b) disable output from `yarn lib`, I think it's best to close this PR.  I'll revert vignesh's commit and release v1.0.6 of the eslint-plugin, then open a new PR upgrading to 1.0.6 without react-hooks linter rules.

app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/ManualExecutionBakeManifest.tsx
caseyhebebrand: That is understandable. I did not look into what these functions did so I wasn't sure what they did.

erikmunson:
HOLY MOLY that github linting thing is SORCERY.

> I'm not convinced the deps rule should be followed blindly. 

This is so true — I regularly run into situations where I'm referencing several things in a hook, and my desired behavior is that I have access to the _latest value_ of one or two deps whenever the other dep _changes_. In observable land it'd be like a `.buffer()` kind of thing. By definition to get that behavior you have to avoid listing some things out as deps. As we do more and more interesting stuff in hooks i find that use case popping up more frequently.

christopherthielen:
I've updated this PR with a 1.0.6 release with the deps rule turned off.  If all goes well I will merge it.

PR #8028, opened 2020-03-11
fix(plugins): better error messages when loading plugin information

bpowell:


bpowell:
@christopherthielen

christopherthielen:
Shouldn't the re-throw cause this to get handled by the global error handler and print the entire error to the console?

bpowell:
Currently it does not, @link108 and I were testing out something and not getting the real error messages

christopherthielen:
I'll take a look at the current behavior and try to find out why it's not getting logged.

christopherthielen:
before your PR:

Screen Shot 2020-03-11 at 10 32 16 AM

after your PR:

Screen Shot 2020-03-11 at 10 32 37 AM


It's possible you were exercising a different catch block.  Which catch block exactly was the problem?  I exercised the manifest validation one.

bpowell:
It was an issue when trying to parse the `plugin-manifest.json` from Gate

christopherthielen:
> It was an issue when trying to parse the plugin-manifest.json from Gate

Is this because of the 404 handler you put in?  If so, then the change should only be applied to that handler.   It should also log the error object itself, (not coercing it to a string).

app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -97,6 +97,7 @@ export class PluginRegistry {
       .then(response => response.data)
       .catch((error: any) => {
         console.error(`Failed to load ${uri} from ${source}`);
+        console.error(`Error message: ${error}`);
```
christopherthielen:
```suggestion
```

app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -111,6 +112,7 @@ export class PluginRegistry {
       .get()
       .catch((error: any) => {
         console.error(`Failed to load ${uri} from ${source}`);
+        console.error(`Error message: ${error}`);
```
christopherthielen:
```suggestion
```

app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -137,6 +139,7 @@ export class PluginRegistry {
       return plugins.map(pluginMetaData => this.registerPluginMetaData(source, pluginMetaData));
     } catch (error) {
       console.error(`Error loading plugin manifest from ${location}`);
+      console.error(`Error message: ${error}`);
```
christopherthielen:
```suggestion
```

app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -175,6 +178,7 @@ export class PluginRegistry {
       return module;
     } catch (error) {
       console.error(`Failed to load plugin code from ${pluginUrl}`);
+      console.error(`Error message: ${error}`);
```
christopherthielen:
```suggestion
```

app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -111,6 +112,7 @@ export class PluginRegistry {
       .get()
       .catch((error: any) => {
         console.error(`Failed to load ${uri} from ${source}`);
+        console.error(`Error message: ${error}`);
         // If we cannot hit the Gate URL, ignore it
         if (error.data.status === 404) {
           return Promise.resolve([]);
```
christopherthielen:
```suggestion
          console.error(error);
          return Promise.resolve([]);
```

PR #8029, opened 2020-03-11
feat(amazon/serverGroupDetails): Make server group's details field overridable

vigneshm:
A simple change that extracts out the details field in to a separate component so that we can make it overridable. This will be used to have netflix specific customizations to enable sharding.


vigneshm:
> Did you test this? LGTM.

Yes, I customized this field in -nflx and the customization did work.

> If you have time, it would be nice to convert this component to FormikFormField(s).
Sure, will give it a shot

PR #8032, opened 2020-03-11
feat(md): Initial draft of Environments view

alanmquach:
~~DRAFT PR DO NOT MERGE~~ 
~~Still contains a bunch of TS and lint errors that need to be corrected~~

Introduces an artifacts and environments view for managed delivery.

Screen Shot 2020-03-12 at 9 02 23 AM

For now this new data source will be optIn via API only (by adding it as an explicitly enabled data source).

### Extra Context
* Components that could be reusable were intentionally kept in `/managed` so we can refine them before graduating them into `/presentation`, little to no thought has been given to their Component APIs for this first draft, we intend to still iterate on them heavily
* Some things that appear unused/useless because they will be used very soon and it didn't make sense to remove it only to add it back shortly


app/scripts/modules/core/src/presentation/index.ts
```diff
@@ -7,6 +7,7 @@ export * from './hooks';
 export * from './HoverablePopover';
 export * from './LabelComponent';
 export * from './labeledValues';
+export * from './layout';
```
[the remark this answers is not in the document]
alanmquach: sounds good to me

app/scripts/modules/core/src/presentation/layout/ObjectRow.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import styles from './ObjectRow.module.css';
+
+interface IObjectRowProps {
+  icon: string;
+  title: string;
+}
+
+export const ObjectRow = ({ icon, title }: IObjectRowProps) => {
+  const depth = 0;
```
caseyhebebrand: Does it make sense to have the depth passed in as an optional prop that defaults to 0? This way the parent component can control the placement of the child component. It also makes the component API more versatile for other use cases of a row that just has an icon and title.

app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -0,0 +1,64 @@
...
+  artifactSelected: (artifact: ISelectedArtifact) => void;
+  selectedArtifact: ISelectedArtifact;
+}
+
+export function ArtifactsList({ artifacts, artifactSelected }: IArtifactsListProps) {
+  return (
+    <div>
+      {artifacts.map(({ versions, name }) =>
+        versions.map(({ version }) => (
+          <>
```
christopherthielen: unnecessary fragment tag?

app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -1,5 +1,59 @@
...
+  version: string;
+}
+
+interface IEnvironmentsProps {
+  app: Application;
+}
+
+export function Environments(props: IEnvironmentsProps) {
+  const { app } = props;
+  const dataSource: ApplicationDataSource<IManagedApplicationSummary<
```
christopherthielen: This caused my brain's syntax scanner to reboot:
```ts
    'resources' | 'artifacts' | 'environments'
  >> = app.getDataSource('environments');
```

This might read a lot nicer if you extract a local type
```ts
type ISummary = IManagedApplicationSummary<'resources' | 'artifacts' | 'environments'>;
```

or even just the union:
```ts
type IArtifactTypes = 'resources' | 'artifacts' | 'environments'
```

erikmunson:
For those reading along at home, a couple notes on this PR:

1. Since all the entry points to this code are behind the `managedDelivery` feature flag, _and_ the routes/data sources are off by default and not visible to users for turning on/off, we're effectively considering this an experimental testing ground that doesn't yet need to be fully production-ready
2. Because of #1, we're adding some new CSS module code even though that technology is being removed from Deck in the near future. expect that to get refactored into `.css` or `.less` files in a subsequent change before this becomes available to turn on outside of experimental purposes

app/scripts/modules/core/src/presentation/layout/ObjectRow.tsx
erikmunson: I don't disagree, but for now we're not worrying too much about component APIs https://github.com/spinnaker/deck/pull/8032#issuecomment-598345196

Expect these things to be up in a PR for proper review when we [decide to graduate them](https://github.com/spinnaker/deck/pull/8032#discussion_r391713997) back into `core/presentation`

app/scripts/modules/core/src/presentation/index.ts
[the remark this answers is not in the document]
alanmquach: done.

app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -1,5 +1,59 @@
...
+    'resources' | 'artifacts' | 'environments'
+  >> = app.getDataSource('environments');
+  const [selectedArtifact, setSelectedArtifact] = useState<ISelectedArtifact>();
```
christopherthielen: I'd prefer to use the data source's observable than manage refreshes manually.


I have a `useObservable` hook in a pending branch I'm working on.  What do you think of this:

`useObservable.hook.ts`:
```ts
/** Subscribes the callback to the observable on mount. Unsubscribes on unmount. */
function useObservable<T>(observable: Observable<T>, callback: (val: T) => void): void {
  useEffect(() => {
    const subscription = observable.subscribe(callback);
    return () => subscription.unsubscribe();
  }, [observable]);
}

function useObservableValue<T>(observable: Observable<T>): void {
  const [value, setValue] = useState<T>(undefined);
  useObservable(observable, setValue);
  return value;
}
```

Then in this file:
```ts
const environments = useObservableValue(dataSource.data$);
```

app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -1,5 +1,59 @@
...
+  const [environments, setEnvironments] = useState(dataSource.data);
+  const [isFiltersOpen] = useState(false);
```
christopherthielen: WIP?

app/scripts/modules/core/src/presentation/layout/ObjectRow.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <>
```
christopherthielen: useless fragment?

app/scripts/modules/core/src/presentation/layout/ObjectRow.tsx
christopherthielen: WIP?

app/scripts/modules/core/src/managed/Environments.tsx
erikmunson: this whole thing is WIP 😄

alanmquach: Yes, we'll eventually add this so I figured i'd just leave it in.

app/scripts/modules/core/src/presentation/layout/ObjectRow.tsx
alanmquach: > Yes, we'll eventually add this so I figured i'd just leave it in.

app/scripts/modules/core/src/managed/Environments.tsx
alanmquach: I like it. I was thinking the same thing and had a brief @erikmunson about creating a `useDataSource` hook, but we didn't think it would end up being used in too many places.

The thinking was that we wouldn't want a ton of components grabbing data directly from the datasources and instead would prefer for them to be managed by a root component and fed through the component tree.

Thoughts?

erikmunson: yeah this is really similar to what we were brainstorming on, and if we ever wanted one that was specific to data sources we could just build it on top of this hook. Let's wait until the hook makes it into a PR + gets merged and then switch over to it?

app/scripts/modules/core/src/managed/Environments.tsx
alanmquach: mine too.

I added the following to the types:
```
export type IManagedApplicationEnvironmentSummary = IManagedApplicationSummary<
  'resources' | 'artifacts' | 'environments'
>;
```

app/scripts/modules/core/src/managed/Environments.tsx
christopherthielen: I think a really thin `useDataSource` could possibly be good if it provides more functionality than the generic observable one, such as the ability to request a refresh.  Otherwise, if it's just streaming values, I think `useObservableValue` is clear enough.  I'd love it if all usage of datasource transitioned towards using the `data$` and away from the `onRefresh` api.

> managed by a root component and fed through the component tree

agreed

christopherthielen: > Let's wait until the hook makes it into a PR + gets merged

I'd just write the hooks now.  That PR I referenced is _way back burner_ and may never even land.

app/scripts/modules/core/src/managed/ArtifactsList.tsx
alanmquach: Yes, removed. thanks.

PR #8038, opened 2020-03-12
fix(core): Do not initialize strategies on mount

alanmquach:
#7967 fixed the missing `maxRemainingAsgs` issue, but didn't realize it was always called on mount, causing `maxRemainingAsgs` to be filled in even when it was originally missing in the config.

I think not running it `onMount` should do the trick.


alanmquach:
I don't think so. We're still running the rest of the logic `selectStrategy` logic. Just not confusingly mutating on render.

It's possible there is a different case where the UX _expectation_ was for the render to fix your invalid config, but 1) I don't think that is likely and 2) I don't think we want to be doing that because users don't expect a render to be dirtying their config.

But who knows? 🤷‍♂

erikmunson:
won't know until we try!

PR #8040, opened 2020-03-12
Add hooks for observables, data sources

erikmunson:
This introduces 3 new hooks, each one builds off the former:

**useObservable**
Calls a callback every time an observable emits a value. When you get that value is determined by the observable — if it immediately emits, you'll receive a value then, but if it only emits when something changes you might not get called for a while (or ever)

**useObservableValue**
Delegates to `useObservable` for subscribing to an observable, but returns the value directly from the hook and takes care of triggering a component re-render when the value changes. Also accepts a default value so you can avoid an initial `undefined` value on renders before the first observable value is emitted

**useDataSource**
Delegates to `useObservableValue`. When given a data source, immediately returns an object (`IDataSourceResult`) which first contains the value of `dataSource.data`, and then triggers updates whenever that value changes. The result object returned by the hook also includes a `refresh` function to trigger data source refreshes,  which has a stable reference (even when the data source changes)

### DIFF
--- a/app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
+++ b/app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
@@ -1,0 +1,32 @@
+import { useObservableValue } from './useObservableValue.hook';
+import { useLatestCallback } from './useLatestCallback.hook';
+
+import { ApplicationDataSource } from '../../application';
+
+export interface IDataSourceResult<T> {
+  data: T;
+  refresh: () => void;
+}
+
+/**
+ * A react hook that returns the current value an ApplicationDataSource
+ * and triggers a re-render when a new value is set. Also provides a function
+ * for refreshing the data source manually.
+ *
+ * @param dataSource the data source to subscribe to
+ * @returns IDataSourceResult<T>
+ */
+export const useDataSource = <T>(dataSource: ApplicationDataSource<T>): IDataSourceResult<T> => {
+  const data = useObservableValue(dataSource.data$, dataSource.data$.value);
+
+  // Memoize to give consumers a stable ref,
+  // but don't return the promise that dataSource.refresh() returns
+  const refresh = useLatestCallback(() => {
+    dataSource.refresh();
+  });
+
+  return {
+    data,
+    refresh,
+  };
+};
--- a/app/scripts/modules/core/src/presentation/hooks/useObservable.hook.ts
+++ b/app/scripts/modules/core/src/presentation/hooks/useObservable.hook.ts
@@ -1,0 +1,18 @@
+import { useEffect } from 'react';
+import { Observable } from 'rxjs';
+
+/**
+ * A react hook that subscribes to an rxjs observable on mount and calls the provided callback when
+ * a new value is emitted. Unsubscribes on unmount or when the observable parameter changes.
+ *
+ * @param observable the rxjs observable to subscribe to
+ * @param callback the function to call when the observable emits a value.
+ * Passed the new value as its only argument.
+ * @returns void
+ */
+export const useObservable = <T>(observable: Observable<T>, callback: (val: T) => void) => {
+  useEffect(() => {
+    const subscription = observable.subscribe(callback);
+    return () => subscription.unsubscribe();
+  }, [observable]);
+};
--- a/app/scripts/modules/core/src/presentation/hooks/index.ts
+++ b/app/scripts/modules/core/src/presentation/hooks/index.ts
@@ -1,6 +1,7 @@
 export * from './useContainerClassNames.hook';
 export * from './useData.hook';
 export * from './usePollingData.hook';
+export * from './useDataSource.hook';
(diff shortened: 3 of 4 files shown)


app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
```diff
@@ -0,0 +1,32 @@
...
+export const useDataSource = <T>(dataSource: ApplicationDataSource<T>): IDataSourceResult<T> => {
+  const data = useObservableValue(dataSource.data$, dataSource.data);
+
+  // Memoize to give consumers a stable ref,
+  // but don't return the promise that dataSource.refresh() returns
+  const refresh = useLatestCallback(() => {
+    dataSource.refresh();
+  });
+
+  return {
```
alanmquach: Should we make this a tuple?

app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A react hook that returns the current value an ApplicationDataSource
+ * and triggers a re-render when a new value is set. Also provides a function
+ * for refreshing the data source manually.
+ *
+ * @param dataSource the data source to subscribe to
+ * @returns IDataSourceResult<T>
+ */
+export const useDataSource = <T>(dataSource: ApplicationDataSource<T>): IDataSourceResult<T> => {
+  const data = useObservableValue(dataSource.data$, dataSource.data);
```
christopherthielen: any reason you didn't use `dataSource.data$.value`?

app/scripts/modules/core/src/presentation/hooks/useObservable.hook.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from 'react';
+import { Observable } from 'rxjs';
+
+/**
+ * A react hook that subscribes to an rxjs observable on mount and calls the provided callback when
+ * a new value is emitted. Unsubscribes on unmount.
```
christopherthielen:
```suggestion
 * a new value is emitted. Unsubscribes on unmount or when the observable parameter changes.
```

app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
erikmunson: yeah, that'd work too. I haven't really worked with behavior subjects much so didn't even come to mind. will update.

app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
alanmquach: Just kinda like
```
const [environments, refreshEnvironments] = useDataSource(dataSource);
```
more than
```
const { data: environments, refresh: refreshEnvironments } = useDataSource(dataSource);
```

alanmquach: Fine either way though

erikmunson: i also have a mild pref for tuples in these APIs, but in this case i wanted to leave room for adding additional nice things onto this API without a big refactor (e.g. what if you want to know what the pending/in flight status of a refresh is? errors?)

alanmquach: A 3-tuple then?

Just kidding, let's just leave it as is then.

app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
erikmunson: done!

app/scripts/modules/core/src/presentation/hooks/useObservable.hook.ts
erikmunson: done!

app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
christopherthielen: Yeah, FWIW I went through this same process on `useLatestPromise`.  Started out with tuples.  Then threeples.  Then returned values that some callers wouldn't care about, (but they might care about a latter fourple and have to destructure the thirdple to a throwaway variable).

app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
alanmquach: It appears that `dataSource.data$.value` is initially undefined whereas `dataSource.data` contains the `defaultData` 🤷‍♂

alanmquach: Looks like it's because
```
public data$ = new BehaviorSubject<T>(this.data);
```
runs before
```
this.data = this.data || this.defaultData;
```
in the `constructor`

PR #8042, opened 2020-03-13
fix(core/application): rearrange init ordering of member variables on data sources

erikmunson:
Some detective work from @alanmquach uncovered that apparently all the member initialization on classes is getting transpiled at the _top_ of the constructor, above any user-supplied constructor code. The way `ApplicationDataSource` initializes is leading to some things we setup in the constructor (`this.data`, `this.defaultData`) being undefined at times where it's important to have their initial value. The most obvious symptom of this issue is that subscribing to the `data$` BehaviorSubject actually emits `undefined` as its first value, _not_ the value of `defaultData` as is intended.

This change continues to define `private` / `public` and type info on the class directly, but moves the actual assignment of an initial value inside the constructor below the code for populating other dependent fields.

### DIFF
--- a/app/scripts/modules/core/src/application/service/applicationDataSource.ts
+++ b/app/scripts/modules/core/src/application/service/applicationDataSource.ts
@@ -273,42 +273,35 @@ export class ApplicationDataSource<T = any> implements IDataSourceConfig<T> {
   public lastRefresh: number;
 
   /** This subject is used to cancel the internal subscription */
-  private destroy$ = new Subject<void>();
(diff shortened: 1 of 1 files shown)


erikmunson:
yeah 🤷‍♂all the stuff that happens above it is pretty static/side effect free, so hopefully the only actual change in behavior is that `status$` and `data$` each return different things initially. The other dependent streams which reference `status$` and `data$` aren't even referencing the data that gets emitted — only `status` and `error` which should remain unchanged.

PR #8047, opened 2020-03-14
feat(md): Generate fallback artifact info using Frigga

alanmquach:
This is temporary.

The following pieces of information is much more meaningful to users of the environments UI than the currently available raw artifact version string:

* A meaningful artifact version number (usually the CI build number)
* Commit hash

Until keel builds proper support for returning rich artifact metadata, Queen of Asgard will parse that information out from the artifact version string for some artifact types.

Before
keeldemo_·_Environments

After
keeldemo_·_Environments

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactsList.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactsList.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import { IManagedArtifactSummary } from '../domain/IManagedEntity';
+import { IManagedArtifactSummary, IManagedArtifactVersion } from '../domain/IManagedEntity';
 import { ISelectedArtifact } from './Environments';
 import { Pill } from './Pill';
 import { parseName } from './Frigga';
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -22,7 +23,7 @@ export function ArtifactsList({ artifacts, artifactSelected }: IArtifactsListPro
             clickHandler={artifactSelected}
             version={version}
             name={name}
-            sha="abc123"
+            sha=""
```
erikmunson: maybe just give this the axe for now? 🔪

app/scripts/modules/core/src/managed/Frigga.ts
```diff
@@ -0,0 +1,28 @@
+export interface IParsedVersionComponents {
+  packageName: string;
+  version: string;
+  buildNumber: string;
+  commit: string;
+}
+
+// Lifted from Netflix/frigga
+// https://github.com/Netflix/frigga/blob/master/src/main/java/com/netflix/frigga/ami/AppVersion.java
+// https://github.com/Netflix/frigga/blob/master/src/main/java/com/netflix/frigga/NameConstants.java
```
erikmunson: ALAN

lol this is great.

app/scripts/modules/core/src/managed/ArtifactsList.tsx
alanmquach: Oh ya, since I'm now passing in the full version object

alanmquach: > passing in the full version object
Whoops left that part and some other things unpushed

erikmunson: still looks great to me

PR #8048, opened 2020-03-14
feat(core/managed): add artifact detail pane to environments

erikmunson:
Most of the changes here are for adding a detail pane to the right column of Environments when a specific artifact version is selected:

Screen Shot 2020-03-13 at 9 12 12 PM

There are some other tweaks I'll call out in comments given they're pretty contextual.

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.less
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.less
@@ -1,0 +1,15 @@
+.ArtifactDetail {
+  background-color: var(--color-white);
+  padding: 16px 0 64px 16px;
+
+  .detail-section-left {
+    flex: 0 0 280px;
+    font-style: italic;
+    font-size: 14px;
+    color: var(--color-charcoal);
+  }
+
+  .detail-section-right {
+    flex: 1 1 auto;
+  }
+}
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+
+import { IManagedArtifactVersion } from '../domain';
+import { useEventListener } from '../presentation';
+
+import { ArtifactDetailHeader } from './ArtifactDetailHeader';
+
+import './ArtifactDetail.less';
+
+export interface IArtifactDetailProps {
+  name: string;
+  version: IManagedArtifactVersion;
+  onRequestClose: () => any;
+}
+
+export const ArtifactDetail = ({ name, version, onRequestClose }: IArtifactDetailProps) => {
+  const keydownCallback = ({ keyCode }: KeyboardEvent) => {
+    if (keyCode === 27 /* esc */) {
+      onRequestClose();
+    }
+  };
+  useEventListener(document, 'keydown', keydownCallback);
+
+  return (
+    <>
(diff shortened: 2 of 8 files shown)


app/scripts/modules/core/src/managed/ArtifactDetail.less
```diff
@@ -0,0 +1,15 @@
...
+    flex: 0 0 280px;
+    font-style: italic;
+    font-size: 14px;
+    color: var(--color-charcoal);
+  }
+
+  .detail-section-right {
+    flex: 1 1 auto;
+  }
+}
```
erikmunson: because i've already been working with stuff in `core/presentation` that uses some of the new colors/styles, it was quick and straightforward for me to move the CSS module code into `.less` files, change the class names, and replace things like flex layout rules + text weights + colors with standard stuff. I don't think that's 100% necessary right now but I did it.

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -0,0 +1,36 @@
...
+  onRequestClose: () => any;
+}
+
+export const ArtifactDetail = ({ name, version, onRequestClose }: IArtifactDetailProps) => {
+  const keydownCallback = ({ keyCode }: KeyboardEvent) => {
+    if (keyCode === 27 /* esc */) {
+      onRequestClose();
+    }
+  };
+  useEventListener(document, 'keydown', keydownCallback);
```
erikmunson: it seemed nice to exit out of this on escape. later I might even add previous/next keyboard shortcuts for versions.

app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -30,6 +31,12 @@ export function Environments(props: IEnvironmentsProps) {
 
   const totalContentWidth = isFiltersOpen ? CONTENT_WIDTH + 248 + 'px' : CONTENT_WIDTH + 'px';
 
+  const selectedArtifactDetails =
+    selectedArtifact &&
+    environments.artifacts
+      .find(({ name }) => name === selectedArtifact.name)
+      ?.versions.find(({ version }) => version === selectedArtifact.version);
```
erikmunson: probably slow and bad to do this on every render, but going to likely change soon anyway.

app/scripts/modules/core/src/managed/EnvironmentsList.tsx
```diff
@@ -24,22 +23,19 @@ interface IEnvironmentsListProps {
...
         icon="search"
         text={undefined}
-        title={
-          selectedArtifact
-            ? `Showing ${selectedArtifact.name} ${selectedArtifact.version}`
-            : `${artifacts.length} artifacts is deployed in 2 environments with no issues detected.`
-        }
+        title={`${artifacts.length} artifacts ${
+          artifacts.length === 1 ? 'is' : 'are'
+        } deployed in 2 environments with no issues detected.`}
```
erikmunson: it was just bothering me.

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
alanmquach: j, k, up, down? 😅

erikmunson: _literally_ what i was thinking of. too many 2000s-era flash games and Gmail threads.

PR #8050, opened 2020-03-17
chore(titus): Reenable previous image rollbacks for titus

jeyrschabu:
- enabled previously disabled and broken titus previous image rollback

### DIFF
--- a/app/scripts/modules/titus/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
+++ b/app/scripts/modules/titus/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
@@ -66,11 +66,14 @@ module(TITUS_SERVERGROUP_DETAILS_ROLLBACK_ROLLBACKSERVERGROUP_CONTROLLER, [SERVE
       $scope.command = {
         rollbackType: rollbackType,
         rollbackContext: {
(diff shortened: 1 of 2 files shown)


jeyrschabu:
Was disabled as part of this change: https://github.com/spinnaker/deck/pull/7688

Orca has been patched to fix the underlying issue.

app/scripts/modules/titus/src/serverGroup/details/rollback/rollbackServerGroup.controller.js
```diff
@@ -66,11 +66,14 @@ module(TITUS_SERVERGROUP_DETAILS_ROLLBACK_ROLLBACKSERVERGROUP_CONTROLLER, [SERVE
       $scope.command = {
         rollbackType: rollbackType,
         rollbackContext: {
+          imageId: previousServerGroup ? previousServerGroup.imageId : undefined,
           rollbackServerGroupName: serverGroup.name,
           restoreServerGroupName: previousServerGroup ? previousServerGroup.name : undefined,
           targetHealthyRollbackPercentage: healthyPercent,
           delayBeforeDisableSeconds: 0,
         },
+        targetGroups: serverGroup.targetGroups,
```
jeyrschabu: My guess is that Titus was never updated when ALBs were introduced, so all titus prev-ami rollbacks just dropped the target groups.

PR #8051, opened 2020-03-17
refactor(amazon/serverGroup): Refactor ServerGroupBasicSettings to use FormikFormField

vigneshm:
`ServerGroupBasicSettings` form now uses `FormikFormField`s only. This PR is split into three commits for easier reviewing.

1. Converted fields in separate modules that are used in `ServerGroupBasicSettings` to use `FormikFormField`. If these fields are used in other areas, then they use `FormikFormField` only when `useStandardLayout` prop is set `true` so that it's existing uses is not affected.

2. Moved notices that are rendered in `ServerGroupBasicSettings` to separate modules for easier maintenance.

3. Converted all fields in `ServerGroupBasicSettings` to use `FormikFormField`.

Before
before

After
after

I tested each field by changing the values and ensured that the backing formik data is updated. For fields that are typically hidden, I explicitly enabled them inline and ensured they rendered correct.

local-testing

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
+++ b/app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
@@ -9,7 +9,7 @@ export interface DetailsFieldProps {
 }
 
 @Overridable('aws.serverGroup.configure.detailsField')
-export default class ServerGroupDetailsField extends React.Component<DetailsFieldProps> {
+export class ServerGroupDetailsField extends React.Component<DetailsFieldProps> {
   private freeFormDetailsChanged = (freeFormDetails: string) => {
     const { setFieldValue, values } = this.props.formik;
     values.freeFormDetails = freeFormDetails; // have to do it here to make sure it's done before calling values.clusterChanged
--- a/app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
+++ b/app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { FormikProps } from 'formik';
 
-import { HelpField, Overridable } from '@spinnaker/core';
+import { FormikFormField, HelpField, Overridable, TextInput } from '@spinnaker/core';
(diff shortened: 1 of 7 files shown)


app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
```diff
@@ -18,29 +18,16 @@ export default class ServerGroupDetailsField extends React.Component<DetailsFiel
...
-            <div className="col-sm-9 col-sm-offset-2 error-message">
-              <span>{errors.freeFormDetails}</span>
-            </div>
-          </div>
-        )}
+        <FormikFormField
+          label="Details"
+          name="freeFormDetails"
+          help={<HelpField id="aws.serverGroup.detail" />}
+          input={({ onChange, ...props }) => (
```
christopherthielen:
```suggestion
          input={({ onChange, props }) => (
```

christopherthielen:
This is a good start, thanks for tackling this.  I have a couple of requests above.

I'd like to eliminate as much manual state management as possible.  Can you check if it's possible to:

1) eliminate all the `fieldChanged()` callbacks
2) see if its possible to migrate to a functional component

vigneshm:
> I'd like to eliminate as much manual state management as possible. Can you check if it's possible to:
> 1. eliminate all the fieldChanged() callbacks
> 2. see if its possible to migrate to a functional component

I'm going to spend a little bit more time on investigating this one. There are a few dependencies that might make it non-trivial, but it would be nice to have a simple functional component with all state managed by formik (or in some other form) and would make it much easier to maintain. Can be used as a model for other forms too.

app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/fields/ServerGroupDetailsField.tsx
vigneshm: Not sure I follow.

christopherthielen: my bad, disregard.

christopherthielen:
@vigneshm are you going to complete this refactor?  If not can we close this PR until you're ready to continue working on it?

vigneshm:
> @vigneshm are you going to complete this refactor? If not can we close this PR until you're ready to continue working on it?

Yes, I want to finish it. Got pulled into other items that I haven't finished yet. Let me close it now and re-open it when I'm ready.

PR #8052, opened 2020-03-17
fix(core): replace default account select option

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5579

This [refactor](https://github.com/spinnaker/deck/pull/7952) removed the first un-selectable `Select...` placeholder option from the `AccountSelectInput`. For newly created stages, this resulted in the first account displaying as selected, but the `onChange` handler never firing with that account value (so the stage JSON not reflecting the UI).

An alternate solution I considered was by default selecting the first account if no valid account is selected, but that would potentially confuse users in cases where the account that was previously selected is no longer an option, and we silently updated the selection to the first valid account.


christopherthielen:
>  An alternate solution I considered was by default selecting the first account if no valid account is selected

I think you made the right choice.  Thanks for catching and fixing!

maggieneterval:

PR #8053, opened 2020-03-18
fix(kubernetes): whitelist helm artifact types for bake manifest stage

maggieneterval:
Related to: https://github.com/spinnaker/spinnaker/issues/5589

Previously, we permitted all artifact types as input to Bake (Manifest) stage with the Helm render engine specified. However, this has led to confusion since only certain artifact types are appropriate. For example, the Git Repo artifact type does not yet support Helm (open feature request [here](https://github.com/spinnaker/spinnaker/issues/5249)), but it was available for selection. Let's explicitly whitelist acceptable artifact types for Helm bakes.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/helm/BakeHelmConfigForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/helm/BakeHelmConfigForm.tsx
@@ -4,7 +4,7 @@ import { StageConfigField } from '../../common/stageConfigField/StageConfigField
 import { CheckboxInput, TextInput } from 'core/presentation';
 import { MapEditor } from 'core/forms';
 import { IArtifact, IExpectedArtifact, IPipeline } from 'core/domain';
-import { StageArtifactSelectorDelegate } from 'core/artifact';
+import { excludeAllTypesExcept, ArtifactTypePatterns, StageArtifactSelectorDelegate } from 'core/artifact';
 import { IFormikStageConfigInjectedProps } from '../../FormikStageConfig';
 
 interface IBakeHelmConfigFormProps {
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/helm/BakeHelmConfigForm.tsx
```diff
@@ -4,14 +4,25 @@ import { StageConfigField } from '../../common/stageConfigField/StageConfigField
...
+  private static readonly excludedArtifactTypes = excludeAllTypesExcept(
+    ArtifactTypePatterns.BITBUCKET_FILE,
+    ArtifactTypePatterns.CUSTOM_OBJECT,
+    ArtifactTypePatterns.GCS_OBJECT,
+    ArtifactTypePatterns.GITHUB_FILE,
+    ArtifactTypePatterns.GITLAB_FILE,
+    ArtifactTypePatterns.S3_OBJECT,
+    ArtifactTypePatterns.HELM_CHART,
+    ArtifactTypePatterns.HTTP_FILE,
+  );
```
maggieneterval: @dbyron0 I'm going to need my resident Helm artifact expert to give this a once-over and make sure I didn't exclude any artifact types that you think folks could be using as input to the Bake (Manifest) stage for Helm. Complete list of artifact types is [here](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/artifact/ArtifactTypes.ts#L5).

Thank you! 🙏

dbyron0: Does EMBEDDED_BASE64 make sense?  Seems like it could.  I'm curious about JENKINS_FILE, IVY_FILE and MAVEN_FILE too, but know even less about those.  Could a helm chart tar ball be living behind them?

maggieneterval: Thanks for the quick reply, @dbyron0! I think you're right that `EMBEDDED_BASE64` could be fair game (just pushed a commit to include that type), but I think Ivy/Maven artifacts are for JAR files so I think we can safely exclude those for now.

PR #8054, opened 2020-03-18
fix(titus/instance): Temporary workaround to get IPv6 for Titus instances

caseyhebebrand:
Currently, Titus instances do not expose IPv6 addresses. As a temporary workaround to unlock other IPv6 work, we will leverage the `agentId` and `awsAccount` to make a secondary call for this information. In the future we will remove the `addIpv6Addresses` function, since the instance will natively have an `ipv6Addresses` attribute.

Also, I have separated DNS headers for Public v4 addreses and Ipv6 Addresses (which are all public) based on some feedback.

### DIFF
--- a/app/scripts/modules/amazon/src/instance/details/instanceDetails.html
+++ b/app/scripts/modules/amazon/src/instance/details/instanceDetails.html
@@ -209,9 +209,7 @@
         <dd ng-if="instance.permanentIps" ng-repeat="ip in instance.permanentIps">
           <link-with-clipboard url="'http://' + ip + ':' + state.instancePort" text="ip"> </link-with-clipboard>
         </dd>
-        <dt ng-if="instance.publicIpAddress || instance.ipv6Addresses.length">
-          Public IP Address<span ng-if="instance.ipv6Addresses.length > 1">es</span>
-        </dt>
+        <dt ng-if="instance.publicIpAddress">Public IP Address</dt>
         <dd ng-if="instance.publicIpAddress">
           <link-with-clipboard
             url="'http://' + instance.publicIpAddress + ':' + state.instancePort"
@@ -219,6 +217,9 @@
           >
           </link-with-clipboard>
         </dd>
+        <dt ng-if="instance.ipv6Addresses.length">
+          IPv6 Address<span ng-if="instance.ipv6Addresses.length > 1">es</span>
+        </dt>
         <dd ng-if="instance.ipv6Addresses.length" ng-repeat="ip in instance.ipv6Addresses">
           <link-with-clipboard url="'http://[' + ip + ']:' + state.instancePort" text="ip"> </link-with-clipboard>
         </dd>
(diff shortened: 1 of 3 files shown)


app/scripts/modules/titus/src/instance/details/instance.details.controller.js
```diff
@@ -118,6 +118,7 @@ module(TITUS_INSTANCE_DETAILS_INSTANCE_DETAILS_CONTROLLER, [
             $scope.instance.externalIpAddress = $scope.instance.placement.host;
             getBastionAddressForAccount(accountDetails, region);
             $scope.instance.titusUiEndpoint = this.titusUiEndpoint;
+            addIpv6Addresses(accountDetails.awsAccount, region, instanceDetails.agentId);
```
christopherthielen: let's mix this into the promise chain, instead of mutating the object on $scope after the overall promise resolves.

app/scripts/modules/titus/src/instance/details/instanceDetails.html
```diff
@@ -166,6 +166,10 @@ <h3 class="horizontal middle space-between flex-1" select-on-dbl-click>
         </dd>
         <dt>Host IP</dt>
         <dd>{{instance.placement.host}}</dd>
+        <dt ng-if="instance.ipv6Addresses.length">IPv6</dt>
```
erikmunson: Would it make sense to restructure this DNS section slightly now that there's technically multiple container IPs? i.e. the structure right now is:
- DNS
  - Container IP
  - Host IP
  - IPv6

What if it became:
- DNS
  - Container IPs
    - IPv4
    - IPv6
  - Host IP

Not suggesting a specific visual treatment for now, just asking about taxonomy.

app/scripts/modules/amazon/src/instance/details/instanceDetails.html
```diff
@@ -209,16 +209,17 @@ <h3 class="horizontal middle space-between flex-1" select-on-dbl-click>
...
         <dd ng-if="instance.publicIpAddress">
           <link-with-clipboard
             url="'http://' + instance.publicIpAddress + ':' + state.instancePort"
             text="instance.publicIpAddress"
           >
           </link-with-clipboard>
         </dd>
+        <dt ng-if="instance.ipv6Addresses.length">
+          IPv6 Address<span ng-if="instance.ipv6Addresses.length > 1">es</span>
+        </dt>
```
erikmunson: similar question as my other comment re: structure — could there be IPv4 and IPv6 sections under "Public IP Address(es)"?

caseyhebebrand: The feedback I had was to keep IPv6 and IPv4 in separate headings for the users since the concept of public vs private does not exist in v6. We started out with exactly what you said. If you look to the changed code, you can see that everything was listed under "Public IP Addresses".  However, we concluded that it is better to isolate v6 since it is its own entity. 

Also, it is extremely unlikely that an instance will have all 3 types of IP addresses. We may see it during this roll out phase, but the common case will be an instance with 1 Private IPv4 and 1 or more IPv6.

app/scripts/modules/titus/src/instance/details/instanceDetails.html
caseyhebebrand: I will restructure a bit:
- Container IPv4
- Container IPv6
- Host IP

I want to keep the heading separate since they are separate concepts.

app/scripts/modules/amazon/src/instance/details/instanceDetails.html
erikmunson: 👍 👍

app/scripts/modules/titus/src/instance/details/instanceDetails.html
erikmunson: 👍 seems very reasonable

PR #8055, opened 2020-03-18
refactor(svg): add SVGR loader for inlined react SVG support

erikmunson:
Diff is **not** as big as it looks — the vast majority of the changes are in the lockfile.

We've had informal discussions of moving from using icon fonts for icons, to inline SVG. There's a few nice things we get from it:
- The ability to properly leverage version control for modifying icons — today all changes are wrapped up in opaque font file binaries, which makes it nearly impossible to do concurrent changes to icons across separate PRs
- The service we use to emit font files (Icomoon) now requires that you pay for a premium subscription to emit modern (woff2) files. While paying for the premium plan is fine (we already do this), it creates a very challenging bottleneck where all icon updates have to go through whoever has the credentials for the Icomoon account
- Inline SVGs carry a variety of nice benefits from a purely technical perspective, like more flexibility in how they're constructed/animated/colored. There's a really good [breakdown on CSS Tricks](https://css-tricks.com/icon-fonts-vs-svg/) that covers most of the important points

This PR prepares us to make the switch by adding the ability to automatically load SVG imports as pre-compiled, optimized React components that render SVG markup inline.  I'm choosing to use the [SVGR library](https://react-svgr.com/) which works really well with webpack/rollup/typescript, and is very actively maintained and updated. It's the default choice in Create React App for handling SVG files, and appears very healthy.

There are two commits, which each handle:
- Adding the loader configuration, and TS module declarations so the compiler knows what should be passed in to the pre-compiled components
- Switching _all_ our `require('./blah/thing.svg')` imports over to ES default import syntax — this is because in order to support both URL/path based imports and inline JSX, SVGR creates a default export that contains the output of our existing `file-loader`, and a named export (`ReactComponent`) which holds the generated component

### DIFF
--- a/app/scripts/modules/amazon/src/svg.d.ts
+++ b/app/scripts/modules/amazon/src/svg.d.ts
@@ -1,0 +1,6 @@
+declare module '*.svg' {
+  import { ComponentType, SVGProps } from 'react';
+  export const ReactComponent: ComponentType<SVGProps<HTMLOrSVGElement>>;
+  const URL: string;
+  export default URL;
+}
--- a/app/scripts/modules/amazon/webpack.config.js
+++ b/app/scripts/modules/amazon/webpack.config.js
@@ -104,6 +104,14 @@ module.exports = {
         use: [{ loader: 'style-loader' }, { loader: 'css-loader' }, { loader: 'postcss-loader' }],
       },
       {
+        test: /\.svg$/,
+        issuer: {
+          test: /\.(tsx?|js)$/,
+        },
+        use: [{ loader: '@svgr/webpack' }],
+        exclude: exclusionPattern,
+      },
+      {
         test: /\.html$/,
         exclude: exclusionPattern,
         use: [
--- a/app/scripts/modules/core/src/artifact/ArtifactIconService.ts
+++ b/app/scripts/modules/core/src/artifact/ArtifactIconService.ts
@@ -1,6 +1,20 @@
 import { ArtifactTypePatterns } from './ArtifactTypes';
 
-const unknownArtifactPath = require('./icons/unknown-type-artifact.svg');
+import unknownArtifactIcon from './icons/unknown-type-artifact.svg';
+import dockerIcon from './icons/docker-image-artifact.svg';
+import kubernetesIcon from './icons/kubernetes-artifact.svg';
+import embeddedBase64Icon from './icons/embedded-base64-artifact.svg';
+import gcsObjectIcon from './icons/gcs-file-artifact.svg';
+import gitHubFileIcon from './icons/github-file-artifact.svg';
+import gitRepoIcon from './icons/git-repo-artifact.svg';
+import gitLabFileIcon from './icons/gitlab-file-artifact.svg';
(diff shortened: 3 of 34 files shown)


app/scripts/modules/amazon/src/svg.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module '*.svg' {
```
erikmunson: I would _really_ like to find a good way of doing this without type declarations. [The docs have a guide on how to create a custom template string that actually creates TS code for the autogenerated components](https://react-svgr.com/docs/typescript/), but when I tried that it was a total mess because the generated code has to then also pass through the `ts-loader` / `thread-loader` etc. etc. path which kept leading to obscure and weird errors. Using a custom template also means we can't make use of some of the really clever optimizations the default template has like creating `const`s for all the paths inside the SVG in order to reuse the same JSX elements across renders.

I think what i really want is a way to do this module declaration in a spot where it doesn't have to be repeated for every single module.

app/scripts/modules/amazon/webpack.config.js
```diff
@@ -103,6 +103,14 @@ module.exports = {
         test: /\.css$/,
         use: [{ loader: 'style-loader' }, { loader: 'css-loader' }, { loader: 'postcss-loader' }],
       },
+      {
+        test: /\.svg$/,
+        issuer: {
+          test: /\.(tsx?|js)$/,
+        },
```
erikmunson: we need this `issuer` block so that CSS/LESS files continue to use the same `file-loader` import — the thing SVGR does of adding a named `ReactComponent` export messes that up.

app/scripts/modules/core/src/artifact/ArtifactIconService.ts
```diff
@@ -1,6 +1,20 @@
 import { ArtifactTypePatterns } from './ArtifactTypes';
 
-const unknownArtifactPath = require('./icons/unknown-type-artifact.svg');
+import unknownArtifactIcon from './icons/unknown-type-artifact.svg';
+import dockerIcon from './icons/docker-image-artifact.svg';
```
christopherthielen: side note: as an opponent of default imports, I might be inclined to add a linter rule that flags this line complaining that `dockerIcon` should be `dockerImageArtifact` or something similar.

Like? Hate?

christopherthielen:
One thing to think about: I'm on a path to eliminate webpack from the module builds.  I'm 99.9999% sure there is an equivalent "loader" for rollup.

erikmunson:
>  I'm 99.9999% sure there is an equivalent "loader" for rollup.

[Indeed there is](https://react-svgr.com/docs/rollup/)! I'm counting the days until all the submodules are rollup 📆

app/scripts/modules/core/src/artifact/ArtifactIconService.ts
erikmunson: Yeah I like loosey goosey default imports less and less over time. Not against a lint rule. Would you think of the check as basically "imported name must equal file name transformed from kebab case to camel"?

I haven't ever written one myself, but given there's a bunch of examples where we're already looking at imports I'm guessing I could whip something up.

PR #8056, opened 2020-03-18
feat(core/presentation): add Icon component w/ inline SVG

erikmunson:
**Depends on #8055**

Per the points in #8055 around switching to inline SVG for icons in place of icon fonts, this change adds a new `Icon` component which is intended to be the central place for Deck code to add icons — this is effectively a replacement for `<i className="fa fas icon-xyz" />`. 

The API is pretty small right now — I'm placing a bet on having more curated/opinionated options for 'standard' colors and sizes ~~using the `color` and `size` props, but with an escape hatch to allow folks to arbitrarily specify colors and sizes via `customColor` and `customSize`. We'll see how that plays out — I'm very open to changing it.~~ After feedback, there's now [a different API than what was initially described here](https://github.com/spinnaker/deck/pull/8056#discussion_r394766394).

I'm also intentionally punting the question of how to _extend_ the set of available icons via things like plugins or internal custom builds — I want to get feedback on the API and direction first before tackling that. For now I'm really liking the autocomplete and static validation for icon names (i.e. you type `name="` into your editor, and an autocomplete list of all the available icons shows up), but I realize we may have to give that up if we want to use some sort of `IconRegistry` for extending the set of icons. Suggestions or ideas on how to retain type info while making icons extensible would be helpful!


erikmunson:
Typical usage looks like this:
```tsx
// "small" is the default size, since that's what we use most often
<Icon name="artifacts" color="alabaster" size="medium" />
```

(cc @gcomstock)

.gitignore
```diff
@@ -2,7 +2,10 @@
 *.swp
 .DS_Store
 ehthumbs.db
-Icon?
+# This "Icon" entry needs its invisible carriage return character preserved
+# to avoid igoring *any* directory that starts with the name 'icon':
+# https://stackoverflow.com/questions/17556250/how-to-ignore-icon-in-git
+Icon
```
erikmunson: this was fun to figure out.

app/scripts/modules/core/src/presentation/icons/Icon.tsx
```diff
@@ -0,0 +1,31 @@
...
+  customSize?: string;
+  className?: string;
+}
+
+const DEFAULT_SIZE = 'small';
+
+const pxDimensionsBySize = {
+  small: '24px',
+  medium: '34px',
+  large: '40px',
```
erikmunson: @gcomstock In thinking about the API for this component I wanted to encourage more standardized/opinionated practices around icons. A couple questions for you:

- Do you feel confident enough about how icons will be used _going forward_ (don't worry about existing icon usage) to make choices about a standard set of 'sizes' like this? It's not impossible to backtrack/change our mind, but it will be much much harder to do so after we've migrated a bunch of font file icons over
- If the answer to the first question is yes, do you feel these _specific_ names and sizes are appropriate? I went off of what I've seen in both mockups and prototype code

app/scripts/modules/core/src/presentation/icons/Icon.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { iconsByName } from './iconsByName';
+
+export type IconNames = keyof typeof iconsByName;
+
+export interface IIconProps {
+  name: IconNames;
+  color: string;
+  customColor?: string;
+  size?: 'small' | 'medium' | 'large';
+  customSize?: string;
```
caseyhebebrand: I think we could reduce the need for some of the props like `customColor` and `customSize`:

- `color`: I like the idea of this component determining if the color is custom or not. I am not sure how feasible that is. Maybe we can just have the user input hex or a common color string, since we are using inline styling. Maybe we lookup to see if that custom string exists. Or maybe we leave it :) Just a suggestion, that could make this simpler.
- `size`: If the `pxDimensionsBySize[size]` does not exist, fallback to input size or default size. Thus reducing the need for `customSize`:   
```js
const width = pxDimensionsBySize[size] || size || pxDimensionsBySize[DEFAULT_SIZE]
```

A third thought, that I liked a lot when working with CSS-in-JS and I think could apply here is a catch-all `customStyles` prop that can override anything.

```js
style={{ width, fill, ...customStyles }}
```

app/scripts/modules/core/src/presentation/icons/Icon.tsx
caseyhebebrand: Are there _types_ of icons that can inform the color?  This could be restrictive as far as colors go, but also worth thinking about. I see component libraries take this approach

gcomstock: @erikmunson The key to making an icon jive with text is to have it hold a similar visual weight. I prefer using the light/"hollow" icons most of the time because the stroke width is similar to a text's stroke width, so your eye doesn't see them as two things smushed together but instead one thing.

Having said that, there are times when an icon also needs to be visually substantial and hold its own, so it's not just going to be light icons, but also bolder ones. So it's not just a matter of size, but also weight. But it's not like we should include multiple weights of every icon. It's circumstantial.

If we were to make some rules, I'd suggest the following:

x-small: 16
small: 20
medium: 24
large: 32
x-large: 40

I think this will cover almost every use case, except for when we want to use an svg icon as an image. Meaning, visually it doesn't read like an icon, but is instead given the prominence and importance of a picture. I think if we get much larger though it's just generally more fun to use more visually rich svgs, like the characters I've used before.

@caseyhebebrand This is a cool idea! I haven't considered it. My goal with the icon color itself is readability, which usually means contrast against its background, vs having the icon color communicate something more. ex: a white icon against a red/green background I feel more quickly communicates bad/good instead of a red/green icon against a white background. We're starting to wade into subjective waters here though.

If we want to get prescriptive with the icon color, we can maybe just have:

dark: #000
neutral: #777
light: #fff

I scanned over my designs and I think it satisfies almost every use case.

Thanks!

gcomstock: actually I think using neutral: rgba(0,0,0,0.5) would make it a bit more foolproof. Thanks!

app/scripts/modules/core/src/presentation/icons/Icon.tsx
erikmunson: I'm just gonna respond to these in order:
- `color`: I actually agree wholeheartedly, but couldn't really figure out a way to make it work without doing inlining of actual CSS colors as you mention. We don't currently have a great story around getting available color values in JS code to check against, though I'm still holding out hope for some kind of CSS-in-JS solution to that problem. While not violently opposed to using CSS colors in the API, I also do believe it's important to make configuring colors as prescriptive and straightforward as possible for things like plugins, or PRs from folks who don't know much of anything about deck or its conventions. What if... we just omitted `customColor` for now and tried to move everything we migrate over to a color that lives in `styleguide`? I don't actually think there's a lot of places we'd run into one-off colors.
- `size`: Also agree, and I think the reason I didn't go this route was because I had already split out `customColor` and thought eh, what's another 'custom' prop 😆. I'm gonna just switch to using your suggested behavior
- `style`: I suspect we'll end up adding this. It's a pretty common convention to have both `className` and `style` passthroughs on these sorts of components, and I already added `className` as I ran into some uses for it. My preference would be to hold off on actually adding it until we try to move a thing that we think needs it — IMO the more we can learn and experiment with the API before adding surface area, the better the quality of those decisions will be.

app/scripts/modules/core/src/presentation/icons/Icon.tsx
erikmunson: I am _super_ into that expanded set of sizes. Thanks for the brainpower.

> I think this will cover almost every use case, except for when we want to use an svg icon as an image.

I agree, and I'm explicitly trying to make SVG images out of scope for the job of this component — IMO being very prescriptive about the API makes it possible to avoid this getting abused for graphics/images (which should really just use the existing pattern we already have in deck today).

Thanks to both of you for the feedback/suggestions around getting even more prescriptive with color — based on the direction of Greg's work overall I am also in favor of providing a dark/neutral/light option. In order to make this component work for the current state of deck in addition to where we want to go, what about providing that more prescriptive API as a first class thing, and demoting `color` to a sort of custom/escape hatch property? For example:

```tsx
<Icon name="artifacts" appearance="light" />
```

```tsx
<Icon name="artifacts" color="alabaster" />
```

We could also make `color` accept both `light`/`neutral`/`dark` _and_ color strings, but I believe I have a slight preference for keeping them separate so that examples, guides, newer code etc. all push folks down a path where they're using those three options, rather than giving them a big bag of strings to pick from.

erikmunson: Ok so the new API I've published here looks more like this:

```tsx
// 'neutral' appearance, 'small' size
<Icon name ="artifact" />

// This would be the new-world, preferred way to use icons
<Icon name ="artifact" appearance="light" size="medium" />

// these explicit color and size props would be legacy ways
// of migrating the rest of deck, we could refactor them away
// over time or keep them
<Icon name ="artifact" color="danger" size="21px" />

```

.gitignore
christopherthielen: o boy...

app/scripts/modules/core/src/presentation/icons/Icon.tsx
```diff
@@ -0,0 +1,47 @@
...
+  light: 'var(--color-white)',
+  neutral: 'rgba(0, 0, 0, 0.5)',
+  dark: 'var(--color-black)',
+};
+
+const throwInvalidIconError = (name: string) => {
+  throw new Error(`No icon with the name ${name} exists`);
+};
+
+export const Icon = memo(({ name, appearance, size, color, className }: IIconProps) => {
```
christopherthielen: `memo` nice...

app/scripts/modules/core/src/presentation/icons/Icon.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { memo } from 'react';
+
+import { iconsByName } from './iconsByName';
+
+export type IconNames = keyof typeof iconsByName;
+
+export type IIconProps = {
+  name: IconNames;
+  appearance?: 'light' | 'neutral' | 'dark';
+  size?: string;
```
christopherthielen:
```suggestion
  size?: 'extraSmall' | 'small' | 'medium' | 'large' | 'extraLarge' | string;
```

erikmunson: ha. You know what's funny is I wrote exactly this union out in my editor, and when I looked over the diff before committing I decided to take it out. At the time it felt like even though I enjoyed having it there we weren't getting much from it given TS coalesces the whole thing down to `string` (e.g. no autocomplete). You independently wishing it was there gives me the motivation to go with it 😄

erikmunson:
> I bet we can use module augmentation

Thought you might say that! I'll probably try going down that path as far as possible — I remember with data sources there were some issues around other modules (e.g. `amazon`) getting ahold of those due to weird deck specialness (I think we theorized it was those subpackage aliases?). Hoping we can figure it out nicely.

My only annoyance with the interface route is that it adds an additional step in modifying the collection of icons. Having to touch SVG import + icon map + interface when making changes still seems way less tedious than the Icomoon / font file flow though.

gcomstock:
Thanks everyone! I'll do a PR with a bunch of separated svg icons

PR #8060, opened 2020-03-19
feat(core/managed): use real data for version progress in Environments

erikmunson:
We were hard-coding all these progress bar steps as green because the API couldn't tell us what to show yet, but now it can!

Screen Shot 2020-03-18 at 11 11 08 PM

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactRow.module.css
+++ b/app/scripts/modules/core/src/managed/ArtifactRow.module.css
@@ -50,11 +50,36 @@
   height: 6px;
   padding-bottom: 2px;
 }
+
 .stage {
-  background-color: #00b237;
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/managed/ArtifactRow.module.css
```diff
@@ -50,11 +50,36 @@
...
+  &.pending {
+    background-color: var(--color-porcelain);
+  }
+
+  &.previous {
+    background-color: var(--color-nobel);
+  }
+
+  &.vetoed {
+    background-color: #be0000;
```
erikmunson: don't worry about these inlined colors, I'll be doing a pass to move all of it into `styleguide` soon.

PR #8063, opened 2020-03-20
feat(core/managed): add resource breakdown to artifact details

erikmunson:
~~**Depends on https://github.com/spinnaker/deck/pull/8060, diff will be bloated and CI will fail until merged**~~

This does a very light refactoring pass to pull out our existing use of `ObjectRow` into a shared spot, adds some data massaging for getting a breakdown of environments + resources for a particular artifact version, and then shows that on the detail view:

Screen Shot 2020-03-19 at 5 26 22 PM

Right now only environments that reference the artifact are displayed, **but** within those environments all resources are shown. Right now on the Keel side proper resource <--> artifact linkage is in the works and when that drops we'll filter down to just the compute resources that use that artifact.

### DIFF
--- a/app/scripts/modules/core/src/managed/Environments.tsx
+++ b/app/scripts/modules/core/src/managed/Environments.tsx
@@ -1,9 +1,10 @@
-import React, { useState } from 'react';
-import { isEqual } from 'lodash';
+import React, { useState, useMemo } from 'react';
+import { isEqual, keyBy } from 'lodash';
 
 import { useDataSource } from '../presentation/hooks';
 import { Application, ApplicationDataSource } from '../application';
-import { IManagedApplicationEnvironmentSummary } from '../domain/IManagedEntity';
+import { IManagedApplicationEnvironmentSummary, IManagedResourceSummary } from '../domain';
+
 import { ColumnHeader } from './ColumnHeader';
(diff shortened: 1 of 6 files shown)


app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -22,20 +23,35 @@ interface IEnvironmentsProps {
...
 
-  const totalContentWidth = isFiltersOpen ? CONTENT_WIDTH + 248 + 'px' : CONTENT_WIDTH + 'px';
+  const resourcesById = useMemo(() => keyBy(resources, 'id'), [resources]);
+  const resourcesByEnvironment = useMemo(
+    () =>
+      environments.reduce((byEnvironment, { name, resources: resourceIds }) => {
+        byEnvironment[name] = resourceIds.map(id => resourcesById[id]);
+        return byEnvironment;
+      }, {} as { [environment: string]: IManagedResourceSummary[] }),
+    [environments, resourcesById],
```
erikmunson: i'd like to:
1) in the immediate term, push this calculation down to `ArtifactDetail` so it's done less frequently/only when necessary
2) in the medium term, move this mostly to the backend by storing a list of resources inside these artifact version summaries

PR #8069, opened 2020-03-21
feat(core/managed): status indicators, explanation cards on artifacts

erikmunson:
This is an early stab at some status indicators for each environment in an artifact's journey, along with explanatory cards for a subset of statuses. It's likely to change significantly in the very near future as @gcomstock and I flesh out the details, and for now I picked some random icons to use in lieu of the real ones.

**Current**
Screen Shot 2020-03-20 at 9 04 13 PM

**Deploying (data is faked for this one so things don't line up between cards/sidebar):**
Screen Shot 2020-03-20 at 9 01 51 PM

**Previous**
Screen Shot 2020-03-20 at 9 04 21 PM

**Vetoed**
Screen Shot 2020-03-20 at 9 04 32 PM

**Pending (never deployed)**
Screen Shot 2020-03-20 at 9 04 50 PM

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -1,10 +1,16 @@
 import React from 'react';
+import classNames from 'classnames';
+import { DateTime } from 'luxon';
 
+import { relativeTime, timestamp } from '../utils';
 import { IManagedArtifactVersion, IManagedResourceSummary } from '../domain';
 import { useEventListener } from '../presentation';
 
 import { ArtifactDetailHeader } from './ArtifactDetailHeader';
+import { NoticeCard } from './NoticeCard';
+import { Pill } from './Pill';
 import { ManagedResourceObject } from './ManagedResourceObject';
+import { parseName } from './Frigga';
 
 import './ArtifactDetail.less';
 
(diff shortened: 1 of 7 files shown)


app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -42,14 +48,97 @@ export const ArtifactDetail = ({
...
+              )}
+              {state === 'vetoed' && (
+                <NoticeCard
+                  className="sp-margin-l-right"
+                  icon="skull"
+                  text={undefined}
+                  title={
+                    <span className="sp-group-margin-xs-xaxis">
+                      Marked as bad{' '}
+                      {deployedAt && (
```
erikmunson: if the version got manually vetoed before ever making it into an environment, we won't have a `deployedAt`

app/scripts/modules/core/src/managed/NoticeCard.tsx
```diff
@@ -3,14 +3,15 @@ import classNames from 'classnames';
 import styles from './NoticeCard.module.css';
 
 interface INoticeCardProps {
-  title: string;
-  text: string;
+  title: JSX.Element | string;
+  text: JSX.Element | string;
   icon: string;
-  noticeType: 'ok'; // TypeScript doesn't like this being arbitrary strings because it can't validate styles[noticeType]
+  noticeType: 'success' | 'neutral' | 'info' | 'error';
```
erikmunson: i ended up needing to do a bit more refactoring on `NoticeCard` in this PR than I had anticipated/hoped as there's a wider variety of states this component needed to support. I'm holding off on doing a full refactor of it as Greg is in the process of tightening up the markup/states.

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -42,14 +48,97 @@ export const ArtifactDetail = ({
...
+          const deployedAtMillis = DateTime.fromISO(deployedAt).toMillis();
+          const replacedAtMillis = DateTime.fromISO(replacedAt).toMillis();
+          const { version: replacedByPackageVersion, buildNumber: replacedByBuildNumber } =
+            parseName(replacedBy || '') || {};
+
+          return (
+            <div key={name}>
+              <h3>{name.toUpperCase()}</h3>
+              {state === 'deploying' && (
+                <NoticeCard
```
erikmunson: i'll probably pull out all this card business into a component soon — needs to stabilize a little bit first.

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
alanmquach: Worth dropping that into a comment?

erikmunson: Yeah good call ➕

erikmunson: after thinking about it, I actually went one step further and explicitly surfaced a user-facing state for this (rather than only giving people looking at the code that context):

Screen Shot 2020-03-23 at 1 37 14 PM

gcomstock:
I forgot about these cloud icons! there's a few of them to work with and I think they'll do the job nicely.

erikmunson:
> I forgot about these cloud icons! there's a few of them to work with and I think they'll do the job nicely.
 
FWIW I like the rocket ship icon you chose for "Deployed" in your prototype — i'd be very happy using that.

PR #8076, opened 2020-03-23
fix(titus): Update host account environment

caseyhebebrand:
Titus instances were using the `awsAccount` field to fetch the host instance network interface information. The best field to use is the `environment` field since this is will always evaluate to a known host account-type (`test` or `prod`), whereas `awsAccount` can evaluate to anything and cause errors (e.g.: `workstation_test` instead of `test`).

### DIFF
--- a/app/scripts/modules/titus/src/instance/details/instance.details.controller.js
+++ b/app/scripts/modules/titus/src/instance/details/instance.details.controller.js
@@ -123,7 +123,7 @@ module(TITUS_INSTANCE_DETAILS_INSTANCE_DETAILS_CONTROLLER, [
             }
 
             // Network interfaces are needed for any IPv6 addresses.
-            return InstanceReader.getInstanceDetails(accountDetails.awsAccount, region, instanceDetails.agentId);
+            return InstanceReader.getInstanceDetails(accountDetails.environment, region, instanceDetails.agentId);
           })
           .then(instance => {
             if (instance && instance.networkInterfaces) {


app/scripts/modules/titus/src/instance/details/instance.details.controller.js
```diff
@@ -123,7 +123,7 @@ module(TITUS_INSTANCE_DETAILS_INSTANCE_DETAILS_CONTROLLER, [
             }
 
             // Network interfaces are needed for any IPv6 addresses.
-            return InstanceReader.getInstanceDetails(accountDetails.awsAccount, region, instanceDetails.agentId);
+            return InstanceReader.getInstanceDetails(accountDetails.environment, region, instanceDetails.agentId);
```
alanmquach: @aravindmd for the longer term, would it make sense for clouddriver to expose an attribute that specifies what account the compute is running? This would be a required field for titus accounts in `clouddriver.yml`.

This way when Titus runs in many accounts we won't be burned by inferring the compute account from the environment.

caseyhebebrand: This secondary call will only be needed for another month or two. Titus is adding an IPv6 attribute that we can directly use for the details panel. This effort was delayed a bit but currently in progress.

aravindmd: Agree , that sounds like a good solution , i can take a look at the clouddriver bits

PR #8080, opened 2020-03-24
feat(core/managed): show relevant resources on version details

erikmunson:
Up until today we did not have the ability to tell whether a resource referenced a given artifact. [Now we do](https://github.com/spinnaker/keel/pull/905), so this change filters out all the non-referenced resources when looking at a particular version's journey. For example, here we only see one cluster resource per environment even though each environment actually contains many other resources (security groups, LBs):

Screen Shot 2020-03-23 at 10 10 54 PM

Later I will likely add some sort of treatment that clarifies why you're only seeing those instead of every resource (e.g. a "4 other resources exist in this environment" link that goes to the overview of that environment).

Because the modeling for a resource's `artifact` field changed I also had to mess with resource rows on `EnvironmentList`, so I stuck a pill in for the current version. Doesn't handle when there's a `deploying` version and probably missing all kinds of other stuff: 

Screen Shot 2020-03-23 at 10 11 03 PM

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
+++ b/app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
@@ -1,12 +1,15 @@
 import React from 'react';
 
-import { IManagedResourceSummary } from '../domain/IManagedEntity';
+import { IManagedResourceSummary, IManagedEnviromentSummary } from '../domain/IManagedEntity';
 
 import { getKindName } from './ManagedReader';
 import { ObjectRow } from './ObjectRow';
+import { Pill } from './Pill';
+import { parseName } from './Frigga';
 
 export interface IManagedResourceObjectProps {
   resource: IManagedResourceSummary;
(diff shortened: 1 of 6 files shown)


app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -20,13 +23,20 @@ function getIconTypeFromKind(kind: string): string {
...
-);
+  artifact,
+}: IManagedResourceObjectProps) => {
+  const { version: currentVersion, buildNumber: currentBuild } = parseName(artifact?.versions.current || '') || {};
+  return (
+    <ObjectRow
+      icon={getIconTypeFromKind(kind)}
+      title={[app, stack, detail].filter(Boolean).join('-')}
+      metadata={
+        artifact?.versions.current && (
```
erikmunson: I don't really know if I love any of the way this is structured/written, but it's a start 🤷‍♂

PR #8082, opened 2020-03-24
fix(core/deploymentStrategy): Handle missing rollback property in command

vigneshm:
This prevents the server group edit/clone modal from crashing if `red-black` strategy is chosen.


app/scripts/modules/core/src/deploymentStrategy/strategies/redblack/AdditionalFields.tsx
```diff
@@ -24,7 +24,7 @@ export class AdditionalFields extends React.Component<IRedBlackStrategyAdditiona
           <label>
             <input
               type="checkbox"
-              checked={command.rollback.onFailure}
+              checked={command.rollback?.onFailure}
```
erikmunson: not essential, but might be worth ensuring this prop is always a boolean so the componet is more clearly controlled (vs uncontrolled)
```suggestion
              checked={command.rollback?.onFailure ?? false}
```

vigneshm: Good call!

PR #8083, opened 2020-03-24
feat(core/presentation): adds svg icons for new ui designs

gcomstock:
Here's a bunch of icons for the new and upcoming design efforts on deck.

Icons pre-fixed with "sp_" are intended for the existing UI. They are visually heavier. This set includes icons for v1 nav restructure and the CI views. These icons will be cleaned up in the distant future when the UI has been visually rolled over. We can clean the font file up at that time as well.

@erikmunson  @caseyhebebrand


app/scripts/modules/core/src/presentation/icons/iconsByName.ts
```diff
@@ -1,9 +1,124 @@
...
+import { ReactComponent as md_diff } from './vectors/md_diff.svg';
+import { ReactComponent as md_error } from './vectors/md_error.svg';
+import { ReactComponent as md_flapping } from './vectors/md_flapping.svg';
+import { ReactComponent as md_paused } from './vectors/md_paused.svg';
+import { ReactComponent as md_resumed } from './vectors/md_resumed.svg';
+import { ReactComponent as md_unknown } from './vectors/md_unknown.svg';
+import { ReactComponent as md } from './vectors/md.svg';
+import { ReactComponent as placeholder } from './vectors/placeholder.svg';
+import { ReactComponent as servergroup_aws } from './vectors/servergroup_aws.svg';
+import { ReactComponent as sp_ci_branch } from './vectors/sp_ci_branch.svg';
```
gcomstock: If you want to use svg icons for the CI views, they'll be available here. Feel free to continue to use the icon font though.

erikmunson:
@gcomstock woohoo this is so cool!

I believe CI is already telling you this in its own very noisy way, but FYI you'll need to change all the `snake_case` stuff here to `camelCase` as we enforce that via the build (i recommend also changing the casing of the svg files themselves to either `kebab-case` or `camelCase`)

app/scripts/modules/core/src/presentation/icons/iconsByName.ts
erikmunson: could we add a line break + big comment block before all the legacy `sp` ones explaining what they are / the intended usage?

Might even be worth doing the same thing for different categories of icons (i.e. but all the MD ones in their own block, etc.)

gcomstock: Great idea

gcomstock:
> @gcomstock woohoo this is so cool!
> 
> I believe CI is already telling you this in its own very noisy way, but FYI you'll need to change all the `snake_case` stuff here to `camelCase` as we enforce that via the build (i recommend also changing the casing of the svg files themselves to either `kebab-case` or `camelCase`)

Ah serves me right for trying to do this without running locally

caseyhebebrand:
This is awesome! Excited to put some of these new svgs to use in the near future.

PR #8086, opened 2020-03-24
chore(mergify): release-* branches require release manager approval

ezimanyi:
The mergify config currently doesn't distinguish between merging to master and
merging to a release branch.  It currently never merges to release branches
because mergify doesn't have permission to merge to these branches (due to
branch protection rules).

I don't want to give mergify that permission without updating the config file
here, as then it would merge to release branches after any approval, whereas we
only want to merge to release branches if a PR is approved by a release manager.

This updates the mergify config to scope the existing rules to only apply to
the master branch, and adds a new rule to merge to release branches upon
approval by a release manager.

Because the 1.17 branch (and likely the 1.18 branch in some repos) is still
exclusively using Travis CI instead of github actions, also add a rule the
merges if Travis CI passes to account for these branches. This rule can be
deleted once all of the active branches in the repo are using Github actions.


ezimanyi:
I didn't also add `ready to rebase` support to the release branches as that seemed like enough of an edge case...the release manager can manually merge cherry-picks in that case (though in any case mostly cherry-picks will be a single commit anyway).

PR #8091, opened 2020-03-25
feat(core/managed): add real icons to Environments view

erikmunson:
Now that @gcomstock has very generously taken the time to move a bunch of icons over to our new SVG-powered `<Icon/>` component, it's time to give Environments UI a more polished look with more finalized icons:

**Before**
Screen Shot 2020-03-25 at 10 49 49 AM

**After**
Screen Shot 2020-03-25 at 10 49 38 AM

As a bonus while I was in here I added a tiny border radius tweak to the column headers (result of a discussion with @gcomstock) and fixed a `key` prop error that was totally my fault and I should've noticed earlier.

### DIFF
--- a/app/scripts/modules/core/src/managed/EnvironmentRow.tsx
+++ b/app/scripts/modules/core/src/managed/EnvironmentRow.tsx
@@ -1,6 +1,8 @@
 import React, { useState } from 'react';
 import classNames from 'classnames';
 
+import { Icon } from '../presentation';
+
 import './EnvironmentRow.less';
 
 interface IEnvironmentRowProps {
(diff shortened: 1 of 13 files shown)


app/scripts/modules/core/src/managed/EnvironmentRow.tsx
```diff
@@ -30,8 +32,8 @@ export function EnvironmentRow({ name, isProd = false, children }: IEnvironmentR
           <span className={envLabelClasses}>{name}</span>
         </span>
         <div className="expand" onClick={() => setIsCollapsed(!isCollapsed)}>
-          {isCollapsed && <i className="ico icon-expand" />}
-          {!isCollapsed && <i className="ico icon-collapse" />}
+          {isCollapsed && <Icon name="accordionExpand" size="medium" />}
+          {!isCollapsed && <Icon name="accordionCollapse" size="medium" />}
```
erikmunson: @gcomstock you'll notice in the "After" screenshot on this PR that these expand/collapse icons changed from 'production' red to the neutral grey icon color — I did this because I'd prefer to stick with the three color options we chose for `<Icon/>` last week until we can come up with a scalable solution to additional color permutations. Alternatively we do have the option to decide that trying to standardize on 3 options was too restrictive for our ideal design system and we should just let people pick any available color 😆

gcomstock: Sounds good!

PR #8096, opened 2020-03-25
refactor(artifacts): consolidate artifacts feature flags checks

maggieneterval:
Related to: https://github.com/spinnaker/governance/pull/111

The `artifacts` and `artifactsRewrite` flags were intended to be [mutually exclusive](https://github.com/spinnaker/deck/pull/6694), but it is possible that this was not fully thought through, as there are several workflows (Find Artifact from Resource stage, Find Artifact from Execution stage) that currently require both flags to be enabled in order to use with the new UI. This PR addresses that bug and consolidates the artifacts feature flag checks to a single service.

- refactor(artifacts): consolidate artifacts feature flags checks

  Adds `ArtifactsModeService`, which is responsible for the mapping between the four possible feature flag configuration states and the three possible UI experiences. Replaces each instance of logic conditional on one or both artifacts feature flags with logic conditional on the UI experience to which they map.

- fix(artifacts): enable artifact-specific stages when only `artifactsRewrite` feature flag is enabled

  Register the Find Artifact from Resource stage and Find Artifact from Execution stage if at least one of the two existing artifacts feature flags is enabled.

Pending a successful audit of the `STANDARD` UI experience, we will:

- Remove the `DISABLED` experience (1.20).
- Remove the `artifacts` and `artifactsRewrite` feature flags (1.20).
- Default to the `STANDARD` experience unless a new, temporary `legacyArtifacts` flag is enabled (1.20).
- Remove the `LEGACY` experience and `legacyArtifacts` flag. With only the `STANDARD` experience remaining, we can remove the `ArtifactsModeService` and all conditional calling code (1.21).

### DIFF
--- a/app/scripts/modules/core/src/artifact/ArtifactsModeService.ts
+++ b/app/scripts/modules/core/src/artifact/ArtifactsModeService.ts
@@ -27,7 +27,7 @@ export enum ArtifactsMode {
 }
 
 export class ArtifactsModeService {
-  public static artifactsMode = ArtifactsModeService.getArtifactsMode();
+  public static readonly artifactsMode = ArtifactsModeService.getArtifactsMode();
 
   private static getArtifactsMode(): ArtifactsMode {
     if (SETTINGS.feature.artifactsRewrite === true) {
--- a/app/scripts/modules/core/src/artifact/ArtifactsModeService.ts
+++ b/app/scripts/modules/core/src/artifact/ArtifactsModeService.ts
@@ -1,0 +1,41 @@
+import { SETTINGS } from 'core/config';
+
+/**
+ * Currently, because there are two artifacts feature flags, there are four
+ * possible configuration states:
+ * 1. `artifacts` disabled, `artifactsRewrite` disabled
+ * 2. `artifacts` enabled, `artifactsRewrite` disabled
+ * 3. `artifacts` disabled, `artifactsRewrite` enabled
+ * 4. `artifacts` enabled, `artifactsRewrite` enabled
+ *
+ * However, only three UI experiences should be possible:
+ * 1. No artifacts UI
+ * 2. The legacy artifacts UI
+ * 3. The standard ("rewrite") artifacts UI
+ *
+ * This service provides a layer of abstraction over the existing feature flag
+ * checks in terms of these three possible experiences. As we deprecate these
+ * feature flags in favor of an enabled-by-default standard artifacts UI, as
+ * described in https://github.com/spinnaker/governance/pull/111, it will be
+ * helpful to have this logic isolated to a single service.
+ */
+
+export enum ArtifactsMode {
+  DISABLED,
+  LEGACY,
+  STANDARD,
+}
+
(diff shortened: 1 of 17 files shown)


app/scripts/modules/core/src/artifact/expectedArtifactMultiSelector.component.ts
```diff
@@ -33,34 +38,33 @@ const expectedArtifactMultiSelectorComponent: IComponentOptions = {
...
-                  height="16"
-                  class="artifact-icon"
-                  ng-src="{{ ctrl.iconPath(expected) }}" />
-                <span>{{ expected.displayName }}</span>
-              </ui-select-choices>
-            </ui-select>
-          </stage-config-field>
-        </ng-form>
-      </render-if-feature>
+      <ng-form name="artifacts" ng-if="ctrl.renderIf()">
```
maggieneterval: This is a big diff because of the re-indentation, but the only actual change is replacement of the `render-if-feature` wrapper with an `ng-if` conditional on `renderIf`.

app/scripts/modules/core/src/artifact/imageSourceSelector.component.ts
```diff
@@ -12,22 +19,20 @@ const imageSourceSelectorComponent: IComponentOptions = {
...
-          <help-field key="{{ ctrl.helpFieldKey }}"></help-field>
-        </div>
-        <div class="col-md-7">
-          <div class="radio" ng-repeat="imageSource in ctrl.imageSources">
-            <label>
-              <input type="radio" ng-model="ctrl.command[ctrl.idField]" value="{{ imageSource }}">
-              {{ imageSource | robotToHuman }}
-            </label>
-          </div>
+    <div class="form-group" ng-if="!ctrl.imageSourceText && ctrl.artifactsEnabled">
```
maggieneterval: Similarly, the only non-indentation diff here is replacement of `render-if-feature` with `ng-if`.

app/scripts/modules/core/src/artifact/expectedArtifactMultiSelector.component.ts
maggieneterval: ⚠️Just discovered that you can select ` Hide whitespace changes` from the ⚙️ dropdown; would recommend for this PR.

app/scripts/modules/core/src/artifact/ArtifactsModeService.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+export enum ArtifactsMode {
+  DISABLED,
+  LEGACY,
+  STANDARD,
+}
+
+export class ArtifactsModeService {
+  public static artifactsMode = ArtifactsModeService.getArtifactsMode();
```
ezimanyi: Should this be `readonly`?

maggieneterval: YES. Thank you.

PR #8097, opened 2020-03-26
refactor(titus/serverGroupDetails): Move details field to core and use that in titus

vigneshm:
- A new `details` field is created under `core` and will be used for `titus` cloud provider
- This field will be overridden later to expose the sharding feature.

NOTE: There will be another PR to make `amazon`'s server group settings to use this new field.


app/scripts/modules/core/src/serverGroup/configure/common/wizard/fields/ServerGroupDetailsField.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import { FormikProps } from 'formik';
+
+import { Application, HelpField, IServerGroupCommand, Overridable } from 'core';
+
+export interface DetailsFieldProps<T extends IServerGroupCommand> {
+  app: Application;
```
christopherthielen: it doesn't look like `app` is being used, but I assume it is there for the override component to use?

vigneshm: Correct, it is used there.

PR #8100, opened 2020-03-27
refactor(amazon/details): Use details field from core and remove amazon-specific implementation

vigneshm:
After #8097  `ServerGroupDetailsField` is part of `core`, so I'm removing the old implementation from `amazon` module.


erikmunson:
nice

PR #8101, opened 2020-03-27
feat(core/managed): remove non-deploying state indicators

erikmunson:
After looking at them in the wild we decided we don't really like having indicators on the left side of every resource for every artifact state. It gets pretty busy:

Screen Shot 2020-03-27 at 9 02 35 AM

Instead we're now removing them entirely for all states except `deploying`. Right now we can't show deployment progress at the resource level, so these indicators are kind of silly and static, but once we get there we'll treat them as pie chart progress indicators and some fanciness will ensue.

I also took the liberty of making the remaining indicator line up better with its surroundings, so the resource + notice card icons are back in proper alignment and the left edge of the indicator lines up with the left edge of environment rows:

Screen Shot 2020-03-27 at 9 00 35 AM

(cc @gcomstock )

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.less
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.less
@@ -18,33 +18,12 @@
     height: 16px;
     border-radius: 16px;
     border-radius: 50%;
-
-    &.current {
-      background-color: #00b237;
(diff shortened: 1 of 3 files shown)


gcomstock:
Thanks!

PR #8106, opened 2020-03-27
chore(core): remove versionedProviders flag

maggieneterval:
Since we already support `unsupportedStageTypes` to differentiate between Kubernetes V1 and V2 stages (and we will be removing V1 for release 1.21 anyway), I do not think there is a need for a `versionedProviders` feature flag to gate V2-specific stages. It is not configurable in Halyard, so this should be the only change necessary to remove the feature flag.


ezimanyi:
(It will greatly simplify things.)

PR #8107, opened 2020-03-28
feat(core/managed): add read-only stateful constraint support

erikmunson:
This adds support for stateful constraints in a _read only_ capacity — so no interactivity like approving/rejecting manual judgements or otherwise overriding. Some examples:

<details>
<summary>NOT_EVALUATED</summary>

Screen Shot 2020-03-28 at 11 00 09 AM
</details>

<details>
<summary>PENDING</summary>

Screen Shot 2020-03-28 at 11 00 36 AM
</details>

<details>
<summary>OVERRIDE_PASS</summary>

Screen Shot 2020-03-28 at 10 59 39 AM
</details>

<details>
<summary>OVERRIDE_FAIL</summary>

Screen Shot 2020-03-28 at 10 59 53 AM
</details>

<details>
<summary>FAIL</summary>

Screen Shot 2020-03-28 at 2 13 48 PM
</details>

I had to mess with the margins on some card-related things to get the right spacing, so that's wrapped up in here too. I also dropped a `pending` state card in for similar spacing/layout reasons even though we haven't quite figured out what we're doing with it yet ("never deployed here"):
Screen Shot 2020-03-28 at 2 15 35 PM


app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -51,105 +53,126 @@ export const ArtifactDetail = ({
           {/* a short summary with actions/buttons will live here */}
           <div className="detail-section-right">{/* artifact metadata will live here */}</div>
         </div>
-        {environments.map(({ name: environmentName, state, deployedAt, replacedAt, replacedBy }) => {
-          const deployedAtMillis = DateTime.fromISO(deployedAt).toMillis();
-          const replacedAtMillis = DateTime.fromISO(replacedAt).toMillis();
-          const { version: replacedByPackageVersion, buildNumber: replacedByBuildNumber } =
-            parseName(replacedBy || '') || {};
+        {environments.map(
+          ({ name: environmentName, state, deployedAt, replacedAt, replacedBy, statefulConstraints }) => {
```
erikmunson: prettier totally re-indented this whole function because of this extra restructuring, so the diff is ridiculous and unreadable. I recommend turning on the "Hide whitespace changes" option in the diff settings menu as it cleans up nicely.

app/scripts/modules/core/src/managed/constraints/constraintRegistry.tsx
```diff
@@ -0,0 +1,65 @@
...
+          );
+        case FAIL:
+          return (
+            <span className="sp-group-margin-xs-xaxis">
+              Something went wrong <span className="text-regular">—</span>{' '}
+              <span className="text-regular">{comment || NO_FAILURE_MESSAGE}</span>
+            </span>
+          );
+        default:
+          return throwUnhandledStatusError(status);
```
vigneshm: Is there a way to let typescript do an exhaustive check of all the possible values of an enum instead of handling the `default` case?

app/scripts/modules/core/src/managed/constraints/ConstraintCard.tsx
```diff
@@ -0,0 +1,47 @@
...
+export interface IConstraintCardProps {
+  constraint: IStatefulConstraint;
+  className?: string;
+}
+
+export const ConstraintCard = ({ constraint, className }: IConstraintCardProps) => {
+  const { type, status } = constraint;
+
+  if (!isConstraintSupported(type)) {
+    throwUnsupportedConstraintError(type);
```
vigneshm: Would be nice to avoid throwing and causing the application to break. Instead we could render a 'Not supported' or 'Invalid' card (or not render anything depending on the use case) and provide a way to alert us of these failing cases through logs.

alanmquach: I agree, I would prefer to see a very crude "unsupported type" card (kind of like the JSON text area for stages without a specific stage config UI), over throwing.

app/scripts/modules/core/src/managed/constraints/constraintRegistry.tsx
erikmunson: There is, but it has some disadvantages.

To my knowledge, the best way to get exhaustiveness here would be to have the handler return an object with functions for each constraint status. Then we could use a mapped type (`{ [status in StatefulConstraintStatus]: (constraint: IStatefulConstraint) => JSX.Element | string }`) to explicitly say "you have to return an object with a key for every value in this enum".

My hesitation in doing that is I'm not yet confident it makes sense to have a handler for every status, for every constraint. In this `manual-judgment` example, `PASS` is actually _not_ a state we expect to ever be possible because it always requires manual interaction (which produces an `OVERRIDE_PASS`/`OVERRIDE_FAIL` status). The backend code _can't_ actually send a manual-judgement into a `PASS` status. We wouldn't know how to handle it on the UI side as a result, and it doesn't seem like it makes much sense to put in UI code specifically for this constraint type intended to handle a case we don't think should exist or know how to represent. Instead I was thinking it'd be ideal for the core code to handle unexpected statuses in some common, generic way that might involve catching this error and showing that an unexpected case was hit with that constraint (i.e. treat it like a failure case, not a one-off thing that every constraint either did well or poorly).

Any opinions on whether that makes sense or not? There's certainly a counter-argument to be made for telling constraint developers they should handle every case and do some sort of best-effort attempt at presenting things in unexpected/impossible cases.

app/scripts/modules/core/src/managed/constraints/ConstraintCard.tsx
erikmunson: I realize now I probably should've explained this better. Here's my thinking, tell me if it's reasonable or not.

At least at the moment, we're taking the approach of not showing constraints at all if they're not explicitly supported by deck. That may change later, but it seems appropriate for now (we have a variety of constraints we expect to keep 'implicit'). This `ConstraintCard` thing is an internal component for use inside the environments view code (i.e. I do not plan on folks being able to render this themselves). In order to properly enforce the contract that we should not be attempting to show unsupported constraints, I expect every parent component rendering this code to filter out things that are unsupported and handle them properly, as opposed to having `ConstraintCard` handle it by rendering `null` and maybe end up with the parent component showing a janky layout due to expecting certain sizing/etc.

So my expectation here was that this is a totally impossible case to even get into, and the purpose of handling it at all is to quickly and efficiently assist whoever is seeing it in fixing their code (i.e. during local development). If we're worried about a bug eventually making it outside of the local dev stage and crashing customer pages, I'm not opposed to switching to rendering `null` and still `console.error`-ing this exception so it gets a nice component stack trace. Though I suspect that will make a little less obvious and harder to fix pre-deployment.

app/scripts/modules/core/src/managed/constraints/constraintRegistry.tsx
erikmunson: I'm gonna think about doing the return-an-object thing in a subsequent PR, even if you have to return _every_ status in that object (i.e. the type can be partial)

app/scripts/modules/core/src/managed/constraints/ConstraintCard.tsx
erikmunson: For now I updated this to `console.error` the same exception, and return `null`. We can see how that goes and if we end up wanting stronger guardrails of some kind to avoid unhandled constraints getting haphazardly rendered

PR #8111, opened 2020-03-30
refactor(core): simplify trigger state management

maggieneterval:
- refactor(core): add tests for TriggersPageContent

  Add test coverage for adding, editing, and removing triggers to ensure nothing breaks after refactoring. Note that currently in the standard UI, an expected artifact that was defined against a trigger (and never associated with any other triggers) persists in `pipeline.expectedArtifacts` even after that trigger is deleted: in this state, there is no way to delete the expected artifact from the pipeline other than manually editing the pipeline JSON.

- refactor(core): simplify trigger state management

  It is currently very difficult to reason about how we make updates to triggers and expected artifacts.  Let's simplify the state management by lifting each type of mutating action for triggers and expected artifacts (add, edit, delete) into individual methods defined on `TriggersPageContent`. Let's also handle the removal of dereferenced expected artifacts as a side effect of trigger removal.

~~Future work: I think we should disambiguate deleting an expected artifact and removing it as a constraint from a trigger (perhaps with a different action button like an "X" with a descriptive tooltip to distinguish it from the trash can?), but would love @justinrlee's input before moving forward with that change since I think he's received the most direct user feedback.~~

Edit: We discussed this at the 4/6 UI SIG and the consensus was that no separate remove-artifact-from-trigger action button is necessary; we will include only the trash can icon for deleting the artifact, which implicitly also removes it from each associated trigger.


justinrlee:
The goal here is to help migration from the old pre-rewrite artifacts UI (and the associated underlying fields) to the new artifact UI, right?  Specifically, the new UI is overall backwards compatible, but you'll end up with fields that are no longer used (or in the wrong places) that can be removed / migrated?

Overall, I'm 100% on board with this, but I'm not enough of a developer to be able to provide actionable feedback on code.

maggieneterval:
@justinrlee thanks for the quick reply! Sorry, should have explained in more detail:
- In the old UI, expected artifacts had a dedicated section in a pipeline config where users could individually delete them.
- In the new UI, expected artifacts do not have a dedicated section and when they are defined as a constraint against the trigger, it's not clear what the trash can icon means: delete the expected artifact or remove it as a constraint from the trigger? For now it means both, which I think is a little confusing.
So my suggestion was to add some new icon next to the trash can for removing the artifact as a constraint, to disambiguate from deletion.

justinrlee:
Ah, yeah.  That makes perfect sense.  Yeah, I like having a UI way of cleaning up expected artifacts, and separating out "remove from trigger" from "remove from pipeline".  Let me know if you need help with testing or something - happy to try to build this locally and see if the workflow makes sense!

Also, sorry for the late reply this time...

PR #8114, opened 2020-03-31
fix(travis): Refresh travis build information on each render

sergiopena:
Probably this could be handled better... the issue is that Travis build info is set in the controller constructor, so any pipeline with more than one travis stage. It only shows execution details for the stage fetched at load time for all the stages.

Just added a refresh method that is called on each render of the execution details.

https://github.com/spinnaker/spinnaker/issues/5601

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/travis/travisExecutionDetails.controller.ts
+++ b/app/scripts/modules/core/src/pipeline/config/stages/travis/travisExecutionDetails.controller.ts
@@ -21,6 +21,7 @@ export class TravisExecutionDetailsCtrl implements IController {
     this.stage = this.$scope.stage;
     this.initialize();
     this.$scope.$on('$stateChangeSuccess', () => this.initialize());
+    this.$scope.$watch('stage.refId', () => (this.stage = $scope.stage));
   }
 
   public initialized(): void {


maggieneterval:
@sergiopena thanks for the PR! Rather than refreshing on every single re-render, would it be possible to instead only refresh when `$scope.stage` changes? Thanks!

sergiopena:
@maggieneterval  I knew this could be done in a better way :) Thanks a lot.
I've pushed the same update using a watch on the stage, is this what you suggested?

app/scripts/modules/core/src/pipeline/config/stages/travis/travisExecutionDetails.controller.ts
```diff
@@ -21,6 +21,13 @@ export class TravisExecutionDetailsCtrl implements IController {
     this.stage = this.$scope.stage;
     this.initialize();
     this.$scope.$on('$stateChangeSuccess', () => this.initialize());
+    this.$scope.$watch(
+      'stage',
+      function() {
+        this.stage = $scope.stage;
+      }.bind(this),
+      true,
+    );
```
maggieneterval, line 30:
```suggestion
    this.$scope.$watch('stage.refId', () => this.stage = $scope.stage);
```

What are your thoughts on simplifying to something like the above? With an arrow function, we can avoid binding `this` since it will use the `this` value of its enclosing lexical scope. Additionally, if we watch the `refId` rather than the entire stage object, we can avoid an expensive deep equality check.

Thanks again for stepping up to resolve this bug!

sergiopena: Awesome! I've changed it and it works flawlessly

Thanks again for your patience @maggieneterval

sergiopena:

PR #8115, opened 2020-03-31
feat(core/managed): Deep link resources to their infrastructure view

vigneshm:
Adding the ability to deep-link resources to their infrastructure view by applying the filters such as `q=name`, `stack`, `detail`. This is done via parsing `kind` and `id` from `resource` object.

[edit: adding screenshots]

image

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
+++ b/app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useSref } from '@uirouter/react';
 
 import { IconNames } from '../presentation';
 import { IManagedResourceSummary, IManagedEnviromentSummary } from '../domain/IManagedEntity';
@@ -21,23 +22,50 @@ const kindIconMap: { [kind: string]: IconNames } = {
   'application-load-balancer': 'loadBalancer',
 };
 
-function getIconTypeFromKind(kind: string) {
-  return kindIconMap[getKindName(kind)] ?? 'placeholder';
-}
+const getIconTypeFromKind = (kind: string) => kindIconMap[getKindName(kind)] ?? 'placeholder';
 
-export const ManagedResourceObject = ({
-  resource: {
+const getResourceName = ({ moniker: { app, stack, detail } }: IManagedResourceSummary) =>
+  [app, stack, detail].filter(Boolean).join('-');
+
+const getResourceRoutingInfo = (
+  resource: IManagedResourceSummary,
+): { state: string; params: { [key: string]: string } } | null => {
+  const {
     kind,
-    moniker: { app, stack, detail },
-  },
-  artifact,
-  depth,
-}: IManagedResourceObjectProps) => {
+    locations: { account },
+  } = resource;
+  const kindName = getKindName(kind);
+  const params = {
+    acct: account,
+    q: getResourceName(resource),
+  };
+
+  switch (kindName) {
+    case 'cluster':
+      return { state: 'home.applications.application.insight.clusters', params };
+
+    case 'security-group':
+      return { state: 'home.applications.application.insight.firewalls', params };
+
+    case 'classic-load-balancer':
+    case 'application-load-balancer':
+      return { state: 'home.applications.application.insight.loadBalancers', params };
+  }
+
+  return null;
+};
+
+export const ManagedResourceObject = ({ resource, artifact, depth }: IManagedResourceObjectProps) => {
   const { version: currentVersion, buildNumber: currentBuild } = parseName(artifact?.versions.current || '') || {};
+  const { kind } = resource;
+  const resourceName = getResourceName(resource);
+  const routingInfo = getResourceRoutingInfo(resource) ?? { state: '', params: {} };
+  const route = useSref(routingInfo.state, routingInfo.params);
+
   return (
     <ObjectRow
       icon={getIconTypeFromKind(kind)}
-      title={[app, stack, detail].filter(Boolean).join('-')}
+      title={route ? <a {...route}>{resourceName}</a> : resourceName}
       depth={depth}
       metadata={
         artifact?.versions.current && (
--- a/app/scripts/modules/core/src/managed/ObjectRow.module.css
+++ b/app/scripts/modules/core/src/managed/ObjectRow.module.css
@@ -30,8 +30,6 @@
   background-image: linear-gradient(to bottom, rgba(160, 180, 220, 0.15), rgba(160, 180, 220, 0.15));
 }
 
-/* "pointer-events: none;" is added to children so the data-key attached to ObjectRow is always captured */
-
 .leftCol {
   flex: 5 0 1px;
   display: flex;
@@ -40,8 +38,6 @@
   text-overflow: ellipsis;
   white-space: nowrap;
   margin-right: 16px;
-
-  pointer-events: none;
   /* background-color: rgba(255, 0, 0, 0.1); */
 }
 
--- a/app/scripts/modules/core/src/managed/ObjectRow.tsx
+++ b/app/scripts/modules/core/src/managed/ObjectRow.tsx
@@ -6,7 +6,7 @@ import styles from './ObjectRow.module.css';
 
 interface IObjectRowProps {
   icon: IconNames;
-  title: string;
+  title: JSX.Element | string;
   metadata?: JSX.Element;
   depth?: number;
 }


app/scripts/modules/core/src/managed/ObjectRow.module.css
```diff
@@ -40,8 +38,6 @@
   text-overflow: ellipsis;
   white-space: nowrap;
   margin-right: 16px;
-
-  pointer-events: none;
```
vigneshm: @alanmquach I deleted this rule since it was preventing the link from being clicked. I didn't quite understand your related comment, so want to check if I'm introducing any issue here.

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -25,19 +26,61 @@ function getIconTypeFromKind(kind: string) {
   return kindIconMap[getKindName(kind)] ?? 'placeholder';
 }
 
-export const ManagedResourceObject = ({
-  resource: {
+const resourceKindParsingRegex = /(?<provider>\w+)\/(?<resourceType>[\w-]+)\@(?<version>\w+)/;
+const resourceNameParsingRegex = /:([\w-]+)$/;
```
vigneshm: Created these regular expressions with the following assumptions.

1. `resource.kind` is of the format `provider/resourceType/version`
2. `resource.id` is of the format `provider:resourceType:account:resourceName`

app/scripts/modules/core/src/managed/ObjectRow.module.css
erikmunson: kinda looks to me like it was left over from some implementation details of Greg's prototype code

vigneshm:
Going to fix the failing builds.

app/scripts/modules/core/src/managed/ObjectRow.tsx
```diff
@@ -7,17 +7,18 @@ import styles from './ObjectRow.module.css';
...
 
-export const ObjectRow = ({ icon, title, metadata, depth = 1 }: IObjectRowProps) => {
+export const ObjectRow = ({ icon, title, link, metadata, depth = 1 }: IObjectRowProps) => {
   return (
     <div className={styles.ObjectRow} style={getStylesFromDepth(depth)}>
       <span className="clickableArea">
         <div className={styles.leftCol}>
           <Icon name={icon} size="medium" appearance="dark" className="sp-margin-s-right" />
-          <span className={styles.rowTitle}>{title}</span>
+          <span className={styles.rowTitle}>{link ? <a href={link}>{title}</a> : title}</span>
```
erikmunson: this `link` prop is a little bit brittle in the sense that it's directly tied to the usage pattern we currently have (wrapping the title in an `<a>` tag for which we already have a fully constructed href). Instead, can we change the type `title` to `JSX.Element | string` and pass in the components/markup we want from _outside_ of the `ObjectRow` code?

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -25,19 +26,61 @@ function getIconTypeFromKind(kind: string) {
...
-  depth,
-}: IManagedResourceObjectProps) => {
-  const { version: currentVersion, buildNumber: currentBuild } = parseName(artifact?.versions.current || '') || {};
+  } = resource;
+
   return (
     <ObjectRow
       icon={getIconTypeFromKind(kind)}
       title={[app, stack, detail].filter(Boolean).join('-')}
+      link={useResourceLink(resource)}
```
erikmunson: calling this function calls hooks, which ideally should all get pulled to the very start of the component to avoid confusion (e.g. if a conditional or other non-deterministic bit of logic gets added later)

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -25,19 +26,61 @@ function getIconTypeFromKind(kind: string) {
...
-  resource: {
+const resourceKindParsingRegex = /(?<provider>\w+)\/(?<resourceType>[\w-]+)@(?<version>\w+)/;
+const resourceNameParsingRegex = /:([\w-]+)$/;
+
+function useResourceLink(resource: IManagedResourceSummary): string | null {
+  const resourceType = (resource.kind.match(resourceKindParsingRegex) as any)?.groups?.resourceType;
+  const acct = resource.locations.account;
+  const { app, stack, detail } = resource.moniker;
+  const resourceNameMatch = resource.id.match(resourceNameParsingRegex);
+  const resourceName = resourceNameMatch ? resourceNameMatch[1] : '';
```
erikmunson: couple things:
- please use [`getKindName()`](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/managed/ManagedReader.ts#L18) for parsing out the `resourceType`, which already uses a regex and is at least a centralized place to improve/keep track of that logic
- please use `app` + `stack` + `detail` from the `moniker` object to compose the resource's name instead of introspecting/parsing `id` ([example](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/managed/ManagedResourceObject.tsx#L40)). `id` is meant to be a blackbox, un-parsed string that's passed around between systems. It may change over time, and may have a different structure for new resource kinds

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -25,19 +26,61 @@ function getIconTypeFromKind(kind: string) {
   return kindIconMap[getKindName(kind)] ?? 'placeholder';
 }
 
-export const ManagedResourceObject = ({
-  resource: {
+const resourceKindParsingRegex = /(?<provider>\w+)\/(?<resourceType>[\w-]+)@(?<version>\w+)/;
```
erikmunson: just for education purposes (it has no behavioral implications here): the first component of the `kind` string is not actually cloud provider — that's just coincidentally what most of them use right now. That's actually a "namespace" that can be anything, roughly equivalent to NPM package namespaces.

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
vigneshm: Thanks! Didn't realize these were already available

app/scripts/modules/core/src/managed/ObjectRow.tsx
vigneshm: Good call. I'll update that

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
vigneshm: @erikmunson I was constructing the params for the route the following way

```
{
    acct: locations.account,
    application: moniker.app,
    detail: moniker.detail,
    stack: moniker.stack,
    q: resourceName,
 };
```

But if `resourceName` is formed the way you mention, then I could just do the following since `detail` and `stack` are just redundant here?

```
{
    acct: locations.account,
    application: moniker.app,
    q: resourceName,
 };
```

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
vigneshm: Yea, the build failure was related to the conditional call made inside the `useResourceLink` function. I'm moving it to the form you described.

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -21,23 +22,50 @@ const kindIconMap: { [kind: string]: IconNames } = {
   'application-load-balancer': 'loadBalancer',
 };
 
-function getIconTypeFromKind(kind: string) {
-  return kindIconMap[getKindName(kind)] ?? 'placeholder';
-}
+const getIconTypeFromKind = (kind: string) => kindIconMap[getKindName(kind)] ?? 'placeholder';
+
+const getResourceName = ({ moniker: { app, stack, detail } }: IManagedResourceSummary) =>
+  [app, stack, detail].filter(Boolean).join('-');
```
vigneshm: If this is common enough, then I can move this to `ManagedReader` as a utility too!

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
erikmunson: Yeah that looks safe to do AFAICT!

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
erikmunson: yeah that's a good point. I have no opinion on whether we need to do it _right now_ (there are 2 existing usages, this will be the third), but if we did I bet we could figure out a nice way to tie it in with [our existing name utilities](https://github.com/spinnaker/deck/tree/master/app/scripts/modules/core/src/naming). I don't love the way the existing ones are structured/named (ironically), but it does seem like the right place for both the kind name extraction that lives in `ManagedReader` and this (there's an existing implementation called `getClusterName()` which is definitely aging poorly and needs a rename/refactor).

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -21,23 +22,50 @@ const kindIconMap: { [kind: string]: IconNames } = {
...
 
-export const ManagedResourceObject = ({
-  resource: {
+const getResourceRoutingInfo = (
+  resource: IManagedResourceSummary,
+): { state: string; params: { [key: string]: string } } | null => {
```
erikmunson: we're not using the `strictNullChecks` flag on TS, so the `| null` union is already done for you (not that I'm saying that's a good thing...)

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -21,23 +22,50 @@ const kindIconMap: { [kind: string]: IconNames } = {
...
+  const {
     kind,
-    moniker: { app, stack, detail },
-  },
-  artifact,
-  depth,
-}: IManagedResourceObjectProps) => {
+    locations: { account },
+  } = resource;
+  const resourceType = getKindName(kind);
```
erikmunson:
```suggestion
  const kindName = getKindName(kind);
```

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -21,23 +22,50 @@ const kindIconMap: { [kind: string]: IconNames } = {
...
+  const params = {
+    acct: account,
+    q: getResourceName(resource),
+  };
+
+  switch (resourceType) {
```
erikmunson:
```suggestion
  switch (kindName) {
```

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -21,23 +22,50 @@ const kindIconMap: { [kind: string]: IconNames } = {
...
+
+  return null;
+};
+
+export const ManagedResourceObject = ({ resource, artifact, depth }: IManagedResourceObjectProps) => {
   const { version: currentVersion, buildNumber: currentBuild } = parseName(artifact?.versions.current || '') || {};
+  const { kind } = resource;
+  const resourceName = getResourceName(resource);
+  const routingInfo = getResourceRoutingInfo(resource);
+  const route = useSref(routingInfo.state, routingInfo.params);
```
erikmunson: is deref-ing `routingInfo` here going to NPE in the case where `getResourceRoutingInfo` returns `null`?

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -21,23 +22,50 @@ const kindIconMap: { [kind: string]: IconNames } = {
...
+
   return (
     <ObjectRow
       icon={getIconTypeFromKind(kind)}
-      title={[app, stack, detail].filter(Boolean).join('-')}
+      title={route ? <a {...route}>{resourceName}</a> : resourceName}
```
erikmunson: nice

erikmunson:
🙏 🚀 

one last thing: can you post screenshots of what it looks like when these are wrapped in links? I assume there's some change in visual appearance?

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
vigneshm: Oops! yes. Thanks for the catch. Fixing it

vigneshm: I wish `strictNullChecks` was enabled to catch these :)

erikmunson:
@vigneshm ready for me to merge this?

vigneshm:
> @vigneshm ready for me to merge this?

Yes please!

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
erikmunson: welp, i was totally wrong about this: https://github.com/spinnaker/deck/pull/8141

sorry!

PR #8116, opened 2020-03-31
fix(core): bump up width on numerical form fields

gcomstock:
This addresses a Netflix customer concern that the width on the pipeline wait stage > wait time form field was too narrow. I took a look and saw the 60px style is used throughout deck, but I don't think it's necessary. I think a placeholder value or tooltips go further to instruct a user on the expected input vs the width of the form field. I wanted to get feedback from folks more familiar with the codebase than me though, so I'm just nudging up the width as a safe first step/conversation starter.

Here's a screen shot of the new width

Screen Shot 2020-03-31 at 12 49 44 PM

### DIFF
--- a/app/scripts/modules/core/src/presentation/main.less
+++ b/app/scripts/modules/core/src/presentation/main.less
@@ -652,7 +652,7 @@ body > .container {
 
 .form-control {
   &.inline-number {
-    width: 60px;
+    width: 100px;
     display: inline-block;
   }
   &.inline-text {


erikmunson:
I dig it. I'm sure there are a few spots where we rely on this explicit width to lay things out, but it's probably not a big deal to mess with it or remove it altogether. A quick search says we only use that `inline-number` class in 14 files as of today anyways.

erikmunson:
Actually yes, upon further inspection we do use it in some places where totally removing an explicit width would have an impact, e.g.:

Screen Shot 2020-03-31 at 1 52 46 PM

erikmunson:
@gcomstock are you ready for me to merge this?

gcomstock:
> @gcomstock are you ready for me to merge this?

Yes! Thank you!

PR #8117, opened 2020-03-31
fix(deck): Evaluate Variables Stage: Update checkbox behavior so 'unchecked' saves correctly

justinrlee:
In the "Evaluate variables" stage, if you uncheck "Fail stage on failed expressions", rather than saving the field `failOnFailedExpressions` to false, it removes it.

So if you then exit the pipeline editor and re-open it, it shows up as checked again.

This PR changes the behavior so that when the checkbox is unchecked, it is set to `false`.

Old behavior:

Failed Expression checkbox before change

New behavior:

Failed Expression checkbox after change

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/failOnFailedExpressions/failOnFailedExpressions.directive.html
+++ b/app/scripts/modules/core/src/pipeline/config/stages/failOnFailedExpressions/failOnFailedExpressions.directive.html
@@ -3,7 +3,7 @@
     <div class="col-md-8 col-md-offset-1">
       <div class="checkbox pull-left">
         <label>
-          <input type="checkbox" ng-false-value="undefined" ng-model="$ctrl.stage.failOnFailedExpressions" />
+          <input type="checkbox" ng-false-value="false" ng-model="$ctrl.stage.failOnFailedExpressions" />
           <strong
             >Fail stage on failed
             <a href="http://spinnaker.github.io/guides/user/pipeline-expressions" target="_blank">expressions</a>


christopherthielen:
two things to note:
1) The pr title says that this is for Evaluate Variables stage.  However, this change will affect all stage types.  This may cause an unexpected side effect where most pipelines will appear "dirty".
2) This checkbox behavior may be symptomatic of a larger issue with the angular/react state synchronization in the new evaluate variables stage.  There may be other angularjs data mutations that are being clobbered by the react stage.  

I'd like if we could look at this behavior through this other lens and see if there's a less impactful fix we could make.

christopherthielen:
1) I verified that this does not cause the "dirty" behavior I was worried about
2) I verified that this behavior occurs because we apply the stage's `default` values each time a stage is selected.  It's not due to angularjs/react shenanigans that I was worried about.

So this fix seems fine by me!

justinrlee:
Hey @christopherthielen thanks for helping me check that!  I don't actually know if this is the best way to address this (I think this may actually be a result of https://github.com/spinnaker/deck/pull/5895, but the logic behind that PR makes sense to me).

I did notice that with this change, I see this behavior, which seems fine to me but I'm not sure what other behavior may result:
* If you have a regular stage (without the custom default), the field isn't populated by default
* If you check the box, the field gets populated with 'true'
* If you then uncheck the box, the field gets populated with 'false'

Is this problematic?  Based on your second comment I'm thinking this behavior is fine, right?  Thanks!

justinrlee:
Also, there's an error in the build, but I don't think that's cause by my PR.

christopherthielen:
>  Based on your second comment I'm thinking this behavior is fine, right? 

Yes, I'm OK with that behavior change because it will only affect people that explicitly toggle that field.  My concern was that it would "dirty" the form for users that didn't perform any actions other than viewing the stage config.

justinrlee:
hey @kevinawoo what does it take to get this cherry-picked into 1.19 and 1.18?

Thanks!

justinrlee:

justinrlee:

ezimanyi:
@Mergifyio backport release-1.18.x

PR #8118, opened 2020-04-01
feat(core/presentation): Expose status in useDataSource hook

alanmquach:
Without tapping into the status stream in the hook, we actually can't handle error or loading states.

This change basically just exposes the `loaded`, `loading`, `loadFailure` convenience properties of the `dataSource`. Tapping into the `status$` allows the hook to run whenever the status changes.

Example of how these are leveraged is in https://github.com/spinnaker/deck/pull/8119

### DIFF
--- a/app/scripts/modules/core/src/application/application.model.ts
+++ b/app/scripts/modules/core/src/application/application.model.ts
@@ -126,10 +126,12 @@ export class Application {
 
     const noneLeftToFetch = statuses.every(({ status }) => ['FETCHED', 'NOT_INITIALIZED'].includes(status));
     const lastFetch = statuses.reduce((latest, status) => Math.max(latest, status.lastRefresh || 0), 0);
+    const allLoaded = statuses.every(({ loaded }) => loaded);
 
     return {
       status: rolledUpStatus,
       lastRefresh: noneLeftToFetch ? lastFetch : this.lastRefresh,
+      loaded: allLoaded,
       data: undefined,
     };
   }
--- a/app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
+++ b/app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
@@ -3,12 +3,8 @@ import { useLatestCallback } from './useLatestCallback.hook';
 
 import { ApplicationDataSource, IFetchStatus } from '../../application';
 
-export interface IDataSourceResult<T> {
+export interface IDataSourceResult<T> extends IFetchStatus {
   data: T;
-  status: IFetchStatus;
-  loaded: boolean;
-  loading: boolean;
-  loadFailure: boolean;
   refresh: () => void;
 }
 
@@ -21,9 +17,7 @@ export interface IDataSourceResult<T> {
  * @returns IDataSourceResult<T>
  */
 export const useDataSource = <T>(dataSource: ApplicationDataSource<T>): IDataSourceResult<T> => {
-  const data = useObservableValue(dataSource.data$, dataSource.data$.value);
   const status = useObservableValue(dataSource.status$, dataSource.status$.value);
-  const { loaded, loading, loadFailure } = dataSource;
 
   // Memoize to give consumers a stable ref,
   // but don't return the promise that dataSource.refresh() returns
@@ -32,11 +26,7 @@ export const useDataSource = <T>(dataSource: ApplicationDataSource<T>): IDataSou
   });
 
   return {
-    data,
-    status,
-    loaded,
-    loading,
-    loadFailure,
+    ...status,
     refresh,
   };
 };
--- a/app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
+++ b/app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
@@ -1,10 +1,14 @@
 import { useObservableValue } from './useObservableValue.hook';
 import { useLatestCallback } from './useLatestCallback.hook';
 
-import { ApplicationDataSource } from '../../application';
+import { ApplicationDataSource, IFetchStatus } from '../../application';
 
 export interface IDataSourceResult<T> {
   data: T;
+  status: IFetchStatus;
+  loaded: boolean;
+  loading: boolean;
+  loadFailure: boolean;
   refresh: () => void;
 }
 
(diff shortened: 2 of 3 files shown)


app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
```diff
@@ -27,6 +33,10 @@ export const useDataSource = <T>(dataSource: ApplicationDataSource<T>): IDataSou
 
   return {
     data,
+    status,
```
erikmunson, line 36: AFAICT `status` will end up being a nested `IFetchStatus` object that has a duplicate copy of `data` which may be out of sync with the one sourced from `data$` depending on the two streams' activity. Can you lift anything you specifically want out of `IFetchStatus` into a top level field? I'd prefer to see this result object remain pretty flat anyway.

christopherthielen: +1 I think we could just destructure `status` here, i.e.: 
`return { ...status, refresh }`

In general I prefer to avoid flag fatigue by using enums (when flags are mutually exclusive, i.e., `loaded, loading, loadFailure`).

christopherthielen:
> Change dataUpdated() to emit a value from status$ with fresh data. I think the only case where status$ misses updates we need here is the dataUpdated() method. Since status$ allegedly contains whatever the latest data is contractually, in theory making it emit a fresh IFetchStatus with all the same values but updated data would be the "correct" behavior. I just don't know for certain whether there are callers that would meaningfully change their behavior as a result. If we did do that, I bet we could stop listening to data$ altogether and pull it out of status$ alone

+1 This seems legit.  Thanks for the analysis and suggestions!

app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
alanmquach: Didn't realize status was a nested object. Also +1 on just destructuring it.

> `loaded`, `loading`, `loadFailure`

Unfortunately they are not all mutually exclusive because `loaded` tracks if it as ever loaded:

https://github.com/spinnaker/deck/blob/b83fcf77fe6d876afd49756d4f6b38922688554a/app/scripts/modules/core/src/application/service/applicationDataSource.ts#L327-L329

I didn't like the flag fatigue either, but I also didn't want this logic ☝️ to just be duplicated everywhere. I'm happy to drop the latter flags, but I think we'll need to keep `loaded`.

christopherthielen: Gotcha.  
Maybe include `loaded` in `IFetchStatus`?  Maybe give it a less ambiguous name like `hasEverLoaded` or `ready` or something?

erikmunson: very 👍 to adding a totally new property to `IFetchStatus` _if we need to_ that's more clear. 

Only question I have before we go down that route: would checking for `lastRefresh > 0` resolve your concern about duplication of logic? Based on reading the code I _think_ `lastRefresh` remains set to `0` until the first fetch has successfully completed. That's similar to how we do 'has ever logic' with `ILatestPromiseResult` which has a monotonically incrementing `requestId`. If that's still going to result in too much spaghetti in this case then let's do a standalone flag.

alanmquach: I'd prefer not to invent a new way of making that calculation. It's even less obvious that checking `lastRefresh > 0` actually means "has ever loaded" and they are equivalent today but we can't guarantee that will hold true in the future (for example, if one decided to update lastRefresh even for errors 🤷‍♂ ).

I'd also rather keep the name `loaded` to be consistent with what it is and means on `dataSource.loaded`. I just don't want to introduce yet another new thing that actually means the same thing and cause someone to think

- what is the difference between `loaded` and `hasEverLoaded`??
- or worse: does `loaded` not actually mean "has ever loaded"? what does it mean then?

It doesn't sound like there are any strong objections, but please let me know if there are. Commit with updates coming shortly. And 🙏 thank you for the thoughtful reviews!

erikmunson: that seems extremely reasonable, I am happy to avoid folks getting confused trying to parse the data source API given it's already a little hard to wrap one's head around

app/scripts/modules/core/src/application/application.model.ts
```diff
@@ -126,10 +126,12 @@ export class Application {
 
     const noneLeftToFetch = statuses.every(({ status }) => ['FETCHED', 'NOT_INITIALIZED'].includes(status));
     const lastFetch = statuses.reduce((latest, status) => Math.max(latest, status.lastRefresh || 0), 0);
+    const allLoaded = statuses.every(({ loaded }) => loaded);
 
     return {
       status: rolledUpStatus,
       lastRefresh: noneLeftToFetch ? lastFetch : this.lastRefresh,
+      loaded: allLoaded,
```
erikmunson, line 134: is this related/required for the other data source changes? I don't fully understand its purpose right now

app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts
```diff
@@ -1,9 +1,9 @@
 import { useObservableValue } from './useObservableValue.hook';
 import { useLatestCallback } from './useLatestCallback.hook';
 
-import { ApplicationDataSource } from '../../application';
+import { ApplicationDataSource, IFetchStatus } from '../../application';
 
-export interface IDataSourceResult<T> {
+export interface IDataSourceResult<T> extends IFetchStatus {
   data: T;
```
erikmunson, line 7: if you feel like getting really fancy you could parameterize `IFetchStatus` with `T = any` so everybody (even outside hooks) gets this optional stronger type. I don't think there's a pressing / important reason to do it right now though so feel free to ignore.

app/scripts/modules/core/src/application/application.model.ts
alanmquach: This is just because we added `loaded` to `IFetchStatus` and this method was using `IFetchStatus` as its return type. Seemed better and simple enough to just give a loaded rollup instead of just doing `loaded: undefined` like what it's already doing for `data`.

erikmunson: AHA ok. i get it now

PR #8120, opened 2020-04-01
feat(core): Add svg name to dataSource

caseyhebebrand:
Given the svg icon loader (https://github.com/spinnaker/deck/pull/8056) and the upcoming Nav restructure, it is a good time to migrate the nav bar icons to svgs.

In order to do this, the `ApplicationDataSource` type now has an `iconSvg` attribute. Once this change propagates to all data sources in `deck` and `deck-kayenta`, a follow-up PR will update the `NavIcon` component to utilize `dataSource.iconSvg` instead of the current `dataSource.icon`, which points to the font awesome icon.

### DIFF
--- a/app/scripts/modules/core/src/application/service/applicationDataSource.ts
+++ b/app/scripts/modules/core/src/application/service/applicationDataSource.ts
@@ -3,7 +3,7 @@ import { $log, $q } from 'ngimport';
 import { BehaviorSubject, Observable, Subject } from 'rxjs';
 
 import { IEntityTags } from 'core/domain';
-import { robotToHuman } from 'core/presentation';
+import { robotToHuman, IconNames } from 'core/presentation';
 import { ReactInjector } from 'core/reactShims';
 import { FirewallLabels } from 'core/securityGroup';
 import { toIPromise } from 'core/utils';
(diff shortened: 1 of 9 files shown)


app/scripts/modules/core/src/application/service/applicationDataSource.ts
```diff
@@ -64,6 +64,11 @@ export interface IDataSourceConfig<T> {
    */
   icon?: string;
 
+  /**
+   * Represents the name of the svg to be used with the svg loader (Icon.tsx)
+   */
+  iconSvg?: string;
```
erikmunson, line 70: Instead of `string` here and in `navigationCategory`, let's use the `IconNames` type from `core/presentation` so we can strongly type this field going forward

caseyhebebrand: Great call out, I did not we had this type but I am a huge fan of strongly typed enums!

caseyhebebrand:
> Exciting! 🎨
> 
> I left a comment about typing that I'd like to see addressed, had one other thought that's just for consideration: we're going to (hopefully) live with this field for a very long time, and once `icon` is gone `iconSvg` may seem a bit over-specific in terms of exposing the implementation. We already have the concept of a 'name' in the `<Icon />` API, thoughts on calling this field `iconName` to match?

@erikmunson I actually had this same debate in my head as I was putting this PR together. I am happy to change it to `iconName`, especially since it will match the `IconName` type I am switching to.

erikmunson:
I had a feeling you had been considering lots of names 😄

PR #7138, opened 2019-06-21
feat(core): Enable new artifacts workflow in bakeManifest

Jammy-Louie:

PR #8121, opened 2020-04-01
fix(core): default and validate Bake (Manifest) produced artifacts

maggieneterval:
Closes https://github.com/spinnaker/spinnaker/issues/5624

* refactor(core): clean up Bake (Manifest) stage 

  Before resolving a long-standing regression, let's give the Bake (Manifest) code some TLC. This commit cleans up import statements per convention, abstracts the manifest renderer constants into an enum, refactors stage validation per new Formik conventions, and simplifies components that are unnecessarily class components into functional components.

* fix(core): default and validate Bake (Manifest) produced artifacts 

  [This PR](https://github.com/spinnaker/deck/pull/7138) unintentionally removed the functionality that added a default base64 expected artifact to Bake (Manifest) pipelines. Let's re-add that functionality, and add an additional validator to ensure that users catch this issue prior to Orca reporting the [error](https://github.com/spinnaker/orca/blob/master/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/manifests/CreateBakeManifestTask.java#L113).

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
@@ -1,74 +1,70 @@
 import React from 'react';
 
+import { SETTINGS } from 'core/config';
 import { IPipeline } from 'core/domain';
+import { ReactSelectInput } from 'core/presentation';
+
 import { StageConfigField } from '../common';
 import { IFormikStageConfigInjectedProps } from '../FormikStageConfig';
 import { BakeKustomizeConfigForm } from './kustomize/BakeKustomizeConfigForm';
 import { BakeHelmConfigForm } from './helm/BakeHelmConfigForm';
-import { ReactSelectInput } from 'core/presentation';
-import { SETTINGS } from 'core/config/settings';
+import { ManifestRenderers, HELM_RENDERERS } from './ManifestRenderers';
 
 interface IBakeManifestStageFormProps {
   updatePipeline: (pipeline: IPipeline) => void;
 }
 
-export class BakeManifestStageForm extends React.Component<
-  IBakeManifestStageFormProps & IFormikStageConfigInjectedProps
-> {
-  public HELM_RENDERERS = new Set(['HELM2', 'HELM3']);
-  public KUSTOMIZE_RENDERER = 'KUSTOMIZE';
+export function BakeManifestStageForm({
+  application,
+  formik,
+  pipeline,
+  updatePipeline,
+}: IBakeManifestStageFormProps & IFormikStageConfigInjectedProps) {
+  const stage = formik.values;
 
-  private templateRenderers = (): string[] => {
-    const renderers = [...this.HELM_RENDERERS];
+  const templateRenderers = React.useMemo(() => {
+    const renderers = [...HELM_RENDERERS];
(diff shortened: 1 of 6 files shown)


app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/ManualExecutionBakeManifest.tsx
```diff
@@ -29,7 +31,7 @@ export function ManualExecutionBakeManifest(props: ITriggerTemplateComponentProp
   React.useEffect(() => {
     if (overrideArtifact === false) {
       removeHelmArtifact();
-    } else {
+    } else if (defaultArtifact !== null) {
```
ezimanyi, line 34: Do we need to also add `defaultArtifact` as a dependency of the `useEffect`?  (Or is that only for things that are part of the react state whereas this is a local variable?)

app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
```diff
@@ -22,6 +24,16 @@ export function BakeManifestStageForm({
...
 
+  React.useEffect(() => {
+    if (!validateProducedArtifacts(stage.expectedArtifacts)) {
+      const defaultProducedArtifact = ExpectedArtifactService.createEmptyArtifact();
+      defaultProducedArtifact.matchArtifact.type = BASE_64_ARTIFACT_TYPE;
+      defaultProducedArtifact.matchArtifact.artifactAccount = BASE_64_ARTIFACT_ACCOUNT;
+      defaultProducedArtifact.matchArtifact.customKind = false;
+      formik.setFieldValue('expectedArtifacts', [defaultProducedArtifact]);
+    }
+  }, []);
```
ezimanyi, line 35: Am I understanding correctly that this empty array makes it run only on the initial render, but not on re-renders?

ezimanyi:
(The couple of comments are mostly just to make sure I'm understanding how `React.useEffect` works as I'd never seen it before!)

app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/BakeManifestStageForm.tsx
maggieneterval: It'll run only once, but _after_ the initial render (can roughly map to `componentDidMount` for class components).

app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/ManualExecutionBakeManifest.tsx
maggieneterval: This question made me start asking 5️⃣ ❓ about this component and ended up on a possibly regrettable journey; I'm going to push another commit that cleans it up a bit and will let you know when this is ready for another look.

(But yes, I think your instinct was right because the general advice is that for effects with explicitly specified dependencies, they should be exhaustive)

maggieneterval: Okay, pushed a cleanup commit for this manual execution component (and tested with both legacy/standard artifacts UI). Still think the overall experience of supplying artifacts to override is pretty confusing but am drawing the line here so this PR doesn't get out of control.

PR #8123, opened 2020-04-02
feat(core/managed): Show deploying state in environments view

vigneshm:
- Added a new `carrotRight.svg` file for the icon separating current build number and the deploying build number.
- Added an opacity-animation feature to `Pill` using css animations.

Screenshot of the feature
md-deploying

Here's a quick gif showing the animation
quick-recording-2


app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -57,20 +57,37 @@ const getResourceRoutingInfo = (
...
+  const currentPill = artifact?.versions.current ? (
+    <Pill text={currentBuild ? `#${currentBuild}` : currentVersion || artifact.versions.current || 'unknown'} />
+  ) : null;
+  const deployingPill = artifact?.versions.deploying ? (
+    <>
+      <Icon appearance="dark" name="caretRight" size="medium" />
+      <Pill
+        animateOpacity={true}
+        bgColor="#61b9ff"
+        text={deployingBuild ? `#${deployingBuild}` : deployingVersion || artifact.versions.deploying || 'unknown'}
```
erikmunson, line 76: since we won't show this component if `artifact?.versions.deploying` is false, do we need a fallback to `'unknown'` here?

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -57,20 +57,37 @@ const getResourceRoutingInfo = (
...
   const resourceName = getResourceName(resource);
   const routingInfo = getResourceRoutingInfo(resource) ?? { state: '', params: {} };
   const route = useSref(routingInfo.state, routingInfo.params);
 
+  const currentPill = artifact?.versions.current ? (
+    <Pill text={currentBuild ? `#${currentBuild}` : currentVersion || artifact.versions.current || 'unknown'} />
+  ) : null;
+  const deployingPill = artifact?.versions.deploying ? (
+    <>
+      <Icon appearance="dark" name="caretRight" size="medium" />
```
erikmunson, line 72: I like how this came out — glad you went with it over a CSS impl.

Given that this is a visually secondary element to other things like the resource icon and version pills, should we go for the `neutral` appearance instead of `dark`?

app/scripts/modules/core/src/managed/Pill.module.css
```diff
@@ -9,3 +9,19 @@
...
+  animation-iteration-count: infinite;
+  animation-name: animate-opacity;
+}
+
+@keyframes animate-opacity {
+  from {
+    opacity: 0.5;
+  }
+  to {
+    opacity: 1;
```
erikmunson, line 25: because this is an opacity adjustment across the entire subtree of the pill, am I correct in thinking the text also fades in/out? If so, can you think of a good way to have the background pulse here without fading the text?

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
vigneshm: Good call!

app/scripts/modules/core/src/managed/Pill.module.css
vigneshm: > because this is an opacity adjustment across the entire subtree of the pill, am I correct in thinking the text also fades in/out?

That is correct.

> If so, can you think of a good way to have the background pulse here without fading the text?

I might have to use JS based animation since the background opacity animation may not be possible with css animations and the background color are dynamically provided as props. Let me try the JS animation approach.

erikmunson: cc @gcomstock in case he has opinions or suggestions from having done this elsewhere — one other way is to animate the background color between two points vs. animating opacity.

erikmunson: another thing to consider: maybe you could just have the background be it's own sibling element alongside the text (i.e. behind it) instead of being applied to the containing element.

vigneshm: I ended up using `react-spring` to perform the animation and now only the [background color's opacity](https://github.com/spinnaker/deck/pull/8123/files#diff-8f60187ca9e9849993494ad0afd80689R71) is animated.

vigneshm: @gcomstock since we have js based animation available, we can potentially change this behavior to any form of animation you prefer. [This](https://codesandbox.io/s/quiet-framework-z4x0p) is the current behavior.

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -55,22 +56,50 @@ const getResourceRoutingInfo = (
...
+
+  const { opacity } = useSpring<{ opacity: number }>({
+    from: { opacity: 0.5 },
+    opacity: 1,
+    reset: true,
+    reverse: reverseAnimation,
+    onRest: () => setReverseAnimation(!reverseAnimation),
+  });
+
+  return <AnimatedPill {...props} bgColor={opacity.interpolate(o => `rgba(97, 185, 255, ${o})`)} />;
```
erikmunson, line 71: this is cool, and I am a fan of having `react-spring` around in our toolbox (especially for future work like animating heights/widths/other tough to animate things).

can you move all this logic around making pills animate-able inside `Pill` rather than making consuming code wire it together? don't worry too much about making the `Pill` API really clean and nice while you're adding it, we'll come back around after this big push to get Environments out and "graduate" this pill component into a central reusable place where it will get more API design attention.

erikmunson: oh also: can you please add some easing to this? the example you linked to was very mechanical in its pulsing due to lack of easing.

vigneshm: > can you move all this logic around making pills animate-able inside Pill rather than making consuming code wire it together? don't worry too much about making the Pill API really clean and nice while you're adding it, we'll come back around after this big push to get Environments out and "graduate" this pill component into a central reusable place where it will get more API design attention.

The challenge with current `Pill` implementation is that `bgColor` is configurable and we can't animate it's opacity unless we take in the r, g, b components separately as props (or convert hex to rgb).

How about the following plan of action ?

1/ I'll move the implementation of `DeployingPill` into the `Pill` module, but export that as a separate component with pre-defined color settings in this PR.
2/ In a separate PR, I can re-implement `Pill` into accepting a [`color`](https://www.npmjs.com/package/color-engine) object and then build the animation feature into `Pill` and will remove `DeployingPill`.

> can you please add some easing to this? the example you linked to was very mechanical in its pulsing due to lack of easing.

Do you like any specific version from [here](https://codesandbox.io/s/quiet-framework-z4x0p). I prefer `wobbly` and I can update to that if you prefer that.

vigneshm: I've completed 1 for now. Let me know if you prefer entire refactor (i.e 2).

app/scripts/modules/core/src/managed/Pill.module.css
gcomstock: Looks fun! I think gentle is fine. We can change our minds if needed as we introduce more animations elsewhere

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
erikmunson: yep that all seems exceedingly reasonable! fwiw I don't think it's urgent/high priority to do #2, if you like we can batch that change up with graduating it into `core/presentation`.

app/scripts/modules/core/src/managed/Pill.module.css
erikmunson: +1 to gentle, might also be worth experimenting with slowing it down by like 15% if that's easy to do (i didn't see a way in the codesandbox?)

btw: that sandbox is 🔥 🔥 🔥

vigneshm: Updated to `gentle` now.

@erikmunson I've updated codesandbox to accept custom values. You can play with `tension` and `friction` values to alter the animation. The values for the presets can be found [here](https://www.react-spring.io/docs/hooks/api)

erikmunson: wowowow this is so cool thanks for setting it up! I am really excited to start using these physics rules on other transitions 💥 

If you're open to it, I think it'd be nice to add some mass and overall sluggishness to the pulse — I personally gravitate toward pulsing animations that feel almost like a calm (but not too slow) breath in and out. After playing with that config maybe something like:
```
{
  mass: 16,
  tension: 390,
  friction: 130
}
```

It'd be really magic if down the road we could get a standard, well-liked "pulse" spring and re-use that across anything that needs to do some kind of animation like this.

vigneshm: I like this behavior, updating it!

vigneshm:
This is ready to be merged!

PR #8124, opened 2020-04-02
feat(core/managed): add support for overriding constraints via Environments

erikmunson:
Today is the day! This is a pretty rough first pass at interactively approving/rejecting manual judgements via the UI 🌈 

Screen Shot 2020-04-01 at 5 52 42 PM

I would show you a nice gif of actually doing a judgement, but sadly Netflix isn't running the version of Gate with the appropriate endpoint yet and I didn't want to go through the hassle of running it locally today. In the case of the endpoint failing we show a pretty lame (but hopefully soon to be improved!) error state with an option to go back and try again:

Screen Shot 2020-04-01 at 5 52 58 PM

There were some choices I made that are either temporary or require a little additional context, for those not already commented I'll add some PR comments.

### DIFF
--- a/app/scripts/modules/core/src/managed/NoticeCard.tsx
+++ b/app/scripts/modules/core/src/managed/NoticeCard.tsx
@@ -6,15 +6,16 @@ import { Icon, IconNames } from '../presentation';
 import styles from './NoticeCard.module.css';
 
 interface INoticeCardProps {
-  title: JSX.Element | string;
-  text: JSX.Element | string;
+  title: React.ReactNode;
+  actions?: React.ReactNode;
+  text?: React.ReactNode;
   icon: IconNames;
   noticeType: 'success' | 'neutral' | 'info' | 'error';
   isActive: boolean;
   className?: string;
 }
 
-export function NoticeCard({ title, text, icon, noticeType, isActive, className }: INoticeCardProps) {
+export function NoticeCard({ title, actions, text, icon, noticeType, isActive, className }: INoticeCardProps) {
   const NoticeCardClasses = classNames({
     [styles.NoticeCard]: true,
     [styles[noticeType]]: noticeType,
@@ -33,6 +34,7 @@ export function NoticeCard({ title, text, icon, noticeType, isActive, className 
         </div>
       )}
       {title && <div className={styles.title}>{title}</div>}
+      {actions && <div className={styles.actions}>{actions}</div>}
       {text && <div className={styles.text}>{text}</div>}
     </div>
   );
--- a/app/scripts/modules/core/src/managed/constraints/ConstraintCard.less
+++ b/app/scripts/modules/core/src/managed/constraints/ConstraintCard.less
@@ -1,0 +1,51 @@
+.ConstraintCard {
+  // TODO (erik): this is a bunch of custom styles for a button.
+  // really it should be a real button component we put somewhere central.
+  // for now it's entirely experimental and might totally go away
+  // or be replaced.
+
+  .constraint-override-action {
+    color: rgb(90, 95, 180);
+    background-color: transparent;
+    font-size: 16px;
+    border: 1px solid rgb(90, 95, 180);
+    border-radius: 2px;
+    padding: 4px 20px;
+    min-width: 120px;
+    transition: all 90ms ease-in;
+  }
+
+  .constraint-override-action:not([disabled]) {
+    cursor: pointer;
+  }
+
+  .constraint-override-action[disabled] {
+    cursor: not-allowed;
+    opacity: 0.66;
+  }
+
+  .constraint-override-action:not([disabled]):hover {
+    transform: scale(1.03);
+    background-color: rgba(90, 95, 180, 0.15);
+  }
+
+  .constraint-override-action:not([disabled]):focus,
+  .constraint-override-action:not([disabled]):active {
+    outline: none;
+    transform: scale(1.04);
+  }
+
+  .constraint-override-action:not([disabled]):focus {
+    box-shadow: 2px 2px 0px rgba(90, 95, 180, 0.66);
+  }
+
+  .constraint-override-action:not([disabled]):active {
+    box-shadow: none !important;
+    background-color: rgba(90, 95, 180, 0.66);
+    color: var(--color-white);
+  }
+
+  .action-error-message {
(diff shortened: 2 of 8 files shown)


app/scripts/modules/core/src/managed/NoticeCard.tsx
```diff
@@ -6,15 +6,16 @@ import { Icon, IconNames } from '../presentation';
 import styles from './NoticeCard.module.css';
 
 interface INoticeCardProps {
-  title: JSX.Element | string;
-  text: JSX.Element | string;
+  title: React.ReactNode;
+  actions?: React.ReactNode;
+  text?: React.ReactNode;
```
erikmunson, line 11: in an exciting twist, turns out `JSX.Element` expects _exactly one_ element (and it has to be an actual element), whereas `React.ReactNode` supports any of the things you can render using `{thingThatIsRenderable}` — including strings. TIL.

app/scripts/modules/core/src/managed/constraints/ConstraintCard.less
```diff
@@ -0,0 +1,51 @@
+.ConstraintCard {
+  // TODO (erik): this is a bunch of custom styles for a button.
+  // really it should be a real button component we put somewhere central.
+  // for now it's entirely experimental and might totally go away
+  // or be replaced.
+
+  .constraint-override-action {
```
erikmunson, line 7: pay no attention to this. I needed to invent a new-looking button but haven't actually built it out properly yet.

app/scripts/modules/core/src/managed/constraints/ConstraintCard.tsx
```diff
@@ -22,26 +27,93 @@ const logUnsupportedConstraintError = (type: string) => {
...
+    const dataSource: ApplicationDataSource<IManagedApplicationEnvironmentSummary> = application.getDataSource(
+      'environments',
+    );
+
+    // Here we wait to say things are fully done until we attempt to refresh, but don't
+    // reject. If the refresh fails things are going to be in a bad state anyway,
+    // so we don't want to wrongly imply that the override didn't take effect
+    // just because we're unable check.
+    return dataSource.refresh().catch(() => null);
+  });
```
erikmunson, line 47: I am open to feedback re: this approach of dereferencing the data source off the app model all the way down here in this child component. It seemed really awkward and mis-aligned to put all this request/state logic _entirely_ up in `ArtifactDetail`, but specifically calling data sources in here felt a little off. I was considering passing down the `refresh` function from `useDataSource` via props into this function, but it does also involve threading another prop all the way down the tree and into these callbacks. Thoughts?

app/scripts/modules/core/src/managed/constraints/constraintRegistry.tsx
```diff
@@ -61,5 +69,17 @@ export const statefulConstraintOptionsByType: { [type: string]: IStatefulConstra
...
+        {
+          title: 'Reject',
+          pass: false,
+        },
+        {
+          title: 'Approve',
+          pass: true,
+        },
+      ],
+    },
```
erikmunson, line 83: i did _not_ want to enable anyone sending up arbitrary statuses to persist on constraints — the whole point of `OVERRIDE_PASS` and `OVERRIDE_FAIL` are to be a way of understanding that a human or other thing external to the constraint's own evaluator decided to 'demand' the a particular status was applied.

app/scripts/modules/core/src/managed/constraints/constraintRegistry.tsx
```diff
@@ -13,13 +13,21 @@ const throwUnhandledStatusError = (status: string) => {
...
+export interface IConstraintOverrideAction {
+  title: string;
+  pass: boolean;
+}
+
 interface IStatefulConstraintConfig {
   iconName: IconNames;
-  shortSummary: (constraint: IStatefulConstraint) => JSX.Element | string;
+  shortSummary: (constraint: IStatefulConstraint) => React.ReactNode;
+  overrideActions: { [status in StatefulConstraintStatus]?: IConstraintOverrideAction[] };
```
erikmunson, line 24: @vigneshm for this new field I went with the map-of-statuses approach we discussed in a previous PR. I ended up really enjoying it and I think I'll probably go back and switch the `shortSummary` over to that API as well. Thanks for pointing it out!

app/scripts/modules/core/src/managed/constraints/ConstraintCard.tsx
```diff
@@ -22,26 +27,93 @@ const logUnsupportedConstraintError = (type: string) => {
...
   constraint: IStatefulConstraint;
   className?: string;
 }
 
-export const ConstraintCard = ({ constraint, className }: IConstraintCardProps) => {
+export const ConstraintCard = ({ application, environment, version, constraint, className }: IConstraintCardProps) => {
   const { type, status } = constraint;
 
+  const [actionPending, setActionPending] = useState(false);
+  const [actionFailed, setActionFailed] = useState(false);
```
vigneshm, line 61: It would be nice to have a single variable represent these two states. Something like the following

```js
const [actionStatus, setActionStatus] = useState<'not-initiated' | 'pending' | 'success' | 'error'>('not-initiated');
```

This would prevent accidentally setting an invalid state such as `actionPending=true` and `actionFailed=true`

If we need additional information for each state, such as error message for `error`, then we could extend it the following way and have typescript help us with refining.

```js
type ActionStatus = {
  status: 'not-initiated'
} | {
  status: 'pending'
} | {
  status: 'success'
} | {
  message: string,
  status: 'error'
};

```

app/scripts/modules/core/src/managed/constraints/ConstraintCard.tsx
vigneshm: One approach I've used in the past is the `useReducer` model and let a high-level parent component take control of such data needs. But that would involve quite a bit of re-write of this component I guess (like the pending status etc)?

app/scripts/modules/core/src/managed/constraints/ConstraintCard.tsx
erikmunson: [done!](https://github.com/spinnaker/deck/pull/8124/files#diff-05e51702972fa4560063447334bc15f9R61)

PR #8130, opened 2020-04-03
fix(amazon/loadbalancer): change default deregistration delay to 300

gcomstock:
Thought this was going to just be a help text edit; this is normally not something I'd touch. We had a Netflix user suggest that the default value should be 90-120 here. @alanmquach observed up the true amazon default value is 300, so I took a pass at editing that value where appropriate. Please review and discuss if this is something we should do. Happy to scrap it. Thanks! @vigneshm @caseyhebebrand


erikmunson:
If I don't get feedback or concerns from anyone by EOD today, I'm going to go ahead and merge this as it's an important quality of life improvement and had been hanging out for almost a month.

PR #8131, opened 2020-04-05
feat(core/managed): add stateless constraint support

erikmunson:
This adds support for the second and final variety of constraint — "stateless". At the moment I'm only adding config to support `depends-on` which requires that a version has _at some point_ been successfully deployed into another environment. Soon we'll add `allowed-times` (the declarative equivalent of execution windows` but there's some additional data work to be done there and for now it's not important enough to invest time in given our other priorities. I've left the types/interfaces in a place where adding a handful more types of stateless constraints should be fine, we'll likely need to come up with a different approach when we get more than that.

I also took the liberty of re-ordering the cards a bit in the interest of clarity after looking at what throwing in more constraints does to the list. The card for overall promotion status is at the top, stateful constraints next, and stateless constraints last. I'm liking the visual hierarchy/flow of that a lot more after playing with it, as we generally reverse chronological ordering on this list of environments and the new card order roughly reflects chronology — i.e. current overall status first, then the constraints in reverse chronological order of when they're expected to evaluate.

<details>
<summary>Before</summary>

Screen Shot 2020-04-05 at 10 01 28 AM

Screen Shot 2020-04-05 at 10 01 57 AM
</details>

<details>
<summary>After</summary>

Screen Shot 2020-04-05 at 9 49 12 AM

Screen Shot 2020-04-05 at 9 49 31 AM
</details>

Right now I'm using the wording `Already deployed to <environment>` and `Deployment to <environment> will be required before promotion` after much iteration/brain hurting. It's quite challenging to come up with clear and unambiguous language that doesn't make it sound like the cared is referring to the _current_ deployment status of the artifact in another environment, or some other weird/confusing relationship.

The other language I gave serious consideration was closer to the actual type of constraint (`depends-on`) and was roughly `Dependency on <environment> satisfied` and `Depends on successful deployment to <environment>`. The only reason I didn't go with that for now was because it felt a bit out of place and/or tough to understand if you've never seen the delivery config for an app before, but it's possible we'll end up flipping to that based on customer feedback.


app/scripts/modules/core/src/domain/IManagedEntity.ts
```diff
@@ -30,6 +30,15 @@ export interface IStatefulConstraint {
...
 
+export interface IDependsOnConstraint {
+  type: 'depends-on';
+  currentlyPassing: boolean;
+  attributes: { environment: string };
+}
+
+// more stateless types coming soon
+export type IStatelessConstraint = IDependsOnConstraint;
+
```
erikmunson, line 41: After working with this data, I'm pretty sure I think it's the wrong modeling for convenient/effective UI code usage. 

Having a totally different structure from stateful constraints means we have to jump through _tons_ of hoops for TS to be happy, do a bunch of weird casting, and constantly duplicate code paths for each type. I don't think that's been a huge deal on the backend because they're generally handled very differently, but in the UI their presentation to users is almost identical with extremely small exceptions so this pain is incredibly acute. You'll notice that in `constraintRegistry` I mostly tried to encapsulate the type differences into generic helpers that take either kind of constraint, and return whatever you need.

I'm going to see if we can move to something that looks structurally more like stateful constraints, perhaps with a `stateful` boolean flag we can use as a type discriminator to make handling a union of `IStatefulConstraint | IStatelessConstraint` productive to work with.

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -160,25 +179,43 @@ export const ArtifactDetail = ({
...
+                )}
+                {statelessConstraints && (
+                  <ConstraintCards
+                    constraints={statelessConstraints}
+                    application={application}
+                    environment={environmentName}
+                    version={version}
+                  />
+                )}
+                <div className="sp-margin-l-top">
```
erikmunson, line 198: nothing changed in this block, just whitespace from wrapping it for proper margins

app/scripts/modules/core/src/presentation/icons/vectors/mdConstraintDependsOn.svg
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 22.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="-1 0 24 24" style="enable-background:new 0 0 24 24;" xml:space="preserve">
```
erikmunson, line 4: @gcomstock I hand-modified the `viewBox` on this to nudge the paths over to the right by... 1/24th of the canvas I guess 🤷‍♂🕵 

While it was technically centered on its 24x24 canvas and sized properly to fill as much of the space as it could, _visually_ it looked way off-center because of the shape and visual weight of the paths. I couldn't figure out what was driving me crazy about it until I went in and started poking at the SVG to move it around, but doing this _really_ pulled the alignment back together in the list of cards. Here's a direct comparison of original/nudged, if you agree this is the right call can you adjust the original SVG to sit in this spot on the canvas so we don't have to worry about losing my `viewBox` hack with it gets modified in the future?

**Without nudge:**
without-nudge

**With nudge:**
with-nudge

app/scripts/modules/core/src/managed/constraints/constraintRegistry.tsx
```diff
@@ -83,3 +122,14 @@ export const statefulConstraintOptionsByType: { [type: string]: IStatefulConstra
     },
   },
 };
+
+const statelessConstraintOptionsByType: { [type: string]: IStatelessConstraintConfig } = {
+  'depends-on': {
+    iconName: 'mdConstraintDependsOn',
+    shortSummary: {
+      pass: ({ attributes: { environment } }) => `Already deployed to ${environment?.toUpperCase()}`,
```
lorin, line 130: I agree with your top-level comments that the wording here is challenging. I lean more towards

What do you think of "Successfully deployed to previous environment $environment..." instead of "Already deployed to $environment"? I think that would give more of a connotation that a constraint has been satisfied.

app/scripts/modules/core/src/presentation/icons/vectors/mdConstraintDependsOn.svg
gcomstock: Good call!

app/scripts/modules/core/src/managed/constraints/constraintRegistry.tsx
gcomstock: I like your suggestion Lorin

erikmunson: I updated this to `Already deployed to previous environment ${environment?.toUpperCase()}` since folks seem to appreciate the extra detail. I am keeping 'already' for now due to explicitly avoiding any statements like 'successfully deployed' or 'previously deployed' — when paired up with the existing cards about versions being currently/previously deployed on each environment, it becomes highly ambiguous and could be easily mistaken for a statement about the _current status_ of the environment being referenced. While there's likely still room to improve this, 'already' has the benefit of not implying present-time status.

PR #8132, opened 2020-04-05
feat(core/managed): use child route for artifact details on environments

erikmunson:
The list/detail view on environments was initially setup with a `useState` hook powering the detail view so we could start iterating fast without worrying about changing the structure if needed. Things have settled a bit and now it's time to use the real router. In addition to being the 'right' way routing wise, this also makes the URL work properly so you can deep link into a given version — e.g. `/applications/keel/environments/deb/keel/keel-0.442.0-h457.297f172`.


app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -37,7 +38,14 @@ export function Environments({ app }: IEnvironmentsProps) {
...
   const loadFailure = status === 'ERROR';
 
-  const [selectedArtifact, setSelectedArtifact] = useState<ISelectedArtifact>();
+  const {
+    stateService: { go },
+  } = useRouter();
+  const { params } = useCurrentStateAndParams();
+  const selectedArtifact = useMemo<ISelectedArtifact>(
+    () => (params.version ? pick(params, ['type', 'name', 'version']) : null),
+    [params.type, params.name, params.version],
```
erikmunson, line 47: doing all this memoized `pick`ing looks and seems kind of gross, but I haven't thought of a better way to do this where:
1. `selectedArtifact` maintains reference equality until one of type/name/version changes.
2. we only get the fields we want out of `params` rather than passing the whole thing all over the place.

vigneshm:
These changes look good to me.

I've noticed in other places that the 'detail view' is typically provided to the router and the parent takes care of providing a placeholder for this view to be injected by the router. I'm curious why this is handled directly within the parent component `Environment.tsx`.

erikmunson:
@vigneshm good q — there are likely some cases/aspects in which doing it via router injection is nice or provides necessary benefits (I can't speak to the nuances of why we chose to do it that way for the older details panes). I personally didn't see much of any benefit here in moving the detail state out of its own parent component, especially since (unlike in other details panes where cloud providers may inject their own content) I do not expect we'll end up ever injecting something ad-hoc or dynamic into this spot — it'll always be the exact same `ArtifactDetail` component.

Generally I find doing dynamic router-based injection to be harder to follow and reason about (much like any sort of dynamic injection can be), whereas using hooks / state to render the appropriate component tree is very straightforward and consistent with how we handle other types of state.

vigneshm:
> especially since (unlike in other details panes where cloud providers may inject their own content)

👍 That makes it clear.

> Generally I find doing dynamic router-based injection to be harder to follow and reason about (much like any sort of dynamic injection can be), whereas using hooks / state to render the appropriate component tree is very straightforward and consistent with how we handle other types of state.

I do agree with this opinion. As someone not familiar with the codebase, it is much easier for me to work through the codebase that has fewer 'dynamic magic' patterns.

PR #8142, opened 2020-04-08
feat(core/managed): add proper support for skipped, approved artifact states

erikmunson:
We recently came to the realization that the `approved` and `pending` states an artifact can have in a particular environment were a bit conceptually overloaded, because they both contained versions that had been skipped over due to later versions becoming available. This meant it was very challenging to explain what `approved` and `pending` meant in the UI because a version could _either_ be in a state of approved-ness / pending-ness, or it could've been skipped and have no chance of ever being deployed.

[We've now split out a totally new state](https://github.com/spinnaker/keel/pull/959) — `skipped` — to help clarify the semantics of the other two states, freeing up `approved` to mean "this version is about to start deploying" and `pending` to mean "this version is actively under consideration for deployment, but hasn't been chosen yet". This change adds full-fledged support to all three of these statuses, as the inherent confusion had prevented us from adding useful UI treatments until this point. Looks like this:

**Pending:**
Screen Shot 2020-04-07 at 7 58 05 PM

**Approved:**
Screen Shot 2020-04-07 at 7 41 52 PM

**Skipped (version info available):**
Screen Shot 2020-04-07 at 7 29 53 PM

**Skipped (no version info available):**
Screen Shot 2020-04-07 at 7 37 56 PM

Icons are coming soon (the `approved` one is also temporary), but wanted to get this out in advance of having iconography.

(cc @gcomstock)

### DIFF
--- a/app/scripts/modules/core/src/domain/IManagedEntity.ts
+++ b/app/scripts/modules/core/src/domain/IManagedEntity.ts
@@ -69,6 +69,7 @@ export interface IManagedEnviromentSummary {
       approved: string[];
       previous: string[];
       vetoed: string[];
+      skipped: string[];
     };
   }>;
 }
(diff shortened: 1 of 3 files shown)


luispollo:
❤️ it!

For another PR, I still think we could change "Already deployed to previous environment X" to "Dependency on environment X passes" or "Dependency on environment X satisfied", though. 😄

erikmunson:
I did [call that out as an alternative](https://github.com/spinnaker/deck/pull/8131) we very well might move to depending on how the overall experience is feeling, but for now I'm sensitive to keeping the voice and tone of the foundational/core constraint cards fairly consistent with each other. I'd likely rework the tone of some other cards along with it if we end up going that direction.

PR #8143, opened 2020-04-08
fix(core): Optimize merging of serverGroups from refresh

alanmquach:
It's not every day one gets a chance to optimize something from O(n^2) to O(n).

This method merges serverGroups from the remote response into the dataSource's local data. This is necessary to preserve referential equality thus preventing unnecessary re-renders of the entire clusters view.

When finding matches (or lack thereof) between the local and remote data, I just replaced the nested iteration of both local and remote with lookups in maps.

While this will cost (marginally) more space, for large datasets we'll benefit from huge savings in time.

And since time is money 💰, this is a very worthwhile tradeoff, despite the fact that space is ridiculously expensive in Silicon Valley's real estate market. 🏚


app/scripts/modules/core/src/cluster/cluster.service.ts
```diff
@@ -84,45 +84,43 @@ export class ClusterService {
...
-          ).length
-        ) {
+        const match = localMap[this.generateServerGroupLookupKey(serverGroup)];
+        if (!match) {
           data.push(serverGroup);
         }
       });
-      return data;
+      // filter out the removed server groups (nullified above)
+      application.serverGroups.data = data.filter(Boolean);
```
christopherthielen, line 122: doesn't this line undo the "update in place"?  Maybe I don't understand what references we need to retain.  In the previous version, the `data` array was `splice`'d so the array reference was the same.

alanmquach: Ah, you're right. Nice catch. Let me bring back the splice.

Since removals should be far less common than matches, I don't think it will be too costly.

alanmquach:
> Nice! I'm curious which component was the performance bottleneck before this fix.

@erikmunson heard that without referential equality, the entire clusters view (all server groups) would re-render, which is why the existing code was put in place. The bottleneck I noticed and fixed here was just that we spent a lot of time (nested iteration) computing it for large datasets (applications with many server groups).

erikmunson:
> > Nice! I'm curious which component was the performance bottleneck before this fix.
> 
> @erikmunson heard that without referential equality, the entire clusters view (all server groups) would re-render, which is why the existing code was put in place.

yeah fwiw I don't think _anybody_ on the team really knows for sure how these components really response to mutations, clones, refreshes, etc. in aggregate. There's some nested/bridged Angular components (which re-render on mutation in ways that make this update-in-place necessary), react components that do various forms of calling `forceUpdate` whenever that updated-in-place data _might_ have changed in some way (e.g. data source refresh or a great many other random things), and probably all kinds of other cases we're not aware of. Something that will hopetully get less unknown as we put a fresh face on the clusters route.

Also:
> While this will cost (marginally) more space, for large datasets we'll benefit from huge savings in time.

While it's not important enough to really worry about/measure in this case, I'd bet money on the increase in memory/GC pressure from this approach being so marginal as to barely even register on a profiler. It's basically however much heap it takes to store the lookup keys twice (one for each map). Truly an impressive find! 💥

PR #8144, opened 2020-04-08
feat(core/managed): use backend-supplied build + git metadata

erikmunson:
Frigga had a good life as a TypeScript library, but now it's time for retirement. This change switches over to using build/git/display name configuration sourced directly from the API, which happens to use a similar (but still better) version parsing strategy for now but will soon get more accurate/detailed.

In addition to making things less brittle and enabling build/commit parsing for most of the docker image artifacts at Netflix (we have some semi-standard tag name conventions), this also gave me an opportunity to generally refactor some things that were intertwined with our assumptions about using client side parsing — there were a ton of values/props getting passed around and transformed that just didn't make sense anymore, and new data flow was also required to enable grabbing `build`/`git`/`displayName` off of a version summary. Sorry that meant the diff ended up pretty large — I didn't see a reasonable way of pulling out multiple commits due to the sweeping data flow/structure changes.

The only thing that's _mostly_ atomic is the refactor to heavily condense/mature the code for the cards that describe a version's state in an environment, but I still ended up needing to do it in the same commit for silly reasons (the changes I made to use the new metadata required I split out a component). That said, you can still sort of review `VersionStateCard` and the related removal of code in `ArtifactDetail` as a mostly separate change.

Quick comparison of the improvements on docker images:

<details>
<summary>Before</summary>

Screen Shot 2020-04-08 at 11 37 26 AM

Screen Shot 2020-04-08 at 11 37 50 AM
</details>

<details>
<summary>After</summary>

Screen Shot 2020-04-08 at 11 37 36 AM

Screen Shot 2020-04-08 at 11 37 41 AM
</details>

Another small visual change to note: I've switched to having the build number pills in the version list/sidebar sit _centered_ when they don't fill their full space. The Titus demo app these screenshots are from revealed that when build numbers are different lengths on the list the left alignment looked very wonky. Keeping them centered with enough room to grow to ~5 digits means the likelihood of having horizontal alignment mis-match is now extremely low.


app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -2,32 +2,31 @@ import React from 'react';
...
+import { ISelectedArtifactVersion } from './Environments';
 import { Pill } from './Pill';
-import { parseName } from './Frigga';
 
 import styles from './ArtifactRow.module.css';
 
 interface IArtifactsListProps {
   artifacts: IManagedArtifactSummary[];
-  artifactSelected: (artifact: ISelectedArtifact) => void;
-  selectedArtifact: ISelectedArtifact;
```
erikmunson, line 14: with the increasing number of places/contexts we refer to things like 'artifact(s)' generally, it's starting to get very ambiguous. Part of the refactor that happened here was moving to be more precise by referring to a _version_, which is really what we're working with (vs. the overall 'artifact' which encompasses all the versions that exist)

app/scripts/modules/core/src/managed/displayNames.ts
```diff
@@ -0,0 +1,7 @@
+import { IManagedResourceSummary, IManagedArtifactVersion } from '../domain';
+
+export const getResourceName = ({ moniker: { app, stack, detail } }: IManagedResourceSummary) =>
+  [app, stack, detail].filter(Boolean).join('-');
+
+export const getArtifactVersionDisplayName = ({ displayName, build, git }: IManagedArtifactVersion) =>
+  build?.id ? `#${build?.id}` : git?.commit || displayName || 'unknown';
```
erikmunson, line 7: I don't know if I really care for this file's taxonomy/naming much, but we needed a good spot to share these functions and I don't think I'm quite ready to go touch `core/naming` yet 😬

app/scripts/modules/core/src/managed/VersionStateCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { NoticeCard } from './NoticeCard';
+import { Pill } from './Pill';
+
+interface CardTitleMetadata {
+  deployedAtMillis?: number;
+  replacedAtMillis?: number;
+  replacedByVersionName?: string;
+}
+
+const cardAppearanceByState = {
```
vigneshm, line 17: Love it!

PR #8150, opened 2020-04-09
feat(core/nav): Components for new navigation categories

caseyhebebrand:
We are moving toward a new UI layout. As part of v1, the navbar will be changing to a one-level vertical nav. This PR introduces some of the components that will make up the final navBar:
-  `navCategory`: This is the nav item that includes running counts, icon, and notifications
- `navRoute`: A ui-router wrapper over the component
- tests
- unused css which will be integrated later

In the future I will introduce the remaining components for the vertical nav, details drawer and minor changes to the filters. Once all the components are merged, I will incorporate them into the layout. 

Preview of vertical nav:
Screen Shot 2020-04-08 at 8 06 15 PM


app/scripts/modules/core/src/application/nav/NavCategory.tsx
```diff
@@ -0,0 +1,68 @@
...
+    tagsSub.unsubscribe();
+  };
+
+  React.useEffect(() => {
+    setRunningCount(badgeSourceData);
+    setTags(alerts || []);
+    clearSubscriptions();
+    configureSubscriptions();
+    return clearSubscriptions;
+  }, [alerts, badgeSourceData]);
```
erikmunson, line 44: There's a lot of manual/imperative subscription and state management going on in this component, which I think we might be able to reduce. If I'm reading it right (please tell me if I'm not!) it appears the intent is roughly:
- If a `badge` is specified, use whatever the current data in the `badge`  data source is, and then subscribe to future updates
- Use whatever the current data in the `category` data source is to lookup `alerts`, and then subscribe to future updates on that `alerts` value

Based on that, I think you could likely make very good use of [the new `useDataSource` and `useObservable` hooks](https://github.com/spinnaker/deck/pull/8040) which handle the vast majority of the subscription + state management and let you mostly write stateless code in your actual component. I recommend playing with them and seeing if they work for you. e.g. for the badge one:

```tsx
// The default value is handled for you (it's whatever the current value of the data source is)
const { data: badgeData } = useDataSource(app.getDataSource(badge));
const runningCount = badgeData?.length || 0;
```

The only catch is I'm not 100% sure the hooks can handle passing an undefined/null value which you'd be doing here for `badge`, so might need to modify the hooks to check for nullish values and do nothing in those cases. If you run into that problem I'm happy to help get the hooks fixed.

Because `alerts` isn't part of the data source's status object, right now you'd probably still need a `useState` for that e.g.:

```tsx
const [alerts, setAlerts] = useState(category.alerts);
useObservable(category.status$, () => {
  setAlerts(category.alerts);
});
```

app/scripts/modules/core/src/application/nav/NavRoute.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { UIRouterContext } from '@uirouter/react-hybrid';
+import { UISref, UISrefActive } from '@uirouter/react';
+
+import { NavCategory } from './NavCategory';
+import { ApplicationDataSource } from '../service/applicationDataSource';
+import { Application } from '../application.model';
```
erikmunson, line 7:
```suggestion
import { Application } from '../application';
```

app/scripts/modules/core/src/application/nav/NavRoute.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export interface INavRouteProps {
+  category: ApplicationDataSource;
+  isActive: boolean;
+  app: Application;
+}
+
+@UIRouterContext
+export class NavRoute extends React.Component<INavRouteProps> {
```
erikmunson, line 16: If you're interested, it's possible to turn this into a functional component by [rendering a `<UIRouterContextComponent>` element](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/modal/showModal.tsx#L84-L86) at the root (vs. using the `@UIRouterContext` decorator)

package.json
```diff
@@ -100,7 +100,7 @@
   },
   "devDependencies": {
     "@spinnaker/eslint-plugin": "1.0.7",
-    "@spinnaker/mocks": "1.0.2",
+    "@spinnaker/mocks": "1.0.3",
```
erikmunson, line 103: in the future, might be cleaner in the commit history to have this package/lockfile bump in a separate PR/commit?

caseyhebebrand: yes I am going to make a separate PR and rebase. I did this temporarily so I wouldn't be blocked, then got side-tracked haha.

app/scripts/modules/core/src/application/nav/NavRoute.tsx
caseyhebebrand: Very interested thanks! I appreciate the context with existing helper components in this PR.

app/scripts/modules/core/src/application/nav/NavCategory.tsx
caseyhebebrand: Yes! Thank you, this is exactly what this needs. To be honest, migrating `DataSourceEntry` to this functional component already reduced a lot of the complexity so I am excited to clean it up a bit more.

app/scripts/modules/core/src/application/nav/NavCategory.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Application } from '../application.model';
+import { IEntityTags } from 'core/domain';
+
+export interface INavCategoryProps {
+  category: ApplicationDataSource;
+  isActive: boolean;
+  app: Application;
+}
+
+export const NavCategory: React.FC<INavCategoryProps> = ({ app, category, isActive }: INavCategoryProps) => {
```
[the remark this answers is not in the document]
caseyhebebrand: No it is not required but I have always been in favor of stricter typing and types for functional components are usually implied (whereas for class components the props and state are clear).

app/scripts/modules/core/src/application/nav/NavCategory.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+
+import { DataSourceNotifications } from 'core/entityTag/notifications/DataSourceNotifications';
+import { Icon } from '../../presentation';
+
+import { ApplicationDataSource } from '../service/applicationDataSource';
+import { Application } from '../application.model';
+import { IEntityTags } from 'core/domain';
```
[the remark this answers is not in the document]
caseyhebebrand: @erikmunson Sounds good. This is still a WIP so I will definitely absorb your import comments into my code cleanup.

app/scripts/modules/core/src/application/nav/NavCategory.spec.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const wrapper = mount(<NavCategory app={app} category={category} isActive={false} />);
+    const nodes = wrapper.children();
+    expect(nodes.find('.badge-running-count').length).toBe(0);
+    expect(nodes.find('.badge-none').length).toBe(1);
+
+    const text = nodes.childAt(0).getDOMNode();
+    expect(text.textContent).toBe('');
+  });
+
+  it('subscribes to runningCount updates', () => {
```
christopherthielen, line 71: What is this testing?  Shouldn't it be asserting something related to runningCount?

app/scripts/modules/core/src/application/nav/NavCategory.tsx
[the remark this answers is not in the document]
christopherthielen: Here's my take:
I'm in favor of stronger typing for FCs.  Yet, I don't want to introduce yet another code style.    We should strive to be consistent.  `<NavRoute />` in this very PR is typed differently.  If we can agree that using `FC` is preferred, I'd love to see a linter rule written for it (with a fixer) to consolidate code style.  
  
I'd also suggest omitting the second/duplicate type declaration:

```suggestion
export const NavCategory: React.FC<INavCategoryProps> = ({ app, category, isActive }) => {
```

app/scripts/modules/core/src/application/nav/NavCategory.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const NavCategory: React.FC<INavCategoryProps> = ({ app, category, isActive }: INavCategoryProps) => {
+  const { alerts, badge, iconName, key, label, status$ } = category;
+
+  const { data: badgeData } = useDataSource(app.getDataSource(badge || key));
+  const runningCount = badge ? badgeData.length : 0;
+
+  const [tags, setTags] = React.useState<IEntityTags[]>(alerts || []);
+  useObservable(status$, () => {
+    setTags(category.alerts || []);
+  });
```
christopherthielen, line 25: Curious if we can use the more specific hook?
```suggestion
  const categoryData = useDataSource(category);
  const tags: IEntityTags[] = categoryData?.alerts || [];
```

app/scripts/modules/core/src/application/nav/NavRoute.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const NavRoute = ({ app, category, isActive }: INavRouteProps) => (
+  <UIRouterContextComponent key={category.key}>
+    <UISref className={isActive ? 'active' : ''} to={category.sref}>
+      <a>
+        <NavCategory app={app} category={category} isActive={isActive} />
+      </a>
+    </UISref>
+  </UIRouterContextComponent>
+);
```
christopherthielen, line 23: ICYMI, there are now uirouter/react hooks that are (IMHO) slightly less awkward than `UISref`:
```suggestion
export const NavRoute = ({ app, category, isActive }: INavRouteProps) => {
  const sref = useSrefActive(category.sref, null, 'active');
  return (
    <UIRouterContextComponent key={category.key}>
      <a {...sref}>
        <NavCategory app={app} category={category} isActive={isActive} />
      </a>
    </UIRouterContextComponent>
  );
};
```

app/scripts/modules/core/src/application/nav/NavRoute.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { Application } from '../../application';
+
+export interface INavRouteProps {
+  category: ApplicationDataSource;
+  isActive: boolean;
+  app: Application;
+}
+
+export const NavRoute = ({ app, category, isActive }: INavRouteProps) => (
+  <UIRouterContextComponent key={category.key}>
```
christopherthielen, line 16: The `<UIRouterContextComponent/>` is only necessary if the thing rendering this component is AngularJS (typically because the React component has been wrapped using `react2angular`).  If the parent components are all React and rendered inside a React `UIView`, then this kludge isn't necessary.  I'm not sure where this is going to be used, but I wanted to point that out.

christopherthielen, line 16: `key` is not necessary here, right?

app/scripts/modules/core/src/application/nav/NavCategory.spec.tsx
caseyhebebrand: Yes, I forgot to push the update to this test. My bad. This is a an error on my part. It is meant to to push a running execution to the application and monitor the change in the runningCount.

app/scripts/modules/core/src/application/nav/NavCategory.tsx
caseyhebebrand: I did this because `alerts` isn't part of the data source's status object, so I did not think that `useDataSource` would subscribe correctly. I will play around though.

app/scripts/modules/core/src/application/nav/NavRoute.tsx
caseyhebebrand: Ok I will remove

app/scripts/modules/core/src/application/nav/NavRoute.tsx
caseyhebebrand: Sweet, I had no idea. I will use this and remove the `UIRouterContextComponent`

app/scripts/modules/core/src/application/nav/NavRoute.tsx
caseyhebebrand: Yes I was clearing a React warning since multiple `NavRoutes` are rendered.

app/scripts/modules/core/src/application/nav/NavCategory.tsx
christopherthielen: oh I see!

app/scripts/modules/core/src/application/nav/NavRoute.tsx
christopherthielen: the `key` should be added to the `NavRoute` when mapping over them `array.map(category => <NavRoute key={category.key} />)`

app/scripts/modules/core/src/application/nav/NavCategory.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export const NavCategory = ({ app, category, isActive }: INavCategoryProps) => {
+  const { alerts, badge, iconName, key, label } = category;
+
+  const { data: badgeData } = useDataSource(app.getDataSource(badge || key));
+  const runningCount = badge ? badgeData.length : 0;
+
+  // useDataSource is enough to update alerts when needed
+  useDataSource(category);
```
alanmquach, line 23: I understand why this works but wonder if it's worth making this more clear/explicit, i.e. pulling alerts off of the result of useDataSource instead of relying knowing (based on the comment) that this is what causes a re-render when that thing over there changes.

caseyhebebrand: Yes I get that. I tried a few options before ending on this because the behavior I observed was a bit strange. The result of use `dataSource` updates the `entityTags`.  When I pulled `alerts` directly off of the result of `useDataSource` the alerts would not update, and I would instead have to map/filter `entityTags` to alerts. I ended up here because it was simplest to cause the re-render. 

I can either go back to this approach with the entityTags, or make a more detailed comment to explain the choice. What do you think?

alanmquach: If it didn't work I'm fine with it as is.

Separately, I would like us to think about what usage of the `useDataSource` hook looks like longer term though, wrt things like `alerts` that are exposed on the `ApplicationDataSource` itself but not through the hook.

Just feels awkward to be able to use the hook unless you need any of these other things, then still use the hook, but still pull it off the datasource directly.

christopherthielen: +1 to what @alanmquach said wrt revisiting this pattern

christopherthielen:
looks great!

app/scripts/modules/core/src/application/nav/NavCategory.tsx
caseyhebebrand: Sounds good. Its more of a `subscribeToDataSource` rather than a `useDataSource`

PR #8153, opened 2020-04-09
fix(core/presentation): Avoid excessive RadioButton remounts by moving the component outside the render function

christopherthielen:
Previously, we were rendering components defined inside the render function.  Since they get a new reference each time, the buttons were getting unmounted/remounted.  This also caused the radio buttons to be unfocusable, so keyboard navigation was impossible.

Now, the radiobutton component is defined as a separate component at the top level.

before: (pressing tab)

before

after: (pressing tab and selecting with spacebar)

after

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/inputs/RadioButtonInput.tsx
+++ b/app/scripts/modules/core/src/presentation/forms/inputs/RadioButtonInput.tsx
@@ -21,36 +21,34 @@ interface IRadioButtonOptions extends Option {
 }
 
 export const RadioButtonInput = (props: IRadioButtonInputProps) => {
-  const { inline, value: selectedValue, validation, inputClassName, options, stringOptions, ...otherProps } = props;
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/presentation/forms/inputs/RadioButtonInput.tsx
```diff
@@ -21,36 +21,34 @@ interface IRadioButtonOptions extends Option {
...
-  );
-
-  const VerticalRadioButtons = ({ opts }: { opts: IRadioButtonOptions[] }) => (
-    <div className={`${elementClassName} vertical left`}>
-      {opts.map(option => (
-        <RadioButton key={option.label} option={option} />
-      ))}
-    </div>
-  );
+  const verticalClassName = inline ? '' : 'vertical left';
```
caseyhebebrand, line 31: Nice way to remove a few lines of code with one className

erikmunson:
nice

PR #8158, opened 2020-04-11
feat(core/managed): use StatusCard for artifact cards, add enter transition

erikmunson:
**Depends on https://github.com/spinnaker/deck/pull/8157**

This PR does three things:

1. Temporarily retire/remove the summary card at the top of the Environments overview component. While we do still believe this card will come back in a different form and be useful, at the moment its information is not that helpful and the 'success' visual treatment makes its relationship with green artifact statuses confusing (it's not related). So basically this card goes away:
Screen Shot 2020-04-10 at 5 47 46 PM

2. Use `StatusCard` for all cards in the Environments view and retire `NoticeCard`. This builds on the work done in https://github.com/spinnaker/deck/pull/8157 and the screenshots from that PR should mostly cover what it looks like

3. Add a "trailing" enter transition for status cards that calls attention to them when opening a version. The effect causes the list of cards to fade/scale in one after the other with and overlapping timeline, so your eye is drawn up the list of cards and lands on the one summarizing the overall state (e.g. 'Deployed here since an hour ago'). While not _strictly_ necessary for usability, I found it helps a lot in building an intuitive sense of the information. I can't embed videos in GitHub, but [take a look at this video of the transition to get a sense of it](https://slack-files.com/T091CRSGH-F0128JEBE4Q-3d0e17929e).


erikmunson:
fyi @gcomstock

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -26,28 +21,95 @@ function shouldDisplayResource(name: string, type: string, resource: IManagedRes
...
+  leave: outStyles,
+  trail: 40,
+  config: { mass: 1, tension: 600, friction: 40 },
+} as UseTransitionProps<JSX.Element, React.CSSProperties>;
+
+type IEnvironmentCardsProps = Pick<IArtifactDetailProps, 'application' | 'version' | 'allVersions'> & {
+  environment: IManagedArtifactSummary['versions'][0]['environments'][0];
+};
+
+const EnvironmentCards = memo(
```
erikmunson, line 56: This whole block of logic w/ `EnvironmentCards` and related stuff is kind of gross and messy, though certainly more organized that it was when I first wrote it. I'm leaving it as-is for now because I want to confirm that this transition is something useful we want to keep. If we decided it is, I'm going to extract a `<CardList />` component that accepts an array of JSX elements and whose _only_ job is to add list margin + do enter/leave transitions. IMO that will clean this code up immensely and likely eliminate the need for this component (or at least heavily simplify it)

PR #8159, opened 2020-04-11
feat(core/managed): add scrolling grid layout, detail pane transition

erikmunson:
A couple _very_ related changes:

**First,** the overall layout for the two-column Environments view is a mix of some prototype flexbox markup/styles, some ad-hoc 'make it work' layout code from porting over the prototype, and then some layering on top of stuff we did to accommodate a particular feature/component. I'd like to simplify that markup, prepare it to become something more re-usable (i.e. some kind of standard list/detail layout we can use elsewhere), and fix the sizing/scroll overflow wonkiness we've deferred polishing. As part of this first chunk of changes I've:

- Swapped out the flexbox rules which implicitly depended on each other's size in favor of a two-column grid layout, which has the pleasant side effect of filling the available space in the route container
- Tweaked the size relationship between the list and detail panes to give more breathing room to the list pane when there's sufficient horizontal space
- Made each column handle overflow properly so you can independently scroll
- Made all the headers sticky so the new scroll behavior doesn't allow them to scroll away
- Fixed the position and stickiness issues with `EnvironmentRow` and `ObjectRow` to fix the wonkiness that was happening on scroll (I intend to add back proper stickiness later)

**Second,** I've added enter/leave transitions for swapping between the overview pane and the detail pane for a specific version. After getting early feedback we decided this should be upgraded from nice-to-have to must-have, as the lack of a context-setting transition made it very unclear what was going on when you clicked on a version. For now it's done in a sort of annoying way (I'll comment on the lines I'm talking about) because I want to experiment with the transition and confirm that folks are feeling good about it, but my plan is to extract a nice layout component like `<ListDetailLayout />` that will do all the transition stuff internally, and expose a render prop for rendering the detail pane so we can avoid the stuff I had to do here.

While this PR (amazingly!) doesn't depend on https://github.com/spinnaker/deck/pull/8158, the usage examples are from locally running both https://github.com/spinnaker/deck/pull/8158 _and_ this change together. I chose to do that because https://github.com/spinnaker/deck/pull/8158 has related visual and motion changes, and their interplay is useful in evaluating these changes.

<details>
<summary>Before</summary>

Screen Shot 2020-04-10 at 6 39 45 PM

Screen Shot 2020-04-10 at 6 39 58 PM

</details>

<details>
<summary>After</summary>

Screen Shot 2020-04-10 at 6 34 46 PM

Screen Shot 2020-04-10 at 6 35 09 PM

[Video of properly scrolling columns and detail pane transitions](https://slack-files.com/T091CRSGH-F011N2QGKPD-0f21d9a110). I recommend making it fullscreen so you can see the smaller components.

</details>

### DIFF
--- a/app/scripts/modules/core/src/managed/Environments.module.css
+++ b/app/scripts/modules/core/src/managed/Environments.module.css
@@ -27,16 +27,27 @@
 }
 
 .mainContent {
-  display: flex;
-  margin-top: 24px;
+  display: grid;
+  grid-template-columns: minmax(300px, 3fr) 8fr;
+  grid-template-rows: auto;
+  width: 100%;
+  overflow: hidden;
 }
 
 .artifactsColumn {
-  flex: 0 1 280px;
-  overflow: hidden;
-  margin-right: 16px;
+  padding-right: 16px;
+  overflow: scroll;
 }
 
 .environmentsColumn {
-  flex: 1 1 400px;
+  display: grid;
+  grid-column-template: 1fr;
+  grid-row-template: auto;
+  overflow: scroll;
+}
+
+.environmentsPane {
+  grid-column: 1;
+  grid-row: 1;
+  overflow: scroll;
 }
--- a/app/scripts/modules/core/src/managed/Environments.tsx
+++ b/app/scripts/modules/core/src/managed/Environments.tsx
@@ -1,6 +1,7 @@
-import React, { useState, useMemo } from 'react';
+import React, { useMemo } from 'react';
 import { pick, isEqual, keyBy } from 'lodash';
 import { useCurrentStateAndParams, useRouter } from '@uirouter/react';
+import { useTransition, animated } from 'react-spring';
 
 import { SETTINGS } from 'core/config/settings';
 import { Spinner } from 'core/widgets';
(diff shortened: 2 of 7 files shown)


app/scripts/modules/core/src/managed/Environments.module.css
```diff
@@ -27,16 +27,27 @@
 }
 
 .mainContent {
-  display: flex;
-  margin-top: 24px;
+  display: grid;
+  grid-template-columns: minmax(300px, 3fr) 8fr;
```
erikmunson, line 31: if you're not familiar with grid, tld;r the sidebar has a minimum width of 300px, but if given the space will attempt to take 3 "fractional units" of the overall width. The detail column will always try to take 8 fractional units (so at larger widths the sidebar is 3/8ths of the detail column) but will go smaller once the sidebar hits the minimum width.

app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -99,46 +140,46 @@ export function Environments({ app }: IEnvironmentsProps) {
...
-              />
-            )}
-          </div>
-        </div>
+              </animated.div>
+            ),
+        )}
+        {detailPaneTransition.map(
+          ({ item, key, props }) =>
+            item.selectedVersion && (
```
erikmunson, line 169: passing _all_ the data we need for rendering `ArtifactDetail` is nasty, and though it works I don't like it. The reason we need it is because referencing the usual `selectedVersion` / etc. would mean that the moment you hit close on the detail pane `selectedVersion` would become `null` and `ArtifactDetail` would have no data to use for rendering during its leave transition. That's fine for experimental purposes but should change for longer term use.

If we decide this transition is useful and worth keeping, I'll be encapsulating both the scrollable two column layout _and_ all this detail pane transition code inside a `<ListDetailLayout />` component that will let `ArtifactDetail` render itself via a render prop. That will mean we can use _older_ render prop callbacks to retain access to the previously selected version data without needing to pass things around through transition hooks. psuedo-example:

```tsx
// Because selectedVersion etc. are held inside the render prop's closure,
// we can hold onto its reference and use it to re-render `ArtifactDetail`
// with old props during the leave transition.
<ListDetailLayout
  renderListPane={() => { /* ... */ }}
  renderDetailPane={() =>
    <ArtifactDetail
      application={app}
      name={selectedVersion.name}
      type={selectedVersion.type}
      version={selectedVersionDetails}
      allVersions={selectedArtifactDetails.versions}
      resourcesByEnvironment={resourcesByEnvironment}
      onRequestClose={() => go('^')}
    />
  }
>
  {/* render initial/overview pane as a child? */}
</ListDetailLayout>
```

app/scripts/modules/core/src/managed/ObjectRow.tsx
```diff
@@ -30,8 +30,5 @@ export const ObjectRow = ({ icon, title, metadata, depth = 1 }: IObjectRowProps)
 const getStylesFromDepth = (depth: number): React.CSSProperties => {
   return {
     marginLeft: 16 * depth,
-    position: 'sticky',
-    top: 104 + 40 * depth,
-    zIndex: 500 - depth,
```
erikmunson, line 35: some of this (in some form) will come back, but for now I'm prioritizing other things.

erikmunson:
fyi @gcomstock

app/scripts/modules/core/src/managed/Environments.module.css
vigneshm: Thanks for the explanation!

PR #8162, opened 2020-04-13
feat(core/nav): Add sections for vertical nav

caseyhebebrand:
Dependent on https://github.com/spinnaker/deck/pull/8150.

A simple component that takes a list of navigation categories and maps them to the `NavCategory`. The list of categories are formatted into a component.

Screen Shot 2020-04-13 at 5 17 05 PM

### DIFF
--- a/app/scripts/modules/core/src/application/nav/NavSection.spec.tsx
+++ b/app/scripts/modules/core/src/application/nav/NavSection.spec.tsx
@@ -6,26 +6,27 @@ import {
   mockPipelineDataSourceConfig,
   mockServerGroupDataSourceConfig,
 } from '@spinnaker/mocks';
-import { Application } from '../../application';
+import { ApplicationModelBuilder } from '../../application';
 import { NavSection } from './NavSection';
 
 describe('NavCategory', () => {
   it('should render multiple categories', () => {
-    const app = new Application('testapp', null, [
+    const app = ApplicationModelBuilder.createApplicationForTests(
+      'testapp',
       mockPipelineDataSourceConfig,
       mockLoadBalancerDataSourceConfig,
       mockServerGroupDataSourceConfig,
-    ]);
+    );
 
-    const wrapper = shallow(<NavSection app={app} categories={app.dataSources} activeCategoryName={'Pipelines'} />);
+    const wrapper = shallow(<NavSection app={app} categories={app.dataSources} activeCategoryName="Pipelines" />);
     const nodes = wrapper.children();
     expect(nodes.length).toEqual(3);
   });
 
   it('should not render if no categories', () => {
-    const app = new Application('testapp', null, []);
+    const app = ApplicationModelBuilder.createApplicationForTests('testapp');
 
-    const wrapper = shallow(<NavSection app={app} categories={[]} activeCategoryName={''} />);
+    const wrapper = shallow(<NavSection app={app} categories={[]} activeCategoryName="" />);
     const nodes = wrapper.children();
     expect(nodes.length).toEqual(0);
   });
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/application/nav/NavSection.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+describe('NavCategory', () => {
+  it('should render multiple categories', () => {
+    const app = new Application('testapp', null, [
+      mockPipelineDataSourceConfig,
+      mockLoadBalancerDataSourceConfig,
+      mockServerGroupDataSourceConfig,
+    ]);
+
+    const wrapper = shallow(<NavSection app={app} categories={app.dataSources} activeCategoryName={'Pipelines'} />);
```
erikmunson, line 20:
```suggestion
    const wrapper = shallow(<NavSection app={app} categories={app.dataSources} activeCategoryName="Pipelines" />);
```

app/scripts/modules/core/src/application/nav/verticalNav.less
```diff
@@ -0,0 +1,71 @@
+.nav-section {
```
erikmunson, line 1: not a must-do suggestion — and not even something you have to adhere to generally — but I usually have a preference for leveraging our [spacing](https://github.com/spinnaker/styleguide/blob/master/src/styles/spacing.css) and [flex layout](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/flex-layout.less) utility classes in JSX vs. encoding margin, padding, and flex rules into stylesheets.

In my opinion (again not authoritative) working with layout and spacing within the component markup is much simpler to reason about than trying to mentally map layout rules to JSX elements based on their classnames (especially when you're not familiar with a component's code or the class names are not self-evident). If you're curious here's a [recent example](https://github.com/spinnaker/deck/pull/8157/files#diff-b42fb40803c15e767e7fe6881f22f1c1R18-R36) of what their usage looks like.

Mostly just bringing up for awareness, I have no strong feelings about this PR's use of layout rules in particular.

app/scripts/modules/core/src/application/nav/NavSection.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+  mockLoadBalancerDataSourceConfig,
+  mockPipelineDataSourceConfig,
+  mockServerGroupDataSourceConfig,
+} from '@spinnaker/mocks';
+import { Application } from '../../application';
+import { NavSection } from './NavSection';
+
+describe('NavCategory', () => {
+  it('should render multiple categories', () => {
+    const app = new Application('testapp', null, [
```
erikmunson, line 14: I think we usually use [this builder method](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/application/applicationModel.builder.ts#L8-L11) in tests? I haven't used it personally before so don't take my word for it.

caseyhebebrand: Yes we do. It is just a simple wrapper around this function but I can swap it out.

app/scripts/modules/core/src/application/nav/verticalNav.less
caseyhebebrand: That is fair. I have a different opinion coming from my old best practices where CSS is more closely tied to each component. I prefer it to be all consolidated in one place, and reused styles can be inherited from the parent class. With the spacing/layouts, components require multiple classes, many with only 1 attribute. This means the style is scattered about making it a hunt to modify styling and determine inheritance. 

That being said, since most of the team is used to the pattern you described I can make some adjustments. I will have another PR coming out soon that consolidates the nav components in the vertical navbar. This will tie everything together so I think it makes the most sense to update it there where I will already be putting final touches on the styles.

caseyhebebrand:
> cool! at some point would love to see screenshots :D

Just uploaded one :)

PR #8169, opened 2020-04-14
feat(core/filters): New search header for filter section

caseyhebebrand:
The search aspect of the filter section is undergoing a layout/design change. 

Old:
Screen Shot 2020-04-14 at 12 52 01 PM

New: 
Screen Shot 2020-04-14 at 12 50 31 PM

This PR will be integrated into the UI once v1 layout is ready for rollout. It accompanies some other layout changes to the filter section.


app/scripts/modules/core/src/presentation/main.less
```diff
@@ -974,6 +974,18 @@ select:invalid {
   }
 }
 
+.filter-search {
```
christopherthielen, line 977: Seems like this should be in `FilterSearch.less`?

app/scripts/modules/core/src/presentation/main.less
```diff
@@ -974,6 +974,18 @@ select:invalid {
...
+  input::placeholder {
+    font-size: 14px;
+    color: var(--color-titanium);
+  }
+  .glyphicon {
```
christopherthielen, line 982: This makes me a little uncomfortable, especially now that we're moving towards SVG icons (sort of like the same queasiness I get when I see new code in AngularJS).  Is this styling making the "?" render nicely?  What does the default styles look like?

app/scripts/modules/core/src/presentation/main.less
```diff
@@ -1067,6 +1079,12 @@ select:invalid {
   }
 }
 
+.help-field-large {
```
christopherthielen, line 1082: What is this?

app/scripts/modules/core/src/presentation/main.less
caseyhebebrand: Noted. I will find more specific non-glyphicon related classnames. This comes from the`HelpField` component, which is defaulted as a small, dark grey, question mark. I can utilize all of the functionality of that component, but need it to visually be bigger and lighter.

christopherthielen: gotcha.  

I bet this use of glyphicon in HelpField will eventually be caught in the great-svg-icon-refactor of 2020 :)

PR #8170, opened 2020-04-14
feat(core): Filter out providers that don't support creating load bal…

allisaurus:
…ancers

While working to address https://github.com/spinnaker/spinnaker/issues/4797, discovered that adding ECS load balancer components results in a broken experience for the `Create Load Balancer` button/modal because no provider-specific `CreateLoadBalancerModal` is defined.

This change allows some providers' load balancer support to be "read only", where viewing existing load balancers is possible but new ones cannot be created through Spinnaker. Builds on precedence set by https://github.com/spinnaker/deck/pull/7370 .

I believe this currently only effects the ECS provider, where load balancers/target groups can be created through the associated AWS account and AWS provider "create load balancer" Ul.

### Testing
Spinnaker with AWS + ECS enabled, `Create Load Balancer` goes straight to AWS LB choice modal (no provider selection)
lb_create_aws

Existing LB page: only AWS LBs shown
lb_view_no_ecs

W/ ECS LB components implemented (PR pending): AWS and ECS LBs shown
lb_view_with_ecs

### DIFF
--- a/app/scripts/modules/core/src/loadBalancer/CreateLoadBalancerButton.tsx
+++ b/app/scripts/modules/core/src/loadBalancer/CreateLoadBalancerButton.tsx
@@ -28,7 +28,11 @@ export class CreateLoadBalancerButton extends React.Component<ICreateLoadBalance
     super(props);
   }
 
-  private createLoadBalancerProviderFilterFn = ({}, {}, provider: ICloudProviderConfig): boolean => {
+  private createLoadBalancerProviderFilterFn = (
+    _app: Application,
+    _acc: IAccountDetails,
+    provider: ICloudProviderConfig,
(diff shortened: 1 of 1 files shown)


clareliguori:
LGTM.  Since this core deck code, I'd like to get another pair of eyes on it
cc @christopherthielen @maggieneterval

maggieneterval:
(@jorgebee65 perhaps this could be helpful in simplifying your work in [this PR](https://github.com/spinnaker/deck/pull/7822) to disable ad-hoc infrastructure creation: maybe it would be possible to only register the relevant components to the Kubernetes V2 provider if your feature flag is enabled. Thought I'd cc you just in case this is useful.)

PR #8177, opened 2020-04-15
feat(core/filters): Update filter section layout

caseyhebebrand:
The v1 layout refactor includes some changes to the filter section:

1. Removing background color from filter section

2. Updating the filter header layout and adding new hover treatments
Screen Shot 2020-04-15 at 12 06 13 PM

3. Moving the hide/show filter button in-line with main content header. This included updating the `FilterCollapse` component, streamlining all dataSources to use this component, and extracting `FilterCollapse` out of the `nav` ui view. [This commit](https://github.com/spinnaker/deck/pull/8177/commits/2d713ea237a12feab709b30d25097786d1d15429) to view these changes. 
Screen Shot 2020-04-15 at 12 10 33 PM

Full filter bar:
Screen Shot 2020-04-15 at 12 06 06 PM


app/scripts/modules/core/src/insight/insightLayout.component.html
```diff
@@ -3,6 +3,7 @@
   ng-if="!$ctrl.app.notFound && !$ctrl.app.hasError"
   ng-class="$ctrl.insightFilterStateModel.filtersExpanded ? 'filters-expanded' : 'filters-collapsed'"
 >
+  <filter-collapse ng-if="!$ctrl.insightFilterStateModel.filtersHidden" class="filter-toggle"></filter-collapse>
```
caseyhebebrand, line 6: I chose to directly place the component here rather than use the ui-view. This breaks the current pattern, but I feel it is much less complex.  The `ng-if` takes care of the conditional rendering for all dataSources. Directly placing the component here is much less complex than adding the new view to every `states` file and having to trace the DOM in a disjointed way.

gcomstock:
Thanks for attaching the screen shots! Looking great.

caseyhebebrand:
Collapsing filter:
Apr-20-2020 12-38-28

Collapsing filter section:
Apr-20-2020 12-38-17

cc: @christopherthielen

app/scripts/modules/core/src/filterModel/FilterCollapse.tsx
```diff
@@ -13,24 +13,31 @@ export class FilterCollapse extends React.Component<{}> {
 
   public render() {
     return (
-      <>
+      <div className="filters-toggle layer-high sp-margin-s-xaxis">
         <h3 className="filters-placeholder">
           <Tooltip value="Show filters">
-            <a className="btn btn-xs btn-default pin clickable" onClick={() => this.onClick(true)}>
+            <a
```
christopherthielen, line 19: while you're in here, is it possible to change this to a `<button>` tag? an `a` tag styled as a button should be used only if it's a link to somewhere else while actions should use a `<button>` tag.  This comment applies to elsewhere in this file.

app/scripts/modules/core/src/loadBalancer/filter/LoadBalancerFilters.tsx
```diff
@@ -190,7 +190,7 @@ export class LoadBalancerFilters extends React.Component<ILoadBalancerFiltersPro
 
     return (
       <div>
-        <FilterCollapse />
+        {/* <FilterCollapse /> */}
```
christopherthielen, line 193: just delete it?

app/scripts/modules/core/src/presentation/main.less
```diff
@@ -159,6 +159,18 @@ html {
       overflow-y: auto;
     }
   }
+  .filters-toggle {
```
christopherthielen, line 162: this PR adds a lot more css to `main.less`.  I'd like to move as much as this into component specific files.  This isn't a blocker for merge, it can be done at any time in the future.

app/scripts/modules/core/src/filterModel/FilterCollapse.tsx
caseyhebebrand: Yes, will do that.

app/scripts/modules/core/src/presentation/main.less
caseyhebebrand: I will just do it before merge, as it is easy enough to pull to a new file.

PR #8180, opened 2020-04-15
feat(plugins): Consolidate typescript config (partially).  Do not strip comments.

christopherthielen:
This should fix official releases of Deck not supporting plugins (due to the magic "webpackIgnore" comment getting stripped)

The `webpackIgnore` comment was being stripped by `ts-loader`, the webpack typescript compiler pipeline.  This is because the package root `tsconfig.json` had `"removeComments": true` while the individual tsconfig in each package had `"removeComments": false`.  

The non-deterministic builds happened because of `thread-loader`.  The main thread queues work and worker threads pick up that work and run with it.  When individual threads initialize, they each find and load a typescript configuration file, and use that configuration for the lifetime of the build (each thread loads typescript configuration once).  The `tsconfig.json` file that is loaded depends on the initial unit of work.  If that unit of work was for a file not in a package (such as `/app/scripts/app.js`, `/settings.js` or `/settings-local.js`), then that thread loads the root `/tsconfig.json` and all transpilation work done on that thread will have comments stripped.  If the initial unit of work for a thread was inside a package, such as `app/scripts/modules/core` or `app/scripts/modules/amazon`, then it will use the `tsconfig.json` from the package and will preserve comments.

The more threads used, the more likely the threads will pick up the nested package `tsconfig.json`, which is why local builds are quite likely to succeed (we generally have beefy developer machines with lots of cores, and thus we spawn lots of threads).  In docker builds, we generally have fewer cores available (2 for travis and github actions) and so *all the threads were loading the root `tsconfig.json`*.

edit: the solution here is to consolidate as much common configuration as possible (including `removeComments`) to the top level so choosing either `tsconfig.json` results in the same output.


app/scripts/modules/amazon/tsconfig.json
```diff
@@ -1,33 +1,11 @@
 {
-  "buildOnSave": false,
-  "compileOnSave": true,
+  "extends": "../../../../tsconfig.json",
```
erikmunson, line 2: did not even know this existed, that's awesome.

tsconfig.json
```diff
@@ -51,7 +51,7 @@
       "huaweicloud": ["huaweicloud/src"]
     },
     "pretty": true,
-    "removeComments": true,
+    "removeComments": false,
```
erikmunson, line 54: should we leave a 'do not change this or you will be fired' comment here?

link108: its a json file, so comments aren't supported, but I agree with the idea 😅

erikmunson: there are [comments](https://github.com/spinnaker/deck/pull/8180/files/1c59e5dba065230a56410d51d79349956dbce6aa#diff-e5e546dd2eb0351f813d63d1b39dbc48R58) in every json file in this PR :)

christopherthielen: `tsconfig.json` is _"special"_ json

christopherthielen:
@ezimanyi I tagged you in case you want to release a patch release (I know... would have been nice to know yesterday, right?)

tsconfig.json
link108: ahhh TIL! :)

PR #8182, opened 2020-04-16
feat(Dockerfile.compile): use node tooling

christopherthielen:


Dockerfile.compile
```diff
@@ -1,5 +1,6 @@
-FROM openjdk:8
+FROM node:12
 MAINTAINER sig-platform@spinnaker.io
-ENV GRADLE_USER_HOME /workspace/.gradle
-ENV GRADLE_OPTS -Xmx2048m
-CMD ./gradlew build --no-daemon -PskipTests
+RUN mkdir -p /home/user
+RUN chmod 777 /home/user
+ENV HOME /home/user
```
christopherthielen, line 5: yarn expects a home directory, so let's give it one?

plumpy:
This won't work because we also use this container to build the Debian packages. Which uses Gradle, so requires Java.

PR #8183, opened 2020-04-16
fix(titus): Fetch ipv6 details from scope, remove additional call

aravindmd:
- Remove additional call made to fetch `ipv6` address , the details are now available in the titus instance details response
- After this is merged i will update the `nflx` version


app/scripts/modules/titus/src/instance/details/instance.details.controller.js
```diff
@@ -122,17 +122,7 @@ module(TITUS_INSTANCE_DETAILS_INSTANCE_DETAILS_CONTROLLER, [
...
-            return InstanceReader.getInstanceDetails(accountDetails.environment, region, instanceDetails.agentId);
-          })
-          .then(instance => {
-            if (instance && instance.networkInterfaces) {
-              $scope.instance.ipv6Addresses = flatMap(instance.networkInterfaces, i =>
-                i.ipv6Addresses.map(a => a.ipv6Address),
-              );
-            }
-          }, autoClose);
+          });
```
caseyhebebrand, line 125: `autoClose` is still needed here: `.then( () => {}, autoClose)`

aravindmd: thanks , fixed it

PR #8184, opened 2020-04-16
feat(core): enable standard artifacts UI by default

maggieneterval:
Related to: https://github.com/spinnaker/governance/blob/master/rfc/legacy_artifacts_ui_removal.md 

This change removes `artifacts` and `artifactsRewrite` feature flags in preparation for release 1.20, in which the standard artifacts UI will be enabled by default. Provides a temporary flag that can be enabled in order to revert to the legacy UI, which will be removed in release 1.21.

Corresponding Halyard and Docs changes coming soon!

### DIFF
--- a/app/scripts/modules/core/src/artifact/ArtifactsModeService.ts
+++ b/app/scripts/modules/core/src/artifact/ArtifactsModeService.ts
@@ -21,7 +21,6 @@ import { SETTINGS } from 'core/config';
  */
 
 export enum ArtifactsMode {
-  DISABLED,
   LEGACY,
   STANDARD,
 }
@@ -30,12 +29,9 @@ export class ArtifactsModeService {
   public static readonly artifactsMode = ArtifactsModeService.getArtifactsMode();
 
   private static getArtifactsMode(): ArtifactsMode {
-    if (SETTINGS.feature.artifactsRewrite === true) {
-      return ArtifactsMode.STANDARD;
-    }
-    if (SETTINGS.feature.artifacts === true) {
+    if (SETTINGS.feature.legacyArtifactsEnabled === true) {
       return ArtifactsMode.LEGACY;
     }
-    return ArtifactsMode.DISABLED;
+    return ArtifactsMode.STANDARD;
   }
 }
--- a/app/scripts/modules/core/src/artifact/expectedArtifactMultiSelector.component.ts
+++ b/app/scripts/modules/core/src/artifact/expectedArtifactMultiSelector.component.ts
@@ -1,7 +1,6 @@
 import { IComponentOptions, IController, module } from 'angular';
 import { IExpectedArtifact } from 'core/domain';
 import { ArtifactIconService } from './ArtifactIconService';
-import { ArtifactsMode, ArtifactsModeService } from './ArtifactsModeService';
 
 import './artifactSelector.less';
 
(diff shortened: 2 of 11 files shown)


erikmunson:
Beyond enabling things by default, am I correct in understanding that this means some form of the artifacts UI will _always_ be enabled now, and you can't turn it off? If so, I think this PR will need to be reverted and modified to still allow totally disabling, even if we come back and add that part a bit later.

To the best of my knowledge currently **none** of the AWS/Titus machinery in Orca is capable of handling artifacts, so even if we at Netflix scrambled to try to understand the feature and fast-follow with AWS + Titus support (we currently have it turned off) there would be long period of time where anybody running on AWS would have a weird/bad time with the default experience. Realistically we are most likely not going to prioritizing adding that support ourselves given our other strategic work like managed delivery, plugins, etc. so I can't say exactly when this feature could become mandatory.

maggieneterval:
@erikmunson sorry for any confusion here; I understood @christopherthielen's approval of [this RFC](https://github.com/spinnaker/governance/pull/111) as the go-ahead from Netflix to remove the artifacts feature flags.

There should be no changes required for cloud providers that do not currently support artifacts. The only noticeable difference when configuring a pipeline should be the addition of an (optional) Artifact Constraints input alongside triggers, which I think @christopherthielen and I had agreed should be minimally intrusive to Netflix's workflow. Let me know if y'all feel otherwise; happy to reconsider!

erikmunson:
Oops, I think there may have been a bit of a miscommunication — that RFC was worded as _enabling the newer version of artifacts by default_. Because that was how the RFC was framed, I'm not sure the nuance in the statements about "removing the flags" came across as intended to readers.  When I read the RFC originally I didn't comment because my understanding based on the text was that the change was default-to-on, not always-on.

Given that we (and many others) use _only_ the cloud providers which do not support artifacts, having those features show up is definitely going to cause disruption for us internally as folks will start trying to use it, hitting failures/issues, and then coming to our team for help.

Not sure if this makes sense technically or not, but is there a world where we revert this, add the ability for _providers_ to declare whether or not they support artifacts, and then hide the UI elements if none of the providers the app uses support artifacts? That might get you the best of both worlds (or it might be a bad idea I'm not sure). Either way I think we'll need to revert this in the immediate term to unblock `master` for folks that are AWS-only.

ezimanyi:
I understand that it will be an additional field for users that don't use artifacts, but I still think this statement from the RFC is correct:
> However, the only change in pipeline configuration experience for users who currently have artifacts disabled will be the addition of the optional Artifacts Constraints input to the Automated Triggers section. Since this is a minimally intrusive addition, the benefits of maintaining and documenting a single configuration experience likely outweigh this risk.

Ultimately there will always be a tradeoff between what should be part of the uniform experience vs. what should be hidden behind a flag.  The disadvantages of a flag are increase maintenance burden (testing things with the flag on vs. off) and difficulty getting Spinnaker set up (ensuring you have the same flags as whatever codelab you're following).

In this case, it feels like the addition of a single drop-down in trigger config that can be ignored is minimally intrusive, and the inconvenience of having users skip over that outweighs the complexity of keeping around another feature flag.  (This seems actually less intrusive than things like pipeline parameters and notifications that take up most of the config screen but are likely not used by all users.)

image

I would definitely lean more towards continuing to support a flag to turn it off if it added a whole bunch of fields all over the UI, but it really is just a single field that can be left empty for users that don't use it (similar to how we show Notifications and Pipeline Parameters on the config screen).

I also really want to see us get better at promoting features through a lifecycle then cleaning up the feature flags controlling them, which will require that Spinnaker users adapt to some changes in the UI around features they aren't using.  I worry that we're being overly cautious here about not wanting to introduce any new UI elements that aren't behind a flag, and would rather see us keep flags for features that are more intrusive.

PR #8188, opened 2020-04-17
fix(ecs): move service discovery configuration to react

pkandasamy91:
Updated the service discovery section of the configure server group wizard to react. Created and updated a server group using both the existing UI and the updated UI and the generated json is the same.
```
       "serviceDiscoveryAssociations": [
        {
          "containerPort": 80,
          "registry": {
            "account": "ecs-my-aws-devel-acct",
            "arn": "arn:aws:servicediscovery:us-west-2:164281909131:service/srv-mmr3glofrvzvtpdv",
            "displayName": "spinnaker-ecs-ec2-demo-svc (srv-mmr3glofrvzvtpdv)",
            "id": "srv-mmr3glofrvzvtpdv",
            "name": "spinnaker-ecs-ec2-demo-svc",
            "region": "us-west-2"
          }
        }
      ],
```
Screen Shot 2020-04-16 at 10 07 41 AM

Screen Shot 2020-04-16 at 10 07 53 AM

Screen Shot 2020-04-16 at 10 08 15 AM

Resolves: https://github.com/spinnaker/spinnaker/issues/5482


pkandasamy91:
@allisaurus and @clareliguori please review

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -97,6 +97,15 @@ export interface IEcsTaskDefinitionArtifact {
   artifactId?: string;
 }
 
+export interface IEcsServiceDiscoveryRegistry {
```
allisaurus, line 100: since this isn't used in this file, why not define it in `ServiceDiscovery.tsx` ?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/serviceDiscovery/serviceDiscovery.html
```diff
@@ -1,10 +1,9 @@
 <div class="clearfix">
-  <div class="row">
```
allisaurus, line 2: can we add back the `row` class to keep section padding consistent

app/scripts/modules/ecs/src/serverGroup/configure/wizard/serviceDiscovery/ServiceDiscovery.tsx
```diff
@@ -0,0 +1,194 @@
...
+      <div className="container-fluid form-horizontal">
+        <div className="form-group">
+          <div className="sm-label-left">
+            <b>Service Registries (optional)</b>
+            <HelpField id="ecs.serviceDiscovery" />
+          </div>
+          <form name="ecsServiceDiscoveryRegistryMappings">
+            <table className="table table-condensed packed tags">
+              <thead>
+                <tr key="header">
```
allisaurus, line 164: is this `key` prop still needed? seems to be only 1 `tr`under this `thead`

app/scripts/modules/ecs/src/serverGroup/configure/wizard/serviceDiscovery/ServiceDiscovery.tsx
```diff
@@ -0,0 +1,194 @@
...
+      arn: '',
+      displayName: '',
+    };
+  };
+
+  private updateServiceDiscoveryRegistry = (index: number, newRegistry: Option<string>) => {
+    const registryMap = this.getNameToRegistryMap();
+    let newServiceDiscoveryRegistry = registryMap.get(newRegistry.value);
+
+    if (!newServiceDiscoveryRegistry) {
```
allisaurus, line 68: since this is firing when the registry drop-down is changed, when would we not find the registry? could `newRegistry` be search input (vs. an existing option)?

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
pkandasamy91: Good catch! Initially thought i might need it here, but didnt end up using it. Moved it.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/serviceDiscovery/ServiceDiscovery.tsx
pkandasamy91: I was trying to maintain consistency with how the other properties are loaded in the configuration wizard currently (artifacts, container mappings, etc)

PR #8189, opened 2020-04-18
feat(core/managed): CURRENTLY_UNRESOLVABLE status, tweak unresolvable event

erikmunson:
backend implementation and further rationale captured in https://github.com/spinnaker/keel/pull/1036

we recently determined that we needed a new resource status — `CURRENTLY_UNRESOLVABLE` — to represent the situations where a resource can't be checked because something Spinnaker needs to perform the check is missing or having transient issues. This PR:
- Adds support for `CURRENTLY_UNRESOLVABLE`
- Tweaks the treatment on the `ResourceCheckUnresolvable` history event which produces that status in order to disambiguate its relationship to the `ResourceCheckError` event

**Before:**

Screen Shot 2020-04-17 at 5 36 57 PM

(there was no treatment for the new status, because it's new)

**After:**

Screen Shot 2020-04-17 at 5 15 50 PM
Screen Shot 2020-04-17 at 5 29 01 PM


erikmunson:
fyi @gcomstock

PR #8191, opened 2020-04-20
Plugin stuff

christopherthielen:
Deck:

- Move some code around
- Define two new extension points:
  - `help`: Help text used by `<HelpField/>` component
  - `search`: Custom search result types
- Made `initialize()` function async, and pass in the plugin object

PluginSDK:

- Exposed more shared libraries to be used by plugin bundles
- Added a base `.huskyrc.js`
- Move dependencies to `peerDependencies`.  Deck plugins should add a script to `package.json`:
```
  "scripts": {
    "postinstall": "npx check-peer-dependencies"
  }
```


app/scripts/modules/core/src/plugins/deck.plugin.ts
```diff
@@ -0,0 +1,29 @@
...
+  /** Help Text for use in <HelpField /> */
+  help?: { [helpKey: string]: string };
+  /** Custom global search types */
+  search?: SearchResultType[];
+
+  initialize?(plugin: IDeckPlugin): void;
+}
+
+/** Given a plugin, registers the plugin's extensions with Deck registries */
+export function registerPluginExtensions(plugin: IDeckPlugin): PromiseLike<any> {
```
caseyhebebrand, line 21: Can we use an arrow function?

christopherthielen: sure, but for what reason?

PR #8192, opened 2020-04-20
fix(mocks): Fix tslint error

caseyhebebrand:


christopherthielen:
I guess this highlights that we could use some CI validation of these packages too.  Just validating that Typescript compiles should be sufficient.

PR #8200, opened 2020-04-22
chore(build): upgrade to spinnaker-gradle 8.x

cfieber:


cfieber:
WIP do not merge, reworking deck gradle build for gradle 6.3, spinnaker gradle 8.x, autobumpability

cfieber:
as far as I can tell, most of the deck workflows don't actually care about the gradle build, it is only the buildtool/release process that ever execute this

build.gradle
```diff
@@ -1,34 +1,20 @@
+plugins {
+  id 'base'
+  id "io.spinnaker.project" version "$spinnakerGradleVersion" apply false
```
cfieber, line 3: grabbing this as a dependency gives us the BintrayPublishPlugin and transitively the ospackage plugin, but we don't want to apply the io.spinnaker.project because this isn't a multiproject jvm app + application-packaged-as-ospackage

build.gradle
```diff
@@ -1,34 +1,20 @@
...
-apply plugin: 'com.github.hierynomus.license'
-
-jar {
-  dependsOn('webpack')
-  from(fileTree("build/webpack")) {
-    into "META-INF/resources"
-  }
-}
+apply plugin: "nebula.ospackage"
+project.tasks.register("publish")
```
cfieber, line 16: the publish lifecycle hook would normally exist from the publishing plugin, but we aren't applying that here. This is our entrypoint to building and publishing the deb.

BintrayPublishPlugin expects `publish` to exist so we have to register it first

build.gradle
```diff
@@ -78,24 +64,28 @@ task functionalTestsFirefox(type: NodeTask) {
...
+  }
 }
 
 webpack.outputs.dir file('build/webpack')
 
 yarn.dependsOn 'generateVersionFile'
 webpack.dependsOn 'yarn'
 karma.dependsOn 'webpack'
-test.dependsOn 'karma'
+project.tasks.register('test') {
```
cfieber, line 86: `test` lifecycle hook doesn't exist because we aren't applying java anymore, so creating it here to tie in to running karma

build.gradle
plumpy: Why don't we just depend on those plugins directly? I guess maybe the advantage of this is it means we use the same versions of bintray and ospackage as the rest of the projects?

plumpy:
Thanks for the comments, they're all super informative... so informative that I'd argue they probably belong as comments in `build.gradle` directly, since they're all kinda weird counter-intuitive things?

build.gradle
plumpy: Oh, I didn't realize `BintrayPublishPlugin` was _our_ plugin... I thought it was one from the main gradle plugins repo. Thanks makes sense...

(Realized from your comment in the monitoring PR)

PR #8202, opened 2020-04-22
fix(UI): add stage summaries to hydrated graphs

AbdulRahmanAlHamali:
When hydrating an execution, we don't add the stage summaries to it. This results in some of the fields not showing, like stage comments.

This PR should fix the problem.


AbdulRahmanAlHamali:
Thank you Erik! Can we merge that before the release please?

PR #8203, opened 2020-04-22
feat(core/nav): Compose vertical nav bar

caseyhebebrand:
Dependant on https://github.com/spinnaker/deck/pull/8150 and https://github.com/spinnaker/deck/pull/8162.

Create the vertical navigation component. This is not integrated yet, but in the future will replace `ApplicationHeader`, `ApplicationNavSection`, and `ThirdLevelNavigation`. 

Screen Shot 2020-04-08 at 8 06 15 PM


app/scripts/modules/core/src/application/nav/ApplicationNavigation.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+    mock.inject((_$uiRouter_: UIRouterReact) => {
+      $uiRouter = _$uiRouter_;
+      // Inititalize base route
+      $uiRouter.stateRegistry.root().name = 'http://localhost:9000';
+    }),
+  );
+  beforeEach(() => {
+    // Initialize current route
+    spyOn(ReactInjector.$state, 'includes').and.callFake((substate: any) => currentStates.includes(substate));
+    spyOn(StateMatcher.prototype, 'resolvePath').and.callFake(() => 'pipelines.executions');
```
christopherthielen, line 38: this is an extremely internal API that shouldn't be relied upon.  I'll see if I can come up with a better way to handle this.

app/scripts/modules/core/src/application/nav/ApplicationNavigation.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+  it('should render header, categories, and pagerduty button', () => {
+    const app = ApplicationModelBuilder.createApplicationForTests(
+      'testapp',
+      mockPipelineDataSourceConfig,
+      mockServerGroupDataSourceConfig,
+      mockAppConfigDataSourceConfig,
+    );
+    const activeDataSource = app.getDataSource('executions');
+    app.dataSources.push({ ...activeDataSource, key: 'runningExecutions' } as ApplicationDataSource<IPipeline>);
+    app.getDataSource(activeDataSource.badge).status$ = new BehaviorSubject({
```
christopherthielen, line 50:
```suggestion
    app.getDataSource(activeDataSource.badge).status$.next({
```

app/scripts/modules/core/src/application/nav/ApplicationNavigation.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+    const app = ApplicationModelBuilder.createApplicationForTests(
+      'testapp',
+      mockServerGroupDataSourceConfig,
+      mockLoadBalancerDataSourceConfig,
+      mockTaskDataSourceConfig,
+      mockAppConfigDataSourceConfig,
+    );
+    const activeDataSource = app.getDataSource('tasks');
+    app.dataSources.push({ ...activeDataSource, key: 'runningTasks' } as ApplicationDataSource<IPipeline>);
+    app.getDataSource(activeDataSource.badge).status$ = new BehaviorSubject({
```
christopherthielen, line 100:
```suggestion
    app.getDataSource(activeDataSource.badge).status$.next({
```

app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
```diff
@@ -0,0 +1,97 @@
...
+  const getActiveCategory = (categories: ApplicationDataSource[]) =>
+    categories.find(c => c.activeState && ReactInjector.$state.includes(c.activeState));
+  const [activeCategory, setActiveCategory] = React.useState(getActiveCategory(_flatten(navSections)));
+
+  const resetActiveCategory = () => {
+    const newActive = getActiveCategory(app.dataSources);
+    setActiveCategory(newActive);
+  };
+
+  const appRefreshSubscription = app.onRefresh(null, () => {
```
christopherthielen, line 49: this should go inside the `useEffect`

app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
```diff
@@ -0,0 +1,97 @@
...
+    if (!_isEqual(app.attributes.dataSources, prevDataSourceAttr)) {
+      const categories = getNavigationCategories(app.dataSources);
+      setNavSections(categories);
+    }
+  });
+
+  React.useEffect(() => {
+    routerSubscription = ReactInjector.$uiRouter.transitionService.onSuccess({}, () => {
```
christopherthielen, line 57: https://ui-router.github.io/react/docs/1.0.1/modules/react_hooks.html#usecurrentstateandparams

app/scripts/modules/core/src/application/nav/ApplicationNavigation.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+describe('ApplicationNavigation', () => {
+  let $uiRouter: UIRouterReact;
+  const currentStates = ['**.pipelines.**', '**.tasks.**'];
+
+  beforeEach(mock.module(REACT_MODULE, OVERRIDE_REGISTRY));
+  beforeEach(
+    mock.inject((_$uiRouter_: UIRouterReact) => {
+      $uiRouter = _$uiRouter_;
+      // Inititalize base route
+      $uiRouter.stateRegistry.root().name = 'http://localhost:9000';
```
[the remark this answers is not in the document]
caseyhebebrand: The injected `$uiRouter` has an undefined base route, which was causing an error when mounting the component. I traced back to this attribute in the `ui-router` code.

app/scripts/modules/core/src/application/nav/ApplicationNavigation.spec.tsx
caseyhebebrand: Sounds good. Maybe moving to `useCurrentStateandParams` hook will help. I had to stub anything relating to the `ReactInjector` for mocking an active category or category change.

christopherthielen: I guess I was wrong.  After looking into the docs, this is indeed marked as public API.
However, this might be a slightly more idiomatic way to mock this:
```suggestion
    spyOn(StateMatcher.prototype, 'find').and.callFake(() => undefined as any);
```

app/scripts/modules/core/src/application/nav/ApplicationNavigation.spec.tsx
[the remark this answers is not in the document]
christopherthielen: I removed this line and your tests still pass.

app/scripts/modules/core/src/application/nav/ApplicationNavigation.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+describe('ApplicationNavigation', () => {
+  let $uiRouter: UIRouterReact;
+  const currentStates = ['**.pipelines.**', '**.tasks.**'];
+
+  beforeEach(mock.module(REACT_MODULE, OVERRIDE_REGISTRY));
+  beforeEach(
+    mock.inject((_$uiRouter_: UIRouterReact) => {
+      $uiRouter = _$uiRouter_;
+      $uiRouter.globals.current.name = '**.tasks.**';
```
christopherthielen, line 30: this sort of thing still seems unnecessary to me, but since it's not hurting anything let's just move on

caseyhebebrand: I don't mind deleting it. In this case, the `uirouter` looks at this value for the initial route, rather than having that be defined.

PR #8205, opened 2020-04-22
chore(build): make buildDeb actually depend on the contents of the deb

cfieber:


cfieber:
only if we update to the new flavorflavicon standard

PR #8209, opened 2020-04-24
fix(stage): fix import delivery config doc link

lorin:
Fix broken doc link in Import Delivery Config stage.


erikmunson:
thanks!

PR #8210, opened 2020-04-26
feat(tencentcloud): init tencentcloud

zexiplus:
This pr is the first commit for the tencentcloud provider and includes the intialization works. The cloud provider proposal is [spinnaker/governance#79](https://github.com/spinnaker/governance/issues/79)


zexiplus:
would you like to take a moment to review this PR, I 'd preciate to take you advice, thx @erikmunson

app/scripts/modules/tencentcloud/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@spinnaker/tencentcloud",
```
caseyhebebrand, line 2: Don't forget to include the version in this file.

zexiplus: > Don't forget to include the package version in this file.

Thanks for your guide. 
But if I add version field in the package.json, I will not pass the actions, cuz the previous version is empty, it's not equal to 0.0.1, how can I fix it ? Do I need to rewrite assert_package_bumps_standalone.sh or ... ?
image
https://github.com/spinnaker/deck/runs/614469005?check_suite_focus=true

zexiplus: > Don't forget to include the package version in this file.

Thanks for your guide. 
But if I add version field in the package.json, I will not pass the actions, cuz the changes in package.json is not only version field but also includes the other fields, in addition, the previous version is empty. It cause that the version change is not only change in package.json and fail the action's pipeline. How could I fix it ? Do I need to rewrite assert_package_bumps_standalone.sh or ... ?
image
https://github.com/spinnaker/deck/runs/614469005?check_suite_focus=true

caseyhebebrand: You can always merge this PR with version `0.0.0` . Then if the action fails, you can follow up with a simple PR that bumps the version to `0.0.1` so the action will be triggered.

zexiplus: > You can always merge this PR with version `0.0.0` . Then if the action fails, you can follow up with a simple PR that bumps the version to `0.0.1` so the action will be triggered.

Thanks for your suggestion, I added version `0.0.0` to package.json

caseyhebebrand: If you keep running into build failures with this, you can also divide this into 2 PRs. The first would include all the module set up and the second would add the config for the workflow.

zexiplus: > If you keep running into build failures with this, you can also divide this into 2 PRs. The first would include all the module set up and the second would add the config for the workflow.

As the first PR,  this time I remove the tencentcloud config for the workflow and version field in package.json. I will add the version field and config for the workflow in the next PR. Is it all right?  If not could you give me some suggestions with more detail?

zexiplus:
@christopherthielen,  would you like to take a moment to review this PR, I 'd preciate to take you advice, thx
additional remarks: 
I did not add version in tencentcloud/package.json cuz I want to pass the action.  If this PR can be merged into master, I will add it in the next PR.

app/scripts/modules/tencentcloud/src/tencentcloud.setting.ts
```diff
@@ -0,0 +1,35 @@
...
+    vpc?: string;
+  };
+  defaultSecurityGroups?: string[];
+  loadBalancers?: {
+    inferInternalFlagFromSubnet: boolean;
+    certificateTypes?: string[];
+  };
+  useAmiBlockDeviceMappings?: boolean;
+  classicLaunchLockout?: number;
+  classicLaunchWhitelist?: IClassicLaunchWhitelist[];
```
christopherthielen, line 22: Does tencent really have AmiBlockDeviceMappings or ClassicLaunch?  Both of these things imply Amazon to me.

halconfig/settings.js
```diff
@@ -113,6 +113,12 @@ var oracle = {
     region: '{%oracle.default.region%}',
   },
 };
+var tencentcloud = {
+  defaults: {
+    account: '{%oracle.default.account%}',
```
christopherthielen, line 118: `oracle`?

zexiplus: sorry, that's was a mistake

zexiplus: > `oracle`?

sorry, that's was a mistake

app/scripts/modules/tencentcloud/src/tencentcloud.setting.ts
zexiplus: > I found a couple references to other cloud providers that look possibly misplaced.

Thanks for your earnest review, I fixed them

christopherthielen:
@zexiplus once this branch is up to date with `master` it will get merged.  We want PRs to have "allow updates" enabled so mergify.io can do its automatic merging.

See: https://github.com/spinnaker/deck/pull/8210/checks?check_run_id=673181817

zexiplus:
> @zexiplus once this branch is up to date with `master` it will get merged. We want PRs to have "allow updates" enabled so mergify.io bot can do its automatic merging.
> 
> See: https://github.com/spinnaker/deck/pull/8210/checks?check_run_id=673181817

Thx, already updated and merged

PR #8211, opened 2020-04-27
fix(amazon/validator): remove the invalid "_" from the regex

antoiner77:
The regex used to perform a check on the bucket name in the Functions menu is wrong. According to AWS documentation a bucket name can have "-" in it but no "_".

https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html

Bug:
S3 bucket:
Capture d’écran 2020-04-27 à 18 07 06
Functions modal:
Capture d’écran 2020-04-27 à 18 07 21


christopherthielen:
lovely, thanks!

PR #8214, opened 2020-04-28
feat(core/managed): Add StatusBubble to denote resource status

vigneshm:
This PR is split into 3 commits for easier reviewing.

- [Commit 1](https://github.com/spinnaker/deck/pull/8214/commits/a8d481870108dbaf43c05f6312227ca099330e79) : The spacing between items in `ObjectRow` is adjusted to fit in the new resource status bubble.

- [Commit 2](https://github.com/spinnaker/deck/pull/8214/commits/927a7b70b1b624bf5953ce298187118cc2c3a93f): `StatusBubble` component now supports rendering the `quantity` subtext. Here's a [demo](https://codesandbox.io/s/status-bubble-with-pill-j5owi).

- [Commit 3](https://github.com/spinnaker/deck/pull/8214/commits/39e547ad26afd85cad48e7ec3f23d9f5eb29950b): Adding resource status bubble

<details>
  <summary>Here's a screenshot of status bubbles showing mock resource status</summary>

image

</details>

### Questions
1. The icons themselves include a managed delivery symbol on the top level area and as a result makes the whole icon not looked center-aligned. Is it worth to create a new set of icons without the managed delivery symbol on the top left?
2. This is something @gcomstock brought up yesterday where we do not want to show the bubble when the resource status is `HAPPY` (which is how I've implemented it now). This follows the principle that we show resource related status only when it needs attention and ignore the happy scenario. Just asking this question here to ensure everyone is in agreement.


erikmunson:
awesome! I'm about to jump in and review the code itself, but wanted to quickly provide some input/general thoughts on the two questions you raised:

> The icons themselves include a managed delivery symbol on the top level area and as a result makes the whole icon not looked center-aligned. Is it worth to create a new set of icons without the managed delivery symbol on the top left?

My 2 cents is that it does make sense to drop the M here. I'm actually starting to wonder if we need the M on those icons _anywhere_, as my observation is that the M iconography itself has ended up being less necessary/important than we had originally anticipated. If we think we can get away without the M on the Infrastructure tabs, then maybe we could just drop them from the icons we already have? Regardless of what we decide I don't think it needs to block us getting this PR out.

> This is something @gcomstock brought up yesterday where we do not want to show the bubble when the resource status is HAPPY (which is how I've implemented it now). This follows the principle that we show resource related status only when it needs attention and ignore the happy scenario. Just asking this question here to ensure everyone is in agreement.

I'm in very full agreement with this, and was happy (pun half intended) to see that's how you and Greg ended up handling it. In the context of the Infrastructure tabs the fact that something is managed _at all_ is 'non-normal' in the sense that it's not something you can assume to be true, so it makes sense to visualize HAPPY there.

The only open question on my mind: does this implementation currently surface the popover UI that exists in the Infrastructure view with explanatory text + buttons for history and pausing? I realize we really want to avoid hover-state madness in newer UIs, but until we're able to implement a drawer component that can show details about a resource when clicked, there's a tradeoff to be made about whether it's worth surfacing those explanations + buttons. If we decided we should do the popovers for now, the only issue would be that the history and pause buttons would not be available on HAPPY resources. I think that's probably ok since the resources are still linkable and we'll be replacing it all with a drawer soon.

cc @gcomstock

erikmunson:
one other random piece of design feedback: is there room to add more padding on the right of that build number pill on the object rows in a way that doesn't clash with the other elements' alignment? maybe i'm just being weird/hyper detailed but for whatever reason my eyes keep getting annoyed about the way the pill doesn't really interact or connect with the things around it in terms of placement, and i'm just not quite sure exactly what's needed to make my eyes happy. maybe not actionable.

erikmunson:
actually re: my design nit: i think what i'm really seeing is the fact that the little expand arrow icon **does not** fill its full svg canvas, which means that even though the pill and the icon's SVG element are lined up with each other, the icon path itself is not actually lined up at all:

Screen Shot 2020-04-28 at 12 58 51 PM

vigneshm:
> The only open question on my mind: does this implementation currently surface the popover UI that exists in the Infrastructure view with explanatory text + buttons for history and pausing?

It doesn't yet. But I'm adding them in subsequent PRs. I'm adding this support directly into the `StatusBubble` component. One point to note though is that I was planning to add a simple popover that just showed the current status text (as shown in one of the mocks). However, it makes more sense to re-use that popover (used in Infrastructure view) directly here. @gcomstock what do you think? This popover has a different look and feel with white background than what we had discussed yesterday.

> is there room to add more padding on the right of that build number pill on the object rows in a way that doesn't clash with the other elements' alignment?

Wow I didn't notice that. You're right that the chevron is causing this issue. I adjusted the margin for the build number pill to make look align and it looks the following way.

image

Even though their right edges are aligned, it looks mildly odd to me. What do you think?

gcomstock:
> > The only open question on my mind: does this implementation currently surface the popover UI that exists in the Infrastructure view with explanatory text + buttons for history and pausing?
> 
> It doesn't yet. But I'm adding them in subsequent PRs. I'm adding this support directly into the `StatusBubble` component. One point to note though is that I was planning to add a simple popover that just showed the current status text (as shown in one of the mocks). However, it makes more sense to re-use that popover (used in Infrastructure view) directly here. @gcomstock what do you think? This popover has a different look and feel with white background than what we had discussed yesterday.
> 
> > is there room to add more padding on the right of that build number pill on the object rows in a way that doesn't clash with the other elements' alignment?
> 
> Wow I didn't notice that. You're right that the chevron is causing this issue. I adjusted the margin for the build number pill to make look align and it looks the following way.
> 
> image
> 
> Even though their right edges are aligned, it looks mildly odd to me. What do you think?

I think perfect visual alignment between the chevron and pill isn't going to happen, and that's OK.  My approach has just been to make as many measurement things as possible in multiples on 8. The chevron feels like the right size for its relative importance.

gcomstock:
> awesome! I'm about to jump in and review the code itself, but wanted to quickly provide some input/general thoughts on the two questions you raised:
> 
> > The icons themselves include a managed delivery symbol on the top level area and as a result makes the whole icon not looked center-aligned. Is it worth to create a new set of icons without the managed delivery symbol on the top left?
> 
> My 2 cents is that it does make sense to drop the M here. I'm actually starting to wonder if we need the M on those icons _anywhere_, as my observation is that the M iconography itself has ended up being less necessary/important than we had originally anticipated. If we think we can get away without the M on the Infrastructure tabs, then maybe we could just drop them from the icons we already have? Regardless of what we decide I don't think it needs to block us getting this PR out.
> 
> > This is something @gcomstock brought up yesterday where we do not want to show the bubble when the resource status is HAPPY (which is how I've implemented it now). This follows the principle that we show resource related status only when it needs attention and ignore the happy scenario. Just asking this question here to ensure everyone is in agreement.
> 
> I'm in very full agreement with this, and was happy (pun half intended) to see that's how you and Greg ended up handling it. In the context of the Infrastructure tabs the fact that something is managed _at all_ is 'non-normal' in the sense that it's not something you can assume to be true, so it makes sense to visualize HAPPY there.
> 
> The only open question on my mind: does this implementation currently surface the popover UI that exists in the Infrastructure view with explanatory text + buttons for history and pausing? I realize we really want to avoid hover-state madness in newer UIs, but until we're able to implement a drawer component that can show details about a resource when clicked, there's a tradeoff to be made about whether it's worth surfacing those explanations + buttons. If we decided we should do the popovers for now, the only issue would be that the history and pause buttons would not be available on HAPPY resources. I think that's probably ok since the resources are still linkable and we'll be replacing it all with a drawer soon.
> 
> cc @gcomstock

My opinion is that for an average user, the status cards are going to get me the level of info I need. The MD-specific statuses seem more appropriate when I'm viewing infra that is both managed and non-managed. Though if we just go back to the philosophy of showing something when it's wrong/needs attention, I'd argue that an unhappy status to prompt a user to view history is enough. the actuating/diff-detected/etc statuses aren't valuable if a card tells me the same thing in a more meaningful way.

erikmunson:
> The chevron feels like the right size for its relative importance.

I agree. Regardless of visual alignment in this specific example, I'm increasingly uncomfortable with the unintended side effects that have been popping up around sizing paths within their SVG canvas to effectively get a different icon size using the standard set of icon sizes we aligned on for `<Icon/>`. As of right now the icon size for those expand/collapse arrows is set to `medium` — that gives us both `small` and `extraSmall` to move down to if we were to make the path fill its canvas to make it standard size. Do we think that's workable?

My anxiety comes from the fact that when we put one of these specially-sized paths into a layout, we're effectively making additional variations of icon sizes that make changing or updating `<Icon/>` and related components less predictable — similar to the challenges we face today with icon fonts where we have very little ability to make holistic changes due to lack of confidence our changes will cleanly apply across all the uses of a given glyph.

gcomstock:
> > The chevron feels like the right size for its relative importance.
> 
> I agree. Regardless of visual alignment in this specific example, I'm increasingly uncomfortable with the unintended side effects that have been popping up around sizing paths within their SVG canvas to effectively get a different icon size using the standard set of icon sizes we aligned on for `<Icon/>`. As of right now the icon size for those expand/collapse arrows is set to `medium` — that gives us both `small` and `extraSmall` to move down to if we were to make the path fill its canvas to make it standard size. Do we think that's workable?
> 
> My anxiety comes from the fact that when we put one of these specially-sized paths into a layout, we're effectively making additional variations of icon sizes that make changing or updating `<Icon/>` and related components less predictable — similar to the challenges we face today with icon fonts where we have very little ability to make holistic changes due to lack of confidence our changes will cleanly apply across all the uses of a given glyph.

Great point! I think icons should fill the bounding box as much as possible, and a developer should rely on the different size props to achieve the different icon sizes. I'll update these.

erikmunson:
> the actuating/diff-detected/etc statuses aren't valuable if a card tells me the same thing in a more meaningful way.

i actually think we should talk verbally about this — i'm realizing that there's some deeper behavioral nuance to how keel works that isn't really apparent yet _precisely because_ we are not yet surfacing resource status alongside those environment cards to make it obvious how environments  and resources differ in their behavior. I think it's maybe worth just explaining the nuance so we all have a shared expectation about how keel behaves, it's not something even i feel like i fully grasp sometimes.

gcomstock:
Sounds good! I'm free until 3:30 or anytime tomorrow.

app/scripts/modules/core/src/managed/StatusBubble.tsx
```diff
@@ -9,6 +10,7 @@ export interface IStatusBubbleProps {
   iconName: IconNames;
   appearance: 'inactive' | 'neutral' | 'info' | 'progress' | 'success' | 'warning' | 'error';
   size: 'extraSmall' | 'small' | 'medium' | 'large' | 'extraLarge';
+  quantity?: string;
```
erikmunson, line 13: does it make sense to allow `number`s here too? (i.e. if you're doing something like `someArray.length`)

app/scripts/modules/core/src/managed/StatusBubble.tsx
```diff
@@ -36,16 +38,47 @@ const transitionConfig = {
...
       {transitions.map(({ item, key, props }) => (
-        <animated.div key={key} className="status-bubble-icon-container flex-container-h center middle" style={props}>
-          <Icon name={item} appearance="light" size={size} />
+        <animated.div className="StatusBubble" key={key} style={props}>
+          <div
+            className={classnames({
+              IconWrapper: true,
+              [`sp-padding-${paddingBySize[size]}`]: true,
+              [`status-${appearance}`]: true,
+              withQuantity,
```
erikmunson, line 71: fwiw `classnames` takes variadic args which can eliminate a lot of the redundant `true`s in object notation when you're mixing static classes with dynamic — e.g.:
```tsx
className={classnames(
  'IconWrapper',
  `sp-padding-${paddingBySize[size]}`,
  `status-${appearance}`,
  { withQuantity }
})}
```

app/scripts/modules/core/src/managed/StatusBubble.tsx
```diff
@@ -1,5 +1,6 @@
-import React, { memo } from 'react';
+import React, { ReactNode, memo } from 'react';
 import { useTransition, animated } from 'react-spring';
+import classnames from 'classnames';
```
erikmunson, line 3: _super_ nit:
```suggestion
import classNames from 'classnames';
```

app/scripts/modules/core/src/managed/StatusBubble.tsx
```diff
@@ -36,16 +38,47 @@ const transitionConfig = {
...
+        })}
+      >
+        {quantity}
+      </span>
+    );
+  }
 
   return (
-    <div className={`StatusBubble status-bubble-${appearance} sp-padding-${paddingBySize[size]}`}>
+    <>
```
erikmunson, line 63: have you tried this out live w/ changing icon names? i can't tell from reading the code but i'm sort of wondering whether the fact that mapping over `transitions` will sometimes generate 2 side-by-side divs here might lead to layout jank during transitions. the grid layout stuff that was previously on `StatusBubble` (now on `IconWrapper`) was originally to make sure that transitioning elements sat atop each other vs. getting stacked vertically or horizontally without constraining them to specific heights/widths

app/scripts/modules/core/src/managed/StatusBubble.tsx
```diff
@@ -36,16 +38,47 @@ const transitionConfig = {
...
 };
 
-export const StatusBubble = memo(({ iconName, appearance, size }: IStatusBubbleProps) => {
+export const StatusBubble = memo(({ appearance, iconName, size, quantity }: IStatusBubbleProps) => {
   const transitions = useTransition(iconName, null, transitionConfig);
+  let quantityPill: ReactNode = null;
+  let withQuantity = false;
+
+  // Render quantity text only for size >= small.
+  if (quantity && (size === 'small' || size === 'medium' || size === 'large' || size === 'extraLarge')) {
```
erikmunson, line 47: small readability tweak: do you care for the idea of pulling the list of sizes out explicitly and then referencing them here via a list lookup? e.g.

```tsx
// up above component near imports
const QUANTITY_SIZES = ['small', 'medium', 'large', 'extraLarge'];

// ...

if (quantity && QUANTITY_SIZES.includes(size)) {
  // ...
}
```

app/scripts/modules/core/src/managed/StatusBubble.tsx
```diff
@@ -36,16 +38,47 @@ const transitionConfig = {
...
+    withQuantity = true;
+    quantityPill = (
+      <span
+        className={classnames({
+          Pill: true,
+          [size]: true,
+          [`status-${appearance}`]: true,
```
erikmunson, line 54: same note re: the variadic arg syntax

app/scripts/modules/core/src/managed/StatusBubble.tsx
```diff
@@ -36,16 +38,47 @@ const transitionConfig = {
...
+        })}
+      >
+        {quantity}
+      </span>
+    );
+  }
```
erikmunson, line 60: _slightly_ subjective, but in my experience makes a meaningful impact on how code is reasoned about: i recommend moving this into a function outside the component (e.g. `getQuantityPill()`) as opposed to using `let` bindings with conditional assignment. occasionally `let` bindings are necessary — sometimes because there's no other solution to the problem, other times for performance reasons like avoiding multiple passes over an array. In most cases, though, opting for a function makes reasoning about the state you'll be in during runtime more straightforward.

if you do decide to refactor, i think tactically it would mean you'd do something like `const quantityPill = getQuantityPill(...);` and then use `!!quantityPill` where you were using `withQuantity`, and `getQuantityPill` would give you back an element or nothing.

app/scripts/modules/core/src/presentation/icons/iconsByName.ts
```diff
@@ -112,7 +112,7 @@ export const iconsByName = {
   mdDeltaResolved,
   mdDiff,
   mdError,
-  mdFlapping,
+  mdUnhappy,
```
erikmunson, line 115: 👍

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -56,6 +57,20 @@ const getResourceRoutingInfo = (
...
 
+const resourceStatusConfig: {
+  [status: string]: Pick<IStatusBubbleProps, 'appearance' | 'iconName'>;
+} = {
+  ACTUATING: { appearance: 'progress', iconName: 'mdActuating' },
+  CREATED: { appearance: 'info', iconName: 'mdCreated' },
+  DIFF: { appearance: 'info', iconName: 'mdDiff' },
+  CURRENTLY_UNRESOLVABLE: { appearance: 'error', iconName: 'mdError' },
+  ERROR: { appearance: 'error', iconName: 'mdError' },
+  PAUSED: { appearance: 'neutral', iconName: 'mdPaused' },
```
erikmunson, line 68: we currently show `PAUSED` as a `warning` state on the Infrastructure views, can we switch to that? While pausing isn't an error state, in terms of the moments where it appears to users the expectation is that in most cases it's something that is non-normal in a probably-not-great way. For example, there's something wrong so you had to pause to fix it, you're not sure what's going on so you paused to figure it out, etc. Even in cases where the reason wasn't necessarily 'bad', it's still important for anyone using the application to be aware that management is paused so they know what to expect.

```suggestion
  PAUSED: { appearance: 'warning', iconName: 'mdPaused' },
```

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -56,6 +57,20 @@ const getResourceRoutingInfo = (
...
+  RESUMED: { appearance: 'progress', iconName: 'mdResumed' },
+  UNHAPPY: { appearance: 'error', iconName: 'mdUnhappy' },
+  UNKNOWN: { appearance: 'inactive', iconName: 'mdUnknown' },
```
erikmunson, line 71: similar to `PAUSED`, this status uses the `warning` appearance on resources today. I think the same context around being 'probably not great' applies — really resources should _never_ enter this state, if they do it means there's almost certainly a bug we need to fix. It's not an error ('bad') but it's also something that's very important to draw attention to.

```suggestion
  UNKNOWN: { appearance: 'inactive', iconName: 'mdUnknown' },
```

app/scripts/modules/core/src/managed/StatusBubble.less
```diff
@@ -1,39 +1,96 @@
...
     background-color: var(--color-status-error);
   }
+
+  .Pill {
+    align-self: flex-end;
+    color: var(--color-icon-light);
+    display: inline-block;
+    font-family: 'Source Sans Pro';
+    font-weight: bold;
+    text-rendering: optimizeLegibility;
```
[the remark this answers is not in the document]
vigneshm: re: `text-rendering`

I'm not sure either. I re-used what was already available in the original `Pill` module.

app/scripts/modules/core/src/managed/StatusBubble.less
```diff
@@ -1,39 +1,96 @@
...
+    z-index: -1;
```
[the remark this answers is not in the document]
vigneshm: The `pill`/`quantity` part of `StatusBubble` comes after the icon and I've used a negative margin and a lower `z-index` to make this pill move back and *under* the icon. Otherwise, the pill would cover some parts of the lower right corner of the icon.

app/scripts/modules/core/src/managed/StatusBubble.tsx
vigneshm: Good question. I've always thought that this would only come into picture during the `actuating` step where we are displaying the `finished/total` string. Is that not the case?

app/scripts/modules/core/src/managed/StatusBubble.tsx
vigneshm: Good question. I'm going to debug this further to see if this is happening

app/scripts/modules/core/src/managed/StatusBubble.less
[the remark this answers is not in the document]
erikmunson: 👍 given that it's an SVG property i suspect it's probably unnecessary (i.e. you could get away with just the `text-bold` class).

app/scripts/modules/core/src/managed/StatusBubble.less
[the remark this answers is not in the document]
erikmunson: ah. yes ok thank you. all makes sense now.

app/scripts/modules/core/src/managed/StatusBubble.tsx
erikmunson: 🤷 medium term it'd be good to graduate this component into `core/presentation`, as it's going to be widely used across spinnaker. might not be worth doing it in this PR, but i bet we'll end up wanting to stick plain numbers on here (i'm already thinking about doing this with my pinning work)

app/scripts/modules/core/src/managed/StatusBubble.tsx
vigneshm: Totally agree. It is way more readable now!

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
vigneshm: Thanks for pointing this out. I didn't realize that we were already doing this in infrastructure view. The explanation makes a lot of sense.

app/scripts/modules/core/src/managed/StatusBubble.tsx
vigneshm: Working on this now and this PR could be merged once I confirm if it is not an issue.

vigneshm: There is an issue, I'm working on fixing it now.

vigneshm: Fixed it and ready for merge.

animating-status-bubble

vigneshm:
Fixed the animation issue and ready for merge.

animating-status-bubble

gcomstock:
> Fixed the animation issue and ready for merge.
> 
> animating-status-bubble

When does the animation play?

gcomstock:
Ah Erik filled me in (we were already on a MD office hours call) very cool!

vigneshm:
> When does the animation play?

When the resource is changing status, say from `DIFF` to `ACTUATING`. In this example though, I'm just animating all of them in a timer.

PR #8219, opened 2020-04-29
fix(buildStatus): Change Deck CI to Branch Build to align with the re…

ttomsu:
…st of the repos

It appears these GHActions are much more integrated, and it's unclear which reference _Actions_ named "Deck CI" and which reference _Jobs_ named "Deck CI". I *think* most of them reference _job names_, but I'm not 100% certain.

This change goes along with https://github.com/spinnaker/spinnaker.github.io/pull/1823

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Deck CI
+name: Branch Build
 
 on:
   pull_request:
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Spinnaker UI
 
-![Deck CI](https://github.com/spinnaker/deck/workflows/Deck%20CI/badge.svg)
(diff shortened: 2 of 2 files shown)


ttomsu:
That's correct - the Build Status page uses the _Action_ name, and this PR changes that name to the same name all the other repos use ("Branch Build").

However, this repo does far more things with Actions, and happens to have a *Job* named "Deck CI" as well. I haven't changed any other places that reference this _Job_ name, as I think that's what the other tools use, like Mergify.

maggieneterval:
@ttomsu, gotcha. Should we verify that's the case by attempting to use the `ready to merge` label on this PR?

ttomsu:
SGTM

christopherthielen:
The *Job* named 'Deck CI' is what Deck's branch protection will block on.  It is a meta-job that depends on all the other jobs passing.

PR #8221, opened 2020-04-29
chore(core): Update ui-router/react

caseyhebebrand:
Bump to @uirouter/react 1.0.3


package.json
```diff
@@ -221,7 +221,7 @@
   "resolutionsPurpose": "This temporarily chooses the right versions of these libs, which are specified in both deck AND deck-kayenta",
   "resolutions": {
     "@uirouter/react-hybrid": "1.0.0",
```
christopherthielen, line 223: Update all 4 of these resolutions, and also update the versions in the `"dependencies"` block

PR #8224, opened 2020-04-30
fix(ecs): set loadBalancedContainer to empty when creating new targetGroupMappings

pkandasamy91:
Clouddriver currently contains validation to ensure that a targetGroup is specified in order to use a loadBalancedContainer(if one is specified).  When using deck, a targetGroupMapping is specified using the previous targetGroup and loadBalancedContainer properties.

This change will reset both of those properties to be empty once the new targetGroupMapping is constructed, containing the values existing in those properties.


allisaurus:
(adding additional context from testing) previously, if you updated from using `targetGroup`/`loadBalancedContainer` to using `targetGroupMappings`, Deck would wipe out `targetGroup` but not `loadBalancedContainer`, causing validation to trigger and failing a deployment with: `targetGroup.not.nullable`. This change correctly removes both.

allisaurus:
@Mergify backport release-1.20.x

PR #8225, opened 2020-04-30
feat(aws/domain): Introduce launch templates to aws domain

caseyhebebrand:
Introduce `IAmazonLaunchTemplate` and necessary `ILaunchTemplateData` for future support of launch templates.


app/scripts/modules/amazon/src/domain/IAmazonLaunchTemplate.ts
```diff
@@ -0,0 +1,87 @@
...
+  httpEndpoint?: 'disabled' | 'enabled';
+  httpPutResponseHopLimit?: number;
+  httpsTokens?: 'required' | 'optional';
+  state?: 'pending' | 'applied';
+}
+
+export interface ITagSpecification {
+  resourceType?: string;
+  tagSet?: Array<{
+    [key: string]: string;
```
christopherthielen, line 49: Wondering if this will be confusing to people down the road.  An array of tag objects, but each tag object can only contain a single key: value pair?  Should we use the `Tag` representation that Amazon documents instead?

```
{
  "Key" : String,
  "Value" : String
}
```

app/scripts/modules/amazon/src/domain/IAmazonLaunchTemplate.ts
```diff
@@ -0,0 +1,87 @@
...
+  }>;
+}
+
+export interface ILaunchTemplateData {
+  [attribute: string]: any;
```
christopherthielen, line 54: It's best to avoid index types if possible because it makes type checking super loose.  I looked up this object type in AWS docs and it doesn't seem to suggest it can contain arbitrary keys.  Can you explain why this is needed?

caseyhebebrand: I definitely agree with you on the fact that it makes type checking loose. I would also say that we do not 'need' this. My think is that this type is dependent on a model that is still having features added to it. I felt like this helped us enforce stronger types than we currently have, while not being too strongly coupled to aws changes. 

The specific attributes are important to Spinnaker, and the rest could evolve over time. If they are important we can specifically include them as needed and if not we do not have to worry about backwards compatibility.

caseyhebebrand:
> I see a LOT of optional fields. Are all these fields really optional?

@christopherthielen Yes most fields are. I will take a second pass just to confirm though. Everything in `IAmazonLaunchTemplate` is required, but the `launchTemplateData` has a lot of optional fields. I worked with a base launch template to determine which fields are always present, and which fields only are passed when the ASG has that data.

PR #8226, opened 2020-04-30
feat(core): remove kubernetes v1 module 

maggieneterval:
Related to: https://github.com/spinnaker/governance/blob/master/rfc/eol_kubernetes_v1.md

Now that the 1.20 release branches are cut, it is safe to remove all Kubernetes V1 code since it will not be supported in 1.21.

* **feat(core): remove kubernetes v1 module**

  Removes the Kubernetes V1 module and the `kubernetes/v1` directory.

* **fix(kubernetes): fix link to logo**

  While clicking around to make sure removing the V1 code didn't break anything in V2, noticed that the Kubernetes V2 logo should be the logo, and not the icon.

* **refactor(kubernetes): merge /v2 and /shared into /src**

  Now that the `/v1` subdirectory is removed, we can refactor the kubernetes module in the style of other provider modules: one top-level `/src` directory. While we're at it, standardize on relative imports.

* **refactor(kubernetes): simplify naming of kubernetes v2 modules**

  Removes `v2` from module names, because the Kubernetes V2 provider would like to be called the Kubernetes provider from now on, please. Does not extract V2 from individual component names; we can change these as we edit these components.

* **refactor(eslint-plugin): remove now-unnecessary kubernetes-specific rules**

  Now that the Kubernetes package structure is no longer ❄️, remove the special eslint rules that were added just for this provider.


christopherthielen:
Screen Shot 2020-04-30 at 12 56 50 PM

packages/eslint-plugin/test/import-relative-within-subpackage.spec.js
```diff
@@ -57,25 +57,5 @@ ruleTester.run('import-relative-within-subpackage', rule, {
         'Do not use an alias to import from core/subpackage from code inside core/subpackage. Instead, use a relative import',
       ],
     },
-    {
```
christopherthielen, line 60: These are the tests for this special case, but there should be changes to the eslint rule itself, I think.

app/scripts/modules/kubernetes/src/index.ts
```diff
@@ -1,2 +1,6 @@
-export * from './v1';
-export * from './v2';
+export { KUBERNETES_MODULE } from './kubernetes.module';
```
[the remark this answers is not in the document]
maggieneterval: Sure!

packages/eslint-plugin/test/import-relative-within-subpackage.spec.js
maggieneterval: DUH sorry. The actual kubernetes-specific package parsing logic is removed now.

PR #8232, opened 2020-05-01
feat(core/presentation): add 'pin' icon

erikmunson:
For the pinning PRs that will be coming in momentarily 

(cc @gcomstock )


gcomstock:
Sweet!

PR #8235, opened 2020-05-01
feat(core/presentation): change <textarea class="code"> font size to default

christopherthielen:
It seems our preference is to move away from teeny fonts.  This proposal changes the `<textarea class="code">` blocks to use the standard font size instead of 80%.

WDYT?  Too much?  @maggieneterval , this would affect k8s manifest screens.

@gcomstock PTAL

Before:

Screen Shot 2020-04-30 at 6 22 53 PM

After:

Screen Shot 2020-04-30 at 6 23 15 PM


maggieneterval:
@erikmunson I also zoom in to at least 125%, even when I'm wearing my glasses ❤️

gcomstock:
YES

PR #8237, opened 2020-05-01
feat(core/presentation): allow custom max widths on Modal, bugfixes

erikmunson:
In moving to `Modal` for some things, we've observed that the current default behavior of trying to grow in width until the max of `1400px` is often not desirable — in particular when working with things like forms or confirmation modals. The challenge is that we want to strongly encourage folks to design their modal content to tolerate flexing from mobile sizes up to larger widths semi-fluidly, so a good solution should ideally go with that instead of against it.

What I've done here is experiment a bit with how we can try to define _maximum_ widths, while still:
1. flexing down to mobile size without choppy breakpoint steps (i.e. we don't want one fixed width per breakpoint)
2. always attempting to fill the space we have up to the max width as opposed to forcing consumers of Modal to 'push' the modal up to whatever size they think is right. This is often challenging with form layouts because they tend to shrink to fix their _minimum_ content, and in a modal context that can mean that if you don't go with a fixed width things shrink down to uncomfortably small sizes

My approach was to use a very simplistic grid layout, because the handy `minmax` row sizing syntax has exactly the behavior we're looking for. If you're not familiar with `minmax` it effectively says "please attempt to fill the available space up to the max, but allow yourself to fluidly flex down to the min when space is unavailable". In this case the min is `min-content` which is the standard DOM layout behavior I described on #2 of finding the minimum size the inner content will tolerate. The max is whatever the new `maxWidth` prop is set to. The default for the prop is the max width we've had all along — `1400px`. You can pass either a number (in pixels) or a string which will work with any valid CSS unit (e.g. you could do `100%`).

One thing I'm seriously considering is augmenting this API to have a couple pre-chosen widths, I think one would be the current default (which we could call `large`) and the other would be roughly the width of our existing smaller modals (~`500px`, we could call this `medium` maybe?). I do think there will always be reasons to use a custom width because the reality is a modal is a container for many types of content. Happy to hear thoughts on any of those ideas if you have them.

There's also some random bugfixes and tidying wrapped up in here, but in separate commits. Nothing exciting.


christopherthielen:
This sounds pretty good to me.  I'm big 👍 for named presets.

christopherthielen:
can we get rid of the css modules?

erikmunson:
> can we get rid of the css modules?

heh, i knew you were going to say that. yes I am very motivated to get rid of it, just still working through the MD stuff I need to ship before I can do a tech debt pass on Environments (which has some module code alan brought in from greg's prototype), at which point I'll also yank this grossness out.

gcomstock:
> > can we get rid of the css modules?
> 
> heh, i knew you were going to say that. yes I am very motivated to get rid of it, just still working through the MD stuff I need to ship before I can do a tech debt pass on Environments (which has some module code alan brought in from greg's prototype), at which point I'll also yank this grossness out.

I'll gladly refactor that prototype to use whatever css solution we land on.

PR #8239, opened 2020-05-01
feat(core): Integrated v1 Layout

caseyhebebrand:
Integrating all new layout components into the existing views:
1. vertical `ApplicationNavigation` replaces  current 2nd and 3rd level navigation
2. `FilterSearch` component integrated into all filter sections ([commit](https://github.com/spinnaker/deck/pull/8239/commits/60c2da9ca748ea62b07ac3b84bb2c149bc8620d3))
3. Converting the details panel to a details drawer
4. A few z-index tweaks for proper layering over the details drawer

Screen Shot 2020-06-01 at 3 38 01 PM


app/scripts/modules/core/src/cluster/filter/FilterSearch.tsx
```diff
@@ -5,27 +5,26 @@ import './FilterSearch.less';
...
+    <form className="horizontal middle space-between" role="form">
+      <div className="form-group nav-search">
+        <input
+          type="search"
+          className="form-control input-med sp-form"
+          value={value}
+          onBlur={onBlur}
+          onChange={onSearchChange}
+          placeholder="Search"
+          ng-model={value}
```
christopherthielen, line 24: This is a surprising addition.  Can you explain how this works?  AngularJS shouldn't be compiling this react template, as far as I understand.

app/scripts/modules/core/src/presentation/main.less
```diff
@@ -293,27 +279,33 @@ html {
   }
 
   .detail-content {
-    overflow: visible;
-    margin: 0;
-    @media (max-width: 768px) {
-      top: 170px;
-      width: 95%;
+    z-index: var(--layer-high);
+    position: fixed;
```
christopherthielen, line 283: have you tested if this works with projects page and/or when a blesk is active?

app/scripts/modules/core/src/presentation/main.less
```diff
@@ -1124,13 +1116,19 @@ textarea.code {
   }
 }
 
+.modal-open {
+  .modal {
+    z-index: var(--layer-critical) !important;
```
christopherthielen, line 1121: obligatory "is `!important` really the best option to use here?" question

app/scripts/modules/core/src/cluster/filter/FilterSearch.tsx
caseyhebebrand: TL;DR This is not needed if I migrate this whole section to React :)

When an Angular template is a parent of this React component, the controller is set up in a way that needs this in order to update the reference model. The `updateClusterGroups` function in the controller has to run after any filter or search is updated. This uses the controller's context of the `FilterSearchModel` which will use the old filter value. `ng-model` will force the context of the `FilterSearch` to reference this value at all times.

app/scripts/modules/core/src/presentation/main.less
caseyhebebrand: No I have not, and will definitely do that.

app/scripts/modules/core/src/presentation/main.less
caseyhebebrand: I hate using `!important`, but there was some overrides happening here at the `bootstrap` level.

caseyhebebrand:
> Nice work. Do you have a feel for the level of effort it would take to migrate the cluster filter and securitygroup filters from Angular to React? Now would be a great time to migrate, and it may even reduce overall complexity by eliminating any angular/react compatibility kludges.
> 
> I have the same question about the insight layout component.

@christopherthielen The filters won't be bad. Insight layout maybe a bit worse. Honestly, I am just going to do it because the are much more pros than cons. The only downside is a few extra days to landing this, but that is not a big deal.

robzienert:
Love it!

Two things:

1. The hamburger control does not work on the Spinnaker landing page (toggle-able, but it doesn't do anything).
2. Wrapping issues on smaller resolutions.

Screen Shot 2020-06-01 at 3 41 21 PM

rahulsom:
+1 I like the ease of getting to menu options.

I’m slightly color-blind and the contrast between the green of the reload icon and the blue of the background is [low](https://contrast-ratio.com/#%238aad6a-on-%23d7e8ed).

This can be compensated by making the green larger or increasing the contrast.

dreynaud:
I don't know if it's because of the custom banner but the property sidebar eats into the top navbar.

image

dreynaud:
If I go from Load Balancers to Clusters, I lose the filter selector:

image

dreynaud:
Even at a respectable 1720 window width, the server group sidebar covers the right hand side of the server group display (with the LB indicator) and the create server group button. Not sure if that's intentional but that's a bit disorienting.

image

caseyhebebrand:
> Even at a respectable 1720 window width, the server group sidebar covers the right hand side of the server group display (with the LB indicator) and the create server group button. Not sure if that's intentional but that's a bit disorienting.

This is intentional. We wanted to the details sidebar to be a drawer, rather than inset on the page. This way it does not affect the width of the main content or the width of the details when the screen is too small (a step towards mobile responsiveness).

Do you feel like this will continue to be a jarring experience or will you get used to treating the details section like a drawer?

dreynaud:
> Do you feel like this will continue to be a jarring experience or will you get used to treating the details section like a drawer?

Got it, I'll see after some usage. Wanted to make sure it was intentional, as that seems to be a departure from current conventions

christopherthielen:
@maggieneterval yeah we're good

vigneshm:
The following questions aren't really a blocker. Just wanted to call them out to bring to your attention and can be handled later.

1. When there are notices/banners at the top, the experience looks like the following which doesn't look like intended.

alerts

2. The whole alignment of the application icon, application name and the refresh icon seems slightly off to me. I can't quite say what is the exact problem though :)

refresh-icon

3. I see that there's an option to hide the left nav in the code. But I can't quite see how to collapse it in the UI? Is the option hidden for now?

4. This is a minor naming nit and this is probably an artifact of the older UI model, but I see the term `category` used to mean different things within navigation that seems confusing to me. `application.dataSources[i].category` seems to yield the `NavSection` which contains `NavRoute`s. However, each `NavRoute` renders a `NavCategory` that takes in a `category` prop which is just the `dataSource`. It would probably make it less confusing if we rename `NavCategory` to something like `NavItem` that takes in the `dataSource` as prop.

app/scripts/modules/core/src/application/nav/NavCategory.tsx
```diff
@@ -1,19 +1,20 @@
...
   isActive: boolean;
-  app: Application;
+  isCollapsed: boolean;
 }
 
-export const NavCategory = ({ app, category, isActive }: INavCategoryProps) => {
+export const NavCategory = ({ app, category, isActive, isCollapsed }: INavCategoryProps) => {
   const { alerts, badge, iconName, key, label } = category;
 
   const { data: badgeData } = useDataSource(app.getDataSource(badge || key));
```
christopherthielen, line 20: this is really confusing to me.  If category is already a DataSource why are we getting the DataSource here?

app/scripts/modules/core/src/application/nav/NavCategory.tsx
```diff
@@ -1,19 +1,20 @@
...
 export interface INavCategoryProps {
+  app: Application;
   category: ApplicationDataSource;
   isActive: boolean;
-  app: Application;
+  isCollapsed: boolean;
 }
 
-export const NavCategory = ({ app, category, isActive }: INavCategoryProps) => {
+export const NavCategory = ({ app, category, isActive, isCollapsed }: INavCategoryProps) => {
```
christopherthielen, line 17: This component is pretty confusing to me.  Is it a Category?  A link/nav item? A data source?  Is a category a collection of items? What is the relationship between the datasource prop and the other datasource inside the component?   

It seems like this might be easier to grok if we separated out the data flow from and extract a purely presentational component, something like:

```js
<DataSourceThing 
  dataSource={datasource} 
  otherDataRelatedProp={otherDataRelatedProp}
  render={navItemMarkupProps => <NavItemMarkupOnly {...navItemMarkupProps} />} 
/>
<PagerDutyThing render={navItemMarkupProps => <NavItemMarkupOnly {...navItemMarkupProps} />} />
``` 
obviously with better names

app/scripts/modules/core/src/application/nav/verticalNav.less
```diff
@@ -92,3 +112,33 @@
     border-top: 1px solid var(--color-white);
   }
 }
+
+.collapsed {
```
christopherthielen, line 116: Careful, a top level selector named `collapsed` seems likely to collide with something else.

app/scripts/modules/core/src/cluster/filter/FilterSearch.tsx
```diff
@@ -5,27 +5,25 @@ import './FilterSearch.less';
...
-            placeholder="Search"
-          />
-        </div>
-        {helpKey && <HelpField id={helpKey} placement="right" />}
-      </form>
-    </div>
-  );
-};
+export const FilterSearch = ({ helpKey, onBlur, onSearchChange, value }: IFilterSearchProps) => (
+  <div className="filter-search sp-margin-s-right">
```
christopherthielen, line 14: as a convention thing, I've been naming the CSS classes that create a component scope the same as the component itself, i.e., `className="FilterSearch"`.  My reasoning is that it's easy to scan for in the DOM and indicates the root element of a component.

app/scripts/modules/core/src/filterModel/FilterCollapse.less
```diff
@@ -1,10 +1,41 @@
...
+    background-color: var(--color-accent-g2);
+    height: 32px;
+    font-size: 16px;
+    margin-top: 6px;
+
+    @media (min-width: 768px) and (max-width: 992px) {
+      width: 104px;
+    }
+
+    @media (min-width: 992px) {
```
christopherthielen, line 18: this is a lot of breakpoints.  If it's trying to utilize more space as the screen gets wider, could we use flexbox instead to get a nice smooth transition?  

Also, do we have CSS variables for breakpoint widths?  If not, we probably should.

app/scripts/modules/core/src/filterModel/FilterCollapse.tsx
```diff
@@ -14,30 +14,35 @@ export class FilterCollapse extends React.Component<{}> {
...
-              className="btn btn-xs btn-default pull-left unpin clickable"
-              onClick={() => this.onClick(false)}
-              style={{ display: ReactInjector.insightFilterStateModel.filtersExpanded ? 'inherit' : 'none' }}
-            >
-              <i className="fa fa-backward" />
-            </button>
-          </Tooltip>
-        </h3>
+      <div className="filters-toggle layer-medium">
+        {!ReactInjector.insightFilterStateModel.filtersExpanded && (
```
christopherthielen, line 18: might as well extract this into a variable

app/scripts/modules/core/src/header/SpinnakerHeader.tsx
```diff
@@ -2,25 +2,45 @@ import React from 'react';
...
     super(props);
     this.state = {
       navExpanded: !this.isDevicePhoneOrSmaller(),
+      verticalNavExpanded:
+        !CollapsibleSectionStateCache.isSet('verticalNav') || CollapsibleSectionStateCache.isExpanded('verticalNav'),
+      isApplicationView: ReactInjector.$uiRouter.globals.current.name.includes('applications.'),
     };
   }
 
+  public componentDidMount(): void {
```
christopherthielen, line 33: I would much prefer that this gets migrated to a functional component

app/scripts/modules/core/src/insight/InsightLayout.tsx
```diff
@@ -20,6 +20,9 @@ export const InsightLayout = ({ app }: IInsightLayoutProps) => {
     setExpandFilters(!expandFilters);
   };
 
+  const { state: currentState } = useCurrentStateAndParams();
+  const showDetailsView = Boolean(Object.keys(currentState.views).find(v => v.indexOf('detail@') !== -1));
```
christopherthielen, line 24: interesting  🤔

app/scripts/modules/core/src/presentation/main.less
```diff
@@ -195,7 +189,7 @@ html {
       margin-left: 0;
     }
     .dl-flex {
-      .dl-wide;
+      .dl-narrow;
```
christopherthielen, line 192: I don't understand what a LESS selector like this does? It doesn't seem to be applying anything. Can you shed some light on it for me?

app/scripts/modules/core/src/search/global/globalSearch.less
```diff
@@ -28,7 +28,7 @@ global-search {
     right: 0;
     margin: 2px 19px;
     border-radius: 0;
-    z-index: 10000;
+    z-index: 10001;
```
christopherthielen, line 31: 🤮

app/scripts/modules/core/src/task/tasks.html
```diff
@@ -43,7 +40,7 @@ <h3>Task History</h3>
           <span class="glyphicon glyphicon-remove" uib-tooltip="clear search filter"></span>
         </a>
       </div>
-      <div class="col-md-3 text-right" style="padding-left: 0">
+      <div class="col-md-3 text-right" style="padding-left: 0;margin-left: auto">
```
christopherthielen, line 43: worth migrating this layout over to flexbox?  Also, I noticed this problem when text is entered

Screen Shot 2020-06-22 at 12 36 35 PM


old behavior:

Screen Shot 2020-06-22 at 12 38 52 PM

app/scripts/modules/core/src/widgets/notifier/notifier.component.less
```diff
@@ -4,16 +4,22 @@
   position: fixed;
   bottom: 5px;
   width: 100%;
-  z-index: 3;
+  z-index: var(--layer-critical);
```
christopherthielen, line 7: 😎

app/scripts/modules/core/src/application/nav/NavCategory.tsx
christopherthielen: funny, I just noticed that @vigneshm had almost the exact same feedback

app/scripts/modules/core/src/application/nav/NavCategory.tsx
caseyhebebrand: For some data sources, the "executable" versions of that data source are gathered differently. For example, a pipeline's data comes from the `executions` dataSources, but the number of running pipelines comes from `runningExecutions`. The `runningExecutions`, are a unique data source in itself.

This confused me at first too, but this is how the dataSource is configured initially. There is an attribute called `badge`, which stores the key of the supplementary data source that has the executable tasks.

app/scripts/modules/core/src/application/nav/NavCategory.tsx
caseyhebebrand: I can change the name to `NavItem`, I was trying to stay consistent with legacy naming conventions, but if it confusing I am happy to change it.  Historically, the category was equivalent to a datasource.

I think it is too much abstraction to separate into 2 components. All this one is doing is subscribing to the data source, and then rendering the item with the running count, icon, name and alerts.

app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
```diff
@@ -50,32 +58,56 @@ export const ApplicationNavigation = ({ app }: IApplicationNavigationProps) => {
     };
   }, []);
 
+  React.useEffect(() => {
```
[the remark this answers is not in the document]
caseyhebebrand: The global state of the cache changes when the toggle button is clicked, however this component will not re-render unless I subscribe to the cache. 

I think context would probably work. They are not in the same component tree, but I can use the cache value and provide that context to the application navigation component

app/scripts/modules/core/src/application/nav/verticalNav.less
caseyhebebrand: I will rename with something more specific

gcomstock:
> The following questions aren't really a blocker. Just wanted to call them out to bring to your attention and can be handled later.
> 
> 1. When there are notices/banners at the top, the experience looks like the following which doesn't look like intended.
> 
> alerts
> 
> 1. The whole alignment of the application icon, application name and the refresh icon seems slightly off to me. I can't quite say what is the exact problem though :)
> 
> refresh-icon
> 
> 1. I see that there's an option to hide the left nav in the code. But I can't quite see how to collapse it in the UI? Is the option hidden for now?
> 2. This is a minor naming nit and this is probably an artifact of the older UI model, but I see the term `category` used to mean different things within navigation that seems confusing to me. `application.dataSources[i].category` seems to yield the `NavSection` which contains `NavRoute`s. However, each `NavRoute` renders a `NavCategory` that takes in a `category` prop which is just the `dataSource`. It would probably make it less confusing if we rename `NavCategory` to something like `NavItem` that takes in the `dataSource` as prop.


I agree that the refresh icon feels orphaned. Thanks for the affirmatioN! I owe casey a new icon of the app + refresh. We can then just have a user click the new icon to perform a refresh.

app/scripts/modules/core/src/presentation/main.less
caseyhebebrand: In this case it was controlling 2 different margins in areas of the DOM based on if the filters were expanded or not. `dl-flex` and `dl-narrow` are used in a lot of places so rather than remove this I just controlled streamlined everything to `dl-narrow` in order to have baseline for the rets of the layout changes.

app/scripts/modules/core/src/task/tasks.html
caseyhebebrand: I have those changes coming in my next push :)

caseyhebebrand:
@vigneshm I wanted to reply to your comments: 
> 1. When there are notices/banners at the top, the experience looks like the following which doesn't look like intended.
> 1. The whole alignment of the application icon, application name and the refresh icon seems slightly off to me. I can't quite say what is the exact problem though :)
> 1. I see that there's an option to hide the left nav in the code. But I can't quite see how to collapse it in the UI? Is the option hidden for now?
> 2. This is a minor naming nit and this is probably an artifact of the older UI model, but I see the term `category` used to mean different things within navigation that seems confusing to me. `application.dataSources[i].category` seems to yield the `NavSection` which contains `NavRoute`s. However, each `NavRoute` renders a `NavCategory` that takes in a `category` prop which is just the `dataSource`. It would probably make it less confusing if we rename `NavCategory` to something like `NavItem` that takes in the `dataSource` as prop.

I think you may be looking at the version that doesn't have the internal overrides, which is probably why questions 1-3 are here. Try this: http://go/spin/index.verticalnav_061720.html
1. That has been fixed, I think you have an older version link. 
2. This could be because you are seeing our internal header, and those changes are part of another PR.
3. There should be a hamburger menu next to `SPINNAKER` in the header. If you are not seeing that, then you probably have the old link since we override the OSS header. You can try running the OSS version locally to see it there. 
4. @christopherthielen also brought up these concerns. I was trying to stay close to the historical naming, but I am in favor of renaming to `NavItem` with a dataSource prop. It definitely makes more sense.

app/scripts/modules/core/src/filterModel/FilterCollapse.less
caseyhebebrand: we currently only have a mobile breakpoint. I can set up a couple more for s, m, l, etc screens as well.

app/scripts/modules/core/src/search/global/globalSearch.less
caseyhebebrand: yeah... unfortunately using the critical variable (aka `z-index: 1000`) was not cutting it.

app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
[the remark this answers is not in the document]
christopherthielen: >They are not in the same component tree

recoiljs looks very useful for cases like this

app/scripts/modules/core/src/presentation/main.less
christopherthielen: my question is about the syntax used.  What does this *do*?
```
.dl-flex {
  .dl-narrow;
}
```

I would expect some rules in side there such as 
```
.dl-flex {
  .dl-narrow {
    width: 12345px
  }
}
```

app/scripts/modules/core/src/search/global/globalSearch.less
christopherthielen: Wondering... can you do:
```  
z-index: calc(var(--layer-critical) + 1);
```
still not good but perhaps easier to understand the intent?

gcomstock:
Screen Shot 2020-06-22 at 2 47 33 PM

@vigneshm @caseyhebebrand  How does this feel?

christopherthielen:
> Does data ever get so stale that it's worth showing as red?

Yes, it sure does.  I think this is critical UI and I see it go red regularly.  If we want to remove the color, then perhaps a high visibility icon could replace it.

caseyhebebrand:
> Screen Shot 2020-06-22 at 2 47 33 PM
> 
> @vigneshm @caseyhebebrand How does this feel? I'm not sure we need the current red/green/spinning states of the refresh button. Does data ever get so stale that it's worth showing as red? Is this an outdated feature or do I just not understand the use cases. @christopherthielen Do you have any context on the history of the app refresh button?

@gcomstock I agree that the color indication for staleness is important. I would suggest either utilizing the background color for the  icon to indicate this, or possibly the use of a secondary icon in the place of the one you have to see once the app's data is stale.

app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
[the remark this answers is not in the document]
caseyhebebrand: Just looked into context and recoiljs. Both have challenges. 

Recoil requires the provider to be at the root of that app. React context requires the provider to be above both components so they are in the same tree (which is the root of the app in this case). This is currently in angular and I am not sure how ready it is for a refactor. I will continue digging into a workaround - a partial refactor of `spinnaker.component` could work. I just need the container contents to be in react

gcomstock:
Screen Shot 2020-06-22 at 3 56 19 PM


Thanks for the context! I think we should use different icons to represent fresh/stale/fetching. The different bg color IMO more strongly implies that something is wrong with the app itself. I reiterated the application name in the hover state so we can show the same hover text when the menu is collapsed. If this looks good I will open a PR with the new icons. Let me know!

caseyhebebrand:
> I reiterated the application name in the hover state so we can show the same hover text when the menu is collapsed. If this looks good I will open a PR with the new icons. Let me know!

That is a good point about the bg color. I think the refresh icon is a bit more clear to me for the stale view (below), especially since there is no color-warning anymore. I love the checkbox for the fresh app. Curious what others think too, bur overall I like the use of 2 different icons.  

image

gcomstock:
Good point!

app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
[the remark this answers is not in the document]
christopherthielen: > This is currently in angular and I am not sure how ready it is for a refactor

ah bummer. yeah, migrating that `spinnaker.component` would be a win.  The one sticky point will be handling the spinnakerHeaderTemplate include.

fwiw, recoil is still in incubation, but it looks pretty slick.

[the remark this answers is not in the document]
christopherthielen: 🤔 actually, this might be really easy.

Since `SpinnakerHeader` is @Overridable I think it wires in the angularjs template override automatically.  It might be as easy as replacing the `ng-include` with the `<SpinnakerHeader />`.  You can check deck-nflx to see if our custom header gets rendered.

The rest of those components used in that template are already downgraded React -> AngularJS components so this might be quick and easy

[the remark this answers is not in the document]
christopherthielen: ... and to continue talking to my self, I see you have a PR out there already to migrate that component :) #8366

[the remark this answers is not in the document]
caseyhebebrand: Recoil looks pretty lightweight, so I got excited to try it out. I think we can make this happen, the overridable is the thing I was missing, but the rest was very easy to migrate. Thanks for all the input

jamroks:
Hi, 

Just a regular spinnaker user here, i see a lot will change visually speaking , i was wondering are those sections : 
- Build , Branch, PullRequest on the sidebar, now part of regular oss spinnaker or are those view internal to netflix ?

caseyhebebrand:
> Hi,
> 
> Just a regular spinnaker user here, i see a lot will change visually speaking , i was wondering are those sections :
> 
> * Build , Branch, PullRequest on the sidebar, now part of regular oss spinnaker or are those view internal to netflix ?

cc: @gal-yardeni

gal-yardeni:
@jamroks unfortunately not yet. This is Netflix specific piece - the UI code is internal, but we have plans to open source it in the future. If you are interested, you can join us at the CI sig and we can talk about it more there.

jamroks:
Okay Got it , thanks @caseyhebebrand 
And thank you @gal-yardeni  for the answer. I might take you up on this CI SIG thing is there any link to scheduled meeting for this sig ?

PR #8241, opened 2020-05-02
fix(core/managed): use smaller icon size for expand/collapse arrows

erikmunson:
itty bitty icon tweak now that Greg has changed the size of the paths for expand/collapse arrows to fill more of their svg canvas.

 i would include screenshots, but in an exciting turn of events switching to the `extraSmall` icon size actually look **_identical_** to the way that things looked before the SVG path itself was changed (back when we were using the `medium` size). The paths did switch to rounded ends, but it's not really that visible at that size. Props to Greg for working his designer magic! 🧙‍♂️ 

cc @gcomstock


gcomstock:
Dumb luck!

PR #8243, opened 2020-05-02
feat(core/managed): hide artifact name when there's only one

erikmunson:
The list of versions on the left side of Environments has a fair amount of information already, and we're about to add even more with some curated status bubbles for specific states. Right now all the apps using Managed Delivery are using a single artifact in their manifest, and I expect a very large percentage will continue to be single-artifact out into the future.

Knowing that, one way we can cut down on visual noise is to hide the artifact name altogether if there's only one. It's still visible in the details if you need it.

Before:
Screen Shot 2020-05-02 at 3 47 07 PM

After:
Screen Shot 2020-05-02 at 3 47 13 PM


erikmunson:
fyi @gcomstock

gcomstock:
Sweet!

PR #8244, opened 2020-05-04
feat(bake): Mark the bake stage as an artifact producer

jervi:
This Just Works™ because the bake stage in orca already puts artifacts in the context and has a `bindProducedArtifacts` task.


maggieneterval:
@jervi just curious how you are using this in practice; to consume produced artifacts in a downstream stage?

jervi:
We are not using this as the feature isn't available, but I discovered that it was supported in Orca, so I saw no reason not to enable it in Deck.

maggieneterval:
@jervi I'm hesitant to enable it without a specific use case in mind because many cloud providers don't yet support artifacts, so this would add a Produces Artifacts section to the stage that for many would not be usable. Happy to reconsider if there's a particular use case you're trying to unblock, but if it's just because we can then I'm leaning toward not enabling it. Let me know if that makes sense!

jervi:
No problem, @maggieneterval 👍 It's not blocking us, I just thought it made sense to enable it.

PR #8249, opened 2020-05-05
feat(core/presentation): Add a formik hook to save/restore data

christopherthielen:
Adds a formik hook to save/restore mutually e…xclusive form fields when toggling between two or more modes in a form
See the .spec.tsx file for example usage

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/hooks/useSaveRestoreMutuallyExclusiveFields.hook.ts
+++ b/app/scripts/modules/core/src/presentation/forms/hooks/useSaveRestoreMutuallyExclusiveFields.hook.ts
@@ -1,0 +1,69 @@
+import { FormikProps } from 'formik';
+import { useEffect, useState } from 'react';
+import { usePrevious } from '../../hooks';
+import { get } from 'lodash';
+
+/**
+ * Sometimes a form allows the user to choose between multiple mutually exclusive sets of fields.
+ * This hook saves and restores mutually exclusive fields when the user interactively toggles between them.
+ *
+ * For example, an order form may have separate fields for ordering a pizza or a sandwich. When the user selects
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/presentation/forms/hooks/useSaveRestoreMutuallyExclusiveFields.hook.ts
```diff
@@ -0,0 +1,69 @@
...
+  // Whenever the user switches the form mode source, clear out the
+  // previous form mode's exclusive field values and restore the new mode's values.
+  useEffect(() => {
+    if (!!previousFormMode && currentFieldSetKey !== previousFormMode) {
+      const fieldsToSave = mutuallyExclusiveFieldSets[previousFormMode] ?? [];
+      const fieldsToRestore = mutuallyExclusiveFieldSets[currentFieldSetKey] ?? [];
+
+      const savedDataCopy = { ...savedData };
+      fieldsToSave.forEach(path => (savedDataCopy[path] = get(formik.values, path)));
+      setSavedData(savedDataCopy);
```
vigneshm, line 63: I generally think of using the following form of data structure to store the state for multiple modes. I'm curious on why you chose to save all the keys in the same level. Is it 'cos when someone uses it accidentally to store non-mutually exclusive fields, we want to restore the latest value for those fields?

```js
const savedData: {[fieldSetId: string]: {[path: string]: any}};
```

christopherthielen: great point!  I'll update the PR with your suggestion.

christopherthielen: With your suggestion, both pizza and sandwich can have their own copy of a `cheese` property.

PR #8252, opened 2020-05-06
feat(mocks): Add Launch Template data mocks

caseyhebebrand:
Introduce base and customizable mock data. This will be used in test fixtures through development of the launch template feature.


packages/mocks/src/amazon/mockAmazonBlockDeviceMapping.ts
```diff
@@ -0,0 +1,13 @@
+import { IEbsBlockDevice, IBlockDeviceMapping } from '@spinnaker/amazon';
+
+export const createMockEbsBlockDevice = (size?: number, type?: string): IEbsBlockDevice => ({
```
christopherthielen, line 3: minor nitpick: when creating apis like this, I prefer to see an object with optional keys over a series of optional function parameters.   It is more future proof and allows the caller to provide (for example) a custom `type` without a custom `size`.

caseyhebebrand: Thats good feedback I will incorporate that before merging.

PR #8265, opened 2020-05-07
feat(amazon): Launch Template details section

caseyhebebrand:
If a server group has a launch template, show the section in the details panel. 

Screen Shot 2020-05-07 at 10 51 50 AM


app/scripts/modules/amazon/src/serverGroup/details/sections/LaunchTemplateDetailsSection.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const tags = (description || '').split(', ');
+    tags.forEach((tag: string) => {
+      const keyVal = tag.split('=');
+      if (keyVal.length === 2 && keyVal[0] === 'ancestor_name') {
+        name = keyVal[1];
+      }
+    });
+
+    return name;
+  };
```
vigneshm, line 19: Wondering if there's a good place for this utility. I saw something [similar earlier](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/amazon/src/serverGroup/details/sections/LaunchConfigDetailsSection.tsx#L24-L34) and they look very similar.

caseyhebebrand: Yeah maybe I will pull it out into a `serverGroup/utils.ts` file.

PR #8269, opened 2020-05-11
refactor(core): Convert InsightLayout to React

caseyhebebrand:
Refactoring `InsightLayout` to a React component. There are angular deps on this component, so the `react2angular` version is also exposed.


app/scripts/modules/core/src/application/ApplicationComponent.tsx
```diff
@@ -68,7 +68,7 @@ export class ApplicationComponent extends React.Component<IApplicationComponentP
           </div>
         )}
         <div className="container scrollable-columns">
-          <UIView className="secondary-panel" name="insight" />
+          <UIView className="secondary-panel" name="insight" app={app} />
```
christopherthielen, line 71: This isn't idiomatic use of UIView.  I'd prefer the `app` be wired using the resolve from the router because that's how we wire all the other resolve data right now

PR #8272, opened 2020-05-12
refactor(core): Convert Cluster filters to react

caseyhebebrand:
Create `ClusterFilter` react component. Remove unused angular code, and ensure backward compatibility. 

Note: `FilterCheckbox` component already existed in two places. This PR just pulls it out into its own file so the load balancers, clusters, and functions can use it in their filter sections.


app/scripts/modules/core/src/cluster/filter/ClusterFilters.tsx
```diff
@@ -0,0 +1,334 @@
...
+    stack: ['(none)'].concat(getHeadingsForOption('stack')),
+  });
+
+  useObservable(ClusterState.filterService.groupsUpdatedStream, () => {
+    setTags(ClusterState.filterModel.asFilterModel.tags);
+    setSortFilter(ClusterState.filterModel.asFilterModel.sortFilter);
+  });
+
+  useOnStateChanged(() => {
+    ClusterState.filterModel.asFilterModel.activate();
```
christopherthielen, line 67: Should this be checking if the new state is actually the Clusters State?

app/scripts/modules/core/src/cluster/filter/ClusterFilters.tsx
```diff
@@ -0,0 +1,334 @@
...
+  const { serverGroups } = app;
+  const { data: serverGroupData, loaded: clustersLoaded } = useDataSource(serverGroups);
+
+  const [tags, setTags] = React.useState(ClusterState.filterModel.asFilterModel.tags);
+  const [sortFilter, setSortFilter] = React.useState<ISortFilter>(ClusterState.filterModel.asFilterModel.sortFilter);
+  const [labelFilters, setLabelFilters] = React.useState<ILabelFilter[]>(
+    trueKeyObjectToLabelFilters(sortFilter.labels),
+  );
+
+  const labelsMap = buildLabelsMap(serverGroupData);
```
christopherthielen, line 45: May want to memoize this for when serverGroupData is large. 
```suggestion
  const labelsMap = useMemo(() => buildLabelsMap(serverGroupData), [serverGroupData]);
```

christopherthielen:
Please fix the commit message `Make input controlled` before rebasing and merging

app/scripts/modules/core/src/cluster/filter/ClusterFilters.tsx
caseyhebebrand: That is a good point. The previous version does not do that, but adding that check makes sense from an efficiency standpoint.

PR #8273, opened 2020-05-12
refactor(core): legacy artifacts cleanup

maggieneterval:
This is a final tidying-up of the code we no longer need, now that we can remove the legacy (pre-"rewrite") artifacts UI. It's been 14 months since the [original PR](https://github.com/spinnaker/deck/pull/6634) to add the "rewrite" UI, so in the future hopefully we can get better about ensuring there's a timely plan to bring new features across the finish line, and prevent the type of technical debt that was incurred in order to support both artifacts experiences simultaneously.

This PR appears huge, but each of the ten commits is self-contained and hopefully well-explained, so I hope it isn't crazy annoying to review. 🙏 🙏 🙏 

* refactor(core): remove `legacyArtifactsEnabled` setting 

  Spinnaker 1.21 will not include the legacy artifacts UI. The `legacyArtifactsEnabled` flag was intended as an escape hatch for one release only while any issues were worked out with the standard artifacts UI. This commit removes the flag and makes STANDARD the only possible artifacts mode. The following commits will clean up consuming code now that only one artifacts mode is possible.

* refactor(core): remove legacy Expected Artifacts components 

  Removes now-unused components and logic from legacy Expected Artifacts and Triggers configuration components.

* refactor(kubernetes): remove legacy artifacts UI from KubernetesV2RunJobStageConfig 

* refactor(core): clean up StageArtifactSelectorDelegate 

  Removes PreRewriteStageArtifactSelector and its corresponding props from StageArtifactSelectorDelegate. Adds a comment marking the delegate as deprecated, since now that there is only one UI for stage artifact selection, it is probably cleaner to just include that component inline in stage configs rather than adding an extra level of indirection that just wraps the component in a StageConfigField, which we're trying to migrate away from anyway.

* refactor(core): stop threading updatePipeline to stage config forms 

  With the removal of the legacy stage artifact selection UI, we no longer need to update the pipeline from individual stage config forms.

* refactor(kubernetes): remove legacy manifest artifact binding component 

  With the removal of the legacy artifacts UI, we can remove the component responsible for deciding which manifest artifact binding UI to render, and replace consumers with the standard component.

* refactor(core): remove legacy Produces Artifacts UI 

* refactor(core): remove ArtifactsModeService 

  There are no remaining clients of ArtifactsModeService.

* refactor(core): remove unused artifact and expected artifact angular components 

  The legacy Produces Artifacts UI was the only remaining consumer of these components.

* refactor(core): remove legacy angular artifacts components 

  With the removal of legacy UI components that were written in Angular, we can now remove each artifact type's Angular templates and controllers. We can also remove the logic that was merging the artifact configs that included these angular components with the standard configs.


app/scripts/modules/core/src/artifact/react/StageArtifactSelectorDelegate.tsx
```diff
@@ -3,23 +3,21 @@ import React from 'react';
...
-import { ArtifactsMode, ArtifactsModeService } from '../ArtifactsModeService';
 
 interface IStageArtifactSelectorDelegateProps {
   helpKey?: string;
   label: string;
   fieldColumns?: number;
 }
 
+// Please consider this component deprecated. Use StageArtifactSelector
+// directly instead.
```
christopherthielen, line 14: This tag will cause the deprecation to show up in IDEs

```suggestion
/** @deprecated use StageArtifactSelector instead */
```

maggieneterval: Great suggestion, thanks for making it inline so I could use the handy "commit suggestion" button 🌮

maggieneterval:
Thanks for the reviews, all! I know this was a hefty one. ❤️

PR #8277, opened 2020-05-13
fix(bake): adding skip region detection checkbox

edgarulg:
This change is related with following issue:  spinnaker/spinnaker#4641
and uses the functionality provided in:  spinnaker/orca#3627

This change is about adding the skip region detection checkbox to the bake stage for AWS provider.
Captura de Pantalla 2020-05-13 a la(s) 9 27 14

When a bake stage runs, orca tries to get all the regions that we use in next stages, this causes that the bake happen on different regions rather that we specified. This checkbox prevent Orca to find out regions in other stages for bake.

The default value is false, so any pipeline already defined with bake won't change its behavior.


app/scripts/modules/amazon/src/pipeline/stages/bake/bakeStage.html
```diff
@@ -6,6 +6,14 @@
     <stage-config-field label="Regions">
       <checklist items="regions" model="stage.regions" inline="true" include-select-all-button="true"></checklist>
     </stage-config-field>
+    <stage-config-field label="Skip Region Detection">
```
marchello2000, line 9: I would add a `help-key`, something to the effect of 

> By default, Spinnaker will detect regions to bake in from downstream deploy stages. To prevent failed deploys from accidentally missed regions during the bake process. This setting will disable this detection mechanism

marchello2000:
(seems reasonable to me, but..) I will defer to one of the `deck` engineers on this since my html/js/ts is effectively non-existent.

app/scripts/modules/amazon/src/pipeline/stages/bake/bakeStage.html
```diff
@@ -6,6 +6,14 @@
...
+      <div class="checkbox" style="margin-bottom: 0">
+        <label>
+          <input type="checkbox" ng-model="stage.skipRegionDetection" />
+          Ensure bake happen only on selected regions.
```
erikmunson:
```suggestion
          Only bake explicitly selected regions
```

app/scripts/modules/amazon/src/pipeline/stages/bake/bakeStage.html
edgarulg: Thanks for your feedback @marchello2000 I agree we can give more context of what the checkbox does

app/scripts/modules/amazon/src/pipeline/stages/bake/bakeStage.html
edgarulg: Thanks for give it a look, i rewrite it.

edgarulg:
I took a look at your comments and made the necessary changes to improve the code.

edgarulg:

edgarulg:
@spinnakerbot cherry-pick 1.20

atyutyunnik:
shouldn't the wording be "Ensure Bake happens only in the selected regions"?

PR #8278, opened 2020-05-13
feat(core/managed): add version unpin flow to Environments

erikmunson:
We recently added the ability to pin a version to an environment manually, and based on early feedback it's clear that unpinning is equally as important _even if_ we don't yet have the API machinery to give us confidence-building data like which version will be deployed when you unpin. You'll note I've kept the explanatory/help text pretty simple and sparse, and it's specifically because I don't want to invest there until we have more data to supply a better experience.

Opening as a draft for now, but the only missing piece is an 'unpin' icon which we don't yet have. Here's what the flow looks like:

**Pinned version:**
Screen Shot 2020-05-13 at 9 45 41 AM

**Clicking on the top level 'unpin' button:**
Screen Shot 2020-05-13 at 9 46 04 AM

**Clicking the quick 'unpin' button on the pinned version card:**
Screen Shot 2020-05-13 at 9 46 24 AM

cc @gcomstock


erikmunson:
@gcomstock a couple non-urgent UX questions for you about this:

- How useful/important is it to have a top-level Unpin button that sits next to the existing Pin button? I was left kind of not sure whether it was useful or just noisy/distracting. I think one thing we've learned from pipeline executions is that the top level 'Start manual execution' button doesn't really get used that much, most people go to the pipeline they want to start and hit the button there. Wondering if we'll see a similar dynamic here where people will naturally gravitate toward the card with the Unpin button?
- If we do decide to keep the top level one, what are your thoughts on hiding it entirely when a version is not current pinned anywhere? that's the behavior i have right now, but for the Pin button today i'm not hiding it when the version is pinned to all environments — I'm just disabling the button. I sort of wonder if that behavior will look weird. Maybe I'm getting too detail oriented about it

app/scripts/modules/core/src/managed/UnpinArtifactModal.tsx
```diff
@@ -0,0 +1,195 @@
+import React, { memo, useCallback } from 'react';
```
erikmunson, line 1: you'll note this file looks... very similar to the `PinArtifactModal` component. I explicitly decided to duplicate instead of trying to re-use chunks because in my experience trying to go to reusable chunks too fast in these cases is often counter productive and leads to headaches and refactors. I also expect things to diverge even further between the two as we add more features to both flows. Either way we can always consolidate in the future if it turns out to be useful

vigneshm:
> How useful/important is it to have a top-level Unpin button that sits next to the existing Pin button? I was left kind of not sure whether it was useful or just noisy/distracting. I think one thing we've learned from pipeline executions is that the top level 'Start manual execution' button doesn't really get used that much, most people go to the pipeline they want to start and hit the button there. Wondering if we'll see a similar dynamic here where people will naturally gravitate toward the card with the Unpin button?

Thanks for pointing it out. The first thought that came to mind when I saw the screenshot was why do I see both pin and unpin buttons at the top? It kind of made sense after I saw that this version was pinned in one of the environments. However, more generally I like to see the options presented at the place where it makes the most sense (in this case having it in the card seems the best place to me).

vigneshm:
The code changes look good to me. I'll approve once greg gives his opinion?

gcomstock:
> @gcomstock a couple non-urgent UX questions for you about this:
> 
> * How useful/important is it to have a top-level Unpin button that sits next to the existing Pin button? I was left kind of not sure whether it was useful or just noisy/distracting. I think one thing we've learned from pipeline executions is that the top level 'Start manual execution' button doesn't really get used that much, most people go to the pipeline they want to start and hit the button there. Wondering if we'll see a similar dynamic here where people will naturally gravitate toward the card with the Unpin button?
> * If we do decide to keep the top level one, what are your thoughts on hiding it entirely when a version is not current pinned anywhere? that's the behavior i have right now, but for the Pin button today i'm not hiding it when the version is pinned to all environments — I'm just disabling the button. I sort of wonder if that behavior will look weird. Maybe I'm getting too detail oriented about it

I think we should ditch the top unpin button, and just have the pin button bring up the pin modal that allows you to add (and remove) environments. I can imagine in the future there might be a need to do some pin management in bulk for some users, but maybe we don't worry about that for now? 

Once we can surface to the user which version will go on which environment before/after, we can make the modal a bit more educational.

How's this sound?

erikmunson:
> have the pin button bring up the pin modal that allows you to add (and remove) environments

I'd be curious to see what you'd do here re: design — I feel like if I did it the flow would be pretty tedious/complicated. I know we decided to hold off on the multi-select bulk flow in the older UI designs, do you feel like an add/remove modal flow would be less challenging to design and develop? I'm trying to understand whether we should invest in it now or save that until we do another more detailed pass on pinning overall.

Depending on how much work is involved there, maybe we kill the top level Unpin, stick with the Pin flow as-is, and then design an add/remove flow after the other in flight work? 🤷‍♂️

gcomstock:
> 
> Depending on how much work is involved there, maybe we kill the top level Unpin, stick with the Pin flow as-is, and then design an add/remove flow after the other in flight work? 🤷‍♂️

Yes! Sorry if it wasn't clear that this was my suggestion for now. Definitely on board with the unpin button in the status card.

erikmunson:
👍 👍 wonderful, thanks for taking a look. I'll yank that flow out for now and we can come back to it.

erikmunson:
@vigneshm alrighty, i've gone ahead and removed the top level unpin flow entirely, thanks to both you and greg for helping make a call there. Because there's no longer any form inputs and whatnot the unpin modal went from a simple formik form to an (imo) even simpler setup, so hopefully easy to do a once-over on it.

erikmunson:
Also, screenshot of the new icon in place:

Screen Shot 2020-05-15 at 9 52 24 AM

PR #8279, opened 2020-05-13
refactor(core/presentation): migrate useData and useLatestPromise to use PromiseLike, not IPromise

christopherthielen:
Let's start migrating away from `IPromise` in our typing wherever we can


erikmunson:
question for future Erik: are there limitations to when/where you can move an `IPromise` to a `PromiseLike`? If so what would you advise looking at to decide when to replace vs. not?

PR #8281, opened 2020-05-14
refactor(core): Convert SecurityGroupFilters to react

caseyhebebrand:
- Reactify `SecurityGroupFilters`
- minor update to `FilterCheckbox`
- backwards compatibility with angular deps
- removed unused modules


caseyhebebrand:
> The state management of the filters is something that is heavily edited based on AngularJS paradigms. When all filter components are react, we could potentially rewrite, detangle, and simplify.

Agreed, I think the state management can be much simpler and then we will not need some of the extra code meant to sync up states with Angular

PR #8287, opened 2020-05-15
fix(pipelines): prevent pipelineConfig pollution caused by retrying a trigger unsuccessfully.

cristhian-castaneda:
When this issue happens https://github.com/spinnaker/spinnaker/issues/5682, there is a not expected behavior that adds unwanted data to `pipelineConfig`. 

These are the steps to reproduce:

1). Click `Retry` on any pipeline execution that doesn't implement `manualExecutionComponent` 

Getting this error:

```
PromiseObservable.js:76 Uncaught TypeError: Cannot read property 'formatLabel' of null
    at Triggers.updateTriggerDescription (Triggers.tsx:30)
    at Triggers.componentDidMount (Triggers.tsx:53)
    at commitLifeCycles (react-dom.development.js:17334)
    at commitAllLifeCycles (react-dom.development.js:18736)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at commitRoot (react-dom.development.js:18948)
    at eval (react-dom.development.js:20418)
    at Object.unstable_runWithPriority (scheduler.development.js:255)
```

2) Click on configure -> Pipeline Actions -> Edit as JSON 

What we have: 
Image 2020-05-14 at 9 58 11 PM

What we should have:
Image 2020-05-14 at 9 59 42 PM

So, this PR avoid this strange behavior ensuring the correct state of pipelineConfig


cristhian-castaneda:
@maggieneterval

maggieneterval:
Hi @cristhian-castaneda, thanks for the PR! Looks like @danielpeach opened the original issue you mentioned, so going to loop him in to review this since he likely has more context. Thanks!

app/scripts/modules/core/src/pipeline/config/pipelineConfig.controller.spec.js
```diff
@@ -29,6 +29,11 @@ describe('Controller: PipelineConfigCtrl', function() {
       },
       app: application,
     });
+    application.pipelineConfigs.activate();
```
danielpeach, line 32: The description of this test case is `'should initialize immediately if pipeline configs are already present'`. This change is the opposite of this - it's saying that "even if the pipeline configs are already loaded, I want to reload them". Would delete this test case and include a different one that demonstrates that behavior.

danielpeach:
@christopherthielen 

We have customers that have the following problem:
- They do something in the executions tab and unintentionally mutate a pipeline config. In the example that Cristhian is giving, it's actually the trigger payload from a previous execution that ends up in the config.
- They go into the pipeline config screen
- They save the pipeline and capture the mutation, which can break their pipeline.

It seems like there are a few code paths that can cause this kind of problem. How would you solve this? Reloading all of the configs doesn't seem ideal, but neither does tracking down the source of state mutation that can happen anywhere in the app.

maggieneterval:
@danielpeach my quick guess is we're mutating the trigger somewhere in [here](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/manualExecution/ManualPipelineExecutionModal.tsx#L295) when we actually want to be mutating a deep copy?

christopherthielen:
Oh wow we’ve seen this at Netflix too! I will try to look more in depth next week.

christopherthielen:
> @cristhian-castaneda Merge branch 'master' into fix/clean-up-pipelineConfigs 4497efe
> @cristhian-castaneda Merge branch 'master' into fix/clean-up-pipelineConfigs db8e138
> @cristhian-castaneda Merge branch 'master' into fix/clean-up-pipelineConfigs

FWIW if you mark the pull request as "allow edits from maintainers" then mergify bot can merge master into your branch automatically

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

maggieneterval:
I'm still a little confused by this solution...were we not able to identify the root cause of pipelines sometimes being mutated when doing a manual execution?

christopherthielen:
We didn't track down the actual mutation in re-run dialog.  If we do track it down and fix it, there is still the potential for some other code doing the same (intentionally or unintentionally).  

I think the safest thing to do is to always get a fresh copy of the pipeline config before editing.  Even if there _wasn't_ code that mutates these configs, I still think fetching the pipeline config before editing is a good practice so you know you're not editing a stale config (somebody else could be editing the pipeline at the same time).

cristhian-castaneda:

cristhian-castaneda:

cristhian-castaneda:
@spinnakerbot cherry-pick 1.20

PR #8288, opened 2020-05-15
Refactor FormValidator API

christopherthielen:
This PR is split up into multiple commits.  It will likely be easier to review single commits as one commit was a sweeping change.

I spent some time thinking about the `FormValidator` API and my complaints about the current API.
First and foremost, I can *never* remember the right incantation to do array validation.  I set out to rectify that and ended up with a fairly major rewrite.

## values

- Users now provide `values` to the `validate()` method instead of the `FormValidator()` constructor.

```diff
- validator = new FormValidator(values)
+ validator = new FormValidator()
validator.withValidators(..)
- return validator.validateForm();
- return validator.validateForm(values);
```

This decouples the `values` from the `FormValidator` object itself.  This will theoretically enable us to re-use pre-configured `FormValidator` objects for pathological cases.

## arrayForEach

- Users now return a `FormValidator` object from `arrayForEach` instead of having one passed into the callback
- `arrayForEach` is now a chainable builder method on `IFormValidator` interface.

```diff
- const { arrayForEach } = validator;
validator.withValidators(
  Validators.arrayNotEmpty(),
-  arrayForEach(builder => {
-    builder.field('fieldOnArrayElement').required();
-  })
)
+.arrayForEach(path => {
+  const itemValidator = new FormValidator(path)
+  itemValidator.field('fieldOnArrayElement').required();
+  return itemValidator;
+}
```

This improves discoverability of `arrayForEach` dramatically (although it introduces a new discoverability problem w.r.t. what to do with the `path` parameter -- you pass it to the new FormValidator).
By asking the `arrayForEach` callback to *return* a `FormValidator`, we can again do some fancy caching stuff to return pre-built FormValidators (if they can be defined statically).

## Consolidated Form/Field API

- I combined the `FormValidator` and `FormValidatorField` APIs

Before, `FormValidator` had `.field()` method which returned a `FormValidatorField`.  Now, `.field()` returns a nested `FormValidator`.  You can theoretically now do crazy things like 
```js
const fooValidator = rootValidator.field('foo');
fooValidator.field('nestedFooProperty').withValidators()`
```

There is no distinction between a field and the top level validator.  The root validator is actually a field validator for `path === null`.

I don't know if this improves API design or not, but it "feels" simpler to me now.  I also feel like I can understand the code paths much easier.  Yeah, well, you know, that's just like, my opinion, man.


christopherthielen:
looks like I epically failed CI

caseyhebebrand:
I like the idea of chaining `arrayForEach`, but I also feel like the API should feel similar to the non-arrray fields. Still thinking through it, but here are my thoughts:

```js
const formValidator = new FormValidator();
formValidator.field('test').isArray();
formValidator.field('pathToTestChild').withValidators(...):
```

Where `isArray` would apply the `DeferredArrayValidator`, and the form validator would know to iterate through the child field validators provided for that path. 

The pros are a more cohesive implementation that leaves the work to the `FormValidator`. The cons are that it loses the nested validator objects and also may be too much of a refactor. (I also don't know if `.field` works with nested fields like that... 😄)

Overall I like the direction you are moving this, but thought I would throw this out there.

vigneshm:
I like casey's idea on simplifying the validation model for arrays. Given the ability to write nested validations, something like the following would be easy to read by using a special notation for arrays in a path.

```js
const values = { lotsastuff: [1, 2, 3, 4, 5] };
const formValidator = new FormValidator();

formValidator
  .field('lotsastuff', 'Array')
  .required()
    .field('[i]')
    .withValidators(
      maxValue(3)
    ),
  );

formValidator
  .field('lotsastuff.[i]', 'Array Items')
  .withValidators(maxValue(3))

formValidator
  .field('lotsastuff.[0]', 'Firt Array Item')
  .withValidators(maxValue(3))
```

I haven't completely gone through the implementation of `FormValidator`, so I'm unsure of the effort required to build something like the above. May be a future work?

christopherthielen:
Thanks for the constructive feedback!  I agree with you both that a simpler array API would be awesome sauce.  The first attempt I made in this refactor was similar to what @caseyhebebrand suggested.  However, the simpler approaches I tried (and the suggestions above) assume that all validation rules for each element are know statically.  The approaches don't support conditional validation rules which we currently support  (and use) via the callback.  The callback signature `ICreateValidatorCallback` provides the array element as the second argument `item` (and a bunch of other things like the index and the array itself):

```
.arrayForEach((path, item, index, array, arrayLabel) => {
  
});
```

I didn't call this out because the requirement and behavior was there previously, and the PR doesn't change it.

Take these examples from the amazon provider:

https://github.com/spinnaker/deck/blob/c3fe935e3a3ad6865e6759f5802bbe2f74021299/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx#L80-L82

https://github.com/spinnaker/deck/blob/c3fe935e3a3ad6865e6759f5802bbe2f74021299/app/scripts/modules/amazon/src/loadBalancer/configure/application/TargetGroups.tsx#L84-L88

We _conditionally add validation for each array element_.  This is similar to how we handle conditional validation at the root level too as seen here:

https://github.com/spinnaker/deck/blob/c3fe935e3a3ad6865e6759f5802bbe2f74021299/app/scripts/modules/core/src/pipeline/config/stages/googleCloudBuild/googleCloudBuildValidators.ts#L4-L9

-------------

So that's the background.  I _do think we need a callback based array forEach API_.  It doesn't necessarily have to be the _only_ API.  We could definitely write a second (simpler) mechanism, perhaps triggered when `.arrayForEach()` is called without any arguments, but I think some form of callback based api needs to exist.  If we were to do that, we would have two different mechanisms for validating child arrays, which is another tradeoff in simplicity.

```
const validator = new FormValidator()
// returns one FormValidator for all the items
const arrayItemValidator = validator.field('array').withValidators(arrayNotEmpty()).arrayItemValidator() 
arrayItemValidator.withValidators(minValue(1));

const arrayItemValidator2 = validator.field('otherArray').withValidators(arrayNotEmpty()).arrayItemValidator() 
arrayItemValidator2.field('nested1');
arrayItemValidator2.field('nested2');

const arrayItemValidator3 = validator.field('otherArray').withValidators(arrayNotEmpty()).arrayItemValidator((path, item) => {
  // do the current style arrayForEach
}) 
```

I'd also like to hear what @erikmunson and @maggieneterval  think as well before we move forward.

christopherthielen:
> My biggest concern is how much future churn we expect in Deck's form API, given that we're now considering it a public API for consumption by plugins. Seems like the migration code was probably a fair bit of work. I don't have anything helpful to add here, just feel bad that you had to write all that migration code in order to do this refactor.

Excellent points, thank you.  It wouldn't be too much trouble to shoehorn backwards compatibility in there (at least for a few releases) at the expense of code clarity. I do think `@spinnaker/eslint-plugin` linter rules are a reasonable path forward when making breaking changes.  They're not terribly difficult to write, but are a nightmare to maintain.  For this style of migration linter rule, I am thinking of them as temporary code that should be deleted after a few quarters.

> One other thing that jumped out at me (feel free to ignore) orthogonal to the changes in this PR is the JavaBeans-esque pattern around creating a mutable FormValidator and then calling various mutating setters. I'm wondering what this would look like as more of a builder pattern where the FormValidator itself is immutable, which might more cleanly separate creating a FormValidator from actually calling validateForm, and prevent validateForm from ever being called when the FormValidator is in an incomplete state. Again, feel free to ignore as I don't think no-argument constructors + setters is considered as problematic in JavaScript, but just thought I'd toss that out there as these setters seem like they're getting fairly complex and I envision a future where we're passing around half-baked FormValidators as a way to compose and augment them, which is a pattern I've seen cause problems in other codebases.

Another great point.  This would be easy enough to rectify by removing `.validate()` and adding a `.build()` method.

christopherthielen:
This PR conversation has me reconsidering spinnaker's roll-your-own validation.  Originally, we chose not to use `yup` because we thought we'd prefer to have a small surface area, purpose built validation API.  All I intended to do with this PR was incrementally improve the array handling, but the conversation shows me that theres appetite to really get the API "right".  At this point, I'm not sure that getting it right _by ourselves_ would be preferable to outsourcing the whole thing to a project that's thought the whole thing through.

caseyhebebrand:
> All I intended to do with this PR was incrementally improve the array handling, but the conversation shows me that theres appetite to really get the API "right". At this point, I'm not sure that getting it right _by ourselves_ would be preferable to outsourcing the whole thing to a project that's thought the whole thing through.

You definitely improved the array handling from before. My opinion is that if its going to take 80% more time/energy to convert the last 10% and get it "right" then the trade off may not be worth it. The rest of the API is great and this improves the array fields greatly, so I would be ok with using your incremental change. Then taking on researching/implementing the outsourcing when we are ready.

erikmunson:
ok so I've looked this code over, but I will add a **very big disclaimer** that I am not super familiar with the validator code and the diff itself is quite nuanced, so while I can give a sort of half-way thumbs up I can't say with certainty that I'm understanding everything correctly. It's possible some of my commentary could be totally detached from reality as a result, apologies if so. with that...

I think that in the context of the current validator API this is a pretty nice simplification, and does improve the overall quality of the API to some degree. At the same time, I'm seeing the things I gave a a lot of constructive feedback about on the original implementation PR become extremely painful, and I remain totally unconfident that this overall API approach is the right one.

My original feedback on this approach was that by using a Java-esque mutable bag of chained method state, we were making it extremely brittle and complicated to do any sort of composition or support new and interesting ways of validating without jumping through incredibly complicated hoops to support it in the first class `FormValidator` API. My observation is that people gravitate toward these types of APIs because they appear familiar to things they've seen in the past, whether that's in a language like Java or just an older less mature JS framework or pattern. That appearance of familiarity hides the fact that trying to model these validators as arguments to a big mutable class (which now can contain inner versions of itself!) is fundamentally complex, not simple.

On the array validation use case, I agree with the feedback that regardless of the specifics (so either the old way or the new way), doing array validation with this `arrayForEach` approach is not user friendly. I do think that the way this PR proposes is significantly better (even with the weird bit about the `path` arg) and if it's a choice between the two I think this PR is the way to go. I still don't think either is really right.

Maggie pointed out the idea of an immutable builder pattern, at the expense of being annoying (sorry) I'd like to resurrect my original feedback that for this API to truly become _simple_ (not just familiar to some people who have used builders before), it would ideally become a function-based composition API where I could create validators and the then compose them as arguments, e.g.:
```ts
import { validator, composeValidators, standardValidators } from 'core/presentation';
import { customAddressValidator } from './myCustomValidators';

// I imagine for discoverability we might want some sort of module that does
// an 'export * from xyz' for our common/standard validators
const { required, number } = standardValidators;

// validators accept a path, and a variadic arg list of functions which are provided
// the value to validate and the path as args (similar to today's IValidator), 
const addressValidator = validator('address', required, customAddressValidator);
const zipValidator = validator('zip', required, number);

// validates against formik values on its own without composition
const zipResult = zipValidator(values);

// composeValidators would take the functions returned by validator and run
// them all together, just like today's FormValidator() class
const validate = composeValidators(addressValidator, zipValidator);

// validates with all the composed validators against formik values
const formResult = validate(values);
```

The reason I'm resurrecting this in the context of array validation is because I _think_ based on the comments so far and the usage examples, we could probably make that functional `validator` API do something useful with array validation by default, because in a world where all validators are functions there's none of these totally wild shenanigans with trying to use both class method chaining and callbacks together, e.g.:

```ts
import { validator, composeValidators, standardValidators } from 'core/presentation';

const { required, number } = standardValidators;

// when a field value turns out to be an array, we could take each value in the array
// and run the list of provided validators on that value. Then if we wanted to support
// really complex array validations and use things like index or other args we could
// do that without forcing people to think about it in the common case
const someArrayValidator = validator('path.to.someArray', required, (...args) => {
  // the signature could be something like value, path, index?
  const [value] = args;
  if (value.type === 'someSpecificType') {
    return required(...args) || number(...args);
  } else {
    return required(...args);
  }
});
```

I'm not saying "drop everything and redo the validator API!", but I do feel pretty strongly that we will continue to hit up against fundamental and largely unsolvable problems with this class-based approach because it just doesn't match how we want to think about the problem. I'm also not saying that the pseudo-code I just wrote in a few minutes is the perfect API, or even that it solves all the problems we were just talking about — but I do think that it's directionally headed toward what 'getting it right' probably looks like.

If getting there seems like too much work to justify the investment, then I think I'd +1 the sentiment about trying to go with an off the shelf tool and composing our own validator logic into it. I had personally hoped this validator API would be a very slim bit of logic, and would prioritize simplicity over the appearance of convenience in common use cases. I think somewhere along the way the appeal of adding smarts to FormValidator in the name of convenience overtook the value of a simple, composable API that doesn't make any assumptions about how you're going to use it.

christopherthielen:
>  we were making it extremely brittle and complicated to do any sort of composition or support new and interesting ways of validating without jumping through incredibly complicated hoops to support it in the first class FormValidator API

I disagree completely with this statement.  We already had composition of `IValidators`.  The FormValidator construct marries one or more `IValidator`s with a formik field and adds some sugar for the common stuff like required, optional, and spelAware.  This refactor actually _enables composition_ of FormValidators (although I didn't implement any code to do so).

> resurrect my original feedback that for this API to truly become simple

I think there are a lot of holes to be poked in that pseudocode, but it's probably not worth doing.   Rewriting the validation api from scratch wasn't really on the table (unless you're signing up for it?)


One thing I think we all agree on is that the API today is clunky at best.   I'm just going to close the PR and move on, because nobody is really happy and it's only an incremental win over the status quo.  I might look into yup some day as an alternative.

erikmunson:
I guess my comment about being detached from reality turned out to be appropriate! Exploring yup seems reasonable given that there's turning out to be more complexity hiding in here. Sorry if I just added noise.

PR #8298, opened 2020-05-19
feat(titus): Additional fields for Run Job and Deploy stage

aravindmd:
- Adding `EFS` , `Command` , `Constraint` fields to make run job on par with deploy
- Also adding `Command` field in titus deploy stage

`EFS`  Mount
Screen Shot 2020-05-19 at 12 07 24 PM

`Constraints` under advance section
Screen Shot 2020-05-19 at 12 06 31 PM

river PR to support `Command` : https://github.com/spinnaker/clouddriver/pull/4609/files

### DIFF
--- a/app/scripts/modules/titus/src/domain/ITitusServerGroup.ts
+++ b/app/scripts/modules/titus/src/domain/ITitusServerGroup.ts
@@ -16,6 +16,7 @@ export interface ITitusServerGroup extends IServerGroup {
   capacityGroup?: string;
   disruptionBudget?: IJobDisruptionBudget;
   entryPoint: string;
+  cmd?: string;
   iamProfile: string;
   id?: string;
   image?: ITitusImage;
--- a/app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
+++ b/app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
@@ -20,6 +20,14 @@ import { ITitusResources } from 'titus/domain';
 
 import { TitusSecurityGroupPicker } from './TitusSecurityGroupPicker';
 import { TitusProviderSettings } from '../../../titus.settings';
+import Select, { Option } from 'react-select';
+import { string } from 'prop-types';
+
+const mountPermOptions = [
+  { label: 'Read and Write', value: 'RW' },
+  { label: 'Read Only', value: 'RO' },
+  { label: 'Write Only', value: 'WO' },
+];
 
 export interface ITitusRunJobStageConfigState {
   credentials: string[];
@@ -30,6 +38,8 @@ export interface ITitusRunJobStageConfigState {
 interface IClusterDefaults {
   application: string;
   containerAttributes: object;
+  constraints: object;
+  efs: object;
   env: object;
   labels: object;
   resources: ITitusResources;
@@ -77,6 +87,16 @@ export class TitusRunJobStageConfig extends React.Component<IStageConfigProps, I
     const clusterDefaults: IClusterDefaults = {
       application: application.name,
       containerAttributes: {},
+      constraints: {
+        soft: {},
+        hard: {},
+      },
+      efs: {
+        efsId: string,
+        mountPoint: string,
+        mountPerm: string,
+        efsRelativeMountPoint: string,
+      },
       env: {},
       labels: {},
       resources: {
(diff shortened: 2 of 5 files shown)


app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
```diff
@@ -77,6 +87,16 @@ export class TitusRunJobStageConfig extends React.Component<IStageConfigProps, I
...
       containerAttributes: {},
+      constraints: {
+        soft: {},
+        hard: {},
+      },
+      efs: {
+        efsId: string,
+        mountPoint: string,
+        mountPerm: string,
+        efsRelativeMountPoint: string,
```
erikmunson, line 98: looks like these `string` values are written like they're in a typescript interface, but this is an object. should they be `''` instead?

aravindmd: Makes sense , moved it to `IClusterDefaults`, let me know if this is better.

erikmunson: `IClusterDefaults` interface looks good, just be aware that if you want this `clusterDefaults` object to have initial values for those fields you'll still need to initialize them to empty stings (i.e. `''`) here too. If you don't care about those fields being present in the object when they're not explicitly added by users, then that's fine!

aravindmd: Yes we don't want to have those by default, so those initial values are not required in this case.

erikmunson: 👍 👍

app/scripts/modules/titus/src/domain/ITitusServerGroup.ts
```diff
@@ -16,6 +16,7 @@ export interface ITitusServerGroup extends IServerGroup {
   capacityGroup?: string;
   disruptionBudget?: IJobDisruptionBudget;
   entryPoint: string;
```
christopherthielen, line 18: looking at this now, I wonder if entryPoint should be optional too?  Are entryPoint and cmd mutually exclusive?

app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
```diff
@@ -30,6 +37,13 @@ export interface ITitusRunJobStageConfigState {
 interface IClusterDefaults {
   application: string;
   containerAttributes: object;
+  constraints: object;
```
christopherthielen, line 40: Can this be typed stronger?

app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
```diff
@@ -218,7 +236,69 @@ export class TitusRunJobStageConfig extends React.Component<IStageConfigProps, I
           onChange={this.dockerChanged}
           deferInitialization={stage.deferredInitialization}
         />
-
+        <h4>
+          <b>
```
christopherthielen, line 240: probably don't need the <b> tags

app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
```diff
@@ -218,7 +236,69 @@ export class TitusRunJobStageConfig extends React.Component<IStageConfigProps, I
           onChange={this.dockerChanged}
           deferInitialization={stage.deferredInitialization}
         />
-
+        <h4>
```
christopherthielen, line 239: It would be nice to extract all the EFS options to a separate component

app/scripts/modules/titus/src/domain/ITitusServerGroup.ts
aravindmd: agree , updated `entryPoint` to be optional , from what i see they are not mutually exclusive, titus api applies both of them if they are present.

app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
aravindmd: fixed

app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
aravindmd: left the ` <b>` tag to make it look consistent with deploy UI , let me know if you think we need to remove that.

app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
aravindmd: Agree and i am thinking to get this out first and re-visit the `EFS` component as part of migration where user can specify multiple mounts. Let me know your thoughts ?

christopherthielen: sounds good!

christopherthielen:
is this still a WIP?

aravindmd:
> is this still a WIP?

clouddriver bits are yet to be merged and deployed, but should be good to go i think.

christopherthielen:
@aravindmd  do we still need this PR?

PR #8299, opened 2020-05-20
feat(core/presentation): adds managed delivery illustrations

gcomstock:
@erikmunson @vigneshm 

Here's a first pass at illustrations that will be used on the env view for different flows. I'll get confirmation on using the "M" before we set this all live. Doing this ahead of new mocks because the pin/unpin illustrations can be put in whenever, but I wanted to make them all as a set.

Screen Shot 2020-05-19 at 5 11 31 PM


gcomstock:
> YES 👯 🕺 💃 👯
> 
> Overall looks phenomenal. I'm not sure what to do about the `M`. I'm personally fond of it brand-wise, but also pretty unsure at this point where it makes sense to use it vs. not.
> 
> Only other thing that caught my eye: I immediately connected the orange color of the pause illustration's circle with the warning color we use for the paused status. In that context it felt really right, but for the mark as bad one I was a little thrown off because the rest of the UI treats that as an error (w/ the red error status color). Not saying the background should be red, that would be pretty jarring, but just wanted to point out there's a little symbolism going on with that color which may not fully connect. It's possible I'm only noticing it because I'm seeing all the different icons together though, probably very safe to disregard.

The intention was to match the background to the status level! I forgot we were treating this is red (which I feel is correct) I'll update!

gcomstock:
Screen Shot 2020-05-19 at 9 11 13 PM

erikmunson:
oh hey that came out quite nicely! Very cool that the tie to status colors was intentional.

PR #8300, opened 2020-05-20
feat(core/managed): switch 'Artifacts' header to 'Versions'

erikmunson:
I've heard a couple different forms of feedback that seemed to come from a bit of ambiguity around what "Artifacts" means in the sidebar. In your app's delivery manifest you define an `artifacts` list that specifies the artifact you care about (oftentimes just a single image/deb), I'm sort of wondering if the name being the same here for something that's a list of lots of versions causes a little bit of a mismatch.

Thoughts? I'm honestly a bit torn but figured it was worth trying out and considering. Looks like this:

Screen Shot 2020-05-19 at 9 31 30 PM

fyi @gcomstock

### DIFF
--- a/app/scripts/modules/core/src/managed/Environments.tsx
+++ b/app/scripts/modules/core/src/managed/Environments.tsx
@@ -140,7 +140,7 @@ export function Environments({ app }: IEnvironmentsProps) {
   return (
     <div className={styles.mainContent}>
       <div className={styles.artifactsColumn}>
-        <ColumnHeader text="Artifacts" icon="artifact" />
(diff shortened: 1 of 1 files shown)


gcomstock:
I dig it! If we move towards "state bundles" and config changes, a more generic title works better anyway.

erikmunson:
yeah that's an interesting point re: moving to versioning more than just artifacts, good call out. Given that you're liking it I'm inclined to just try it out in main and see what people's reaction is — only a one-line revert if we change our minds! 😄

PR #8312, opened 2020-05-28
refactor(halyard): clean up halconfig/settings and mark as deprecated

maggieneterval:
* docs(halyard): mark halconfig/settings as deprecated 

* refactor(halyard): remove unneeded authTtl default 

  We are already defaulting SETTINGS.authTtl to 600000 everywhere this is read.

* refactor(halyard): remove unneeded checkForUpdates default 

  We do not need to default this to false, since the logic it gates will only be executed if the flag is explicitly truthy.

* refactor(halyard): remove defaultProviders from halconfig/settings 

  Halyard users are currently supplied with a list of defaultProviders that include every possible cloud provider. The AccountService reads from defaultProviders when trying to list all providers for a given application only if there the application does not have an explicit list of configured cloud providers, and then will take the intersection of available providers with these defaultProviders, but since we're supplying every possible provider in the list of defaultProviders, this is essentially the same as falling back to available providers, which is what we do when SETTINGS.defaultProviders is undefined.

* refactor(halyard): remove unneeded gitSources default 

  We already default gitSources to these four sources everywhere SETTINGS.gitSources is read.

* refactor(halyard): remove newApplicationDefaults default 

  Instead of defaulting newApplicationDefaults.chaosMonkey to false, let's just more safely read from this field, which may be undefined for non-Halyard users anyway.

* refactor(halyard): remove unneeded pagerDuty default 

  pagerDuty.required already defaults to false everywhere it is read.

* refactor(halyard): push default pollSchedule to code 

  We have never exposed a Halyard parameter to configure pollSchedule; let's push the default of 30000 into consuming code. This won't affect non-Halyard users who are setting this parameter.

* refactor(halyard): remove unneeded appengine properties 

  These properties are no longer read in code.

* refactor(halyard): remove unneeded aws defaults 

  These defaults are unnecessary as they are already present in consuming code.

* refactor(halyard): remove unnecessary GCE property 

  IGceProviderSettings no longer includes associatePublicIpAddress, so we can remove this property.

* refactor(halyard): remove unneeded kubernetes defaults 

  These fields are no longer consumed by the provider.

* refactor(halyard): remove titus defaults from halconfig 

  Titus is not configurable with Halyard; there is no need for these defaults to live here.

* refactor(halyard): remove Halyard-defaulted pubsubProviders 

  As documented on [this PR](https://github.com/spinnaker/deck/pull/7908), Amazon pubsub works, so there's not a great reason to default to only 'google.' Let's include 'amazon' in the default list in consuming code and remove the default from the halconfig.

* refactor(halyard): remove triggerTypes from halconfig 

  Including an exhaustive list of triggerTypes has the same effect as not configuring triggerTypes, since if this is unset we will register every available trigger type.

* refactor(halyard): clean up features config 

  Removes unneeded fields from the features config: many of these are no longer supported or were never supported by Halyard. Several are unneeded defaults.

* refactor(halyard): remove final unneeded properties from halconfig/settings.js 

  * SETTINGS.feature.notifications is unused; remove from halconfig/settings.js
  * maxPipelineAgeDays is not set in Halyard's DeckProfileFactory; remove from halconfig/settings.js
  * onDemandClusterThreshold is not set in Halyard's DeckProfileFactory; remove from halconfig/settings.js


maggieneterval:
Sorry @ezimanyi before we merge this, could you take a look at a final commit, which I just pushed up and should include removing the last of the unneeded properties?

PR #8317, opened 2020-05-29
fix(core): Details panel height when in multiselect

caseyhebebrand:

### DIFF
--- a/app/scripts/modules/core/src/presentation/details.less
+++ b/app/scripts/modules/core/src/presentation/details.less
@@ -22,7 +22,6 @@
   .details-panel {
     display: block;
     width: 100%;
-    height: 100%;
     border-bottom-width: 1px;
     .header {
       padding-top: 0;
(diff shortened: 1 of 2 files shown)


christopherthielen:
@caseyhebebrand  Interesting.  
1) Do you know what caused this as a regression?  
2) Do you have any concerns about adding this height: 100% to all those selectors?  
3) Will this CSS change completely when your new layout code is complete?

caseyhebebrand:
> 1. Do you know what caused this as a regression?
I do not know. My git blame shows that this styling hasn't been touched in a long time
> 2. Do you have any concerns about adding this height: 100% to all those selectors?
Not at the moment. It is a temporary change, and the non-mutli-selected details panels all take up the height of the parent div `nav` which is set to 100%.
> 3. Will this CSS change completely when your new layout code is complete?
This height will actually be enforced full-time with the new layout. The overlayed details drawer will always be 100% height, no matter the content or route.

gcomstock:
> > 3. Will this CSS change completely when your new layout code is complete?
> >    This height will actually be enforced full-time with the new layout. The overlayed details drawer will always be 100% height, no matter the content or route.

Yessss!

PR #8324, opened 2020-06-02
fix(ecs): disable terminate instance

pkandasamy91:
Currently terminating tasks fails as it tries to use the instance id to terminate the task. As discussed in this [issue](https://github.com/spinnaker/spinnaker/issues/5383), removing the terminate instance button as it doesn't currently work.


pkandasamy91:
Ready for review @allisaurus

PR #8330, opened 2020-06-04
feat(core): Combine General Exception Messages with Kato's

evansb:
This **partially** solves https://github.com/spinnaker/spinnaker/issues/5782

In this issue, there are two exceptions (General & Kato), but only the General exceptions are shown to the user.

This PR changes the error message extraction logic in `orchestratedItem.transformer` such that Kato exceptions are also shown to the users.

Rationale: The general exception might not be useful sometimes and/or is false positive, I do believe Kato error messages should still be shown instead of being superseded.


christopherthielen:
I think I'm 👍 with this change.  Let's try it for a while and see if it surfaces any undesirable errors.

PR #8336, opened 2020-06-08
feat(core/presentation): add <Illustration/>, supporting files

erikmunson:
**Note:** there's only ~30 lines of code changes here, the rest is SVG files.

A Managed Delivery PR that will be opened momentarily includes repeated use of some illustrations, and after playing around with the SVGs I decided that we should consider having _some_ kind of formalized way of referencing illustrations without hand-importing via the clunky SVGR static import syntax. I decided to try out a near-identical approach to the one we took with `<Icon/>` and see how it goes.

One big difference is that — unlike icons — illustrations are really meant to be something you as the engineer + designer put one-off thought and judgement into. Whereas icons have a pretty prescriptive set of options to choose from for size + appearance, with an illustration we want to embrace the human judgement involved in visually treating a specific spot. It's possible that _within specific re-usable components_ we'll add more prescriptive opinions about how to place and size these, but at the lowest level we want illustrations to be as flexible as possible. As a result this proposed implementation of `<Illustration/>` does not have any props and instead allows itself to be sized based on any layout rules coming down from its DOM ancestors. **This will mean it takes more work to lay out new illustrations**, because you'll need to actually decide how to make the `<Illustration/>` element fill out the space you're looking to use. My experience so far has been that this behavior is what I want anyway, since it's all custom to the specific component I happen to be in.

cc @gcomstock


gcomstock:
Well said!

erikmunson:
> I like the thinking behind this. This component is a good candidate for Emotion.js :)

hah, yeah i thought the same. one of the _hopefully_ very rare cases where being able to apply arbitrary styling feels useful/appropriate.

PR #8238, opened 2020-05-01
feat(core/managed): add support for pinning artifacts to environments

erikmunson:
~**Depends on #8231 and #8237**~

Pinning is a cool new concept we're building in Managed Delivery — a way to tell Spinnaker it should keep an environment on a specific version no matter what happens (until you manually unpin). Pinning is useful for all kinds of different things: rolling back, bypassing tests/canaries/other deployments to get a hotfix out quickly, and halting deployments until some external dependency is satisfied are just some examples.

While the diff is sizable, the commits themselves are very atomic and very tightly scoped. If it's convenient I think you could reasonably do it in multiple passes/sittings without losing too much context. The full thing breaks down into:

- Adding indicators that things are pinned (including comments and the who/when)
- Pulling out a couple existing things into their own components, because now they're going to be used in two places instead of one (`Button` and `EnvironmentBadge`)
- Adding a modal, available while looking at a specific artifact version, that allows you to pick an environment to pin to along with an explanation which supports markdown.

Some screenshots:

<details>
<summary>Bubbles on environments/versions</summary>

Screen Shot 2020-04-30 at 4 56 09 PM
</details>
<details>
<summary>Pin button, card with metadata</summary>

Screen Shot 2020-04-30 at 4 56 19 PM
</details>
<details>
<summary>Pin button disabled because version is pinned everywhere, markdown support</summary>

Screen Shot 2020-04-30 at 5 00 38 PM
</details>
<details>
<summary>Pin modal</summary>

Screen Shot 2020-04-30 at 4 58 20 PM
</details>
<details>
<summary>Validation</summary>

Screen Shot 2020-04-30 at 4 59 20 PM
</details>
<details>
<summary>Environments with 'prod' resources show up red</summary>

Screen Shot 2020-04-30 at 5 00 55 PM
</details>

<details>
<summary>State where an environment is already pinned</summary>

Screen Shot 2020-04-30 at 4 59 58 PM
</details>
<details>
<summary>Request error handling (error/message are mock data)</summary>

Screen Shot 2020-04-30 at 5 06 24 PM
</details>

PR #8337, opened 2020-06-08
feat(core/managed): support for marking a version 'bad', add illustrations

erikmunson:
~~**Depends on https://github.com/spinnaker/deck/pull/8336**~~

We previously added a workflow for [pinning an artifact version to an environment](https://github.com/spinnaker/deck/pull/8238), which is helpful for things like rollbacks/rollforwards and temporarily halting deployments to an environment. The next step in supporting manual override workflows is _marking a version as 'bad'_.

This is useful for times where you don't care about which version specifically runs in an environment, but you want to make sure a broken version is never chosen. Today we'll already mark a version as bad automatically if we try to deploy it and it fails multiple times, but now you can choose to do the same thing manually.

A _ton_ of this code is copy-pasta from the pinning flow, which is generally a good thing. I'm not at all convinced these flows will remain the same under the hood and I'd like to retain the flexibility to do large refactors on one or both. Alongside the new flow, I've also updated the pin + unpin flows to use the new illustrations @gcomstock put together, along with tweaking the layouts + explanatory copy so they match our updated mockups more closely.

Screenshots:

<details>
<summary>New action button on versions</summary>

Screen Shot 2020-06-08 at 4 35 50 PM

</details>

<details>
<summary>Modal flow for marking as bad</summary>

**Normal:**
Screen Shot 2020-06-08 at 1 33 35 PM

**When an environment is already marked bad:**
Screen Shot 2020-06-08 at 1 33 45 PM

**When an environment is already pinned:**
Screen Shot 2020-06-08 at 1 47 52 PM

</details>

<details>
<summary>Version / environment card after pinning</summary>

Screen Shot 2020-06-08 at 3 47 38 PM

</details>

<details>
<summary>New pin illustration + layout</summary>

**Normal:**
Screen Shot 2020-06-08 at 1 32 25 PM

**When an environment is already marked bad:**
Screen Shot 2020-06-08 at 1 33 11 PM

</details>

<details>
<summary>New unpin illustration + layout</summary>

Screen Shot 2020-06-08 at 1 34 28 PM

</details>

cc @gcomstock


luispollo:
Looks gorgeous!

gcomstock:
Screen Shot 2020-06-08 at 5 03 49 PM

Beautiful! Minor nit on "Check out our documentation" and the previous sentence: I think there should either be an extra line of whitespace, or there shouldn't be a full return.

erikmunson:
> I think there should either be an extra line of whitespace, or there shouldn't be a full return.

@gcomstock argh yeah you caught me. I 'settled' on this (a.k.a. I gave up) because I felt like both were awkward in context. let me just show you what each looks like and we can discuss:

Screen Shot 2020-06-08 at 5 11 32 PM
Screen Shot 2020-06-08 at 5 11 52 PM
Screen Shot 2020-06-08 at 5 09 52 PM
Screen Shot 2020-06-08 at 5 11 05 PM

Can you just tell me which to use?? I've already been looking at these layouts for too long to feel like I'm evaluating them coherently 😆😭😎

gcomstock:
😆  Let's go with extra line of whitespace!

erikmunson:
Excellent, sold, I'm doing it ❤️

erikmunson:
> Does it make sense to allow marking a version bad across all environments through a single action in the modal?

YES thanks for bringing this up — greg and I discussed it verbally but I neglected to cover that idea in the PR description. Completely agree that in many cases marking bad everywhere is what folks will truly want to do. I'm doing this now because it's an easy copy-paste job and there's a workaround (it's just tedious), but if @gcomstock has ideas for a quick way to be able to select "all" the available options that doesn't take much UI engineering work i bet the API side could also be modified to take a list of environments (right now it only accepts one).

PR #8338, opened 2020-06-11
fix(core): Validation for max luxon duration

caseyhebebrand:
Upon investigation, the module we use to calculate stage duration has a maximum value that we do not currently impose. It only honors up to 100 million days after epoch start, or 8.64x10^15. 

Fix: Add max value to the validator so a large value does not impact rendering.

### DIFF
--- a/app/scripts/modules/core/src/utils/timeFormatters.ts
+++ b/app/scripts/modules/core/src/utils/timeFormatters.ts
@@ -7,7 +7,9 @@ import { SETTINGS } from 'core/config/settings';
 
 import { SystemTimezone } from './SystemTimezone';
 
-const isInputValid = (input: any) => !(!input || isNaN(input) || input < 0);
+// Luxon supports up to 100 million days after epoch start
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/utils/timeFormatters.ts
```diff
@@ -7,7 +7,9 @@ import { SETTINGS } from 'core/config/settings';
 
 import { SystemTimezone } from './SystemTimezone';
 
-const isInputValid = (input: any) => !(!input || isNaN(input) || input < 0);
+// Luxon supports up to 100 million days after epoch start
```
marchello2000, line 10: only 100 million days? 😞

marchello2000:
thank you!

app/scripts/modules/core/src/utils/timeFormatters.ts
caseyhebebrand: sadly not infinity days :(

PR #8341, opened 2020-06-15
feat(core): link to specific version changelog

maggieneterval:
Since [this change](https://github.com/spinnaker/buildtool/pull/48) to spinnaker/buildtool, the OSS release changelog gists have been published with one top-level gist per minor release and individual files for each patch release. Since we can now stably link to each patch release changelog on spinnaker.io, let's point the Help --> Version link there rather than the general Releases page, which just links out ot each individual changelog anyway.

### DIFF
--- a/app/scripts/modules/core/src/help/HelpMenu.tsx
+++ b/app/scripts/modules/core/src/help/HelpMenu.tsx
@@ -34,6 +34,21 @@ const AdditionalHelpLinks = () =>
     </>
   ) : null;
 
+const Version = () => {
+  if (!SETTINGS.version) {
+    return null;
+  }
+
+  const CHANGELOG_PATH = `${SETTINGS.version.replace(/\./g, '-')}-changelog`;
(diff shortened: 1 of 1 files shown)


maggieneterval:
It's worth noting that for OSS users, the "What's New" changelog has been broken for the entirety of the 1.20 release, since Halyard continued to supply the file name as `changelog.md` to `settings.js` but `buildtool` began naming each file according to the patch release. I haven't heard any bug reports though, so perhaps this feature isn't getting a ton of use. Since we can now link directly to each release's changelog on spinnaker.io, I would not be opposed to removing the What's New changelog modal if Netflix is not using this either. For now, let's provide post-Halyard users with quick access to the changelog.

christopherthielen:
worth noting that netflix uses a custom what's new widget

maggieneterval:
@christopherthielen Good to know; in that case, would you be opposed to me removing the What's New widget altogether?

christopherthielen:
@maggieneterval I will defer to whatever you think is best.

PR #8342, opened 2020-06-15
feat(core/managed): Application pause/resume functionality

vigneshm:
Adding application level pause/resume functionality in the environments view by introducing

- A new header section that contains some high-level description of the current state along with an actions dropdown.
- A modal for completing the pause/resume action.

Note that the texts used in the header and the modal are accurate but temporary and will be updated to reinforce the idea that pause functionality should be used sparingly.

<details>
  <summary>Header in running state</summary>

md-env-header-running

</details>

<details>
  <summary>Header in paused state</summary>

md-env-header-paused

</details>

<details>
  <summary>Modal for resuming</summary>

Note: A more detailed list of artifact versions that will be deployed to each environment will be added in a different PR.

md-env-resume-modal

</details>

<details>
  <summary>Modal for pausing</summary>

md-env-pause-modal

</details>

### DIFF
--- a/app/scripts/modules/core/src/managed/Environments.tsx
+++ b/app/scripts/modules/core/src/managed/Environments.tsx
@@ -13,6 +13,7 @@ import { ColumnHeader } from './ColumnHeader';
 import { ArtifactsList } from './ArtifactsList';
 import { EnvironmentsList } from './EnvironmentsList';
 import { ArtifactDetail } from './ArtifactDetail';
+import { EnvironmentsHeader } from './EnvironmentsHeader';
 
 import styles from './Environments.module.css';
 
@@ -157,6 +158,13 @@ export function Environments({ app }: IEnvironmentsProps) {
             show && (
               <animated.div key={key} className={styles.environmentsPane} style={props}>
                 <ColumnHeader text="Environments" icon="environment" />
+                <EnvironmentsHeader
+                  app={app}
+                  resourceInfo={{
+                    managed: resources.filter(r => !r.isPaused).length,
+                    total: resources.length,
+                  }}
+                />
                 <EnvironmentsList application={app} {...{ environments, artifacts, resourcesById }} />
               </animated.div>
             ),
--- a/app/scripts/modules/core/src/managed/EnvironmentsHeader.tsx
+++ b/app/scripts/modules/core/src/managed/EnvironmentsHeader.tsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import { Dropdown } from 'react-bootstrap';
+
+import { Application } from '../application';
+import { Illustration, IllustrationName } from '../presentation';
+import { showToggleManagedResourceModal } from './ToggleManagedResourceForApplication';
+
+import './EnvironmentsHeader.less';
+
+interface IEnvironmentsHeaderProps {
+  app: Application;
+  resourceInfo: { managed: number; total: number };
+}
+
+export const ToggleManagedResourceAction = ({ app }: { app: Application }) => (
+  <li>
+    <a
+      onClick={() => {
+        showToggleManagedResourceModal({ application: app });
+      }}
+    >
+      {app.isManagementPaused ? 'Resume management' : 'Disable management'}
+    </a>
+  </li>
+);
+
+const environmentHeaderInfo: {
+  [key: string]: { description: string | null; icon: IllustrationName; title: (info: string) => string };
+} = {
+  running: {
+    description: null,
+    icon: 'runManagement',
+    title: (info: string) => `Spinnaker is continuously managing ${info} resources.`,
+  },
+  paused: {
+    description: 'Any actions made will take effect once management is resumed.',
+    icon: 'disableManagement',
+    title: () => 'Management is disabled for this application.',
+  },
+};
+
+export const EnvironmentsHeader = ({ app, resourceInfo: { managed, total } }: IEnvironmentsHeaderProps) => {
+  const { description, icon, title } = environmentHeaderInfo[app.isManagementPaused ? 'paused' : 'running'];
+
+  return (
+    <div className="EnvironmentsHeader">
+      <div className="flex-container-h sp-padding-l">
+        <div style={{ minWidth: 145 }}>
+          <Illustration name={icon} />
+        </div>
+        <div className="flex-container-v sp-padding-xl-top sp-margin-m-left sp-margin-m-top">
+          <div className="heading-3 bold">{title(managed === total ? `${total}` : `${managed}/${total}`)}</div>
+          {description && <div className="sp-margin-s-top">{description}</div>}
+          <Dropdown id="application-actions" className="sp-margin-l-top">
+            <Dropdown.Toggle className="dropdown-toggle-btn">Application Actions</Dropdown.Toggle>
+            <Dropdown.Menu className="dropdown-menu">
+              <ToggleManagedResourceAction app={app} />
+            </Dropdown.Menu>
+          </Dropdown>
+        </div>
+      </div>
+    </div>
+  );
+};
--- a/app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
+++ b/app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
@@ -1,0 +1,145 @@
+import React, { memo, useState } from 'react';
+import ReactGA from 'react-ga';
+
+import { Application } from '../application';
+import { Button } from './Button';
+import {
+  Illustration,
+  IllustrationName,
+  IModalComponentProps,
+  ModalBody,
+  ModalFooter,
+  ModalHeader,
+  showModal,
+  ValidationMessage,
+} from '../presentation';
+import { ManagedWriter } from './ManagedWriter';
+
+const logClick = (label: string, application: string) =>
+  ReactGA.event({
+    category: 'Environments - toggle application management modal',
+    action: `${label} clicked`,
+    label: application,
+  });
+
+export interface IToggleManagedResourceForApplicationModalProps extends IModalComponentProps {
+  application: Application;
+}
+
+interface Error {
+  data: { error: string; message: string };
+}
+
+export const MDRunningStateDescription = () => (
+  <>
+    <p>
+      <span className="bold">
+        Careful! You’re about to stop Spinnaker from managing all resources in your application.
+      </span>
+      This feature should only be used if management is not working properly and manual intervention is required.{' '}
+      <a href="https://www.spinnaker.io/guides/user/managed-delivery" target="_blank">
+        Check our documentation for more information
+      </a>
+      .
+    </p>
+    <p>
+      Need to rollback?{' '}
+      <a href="https://www.spinnaker.io/guides/user/managed-delivery/pinning/" target="_blank">
+        Try pinning a version instead
+      </a>
+      .
+    </p>
+  </>
+);
+
+export const MDPausedStateDescription = () => {
+  return (
+    <p>
+      You’re about to resume management for this application. The latest good version approved for deployment will be
+      deployed to each environment, and any configuration changes made while disabled will take effect.
+    </p>
+  );
+};
+
+const modalViewInfo = {
+  paused: {
+    actionToBeTaken: 'Resume',
+    description: <MDPausedStateDescription />,
+    icon: 'runManagement' as IllustrationName,
+  },
+  running: {
+    actionToBeTaken: 'Disable',
+    description: <MDRunningStateDescription />,
+    icon: 'disableManagement' as IllustrationName,
+  },
+};
+
+export const showToggleManagedResourceModal = (props: IToggleManagedResourceForApplicationModalProps) =>
+  showModal(ToggleManagedResourceForApplicationModal, props, { maxWidth: 800 });
+
+export const ToggleManagedResourceForApplicationModal = memo(
+  ({ application, closeModal, dismissModal }: IToggleManagedResourceForApplicationModalProps) => {
+    const [isSubmitting, setSubmitting] = useState<boolean>(false);
+    const [actionError, setActionError] = useState<{ error: string; message: string } | null>(null);
+    const dataSource = application.getDataSource('environments');
+
+    const { actionToBeTaken, description, icon } = modalViewInfo[application.isManagementPaused ? 'paused' : 'running'];
+
+    const handleActionInitiation = () => {
+      setActionError(null);
+      setSubmitting(true);
+      logClick(`${actionToBeTaken} Management`, application.name);
+      const call = application.isManagementPaused
+        ? ManagedWriter.resumeApplicationManagement(application.name)
+        : ManagedWriter.pauseApplicationManagement(application.name);
+      call
+        .then(() => dataSource.refresh(true).catch(() => null))
+        .then(() => {
+          closeModal();
+        })
+        .catch((error: Error) => {
+          setActionError(error.data);
+          setSubmitting(false);
+        });
+    };
+    return (
+      <>
+        <ModalHeader>{actionToBeTaken} management</ModalHeader>
+        <ModalBody>
+          <div className="flex-container-h sp-padding-xl-yaxis">
+            <div style={{ minWidth: 145 }}>
+              <Illustration name={icon} />
+            </div>
+            <div className="sp-padding-xl">
+              <div>{description}</div>
+              {actionError && (
+                <div className="sp-margin-xl-top">
+                  <ValidationMessage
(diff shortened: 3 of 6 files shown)


vigneshm:
cc @gcomstock

gcomstock:
Thanks Vignesh! Looks solid overall. After our conversation in the channel the other day, let me update the wording in the mocks to not use "managed delivery." Should just be a few changes. I also owe you the more serious "disable" illustration

gcomstock:
Here's my pass at scrubbing "managed delivery" from the UI. See what you think then we can fire up a discussion with the team:

https://gallery.io/projects/MCHbtQVoQ2HCZSOv94xJXrgW/files/MCEJu8Y2hyDSccycEWfOZ-cmQQTkGAKyZI0

and here's some visual bits I owe you:

https://github.com/spinnaker/deck/pull/8344

vigneshm:
> https://gallery.io/projects/MCHbtQVoQ2HCZSOv94xJXrgW/files/MCEJu8Y2hyDSccycEWfOZ-cmQQTkGAKyZI0

I've implemented everything from the above except the following

1. Showing current deployed version and to-be deployed version for each environment in resume modal. This needs support from keel, so I've added a [request](https://github.com/spinnaker/keel/issues/1305). 
2. Red colored modal header and button in pause modal. (This needs some changes in `Modal` and `Button` components. I'll update them in a separate PR)

app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
```diff
@@ -0,0 +1,130 @@
...
+      .
+    </p>
+  </>
+);
+
+export const MDPausedStateDescription = () => {
+  return (
+    <p>
+      You’re about to resume management of cloud resources for your application. The latest good version approved for
+      deployment will be deployed to each environment.
```
erikmunson, line 58: small wording tweak/suggestion:

> You’re about to resume management for this application. The latest good version approved for deployment will be deployed to each environment, and any configuration changes made while disabled will take effect.

app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -157,6 +159,13 @@ export function Environments({ app }: IEnvironmentsProps) {
             show && (
               <animated.div key={key} className={styles.environmentsPane} style={props}>
                 <ColumnHeader text="Environments" icon="environment" />
+                <EnvironmentsHeader
+                  app={app}
+                  clusterInfo={{
+                    managed: resources.filter(r => getKindName(r.kind) === 'cluster' && !r.isPaused).length,
+                    total: resources.filter(r => getKindName(r.kind) === 'cluster').length,
```
erikmunson, line 166: as we move forward with our plans to evolve and add to the available kinds of resources, focusing only on cluster resources will become less and less relevant (and many apps may end up with _no_ cluster resources). Thoughts on shifting this to focus on resource in general of any kind vs. just clusters?

app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const dataSource = application.getDataSource('environments');
+    let actionToBeTaken = 'Disable';
+    let description = <MDRunningStateDescription />;
+    let icon: IllustrationName = 'disableManagement';
+
+    if (application.isManagementPaused) {
+      actionToBeTaken = 'Resume';
+      description = <MDPausedStateDescription />;
+      icon = 'runManagement';
+    }
```
erikmunson, line 79: similar question as in the other component re: turning this conditional assignment into a small function.

app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
```diff
@@ -0,0 +1,130 @@
...
+      setActionError(null);
+      setSubmitting(true);
+      logClick(`${actionToBeTaken} Management`, application.name);
+      const call = application.isManagementPaused
+        ? ManagedWriter.resumeApplicationManagement(application.name)
+        : ManagedWriter.pauseApplicationManagement(application.name);
+      call
+        .then(() => dataSource.refresh(true).catch(() => null))
+        .then(() => {
+          application.isManagementPaused = actionToBeTaken === 'Disable';
```
erikmunson, line 91: in other places, we're not doing any in-memory updates and relying entirely on refreshing the data source to get fresh app state. Are you seeing a lag here? (i.e. you don't see an updated `applicationPaused` value in the data from keel on the first refresh?)

app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
```diff
@@ -0,0 +1,130 @@
...
+          closeModal();
+        })
+        .catch((error: Error) => {
+          setActionError(error.data);
+          setSubmitting(false);
+        });
+    };
+    return (
+      <>
+        <ModalHeader>{actionToBeTaken} Managed Delivery</ModalHeader>
```
erikmunson, line 102:
```suggestion
        <ModalHeader>{actionToBeTaken} management</ModalHeader>
```

app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <ModalBody>
+          <div className="flex-container-h sp-padding-xl-yaxis">
+            <div style={{ minWidth: 145 }}>
+              <Illustration name={icon} />
+            </div>
+            <div className="sp-padding-xl">
+              <div>{description}</div>
+              <div>
+                {actionError &&
```
erikmunson, line 111: Not saying it's better or worse than this, but FWIW I have been using `ValidationMessage` in other MD modal flows ofr handling submit errors: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/managed/PinArtifactModal.tsx#L157-L170

app/scripts/modules/core/src/managed/EnvironmentsHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <a {...useSref('home.applications.application.config', { application: app.name })}>View Config</a>
+  </li>
+);
+
+export const EnvironmentsHeader = ({ app, clusterInfo: { managed, total } }: IEnvironmentsHeaderProps) => {
+  let title = `Spinnaker is continuously managing ${managed}/${total} clusters.`;
+  let icon: IllustrationName = 'runManagement';
+  let description = null;
+
+  if (app.isManagementPaused) {
```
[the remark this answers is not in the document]
vigneshm: How does the following look like? 

```js
const EnvironmentHeaderInfo: {
  [key: string]: { description: string | null; icon: IllustrationName; title: (info: string) => string };
} = {
  running: {
    description: null,
    icon: 'runManagement',
    title: (info: string) => `Spinnaker is continuously managing ${info} clusters.`,
  },
  paused: {
    description: 'Any actions made will take effect once management is resumed.',
    icon: 'disableManagement',
    title: () => 'Management is disabled for this application.',
  },
};

export const EnvironmentsHeader = ({ app, clusterInfo: { managed, total } }: IEnvironmentsHeaderProps) => {
  const { description, icon, title } = EnvironmentHeaderInfo[app.isManagementPaused ? 'paused' : 'running'];

  return (
    <div className="EnvironmentsHeader">
      <div className="flex-container-h sp-padding-l">
        <div style={{ minWidth: 145 }}>
          <Illustration name={icon} />
        </div>
        <div className="flex-container-v sp-padding-xl-top sp-margin-m-left sp-margin-m-top">
          <div className="heading-3 bold">{title(managed === total ? `${total}` : `${managed}/${total}`)}</div>
          {description && <div className="sp-margin-s-top">{description}</div>}
          <Dropdown id="application-actions" className="sp-margin-l-top">
            <Dropdown.Toggle className="dropdown-toggle-btn">Application Actions</Dropdown.Toggle>
            <Dropdown.Menu className="dropdown-menu">
              <ToggleManagedResourceAction app={app} />
              <ViewAppConfigAction app={app} />
            </Dropdown.Menu>
          </Dropdown>
        </div>
      </div>
    </div>
  );
};
```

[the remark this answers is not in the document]
vigneshm: I like the idea of keeping the body of a react component minimal. I'm not sure how I feel about making `title` into a function to accommodate dynamic values. The other way of doing this is just having another function, something like the following.

```js
function getEnvironmentHeaderInfo(
  app,
  managed,
  title,
): {
  description: string | null;
  icon: IllustrationName;
  title: string;
} {
  if (app.isManagementPaused) {
    return {
      title: 'Resource management is disabled for this application.',
      icon: 'disableManagement',
      description: 'Any actions made will take effect once managed delivery is resumed.',
    };
  } else {
    return {
      title: `Spinnaker is continuously managing ${managed}/${total} clusters.`,
      icon: 'runManagement',
      description: null,
    };
  }
}
```

Given the above, I kind of like the 'config' way.

app/scripts/modules/core/src/managed/Environments.tsx
vigneshm: Good call. I'm removing the filtering.

app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
vigneshm: The lag was significant and it wasn't due to `keel` not responding in time. As I was quickly debugging to explain the issue, I actually found the cause of the problem 🤦 .

`application.isManagementPaused` is only [set](https://github.com/spinnaker/deck/blob/05ec5eff04ad4d114ed21dbaafad1adbd4b23e42/app/scripts/modules/core/src/managed/managed.dataSource.ts#L27) when we refresh `managedResources` data source.

If I updated the data source in this component to `managedResources`, it would set `application.isManagementPaused` right after keel updates it. But the data source in the top level `Environments` component is for [`environments`](https://github.com/spinnaker/deck/blob/05ec5eff04ad4d114ed21dbaafad1adbd4b23e42/app/scripts/modules/core/src/managed/Environments.tsx#L69). So the top-level component is not updated for a while, so we continue to see stale information.

I've fixed it by adding `application.isManagementPaused` in the [environments](https://github.com/spinnaker/deck/blob/05ec5eff04ad4d114ed21dbaafad1adbd4b23e42/app/scripts/modules/core/src/managed/managed.dataSource.ts#L45) data source. Let me know if that is the right fix.

app/scripts/modules/core/src/managed/EnvironmentsHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+      {app.isManagementPaused ? 'Resume Managed Delivery' : 'Disable Managed Delivery'}
+    </a>
+  </li>
+);
+
+export const ViewAppConfigAction = ({ app }: { app: Application }) => (
+  <li>
+    <a {...useSref('home.applications.application.config', { application: app.name })}>View Config</a>
+  </li>
+);
```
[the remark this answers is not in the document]
vigneshm: Makes sense. I'll remove it for now. I'll sync with @luispollo to get the delivery config url and add this link back.

app/scripts/modules/core/src/managed/Environments.tsx
```diff
@@ -157,6 +158,13 @@ export function Environments({ app }: IEnvironmentsProps) {
             show && (
               <animated.div key={key} className={styles.environmentsPane} style={props}>
                 <ColumnHeader text="Environments" icon="environment" />
+                <EnvironmentsHeader
+                  app={app}
+                  clusterInfo={{
```
erikmunson, line 163: `resourceInfo` to reflect the newly refactored behavior?

app/scripts/modules/core/src/managed/EnvironmentsHeader.tsx
```diff
@@ -0,0 +1,64 @@
...
+      onClick={() => {
+        showToggleManagedResourceModal({ application: app });
+      }}
+    >
+      {app.isManagementPaused ? 'Resume management' : 'Disable management'}
+    </a>
+  </li>
+);
+
+const EnvironmentHeaderInfo: {
```
erikmunson, line 27: IMO going the config route here was a good decision, much simpler to mentally process. 🔍 🔍 🔍 SUPER nit: consider changing the casing of this to `environmentHeaderInfo` since it's not a component

app/scripts/modules/core/src/managed/EnvironmentsHeader.tsx
```diff
@@ -0,0 +1,64 @@
...
+  [key: string]: { description: string | null; icon: IllustrationName; title: (info: string) => string };
```
erikmunson, line 28: a cool TS trick i have exploited a few times for objects like this is to use [the `as const` thing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) to basically say 'just look at what's in the object and type it as that'. Helps avoid needing to do this type of boilerplate type def. Only nuance is you'll probably have to make all the `title` functions take the exact same args, which you can do by using an underscore to tell TS you don't care about using the argument in that particular function ([example](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/managed/VersionStateCard.tsx#L74)).

app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
erikmunson: wow nice find! geez. I remember back when Alan and I discussed having these two data sources we were pretty sure there would be some obnoxious state-syncing silliness going on as we built more UI surfaces out, apparently it's starting to happen. I'm hopeful as we start experimentally moving this area of the UI over to GraphQL we'll get away from problems like this, but if it continues to prove problematic we might need to do some sort of data source joinery between the two. I think the way you went about it is totally reasonable for now, but I imagine in the future it's possible we'll end up with racing or some other random annoyances.

app/scripts/modules/core/src/managed/ToggleManagedResourceForApplication.tsx
```diff
@@ -0,0 +1,145 @@
...
+export const MDPausedStateDescription = () => {
+  return (
+    <p>
+      You’re about to resume management for this application. The latest good version approved for deployment will be
+      deployed to each environment, and any configuration changes made while disabled will take effect.
+    </p>
+  );
+};
+
+const ModalViewInfo = {
```
erikmunson, line 64: if you decide to take action on either of the comments I left on `EnvironmentHeaderInfo`, probably worth applying those changes here too.

app/scripts/modules/core/src/managed/EnvironmentsHeader.tsx
vigneshm: Woah, that's a nice trick. TIL

vigneshm: I'm seeing the following error though. I was kind of expecting it to automatically set the type of `description` to `string | null`

image


But I'm unable to repro that in [typescript playground](https://www.staging-typescript.org/play?ts=3.7.5#code/MYewdgzgLgBApmAbgSwE7gLYKgCTgQwBM5UBJMAMxBgF4YBvAKBhlQFcwxkwBzALgbMWMYhGCpkAByjJwAsGwA2igDRCWyUGAEBydmACy+MPh5wsYKDrXCYMqIrgCAFNyoDoE3gEpaAPhgAAwBlSW4TAGsSGGQIGC0ZBRA2CEUATxgMY1NuHhgAEno3EABfVjgIZNRgCoA6QJsYEsbJfBS4QgEmW1FxKRk5GB0AQTAM-GAByEyiOBgAd2RlO3wo+AoKOEmYcBqZkzMLWFjyiDYsQlrrdRitXUJY-AAjRyMD82xr23tHF18aAI6N6mD6WGJxB4QZ6OQgwKioOwACxO+EkkkUmnwUyujWajDK+DiWmgAG5GIwKBxJrIwHCQCBnFE0rp9FxeDoYAAfIatdqEHTeDxQLx5bkKZbdcpQNioWkIFDoMBHPCzMiUEAAbSZAF0APy1XoSaQ0sklIA) though. I'm going to switch back until I figure out the problem.

erikmunson: ah yeah that's annoying. i wonder if it's a `tsconfig.json` thing that does that. I often do things like `description: null as string` since we don't have `strictNullChecks` enabled in our config.

PR #8351, opened 2020-06-17
fix(*): Use allowlist and denylist

caseyhebebrand:
Replace `blacklist` with `denylist` and `whitelist` with `allowlist` everywhere Deck controls. 

The only remaining occurrence is external (i.e. `entityTag.category === 'blacklist'`), and we will replace this once the model is updated.


app/scripts/modules/amazon/src/aws.settings.ts
```diff
@@ -1,13 +1,13 @@
...
-export interface IClassicLaunchWhitelist {
+export interface IClassicLaunchAllowlist {
   region: string;
   credentials: string;
 }
 
 export interface IAWSProviderSettings extends IProviderSettings {
   classicLaunchLockout?: number;
-  classicLaunchWhitelist?: IClassicLaunchWhitelist[];
+  classicLaunchAllowlist?: IClassicLaunchAllowlist[];
```
caseyhebebrand, line 10: Does anyone know who else uses this optional setting? I will tag them to inform them of the change.

PR #8353, opened 2020-06-17
refactor(core): remove What's New module and changelog config

maggieneterval:
* refactor(core): remove What's New module 

  For OSS users, the only purpose of the What's New module was to open a modal to display the most recent changelog gist. Since [this change](https://github.com/spinnaker/deck/pull/8341), we will instead link directly to the changelog gist on spinnaker.io. This prevents Deck from needing to be configured with the ID and name of the gist. In the next commit, we will remove those config fields.

* refactor(core): remove SETTINGS.changelog 

  Since we can now statically construct a link to the appropriate changelog gist using top-level OSS version only, we no longer need changelog metadata to be supplied as part of Deck's config.


maggieneterval:
@christopherthielen would you mind confirming that Netflix doesn't need anything from the OSS What's New module as a hook for Netflix's custom What's New component?

christopherthielen:
@maggieneterval  yeah, we're good

maggieneterval:
Thank you 🙏 💯

PR #8361, opened 2020-06-19
feat(core/page): Add support to add banner to pager view

vigneshm:
A new banner needs to be added to pager page, so making it overridable here and later adding it in `-nflx` repo.


app/scripts/modules/core/src/presentation/illustrations/illustrationsByName.ts
```diff
@@ -1,8 +1,10 @@
+import { ReactComponent as announcement } from './vectors/announcement.svg';
```
gcomstock, line 1: Thanks for adding this. I'll be sure to do this when I add more in the future.

PR #8366, opened 2020-06-23
refactor(core): Convert spinnaker container to React

caseyhebebrand:
Refactor most of `spinnaker.component.ts` to React. Doing so allows us to directly use the react version of the `SpinnakerHeader` so we can delete the extra angular code needed. 

**_Motivation:_** As we move to React, some of the angular managed services with broader state management are injected into react components. In order to keep the React component in sync with the injected model and force a re-render, a lot of hacks and duplicate state setting is needed. Refactoring this section to React will allow us to utilize `recoiljs` or React `context` for more clean UI state management for components at any part of the tree.


app/scripts/modules/core/src/bootstrap/SpinnakerContainer.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { Spinner } from '../widgets/spinners/Spinner';
+
+export interface ISpinnakerContainerProps {
+  authenticating: boolean;
+  routing: boolean;
+}
+
+export const SpinnakerContainer = ({ authenticating, routing }: ISpinnakerContainerProps) => {
+  // eslint-disable-next-line
+  console.log(authenticating, routing);
```
erikmunson, line 17: intentional?

app/scripts/modules/core/src/bootstrap/SpinnakerContainer.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const HeaderComponent = ReactInjector.overrideRegistry.getComponent('spinnakerHeader') || SpinnakerHeader;
```
christopherthielen, line 18: We will have to support existing overrides that are template based

PR #8367, opened 2020-06-23
refactor(core): Convert spinnaker container to React

caseyhebebrand:
Refactor most of `spinnaker.component.ts` to React. Doing so allows us to directly use the react version of the `SpinnakerHeader` so we can delete the extra angular code needed. 

**_Motivation:_** As we move to React, some of the angular managed services with broader state management are injected into react components. In order to keep the React component in sync with the injected model and force a re-render, a lot of hacks and duplicate state setting is needed. Refactoring this section to React will allow us to utilize `recoiljs` or React `context` for more clean UI state management for components at any part of the tree.


app/scripts/modules/core/src/bootstrap/SpinnakerContainer.tsx
```diff
@@ -0,0 +1,32 @@
...
+        {!authenticating && routing && (
+          <div className="transition-overlay">
+            <Spinner size="medium" />
+          </div>
+        )}
+        <div className="navbar-inverse grid-header">
+          <CustomBanner />
+          <SpinnakerHeader />
+        </div>
+        <div className="spinnaker-content grid-contents">{!authenticating && <UIView name="main" />}</div>
```
caseyhebebrand, line 27:
```suggestion
        <div className="spinnaker-content grid-contents">{!authenticating && <UIView name="main" className="flex-1" />}</div>
```

PR #8369, opened 2020-06-23
feature(core): Setup recoil for state management

caseyhebebrand:
Wrap `SpinnakerContainer` in `RecoilRoot` to allow access to all branches of the component tree. Add the first Atom as an example to initialize the state (this atom will be used shortly). 

Atoms can access from any **_functional component_**  that is a child under the `RecoilRoot`. Basic usage:

```js

// Initialize the atom with a default value
const myAtom = atom({
  key: 'testAtom',
  default: false
});

// Update an atom
const compA = () => {
  const [testAtom, setTestAtom] = useRecoilState(myAtom);

  return <div onClick={ ()=> setTestAtom(!testAtom) }></div>
}

// Consume the atom's value
const compB = () => {
   const showDiv = useRecoilValue(myAtom);

  return (showDiv && <div>Hello</div>)
}
```


vigneshm:
I'm excited to see us adding `recoiljs` to deck.  A big 👍  from me!

caseyhebebrand:
> I'm excited to see us adding `recoiljs` to deck. A big 👍 from me!

Shout out to @christopherthielen for the suggestion! So far, it has been very awesome to work with. I have always felt that we needed something redux-like for UI state management, and this much more lightweight which is a huge bonus.

app/scripts/modules/core/src/state/recoilAtoms.ts
```diff
@@ -0,0 +1,7 @@
+import { atom } from 'recoil';
+import { CollapsibleSectionStateCache } from '../cache/collapsibleSectionStateCache';
+
+export const verticalNavExpandedAtom = atom({
```
christopherthielen, line 4: I think we should put atom definitions closer to where they will be used.  A global `state` package doesn't smell right to me.   @vigneshm @caseyhebebrand @erikmunson thoughts?

app/scripts/modules/core/src/state/recoilStateInitializer.ts
```diff
@@ -0,0 +1,11 @@
+import { MutableSnapshot } from 'recoil';
+
+import * as atoms from './recoilAtoms';
+import { CollapsibleSectionStateCache } from '../cache/collapsibleSectionStateCache';
+
+export const initializeRecoilState = ({ set }: MutableSnapshot): void => {
```
christopherthielen, line 6: I don't think we need this `initializeRecoilState`.    What is the purpose?  This seems like it would become an unmaintainable mess of state objects.  AFAICT the atom already manages its default value internally?

caseyhebebrand: The docs mention that there is a period of time where the state is unset before initializing all the atoms if we do not use this: https://recoiljs.org/docs/guides/persistence/#restoring-state

`Note that it is important to use this prop instead of just manually setting atom values in an effect. Otherwise there will be an initial render without the restored state which can cause flicker or be invalid.`

I don't think everything needs this, but felt for this case it was important so that the navigation doesn't flicker.

app/scripts/modules/core/src/state/recoilAtoms.ts
caseyhebebrand: I am personally a fan of a centralized state, especially considering that the point of atoms is to be used in multiple places. 

 This atom will be read from the navigation, but written to in the header component.  So I could put this atom with the `verticalNav` since the key references that.

caseyhebebrand: Open to the majority opinion. When working with redux, it was nice to have all the selectors, reducers, etc in one folder since many times they were not tied to one particular scope of the app or data.

app/scripts/modules/core/src/state/recoilStateInitializer.ts
christopherthielen: > Note that it is important to use this prop instead of just manually setting atom values in an effect. Otherwise there will be an initial render without the restored state which can cause flicker or be invalid.

But you're not doing what the docs warn against (setting the recoil state in a `useEffect`). Did you observe it actually flickering before hand?  The warning in the docs is against reading and rendering a _stale value_ in the first render, but the value in the first render should be the default value.

app/scripts/modules/core/src/state/recoilAtoms.ts
christopherthielen: > I am personally a fan of a centralized state

I am _very strongly_ opposed to a redux-like global centralized state in spinnaker.  The reason recoiljs appeals to me is that we can share state between arbitrary component trees without introducing a centralized global state.

vigneshm: I spent sometime working on a react state management framework and I _think_ (although not sure) `recoiljs` may have been influenced by some of the design decisions.

The framework's primary focus was to let domain owners define their own state and let it be shared across react component trees. Usually the modules related to a domain such as react components, utilities, pages etc are co-located and this framework let state management related modules to be co-located as well. This makes it possible to automatically route bugs/runtime errors by looking at the stack trace and finding the owner of the module (via some metadata associated with each module).

Another advantage of splitting state management is that it allows lazy loading of modules. If there's some page (or domain) of an application that is not frequently visited, then you can load the related modules on-demand when the page is requested. With centralized state, all the state initializers and mutators/reducers would be loaded during application bootstrap even though many may not be needed or you would need to explicitly manage the lazy loading of these modules.  

I generally prefer splitting state management for these reasons.

app/scripts/modules/core/src/state/recoilStateInitializer.ts
caseyhebebrand: Not yet. I haven't implemented any atom functionality, since I wanted this PR to be about the setup. I will just remove for now until I am actually utilizing this first atom.

PR #8371, opened 2020-06-24
fix(core/application): ensure disable cluster tab waits for task to finish

zachsmith1:
In application configuration, the cluster option would allow an unauthorized user the ability to enable/disable clusters without waiting for the task to finish. In reality, it should only enable/disable the clusters tab if the task successfully finishes. Even if the task ends up being rejected because the user doesn't have write access, there would still be a window that would allow the user to see the cluster tab under infrastructure. This fix ensures that the task would only show the clusters tab once the task has successfully finished. Otherwise, it will remain its current state. The inverse is also true here which will now not allow an unauthorized user the ability disable the clusters tab momentarily.


fieldju:
@Mergifyio backport release-1.20.x

maggieneterval:
@fieldju @kevinawoo @zachsmith1 I'm not sure this PR meets the new [criteria](https://spinnaker.io/community/contributing/releasing/#release-branch-patch-criteria) to be backported. If it does, can you please add the required tests & documentation to demonstrate that it does? I'm going to close the backport PR for now. Thanks!

kevinawoo:
looking at the [git blame](https://github.com/spinnaker/deck/blame/b2b2f034d37eb4515a67add183c1b05bc9f806e0/app/scripts/modules/core/src/application/config/dataSources/applicationDataSourceEditor.component.ts#L70-L80), it's 2 years old... so it's been working like that for 2 years. Not technically a regression 😞 

It is a bug tho XD

PR #8373, opened 2020-06-24
feat(core/managed): Add status bubble in artifact list

vigneshm:
In this PR, I'm adding support to show the following if an artifact matches the following state in any of the environments.

- Manual Judgement Required
- Is Pinned
- Is Vetoed
- ~~Is Deploying~~

This can show a total of 3 states in a row. If there's more than 3, then we show two 2 states and show a count bubble to denote the number of states that are hidden.

Note that the priority order of the statuses is not final yet and I’m looking for opinions. If there are more states that could be shown other than the 4 I listed above, please let me know.

<details>
  <summary>Example</summary>

image
</details>

<details>
  <summary>Made-up Example</summary>

image
</details>

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactRow.module.css
+++ b/app/scripts/modules/core/src/managed/ArtifactRow.module.css
@@ -3,8 +3,9 @@
   display: block;
   cursor: pointer;
 }
+
 .ArtifactRow:not(.selected):hover {
-  background-image: linear-gradient(to bottom, rgba(160, 180, 220, 0.15), rgba(160, 180, 220, 0.15));
+  background-color: #f1f4fa;
 }
 
 .selected {
--- a/app/scripts/modules/core/src/managed/ArtifactsList.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactsList.tsx
@@ -1,11 +1,15 @@
-import React from 'react';
 import classNames from 'classnames';
+import React, { useState } from 'react';
 
-import { Pill } from './Pill';
-import { StatusBubble } from './StatusBubble';
-
-import { IManagedArtifactSummary, IManagedArtifactVersion } from '../domain/IManagedEntity';
 import { ISelectedArtifactVersion } from './Environments';
+import {
+  IManagedArtifactSummary,
+  IManagedArtifactVersion,
+  IStatefulConstraint,
+  StatefulConstraintStatus,
+} from '../domain/IManagedEntity';
+import { Pill } from './Pill';
+import { IStatusBubbleStackProps, StatusBubbleStack } from './StatusBubbleStack';
 
 import styles from './ArtifactRow.module.css';
 
@@ -44,19 +48,16 @@ interface IArtifactRowProps {
   name?: string;
 }
 
-export const ArtifactRow = ({
-  isSelected,
-  clickHandler,
-  version: { version, displayName, environments, build, git },
-  reference,
-  name,
-}: IArtifactRowProps) => {
-  const pinnedEnvironments = environments.filter(({ pinned }) => pinned).length;
+export const ArtifactRow = ({ isSelected, clickHandler, version: versionInfo, reference, name }: IArtifactRowProps) => {
+  const { version, displayName, environments, build, git } = versionInfo;
+  const [isHovered, setIsHovered] = useState(false);
 
   return (
     <div
       className={classNames(styles.ArtifactRow, { [styles.selected]: isSelected })}
       onClick={() => clickHandler({ reference, version })}
+      onMouseOver={() => setIsHovered(true)}
+      onMouseOut={() => setIsHovered(false)}
     >
       <div className={styles.content}>
         {build?.id && (
@@ -68,16 +69,11 @@ export const ArtifactRow = ({
           <div className={styles.sha}>{git?.commit || displayName}</div>
           {name && <div className={styles.name}>{name}</div>}
         </div>
-        {pinnedEnvironments > 0 && (
-          <div className="sp-margin-s-right">
-            <StatusBubble
-              iconName="pin"
-              appearance="warning"
-              size="small"
-              quantity={pinnedEnvironments > 1 ? pinnedEnvironments : null}
-            />
-          </div>
-        )}
+        <StatusBubbleStack
+          borderColor={isSelected ? 'var(--color-white)' : isHovered ? '#f1f4fa' : 'var(--color-alabaster)'}
+          maxBubbles={3}
+          statuses={getArtifactStatuses(versionInfo)}
+        />
       </div>
       <div className={styles.stages}>
         {environments
@@ -87,3 +83,26 @@ export const ArtifactRow = ({
     </div>
   );
 };
+
+type ArtifactStatusList = IStatusBubbleStackProps['statuses'];
+function getArtifactStatuses({ environments }: IManagedArtifactVersion): ArtifactStatusList {
+  const statuses: ArtifactStatusList = [];
+  // NOTE: The order in which entries are added to `statuses` is important. The highest priority
+  // item must be inserted first.
+
+  const isConstraintPendingManualJudgement = (constraint: IStatefulConstraint) =>
+    constraint.type == 'manual-judgement' && constraint.status == StatefulConstraintStatus.PENDING;
+  const requiresManualApproval = environments.some(environment =>
+    environment.statefulConstraints?.some(isConstraintPendingManualJudgement),
+  );
+  if (requiresManualApproval) {
+    statuses.push({ appearance: 'progress', iconName: 'manualJudgement' });
+  }
+
+  const isPinned = environments.some(({ pinned }) => pinned);
+  if (isPinned) {
+    statuses.push({ appearance: 'warning', iconName: 'pin' });
+  }
+
+  return statuses;
+}
--- a/app/scripts/modules/core/src/managed/StatusBubbleStack.tsx
+++ b/app/scripts/modules/core/src/managed/StatusBubbleStack.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import { take } from 'lodash';
+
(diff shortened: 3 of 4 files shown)


app/scripts/modules/core/src/managed/ArtifactRow.module.css
```diff
@@ -3,8 +3,26 @@
   display: block;
   cursor: pointer;
 }
-.ArtifactRow:not(.selected):hover {
-  background-image: linear-gradient(to bottom, rgba(160, 180, 220, 0.15), rgba(160, 180, 220, 0.15));
+
+.ArtifactRow:not(.selected):hover,
+.ArtifactStatus:not(.selected).hovered {
+  /*TODO: Change the color and remove opacity as it is affecting the border around status bubble*/
+  background-color: rgba(160, 180, 220, 0.15);
```
vigneshm, line 10: @gcomstock, Could we get a color here without the opacity? The opacity is messing up the border around status bubble in the hover state as you can see below.

image

gcomstock: Yes! #f1f4fa

gcomstock: Some feedback on these:

you can reduce the border width to 2px for each bubble.
please add some margin to the right of the +2 bubble, so that bubble takes up the same width as a normal bubble. Maybe a mix of margin and making the gray bubble a bit larger. This way, the bubbles form columns that line up visually:

Screen Shot 2020-06-24 at 4 53 47 PM

@erikmunson , @vigneshm do you think a status bubble for an active deployment is necessary when the environment bar will also be flashing? Im inclined to think less is more, and just the environment bar is enough. But it does break a bit of a pattern we're trying to establish. Let me know.

Thanks for all the work Vignesh! Looks great otherwise.

vigneshm: > do you think a status bubble for an active deployment is necessary when the environment bar will also be flashing? Im inclined to think less is more, and just the environment bar is enough. But it does break a bit of a pattern we're trying to establish.

Good point. The flashing bar must be enough then. I can remove the deploying status icon.

I've reduced each status bubble's width, increased the size of the `count` bubble and added some right margin to make them all aligned. Let me know if this looks fine.

image

image

gcomstock: Looks perfect, thank you!!

app/scripts/modules/core/src/managed/ArtifactRow.module.css
```diff
@@ -3,8 +3,26 @@
   display: block;
   cursor: pointer;
 }
-.ArtifactRow:not(.selected):hover {
-  background-image: linear-gradient(to bottom, rgba(160, 180, 220, 0.15), rgba(160, 180, 220, 0.15));
+
+.ArtifactRow:not(.selected):hover,
+.ArtifactStatus:not(.selected).hovered {
```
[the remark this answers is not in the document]
vigneshm: Good catch. That definitely works and didn't occur to me. I'm updating it.

app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -87,3 +80,85 @@ export const ArtifactRow = ({
...
+  if (requiresManualApproval) {
+    statuses.push({ appearance: 'progress', iconName: 'manualJudgement', kind: 'status' });
+  }
+
+  const isPinned = environments.some(({ pinned }) => pinned);
+  if (isPinned) {
+    statuses.push({ appearance: 'warning', iconName: 'pin', kind: 'status' });
+  }
+
+  const isVetoed = environments.some(({ vetoed }) => vetoed);
```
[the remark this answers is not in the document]
vigneshm: Good call. I'm totally onboard in reducing the clutter here and show fewer, relevant statuses.

app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -87,3 +80,85 @@ export const ArtifactRow = ({
     </div>
   );
 };
+
+const MAX_STATUS_BUBBLE_COUNT = 3;
+
+const ArtifactStatuses = ({
```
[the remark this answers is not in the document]
vigneshm: Thanks for bringing it up. Although I was leaning towards doing the refactor later, I think the potential future of this component leads to some minor but important changes in how this component is currently designed. So I went ahead and actually pulled it out into a separate generic component.

app/scripts/modules/core/src/managed/ArtifactRow.module.css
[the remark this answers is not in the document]
vigneshm: Unfortunately, I had to revert this one due to the refactor into `StatusBubbleStack`.

app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -87,3 +80,85 @@ export const ArtifactRow = ({
...
+    <div className="flex-container-h middle sp-margin-s-right">
+      {statusBubblesToRender.map((status, index) => {
+        if (status.kind === 'status') {
+          return (
+            <div
+              className={classNames(styles.ArtifactStatus, {
+                [styles.selected]: isSelected,
+                [styles.hovered]: isHovered,
+              })}
+              key={status.iconName}
```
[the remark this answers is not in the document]
vigneshm: The way it is currently implemented, there will be only one status bubble 'cos we're running through all environments to see if a condition is satisfied in any of the environment. This could be modified to add the count at the bottom right of the bubble, but that might affect the layout a little bit with all the stacking. Would definitely need @gcomstock's input on this one :)

app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -87,3 +80,85 @@ export const ArtifactRow = ({
...
+type ArtifactStatusList = Array<
+  (Pick<IStatusBubbleProps, 'iconName' | 'appearance'> & { kind: 'status' }) | { count: number; kind: 'count' }
+>;
+function getArtifactStatuses({ environments }: IManagedArtifactVersion): ArtifactStatusList {
+  const statuses: ArtifactStatusList = [];
+  // NOTE: The order in which entries are added to `statuses` is important. The highest priority
+  // item must be inserted first.
+
+  const isConstraintPendingManualJudgement = (constraint: IStatefulConstraint) =>
+    constraint.type == 'manual-judgement' && constraint.status == StatefulConstraintStatus.PENDING;
```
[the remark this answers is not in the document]
vigneshm: Good call. I'm going to add these additional conditions in a separate PR.

vigneshm:
> In addition to PENDING constraints, I would propose that stateful constraints with a FAILED status should also show up. My hunch is that folks will want to see things like failed canaries, smoke tests, or rejected manual judgements clearly highlighted. Is that appropriate? Does it increase the risk of showing too many statuses?

Yes, that does make sense to me. These are definitely important constraints to show. I'm leaning towards ranking the `PENDING` constraints higher than `FAILED` status though. What do y'all think? Also, as I mentioned in another comment, I'm going to address these in a separate PR.

app/scripts/modules/core/src/managed/StatusBubbleStack.tsx
```diff
@@ -0,0 +1,44 @@
...
+          style={{
+            backgroundColor: borderColor,
+            zIndex: statusBubblesToRender.length - index,
+          }}
+        >
+          <StatusBubble iconName={status.iconName} appearance={status.appearance} size="small" quantity="2" />
+        </div>
+      ))}
+      {hiddenStatusBubbleCount && (
+        <div className="HiddenStatusCount" key="count">
```
erikmunson, line 38: is the `key` necessary here now that it's not being rendered inside a `map()` call?

app/scripts/modules/core/src/managed/StatusBubbleStack.tsx
```diff
@@ -0,0 +1,44 @@
...
+      {statusBubblesToRender.map((status, index) => (
+        <div
+          className="StatusBubbleContainer"
+          key={status.iconName}
+          style={{
+            backgroundColor: borderColor,
+            zIndex: statusBubblesToRender.length - index,
+          }}
+        >
+          <StatusBubble iconName={status.iconName} appearance={status.appearance} size="small" quantity="2" />
```
erikmunson, line 34: what's the purpose of using `quantity="2"` here?

app/scripts/modules/core/src/managed/StatusBubbleStack.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { take } from 'lodash';
+
+import { IStatusBubbleProps, StatusBubble } from './StatusBubble';
+
+import './StatusBubbleStack.less';
+
+export interface IStatusBubbleStackProps {
+  borderColor: string;
+  maxStatusBubbleCount: number;
```
erikmunson, line 10: Is there a shorter/more concise name we could give this prop? at a glance it seems a little verbose compared to other prop names nearby.

app/scripts/modules/core/src/managed/StatusBubbleStack.tsx
vigneshm: Oh no! I added that for testing. I'm stopping the merge now.

erikmunson:
> I'm leaning towards ranking the `PENDING` constraints higher than `FAILED` status though. What do y'all think? Also, as I mentioned in another comment, I'm going to address these in a separate PR.

My gut feel is that you're probably on the right track here — failed/errored things are very important (and sometimes critical) but things that are currently running/in progress have the potential to _actively change_ the state of things, so probably require an even closer eye. That's also consistent with the current Spinnaker UI paradigms where in-progress stuff gets surfaced in the nav as little badges.

PR #8375, opened 2020-06-24
refactor(core): Refactor SpinnakerHeader to functional component

caseyhebebrand:
The `SpinnakerHeader` needs to be a functional component to use recoil hooks, but `@Overrides` only works on a class component. This wrapper is temporary to enable what we need until we have time to refactor the override component to work for functional components.

### DIFF
--- a/app/scripts/modules/core/src/header/SpinnakerHeader.tsx
+++ b/app/scripts/modules/core/src/header/SpinnakerHeader.tsx
@@ -9,87 +9,82 @@ import { HelpMenu } from 'core/help/HelpMenu';
 
 import './SpinnakerHeader.css';
 
-export interface ISpinnakerHeaderState {
-  navExpanded: boolean;
-}
-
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/header/SpinnakerHeader.tsx
```diff
@@ -9,87 +9,82 @@ import { HelpMenu } from 'core/help/HelpMenu';
...
+          <ul className="nav nav-items flex-1 page-nav">
+            <li key="navHome">
+              <UISrefActive class="active">
+                <UISref to="home.infrastructure">
+                  <a>Search</a>
+                </UISref>
+              </UISrefActive>
+            </li>
+            <li key="navProjects">
+              <UISrefActive class="active">
```
christopherthielen, line 60: let's migrate to `useSrefActive()`

PR #8378, opened 2020-06-24
fix(core/managed): Prevent the resource history modal from growing beyond viewport

vigneshm:
Setting a max-width for the table contents to prevent it from growing beyond viewport

<details>
<summary>Current Broken UI</summary>

image

</details>

<details>
<summary>Fix</summary>

image

</details>


app/scripts/modules/core/src/managed/ManagedResourceHistoryModal.tsx
```diff
@@ -146,7 +146,7 @@ const renderExpandedRowContent = (
   dismissModal: () => any,
 ) => {
   return (
-    <div className="flex-container-v left">
+    <div className="flex-container-v left" style={{ maxWidth: '90vw' }}>
```
erikmunson, line 149: nothing wrong with this per se, but you could also use the [built in max width controls](https://github.com/spinnaker/deck/pull/8237) on `Modal` if you like. I don't have a strong pref, the `Modal` API is a little more first class but this approach lives close to the content which is nice.

erikmunson:
@vigneshm anything you want to do with this PR before we merge it?

vigneshm:
I'm going to merge this for now while we debug why `Modal`'s `maxWidth` limitation didn't kick in to prevent the issue in the first place.

PR #8382, opened 2020-06-25
feat(core): New App Icon with refresher

caseyhebebrand:
This will replace `ApplicationRefresher`, `ApplicationIcon` and `Refresher` once the new navigation rolls out. 

The icon will toggle between `ApplicationFreshIcon` (or overridden Icon) and the stale icon. When it clicks it will refresh the page.

Fresh (not overridden):
Screen Shot 2020-06-25 at 4 18 12 PM
Stale w/ Tooltip:
Screen Shot 2020-06-25 at 4 18 35 PM

### DIFF
--- a/app/scripts/modules/core/src/application/nav/AppRefresher.tsx
+++ b/app/scripts/modules/core/src/application/nav/AppRefresher.tsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import { Observable, Subject } from 'rxjs';
+
+import { Application } from '../application.model';
+import { ApplicationDataSource, IFetchStatus } from '../service/applicationDataSource';
+
+import { AppRefresherIcon } from './AppRefresherIcon';
+
+export interface IAppRefresherProps {
+  app: Application;
+}
+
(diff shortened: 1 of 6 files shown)


app/scripts/modules/core/src/application/nav/AppRefresher.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+import { AppRefresherIcon } from './AppRefresherIcon';
+
+export interface IAppRefresherProps {
+  app: Application;
+}
+
+export const AppRefresher = ({ app }: IAppRefresherProps) => {
+  const app$ = new Subject<Application>();
+  const destroy$ = new Subject();
```
christopherthielen, line 15: This won't work the way you want because a new `app$` and `destroy$` is created for every render.

christopherthielen: My instinct is to remove most of the rxjs and just use hooks.  I think it would simplify the code.

caseyhebebrand:
> looks way better!

Agreed! `useObservable` is a handy hook.

PR #8386, opened 2020-06-29
fix(core/rosco): Do not default roscoMode if it has been already set

vigneshm:


vigneshm:
Looping in @maggieneterval to ensure this isn't affecting the intended behavior behind #8333.

app/scripts/modules/core/src/config/settings.ts
```diff
@@ -131,7 +131,7 @@ export const SETTINGS: ISpinnakerSettings = (window as any).spinnakerSettings ||
 
 // Make sure to set up some reasonable default settings fields so we do not have to keep checking if they exist everywhere
 SETTINGS.feature = SETTINGS.feature || {};
-SETTINGS.feature.roscoMode = SETTINGS.feature.roscoMode || true;
+SETTINGS.feature.roscoMode = 'roscoMode' in SETTINGS.feature ? SETTINGS.feature.roscoMode : true;
```
christopherthielen, line 134:
```suggestion
SETTINGS.feature.roscoMode = SETTINGS.feature.roscoMode ?? true;
```

maggieneterval:
🙈 So embarrassing, I forgot how javascript works, thank you so much for fixing!!

PR #8391, opened 2020-06-30
feat(core/pipeline): Add sort pipelines alphabetically button

christopherthielen:
reorder


christopherthielen:
Note: I put the "sort alphabetically" at the top of the UI because it's effectively a destructive action.  Previously, the "Done" button appeared in the exact same spot as the "Reorder" button was.   I didn't want people to quickly click in that area and accidentally sort their pipelines (with no way to undo).

PR #8392, opened 2020-07-01
fix(core/presentation): protect against width overflow on Modal + StandardGridTableLayout

erikmunson:
Recently an issue was exposed with how both `Modal` and `StandardGridTableLayout` handle content that overflows their intended width ([temporary hotfix PR](https://github.com/spinnaker/deck/pull/8378)). This PR attempts to protect against content overflow on both components, so that regardless of what's being displayed inside a table or modal overflowing works as one would expect:

1. Unclear to me whether it's always been this way vs. it being a regression from [my recent changes](https://github.com/spinnaker/deck/pull/8237), but as of this moment `Modal` will grow beyond both the viewport width and its own intended max width if something inside the modal gets greedy and forces itself to be a larger width. While the grid layout stuff I added in #8237 generally works, it turns out the downside is an inability to forcefully cap the width while retaining flexibility in the way we were aiming for. I took the approach here of moving back to a combination of `width` and `max-width` while keeping the rest of #8237 which gets us the behavior we want: modal content that is too wide for the modal's available width becomes horizontally scrollable without affecting the parent container, and the modal itself will flex up to the maximum available space until it hits the user defined (or default) max width.
2. The other discovery was that expanded row content in `StandardGridTableLayout` is not currently confined to whatever space is available to the table — if row content (in this case the diff tables in the history UI) decides it has to be a particular width, it is free to force the entire table to grow in width. With `Modal` fixed we at least get a graceful handling of that problem (you can scroll around within a properly sized modal to look at the now-giant table), but it's still a problem. Turns out it was easy enough to retain flexibility by taking the flex layout that housed the row and its expanded content and turning it into a grid, which combined with the already existing `overflow-x: scroll` gives us the right behavior.

In addition to getting sensible overflow handling for anything using either of these components in the future, the immediate-term result of 1 + 2 is that the history UI shows a properly sized modal, with a properly sized table, and a big horizontally scrollable diff table nested in the right spot:

horizontal-table-scroll

### DIFF
--- a/app/scripts/modules/core/src/presentation/modal/Modal.module.css
+++ b/app/scripts/modules/core/src/presentation/modal/Modal.module.css
@@ -122,15 +122,18 @@
 
 .dialogSizer {
   display: grid;
+  grid-template-columns: 1fr;
   grid-template-rows: 1fr;
   max-height: calc(100vh - 64px);
   min-height: 320px;
+  width: calc(100vw - 128px);
   margin: 0 64px;
 }
 
 @media belowMobileSize {
   .dialogSizer {
     max-height: calc(100vh - 32px);
+    width: calc(100vw - 48px);
     margin: 0 24px;
   }
 }
@@ -145,8 +148,8 @@
 .dialog {
   display: flex;
   position: relative;
-  grid-column: 1;
-  grid-row: 1;
+  grid-column: 1 / -1;
+  grid-row: 1 / -1;
   flex-direction: column;
   overflow: hidden;
   border-radius: 8px;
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/presentation/modal/Modal.module.css
```diff
@@ -145,8 +148,8 @@
 .dialog {
   display: flex;
   position: relative;
-  grid-column: 1;
-  grid-row: 1;
+  grid-column: 1 / -1;
+  grid-row: 1 / -1;
```
vigneshm, line 152: 🤯

erikmunson: grid is WILD. IMO the best thing to come out of the standards bodies in a very very long time.

PR #8397, opened 2020-07-07
feat(gce): Add GCP internal http(s) load balancer. 

takaaki7:
spinnaker/spinnaker#5042
This PR depends https://github.com/spinnaker/clouddriver/pull/4725


app/scripts/modules/google/src/loadBalancer/configure/internalhttp/createInternalHttpLoadBalancer.controller.js
```diff
@@ -0,0 +1,162 @@
+'use strict';
```
maggieneterval, line 1: All new files should be added in Typescript; please convert this file to TS.

app/scripts/modules/google/src/loadBalancer/configure/http/commandBuilder.service.js
```diff
@@ -44,52 +47,83 @@ module(GOOGLE_LOADBALANCER_CONFIGURE_HTTP_COMMANDBUILDER_SERVICE, [
     gceHttpLoadBalancerTransformer,
     loadBalancerReader,
     gceAddressReader,
+    gceXpnNamingService,
   ) {
-    function buildCommand({ originalLoadBalancer, isNew }) {
+    function buildCommand({ originalLoadBalancer, isNew, isInternal }) {
```
maggieneterval, line 52: Do we need to be passing around an `isInternal` boolean? I'm not very familiar with this code, but would prefer we find another way since none of the other load balancer types required an additional parameter to these methods.

app/scripts/modules/google/src/loadBalancer/configure/http/backendService/backendService.component.html
```diff
@@ -130,7 +128,7 @@
       </div>
     </div>
 
-    <div class="form-group">
+    <div ng-if="!$ctrl.loadBalancer.isInternal" class="form-group">
```
maggieneterval, line 131: This is a little confusing because Spinnaker already supports internal TCP/UDP load balancers, but I'm guessing the `isInternal` here means internal HTTP(S)? Could we check the type explicitly here instead?

app/scripts/modules/google/src/loadBalancer/configure/http/backendService/backendService.component.js
```diff
@@ -46,6 +46,14 @@ module(GOOGLE_LOADBALANCER_CONFIGURE_HTTP_BACKENDSERVICE_BACKENDSERVICE_COMPONEN
           .value();
       };
 
+      this.getSessionAffinitySuggestions = () => {
+        if (!this.loadBalancer.isInternal) {
```
maggieneterval, line 50: Similar to the comment in `app/scripts/modules/google/src/loadBalancer/configure/http/backendService/backendService.component.html`, can we check the load balancer type explicitly here rather than using the `isInternal` bool?

maggieneterval: (Same comment to everywhere reading `isInternal`.)

app/scripts/modules/google/src/loadBalancer/configure/http/basicSettings/basicSettings.component.html
```diff
@@ -39,12 +39,27 @@
           </account-select-field>
         </div>
       </div>
-      <div class="form-group">
```
maggieneterval, line 42: Is this safe to remove? I see we're only conditionally rendering the `gce-region-select-field` if the load balancer `isInternal`.

app/scripts/modules/google/src/loadBalancer/configure/http/httpLoadBalancer.write.service.js
```diff
@@ -33,9 +33,9 @@ angular
       const job = {
         type: 'deleteLoadBalancer',
         loadBalancerName: loadBalancer.listeners[0].name,
-        regions: ['global'],
-        region: 'global',
-        loadBalancerType: 'HTTP',
+        regions: [loadBalancer.region || 'global'],
+        region: loadBalancer.region || 'global',
+        loadBalancerType: loadBalancer.loadBalancerType,
```
maggieneterval, line 38: Sorry, again I'm pretty unfamiliar with this code...but did you make these changes in order for this service to be shared between external and internal HTTP(S) LBs?

app/scripts/modules/google/src/loadBalancer/loadBalancer.setTransformer.ts
```diff
@@ -15,10 +15,12 @@ export class GceLoadBalancerSetTransformer {
         name: loadBalancer.name,
         certificate: loadBalancer.certificate,
         ipAddress: loadBalancer.ipAddress,
+        subnet: loadBalancer.subnet,
       };
     });
 
     normalized.name = normalized.urlMapName;
+    delete normalized.subnet;
```
maggieneterval, line 23: What is the purpose of this?

app/scripts/modules/google/src/address/address.reader.ts
```diff
@@ -29,9 +29,9 @@ export interface IGceAddress {
 class GceAddressReader {
   public listAddresses(region?: string): IPromise<IGceAddress[]> {
     if (region) {
-      return this.listAddresses(null /* region */).then(addresses =>
-        addresses.filter(address => address.region === region),
-      );
+      return this.listAddresses(null /* region */).then(addresses => {
+        return addresses.filter(address => address.region === region);
+      });
```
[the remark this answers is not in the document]
takaaki7: No, I'll fixt it.

app/scripts/modules/google/src/healthCheck/healthCheck.read.service.ts
```diff
@@ -9,6 +9,7 @@ interface IHealthCheckSearchResults {
   name: string;
   account: string;
   healthCheck: string; // JSON encoded string containing the real health check (it is missing the account).
+  region: string;
```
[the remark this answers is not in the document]
takaaki7: OK, I'll fix it.

app/scripts/modules/google/src/loadBalancer/configure/choice/loadBalancerTypeToWizardMap.constant.ts
```diff
@@ -24,6 +24,12 @@ const loadBalancerTypeToWizardMap: IGceLoadBalancerToWizardMap = {
     editTemplateUrl: require('../http/editHttpLoadBalancer.html'),
     controller: 'gceCreateHttpLoadBalancerCtrl',
   },
+  INTERNAL_MANAGED: {
```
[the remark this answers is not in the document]
takaaki7: UI and backing data needed for this LB is different from other load balancers. So I think this is the simplest way.

app/scripts/modules/google/src/loadBalancer/configure/http/backendService/backendService.component.html
takaaki7: Yes, this field meas internal HTTP(S). But ok, I'll fix it.

app/scripts/modules/google/src/loadBalancer/configure/http/basicSettings/basicSettings.component.html
takaaki7: Yes, External HTTP LB is always global, so I think this form is meaningless.
Sorry for unrelated change for this PR. Should I revert it?

app/scripts/modules/google/src/loadBalancer/configure/http/commandBuilder.service.js
takaaki7: This is difficult problem.
I believe this way is better than others(e.g. copy and pasting)...
Do you have a better idea?

app/scripts/modules/google/src/loadBalancer/configure/http/httpLoadBalancer.write.service.js
takaaki7: Yes, I believe sharing these code is easier than other ways at deck...

app/scripts/modules/google/src/loadBalancer/configure/internalhttp/createInternalHttpLoadBalancer.controller.js
takaaki7: OK, I'll fix it.

app/scripts/modules/google/src/loadBalancer/httpLoadBalancerUtils.service.ts
```diff
@@ -8,8 +8,7 @@ export class GceHttpLoadBalancerUtils {
   public isHttpLoadBalancer(lb: IGceLoadBalancer): lb is IGceHttpLoadBalancer {
     return (
       (lb.provider === 'gce' || lb.type === 'gce') &&
-      lb.loadBalancerType === 'HTTP' &&
-      lb.region === GceHttpLoadBalancerUtils.REGION
```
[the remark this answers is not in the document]
takaaki7: Because External HTTP LB is always global, this condition is meaningless.

app/scripts/modules/google/src/loadBalancer/loadBalancer.setTransformer.ts
takaaki7: Without `delete`, the value is shown at here. (basic section of lb details)
https://github.com/spinnaker/deck/blob/master/app/scripts/modules/google/src/loadBalancer/details/loadBalancerDetails.html#L83

But actually subnet is value of each listeners for Internal HTTP LB.

takaaki7:
Thank you for review. And I also need review for spinnaker/clouddriver#4725

takaaki7:
I've fixed some points you commented.

takaaki7:
I've converted createInternalHttpLoadBalancer.controller.js  to ts.

app/scripts/modules/google/src/loadBalancer/configure/http/basicSettings/basicSettings.component.html
maggieneterval: Oh I see, thanks for the explanation! You don't need to revert it, but in future PRs it's helpful to separate out refactors not related to the main change into their own commits if possible!

app/scripts/modules/google/src/loadBalancer/configure/internalhttp/createInternalHttpLoadBalancer.controller.ts
```diff
@@ -0,0 +1,172 @@
+'use strict';
```
maggieneterval, line 1:
```suggestion

```
I think we can remove this now that this file is in TS.

app/scripts/modules/google/src/loadBalancer/configure/internalhttp/createInternalHttpLoadBalancer.controller.ts
```diff
@@ -0,0 +1,172 @@
...
+
+import { IScope, module } from 'angular';
+
+import { Application, TaskMonitor } from '@spinnaker/core';
+
+import { GCE_CACHE_REFRESH } from 'google/cache/cacheRefresh.component';
+import { BackendServiceTemplate, HealthCheckTemplate, HostRuleTemplate, ListenerTemplate } from '../http/templates';
+
+import '../http/httpLoadBalancerWizard.component.less';
```
maggieneterval, line 10: Could we please rearrange these imports to adhere to Deck's [import conventions](https://github.com/spinnaker/deck/wiki/Conventions#imports)?

app/scripts/modules/google/src/loadBalancer/configure/http/commandBuilder.service.js
maggieneterval: Definitely prefer this approach to copy/pasting! I didn't realize until now that this service was specific to HTTP LBs. I thought this was general to all LBs, but in the context of just HTTP LBs, the `isInternal` boolean makes a lot more sense! Sorry about that!

maggieneterval:
Thanks for the quick updates, @takaaki7! Just two more small change requests to the new TS file and this should be good to merge once the corresponding Clouddriver change is merged. 💯 🎉 👍

takaaki7:
Thank you for review! I've addressed your comments.

PR #8404, opened 2020-07-09
fix(core): Layout v1 feedback

caseyhebebrand:
A batch of fixes from the layout rollout: 

1. Update `collapsibleSectionStateCache` to preserve toggling nav when refreshing
2. Set pipeline graph width, and make container scrollable
3. Make pipeline stage container flex
4. Prevent task exception from overflow by adding a word break
5. Add padding to details panel so text isn't truncated when there is overflow
6. `cursor: pointer` on new app refresh icon

### DIFF
--- a/app/scripts/modules/core/src/application/nav/navAtoms.ts
+++ b/app/scripts/modules/core/src/application/nav/navAtoms.ts
@@ -3,5 +3,7 @@ import { CollapsibleSectionStateCache } from '../../cache/collapsibleSectionStat
 
 export const verticalNavExpandedAtom = atom({
   key: 'verticalNavExpanded',
-  default: !CollapsibleSectionStateCache.isSet('verticalNav') || CollapsibleSectionStateCache.isExpanded('verticalNav'),
+  default:
+    !CollapsibleSectionStateCache.isSet('verticalNav') ||
+    (CollapsibleSectionStateCache.isSet('verticalNav') && CollapsibleSectionStateCache.isExpanded('verticalNav')),
 });
--- a/app/scripts/modules/core/src/application/nav/navAtoms.ts
+++ b/app/scripts/modules/core/src/application/nav/navAtoms.ts
@@ -4,6 +4,5 @@ import { CollapsibleSectionStateCache } from '../../cache/collapsibleSectionStat
 export const verticalNavExpandedAtom = atom({
   key: 'verticalNavExpanded',
   default:
-    !CollapsibleSectionStateCache.isSet('verticalNav') ||
-    (CollapsibleSectionStateCache.isSet('verticalNav') && CollapsibleSectionStateCache.isExpanded('verticalNav')),
+    !CollapsibleSectionStateCache.isSet('verticalNav') || CollapsibleSectionStateCache.isExpanded('verticalNav'),
 });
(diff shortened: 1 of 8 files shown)


app/scripts/modules/core/src/pipeline/config/graph/PipelineGraph.tsx
```diff
@@ -272,6 +274,12 @@ export class PipelineGraph extends React.Component<IPipelineGraphProps, IPipelin
     }
     newState.maxLabelWidth = Math.max(this.minLabelWidth, newState.maxLabelWidth);
 
+    if (newState.maxLabelWidth === this.minLabelWidth) {
+      newState.graphWidth = (newState.phaseCount + 1) * (newState.maxLabelWidth + phaseOffset) + 5 + 'px';
```
erikmunson, line 278: I have zero context about any of this graph layout code so I apologize for my confusion — are there specific cases/flows where this condition applies? And what's different behaviorally between setting this pre-calculated fixed width and using `100%`?

math

app/scripts/modules/core/src/application/nav/navAtoms.ts
```diff
@@ -3,5 +3,7 @@ import { CollapsibleSectionStateCache } from '../../cache/collapsibleSectionStat
 
 export const verticalNavExpandedAtom = atom({
   key: 'verticalNavExpanded',
-  default: !CollapsibleSectionStateCache.isSet('verticalNav') || CollapsibleSectionStateCache.isExpanded('verticalNav'),
+  default:
+    !CollapsibleSectionStateCache.isSet('verticalNav') ||
+    (CollapsibleSectionStateCache.isSet('verticalNav') && CollapsibleSectionStateCache.isExpanded('verticalNav')),
```
christopherthielen, line 8:
```suggestion
    !CollapsibleSectionStateCache.isSet('verticalNav') || CollapsibleSectionStateCache.isExpanded('verticalNav'),
```

app/scripts/modules/core/src/pipeline/config/graph/PipelineGraph.tsx
christopherthielen: I'm wondering the same thing as Erik.  How does this work?  When is it triggered?
making any changes to this code makes me really nervous 😆 are there dragons here?

Is this to allow the horizontal scrolling?  It's a pretty minimal change, but I wonder if css could be applied to the outer element instead?  Either way, we should merge these changes as-is to unblock any users, and come back to it in another PR if there's a better approach landed upon.

caseyhebebrand: @christopherthielen @erikmunson I am actually putting this code back in if that makes you feel any better :) I originally took it out, set the width to 100% and made the containing div responsive. 

The reason I added this back is because of how the svg behaves. With just setting the width to `100%` it truncates the graph. I ended up appling the css to the outer element to enable that to be flex w/ scroll, but without explicitly setting the width the image is usually defaulted to a very small size.

christopherthielen: > I am actually putting this code back in if that makes you feel any better :)

HAHA 👍  🤣  ok! 🌮

erikmunson: ✅✅✅ WOW so glad we had this chat

caseyhebebrand: LOL Unclear if there are now old dragons back in business 🐲 🤣

PR #8419, opened 2020-07-18
fix(project dashboard) columns not overlapping

Chinikins:
When adding clusters to the project dashboard, it often ends up overlapping on the application pipelines side as discussed here:  https://github.com/spinnaker/spinnaker/issues/4925

Added a subclass to make the columns on the project dashboard scrollable.


christopherthielen:
Thanks for fixing this!  I think we could scope the CSS a little tighter like this:

```
.project-dashboard {
  overflow-y: auto;
  h3 {
    font-weight: 600;
  }

  .project-column {
    overflow-x: scroll;
  }
}
```

Second, the spinnaker bot is complaining about the commit message because it's missing a colon.  We generally use `module`/`subdirectory` as the scope:
```
fix(core/projects): columns not overlapping
```

PR #8423, opened 2020-07-22
fix(core): CSS for details panel

caseyhebebrand:
Two css fixes: 
1. The project header being full-width was being overwritten by custom styles. This moves the `width: 100%` to a more specific set of classes to ensure there are no issues.
2. Tweaking the height of the details panel so there are no scrolling issues.

### DIFF
--- a/app/scripts/modules/core/src/presentation/main.less
+++ b/app/scripts/modules/core/src/presentation/main.less
@@ -286,12 +286,11 @@ html {
 
   .detail-content {
     z-index: var(--layer-high);
-    position: fixed;
+    position: absolute;
     margin-top: -20px;
     right: 0;
-    height: 98%;
+    height: 100%;
     width: 390px;
-    padding-bottom: 28px;
 
     @media (max-width: 390px) {
       width: 100%;
--- a/app/scripts/modules/core/src/presentation/main.less
+++ b/app/scripts/modules/core/src/presentation/main.less
@@ -289,9 +289,9 @@ html {
     position: fixed;
     margin-top: -20px;
     right: 0;
-    height: 100%;
+    height: 98%;
     width: 390px;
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/presentation/main.less
```diff
@@ -289,9 +289,9 @@ html {
     position: fixed;
     margin-top: -20px;
     right: 0;
-    height: 100%;
+    height: 98%;
```
christopherthielen, line 292: this seems odd, why?

christopherthielen:
what custom styles were overriding the width, and how were the selectors getting overridden?

caseyhebebrand:
> what custom styles were overriding the width, and how were the selectors getting overridden?

There are generic classes in the FP css that were overriding the use of these classes in core. I also am going to make a supplementary PR that removes styling from the general classes so it can be maintained from one place.

app/scripts/modules/core/src/presentation/main.less
caseyhebebrand: I noticed that the width at 100% was pushing the drawer out of the screen height, which is why the scrolling issues were cropping up. This was more dramatic from a projects view or with a custom header.

app/scripts/modules/core/src/projects/project.less
```diff
@@ -57,6 +58,12 @@
       margin-top: -10px;
     }
   }
+
+  .insight {
+    .detail-content {
+      height: calc(~'100% - 54px');
```
gcomstock, line 64: Is it worth storing 54px as a variable that's shared between here and the header height?

caseyhebebrand: 👍

PR #8424, opened 2020-07-22
feat(core): refactor insight menu from dropdown to buttons

dogonthehorizon:
The motivation for this change is to better surface an action for the user to
take when they are on the search, project, and application pages. The current
UX obscures the next appropriate action behind a drop down menu that visually
blends into the page, making it difficult for users to know what to do next.

The proposed change is to replace the drop down menu with buttons rendered
appropriately for each page. When on the search page, both create project and
application buttons are rendered, with a call to action for create application.
When on project or application pages, only render those buttons. Finally, if a
cache refresh is needed, render this on all pages when appropriate.

At a technical level, this unifies the three dropdowns which currently depend
on 2 separate angular controllers and a react component. The component has been
used in favor of the angular implementations on both search and project pages.

The only thing I'm not really sure of is if I've done the `react2angular` thing correctly 🤷 I'm sure someone will this out in the PR :)

Here are a few screenshots of how this looks:
<kbd></kbd>
<kbd></kbd>
<kbd></kbd>

### DIFF
--- a/app/scripts/modules/core/src/search/infrastructure/infrastructure.controller.js
+++ b/app/scripts/modules/core/src/search/infrastructure/infrastructure.controller.js
@@ -4,6 +4,8 @@ import _ from 'lodash';
 
 import * as angular from 'angular';
 
+import { react2angular } from 'react2angular';
+
 import { FirewallLabels } from 'core/securityGroup/label';
 import { SEARCH_RANK_FILTER } from '../searchRank.filter';
 import { OVERRIDE_REGISTRY } from 'core/overrideRegistry/override.registry';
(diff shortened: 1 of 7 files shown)


dogonthehorizon:
cc: @ncknt @noahi

app/scripts/modules/core/src/search/infrastructure/infrastructure.controller.js
```diff
@@ -17,6 +19,7 @@ import { ClusterState } from 'core/state';
 
 import { SearchService } from '../search.service';
 import { ConfigureProjectModal } from 'core/projects';
+import {InsightMenu as SearchInsightMenu} from 'core/insight/InsightMenu';
```
christopherthielen, line 22: This looks like it doesn't match the formatting style. We have a post-commit hook that runs `prettier`, did it somehow fail to run?

From the command line, you can run ```npx prettier --write app/scripts/modules/core/src/search/infrastructure/infrastructure.controller.js``` and see if it updates the formatting.  If so, it'd be great to run prettier on all the files in this PR.

dogonthehorizon: Yep, just ran and pushed up a squashed commit w/ the prettified files. Thanks!

dogonthehorizon:
TAL at the functional test failure, looks like some need to be updated

PR #8427, opened 2020-07-24
fix(bake): show advanced options

karlomendozaarmory:
Show Checkbox so users can select the view the "Advanced options" section.


karlomendozaarmory:
For the moment, and in the foreseeable future we would not requiere any new fields in this section.
I think we are good with this. I went with the same patterns the AWS stage uses.

PR #8428, opened 2020-07-24
fix(environments): Allow environments view to render when there are managed artifacts and/or resources

luispollo:
Currently the code looks at the `hasManagedResources` flag from the API response to determine if an app is managed, however the app may have a delivery config with just artifacts, which is a valid state, but the page won't render. This PR makes a tiny change such that we look at the sizes of the `resources` and `artifacts` arrays instead of that flag to avoid that problem.

### DIFF
--- a/app/scripts/modules/core/src/managed/Environments.tsx
+++ b/app/scripts/modules/core/src/managed/Environments.tsx
@@ -67,7 +67,7 @@ interface IEnvironmentsProps {
 export function Environments({ app }: IEnvironmentsProps) {
   const dataSource: ApplicationDataSource<IManagedApplicationEnvironmentSummary> = app.getDataSource('environments');
   const {
-    data: { environments, artifacts, resources, hasManagedResources },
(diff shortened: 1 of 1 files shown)


vigneshm:
I do see some other uses of `hasManagedResources` being used in deck like [here](https://github.com/spinnaker/deck/blob/58ae56714db4e6655debb45aa17d3986b4e3da1c/app/scripts/modules/core/src/application/config/applicationConfig.view.html#L9) which seem like should be enabled like this scenario as well?

luispollo:
> I do see some other uses of `hasManagedResources` being used in deck like [here](https://github.com/spinnaker/deck/blob/58ae56714db4e6655debb45aa17d3986b4e3da1c/app/scripts/modules/core/src/application/config/applicationConfig.view.html#L9) which seem like should be enabled like this scenario as well?

Yeah, I saw that but wasn't sure where that `config` object was coming from or what its shape was there... 😬

PR #8430, opened 2020-07-24
fix(core/pipeline): Reformat fixed footer

caseyhebebrand:
The fixed footer for pipelines is now partially hidden by the new nav, preventing some buttons from being used. This PR reformats the footer so that the buttons are all on the right side of the footer. 
Screen Shot 2020-07-24 at 3 34 31 PM


christopherthielen:
ok

PR #8431, opened 2020-07-24
fix(core/pipeline): Pipeline param execution details should handle Ob…

caseyhebebrand:
Many pipelines depend on params that are not strings. Deck crashes if the param is an object, so let's make it handle that better.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/pipeline/PipelineParametersExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/pipeline/PipelineParametersExecutionDetails.tsx
@@ -10,13 +10,7 @@ export function PipelineParametersExecutionDetails(props: IExecutionDetailsSecti
   } = props;
 
   const { pipelineParameters: parameters = {} } = context;
-  // eslint-disable-next-line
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/pipeline/config/stages/pipeline/PipelineParametersExecutionDetails.tsx
```diff
@@ -22,7 +22,7 @@ export function PipelineParametersExecutionDetails(props: IExecutionDetailsSecti
               .map(key => (
                 <React.Fragment key={key}>
                   <dt>{key}</dt>
-                  <dd>{parameters[key]}</dd>
+                  <dd>{parameters[key].toString()}</dd>
```
vigneshm, line 25: We should probably plan for a long-term fix to show the object values instead of '[object Object]'?

caseyhebebrand: I agree. Weirdly enough, it has always shown '[object Object]' , so at least we are back to ground zero 😄

PR #8432, opened 2020-07-27
feat(notifications): add support for dynamically loaded extension not…

danielpeach:
…ifications

This is a follow-up to https://github.com/spinnaker/gate/pull/1288 and https://github.com/spinnaker/echo/pull/981.

### DIFF
--- a/app/scripts/modules/core/src/domain/INotificationTypeConfig.ts
+++ b/app/scripts/modules/core/src/domain/INotificationTypeConfig.ts
@@ -5,7 +5,6 @@ export interface INotificationTypeConfig {
   key: string;
   config?: INotificationTypeCustomConfig;
   component?: React.ComponentType;
-  isExtensionNotification?: boolean;
 }
 
 export interface INotificationTypeCustomConfig {
--- a/app/scripts/modules/core/src/domain/INotificationTypeConfig.ts
+++ b/app/scripts/modules/core/src/domain/INotificationTypeConfig.ts
@@ -1,11 +1,11 @@
 import React from 'react';
-import { INotificationSettings } from 'core/config';
 
 export interface INotificationTypeConfig {
   label: string;
-  key: keyof INotificationSettings;
+  key: string;
   config?: INotificationTypeCustomConfig;
   component?: React.ComponentType;
+  isExtensionNotification?: boolean;
 }
 
 export interface INotificationTypeCustomConfig {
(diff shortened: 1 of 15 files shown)


app/scripts/modules/core/src/domain/INotificationTypeConfig.ts
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
-import { INotificationSettings } from 'core/config';
 
 export interface INotificationTypeConfig {
   label: string;
-  key: keyof INotificationSettings;
+  key: string;
   config?: INotificationTypeCustomConfig;
   component?: React.ComponentType;
+  isExtensionNotification?: boolean;
```
christopherthielen, line 8: This moves knowledge of plugins *into* the registry, which doesn't feel quite right to me. What other options had you considered?

christopherthielen:
Is this PR intended to support notifications as plugins? With the preconfigured job stage, we can wire in a custom UI for the stage config.  I don't see a similar mechanism in this PR, but think it would be nice to support.

app/scripts/modules/core/src/domain/INotificationTypeConfig.ts
danielpeach: I wanted to avoid making users have to configure the plugin AND opt in to the notification in `SETTINGS.notifications`. But if that had to happen, that's OK.

Another option, which is definitely better than what I wrote, would be to include an optional `enabled` field on `INotificationTypeConfig`. This would effectively be the same thing but less plugin-specific.

danielpeach:
Sure, that makes sense. I'll update this to include a `uiType` field and make a corresponding change in Echo.

app/scripts/modules/core/src/domain/INotificationTypeConfig.ts
christopherthielen: I think we could move the registration of the built-in notifications into a separate function.  That way, the registry itself doesn't check if something is `.enabled`. 

For example:
```diff
import { Registry } from 'core/registry';

import { BearychatNoficationType } from './BearychatNotificationType';

-Registry.pipeline.registerNotification({
+export const bearyChatNotification = {
  component: BearychatNoficationType,
  key: 'bearychat',
  label: 'Bearychat',
};
```

Then, somewhere else like `notification/notification.types.ts` you could do:

psuedocode
```js
import { bearyChatNotification } from './types'
import { slackNotification } from './types'
...

const builtInNotificationTypes = [bearyChatNotification, slackNotification, ...other types].forEach(type =>  {
  if (SETTINGS. /* check settings.enabled for the built in ones */) {
    Registry.registerTheEnabledNotification(type);
  }
});
```

danielpeach: done

christopherthielen:
nice work

PR #8438, opened 2020-07-28
refactor(kubernetes): exclude manifest from infrastructure interfaces

maggieneterval:
* refactor(kubernetes): exclude manifest from IKubernetesLoadBalancer 

  As part of ongoing Kubernetes performance improvements, Clouddriver will stop caching the entire Kubernetes resource YAML. In the immediate term, we will stop returning the manifest as part of top-level /loadBalancers, /serverGroups, and /securityGroups endpoints. Let's instead block LB details view loading on a separate /manifests request, and use that to hydrate the template.

* fix(kubernetes): fix alignment of Host Rules and Ingress sections 

* refactor(kubernetes): exclude manifest from IKubernetesServerGroup 

  Also refactors ManifestTrafficService to check for associated services in the `loadBalancers` field of a server group rather than check the server group's manifest's annotations, and removes unnecessary extractEntityTags method which is just re-setting entityTags on the same serverGroup from which it was reading it.

* refactor(kubernetes): exclude manifest from IKubernetesServerGroupManager 

* refactor(kubernetes): remove manifest from IKubernetesInstance 

* refactor(kubernetes): exclude manifest from IKubernetesSecurityGroup 

  Also add `account` to interface, because unlike the other infra types, the base type does not include `account` (but we are always getting this back from Clouddriver for Kubernetes SGs).

* fix(kubernetes): include manifest as $interpolation context for annotation-driven UI 

  The Kubernetes [annotation-driven UI](https://spinnaker.io/guides/user/kubernetes-v2/annotations-ui/) expects the full manifest to be available to hydrate custom UI sections specfied as annotations. Since we already have access to the full manifest in this component off of which we read each annotation, simply include it as part of the context we pass to $interpolate.


app/scripts/modules/kubernetes/src/loadBalancer/details/details.controller.ts
```diff
@@ -70,35 +57,33 @@ class KubernetesLoadBalancerDetailsController implements IController {
...
+      return this.autoClose();
     }
+
+    ManifestReader.getManifest(accountId, region, name).then((manifest: IManifest) => {
+      this.manifest = manifest;
+      this.loadBalancer = {
+        ...rawLoadBalancer,
+        apiVersion: manifest.manifest.apiVersion,
+        displayName: manifest.manifest.metadata.name,
+        namespace: rawLoadBalancer.region,
```
maggieneterval, line 83: @ezimanyi I didn't touch this because this diff was already getting out of control, but since we add these three fields to most of the infrastructure types, feels like we should either:

(1) Just read them directly off the manifest in the templates that need them
(2) Have Clouddriver include them in the payload

Originally, I would have said that going with (1) precludes us from future refactors in which the details sections that actually need the full manifest ("Edit" action button, for instance) could just fetch it for themselves on-demand, but since we need to respect the contract of the annotations-driven UI and so will need the manifest immediately to render any number of custom sections, I am now leaning toward (1).

app/scripts/modules/kubernetes/src/loadBalancer/details/details.controller.ts
```diff
@@ -29,25 +27,14 @@ class KubernetesLoadBalancerDetailsController implements IController {
...
-            name: this.loadBalancerFromParams.name,
-          },
-          this,
-        );
-        this.$scope.$on('$destroy', () => {
-          unsubscribe();
-        });
-      });
+        this.extractLoadBalancer(loadBalancer);
+        dataSource.onRefresh(this.$scope, () => this.extractLoadBalancer(loadBalancer));
```
ezimanyi, line 35: This is mostly just out of curiosity, but is there no way to collapse these two calls into one?  It looks like everywhere we do X when the data source is ready, then also subscribe on refresh to do X again.  I guess the data source probably just doesn't expose a "do X when ready and every time it refreshes"...

app/scripts/modules/kubernetes/src/loadBalancer/details/details.controller.ts
ezimanyi: Assuming that all those templates will need the manifest anyway (which it seems they will based on the annotation-driven UI) I agree that going with (1) makes more sense.

app/scripts/modules/kubernetes/src/loadBalancer/details/details.controller.ts
maggieneterval: We have a [custom React hook](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/hooks/useDataSource.hook.ts) for (I think) exactly that now! We can use them when we convert these details views to React, but would likely want to start by ensuring all leaf node components in each details template are React first. But yeah, great question, my understanding of one of the primary reasons for the introduction of React hooks is that the pattern of "fetch something once after first render and then set up a subscription or re-fetch in response to other things changing" in components written as classes had duplicated and convoluted logic in `componentDidMount` and `componentDidUpdate` methods, so they made a more first-class way with hooks. Better explanation [here](https://reactjs.org/docs/hooks-intro.html#complex-components-become-hard-to-understand) if you are interested.

ezimanyi: Thank you for the excellent explanation! 💯

PR #8441, opened 2020-07-29
config(auth): update default value of authEnabled

micnncim:
This PR suggests `authEnabled`'s default value should be changed from `true` to `false`. The current one, `true`, wouldn't be intuitive for users as a boolean default value is `false` (expect something like `debug`) in general. It becomes `true` even if users don't configure it.


settings.js
```diff
@@ -5,7 +5,7 @@
 var apiHost = process.env.API_HOST || 'http://localhost:8084';
 var atlasWebComponentsUrl = process.env.ATLAS_WEB_COMPONENTS_URL;
 var authEndpoint = process.env.AUTH_ENDPOINT || apiHost + '/auth/user';
-var authEnabled = process.env.AUTH_ENABLED === 'false' ? false : true;
+var authEnabled = process.env.AUTH_ENABLED === 'true' ? true : false;
```
maggieneterval, line 8:
```suggestion
var authEnabled = process.env.AUTH_ENABLED === 'true';
```
Would this be equivalent?

micnncim:
@maggieneterval Thanks for your review!

> this settings.js file is mainly for local development for non-Halyard users

Yeah, of course, I know about that. While running deck locally to experiment something, I've found it doesn't make sense well.

settings.js
micnncim: That's quite true. I'll change it as you expect. But how about the others, like l20 (`const entityTagsEnabled = process.env.ENTITY_TAGS_ENABLED === 'true' ? true : false;`)?

maggieneterval: No need to update the rest of the file; just thought it might be more readable without the ternary since we're making an edit to this line anyway!

micnncim: I agree with you. Have updated that. Thank you 😄

christopherthielen:
@maggieneterval I see ready to merge, but no approval.  Is this good to go?

maggieneterval:
@christopherthielen thanks, thought I had approved before adding that label but must not have gone through!

maggieneterval:
@micnncim could you please make sure the ""Allow edits from maintainers" checkbox is checked for this PR? The Mergify bot is trying to merge latest master into your branch, which requires you to permit external edits. Thank you so much!

micnncim:
@maggieneterval Sorry for the late response :bow: I've done.

PR #8444, opened 2020-07-29
fix(amazon): Ensure successful security group decompression

caseyhebebrand:
Sometimes security groups from the cache are already partially decompressed. This caused an error in the decompression methods. 

The cached groups start off looking like this: 
```js
{ account: { provider: { region: { vpcId:  Array[] } } } }
```

But sometimes are returned from the cache like this: 
```js
{ account: { provider: { region: Array[]  } } }
```

In the latter case, the security groups are already partially decompressed and this error prevents the user from being able to open the edit modal.

Screen Shot 2020-07-29 at 1 59 34 PM


caseyhebebrand:
I was not sure if this is the best way to fix this, so if anyone is familiar with the cache that would be great extra context. 

It almost seems like during compression, sometimes `service.supportsCompression` is true and other times it is not, but I could never confirm this with my debugging.

PR #8447, opened 2020-07-30
fix(core/securityGroup): Clone cached data before mutating it

christopherthielen:
Closes #8444


christopherthielen:
This was a pain in the 🌹 to debug.  We were mutating the cached object in the decompress function.  Now, we clone the cached object before decompressing it.

caseyhebebrand:
> This was a pain in the 🌹 to debug. We were mutating the cached object in the decompress function. Now, we clone the cached object before decompressing it.

Nice work! I know it had something to do with the compress function. This explains why my simple logging before and after the cache returned different things.

PR #8451, opened 2020-07-30
fix(amazon/rollback): Wait 500 ms before opening a second modal

christopherthielen:
When the react-bootstrap based confirmDialog closes, it removes modal-open class from the body tag after a 300ms fade out animation.  this causes the ui-bootstrap modal to not be styled properly.  Adding this delay allows the animation to complete and the modal-open class removed from the body tag.  Then, ui-bootstrap re-adds modal-open when it shows the rollback dialog.

As much as I hate these hacks, the react modal code doesn't currently offer a reasonable way to hook into the lifecycle of the animation, only of the ok/cancel buttons being clicked.


christopherthielen:
Jul-30-2020 10-39-33

PR #8459, opened 2020-08-03
fix(core/cluster): Prevent cluster filter from submitting when enter key is pressed

vigneshm:
The cluster filter already filters by responding to key changes and hitting enter refreshes the whole page. Preventing the form submit so that whole page doesn't get refreshed.

### DIFF
--- a/app/scripts/modules/core/src/cluster/filter/FilterSearch.tsx
+++ b/app/scripts/modules/core/src/cluster/filter/FilterSearch.tsx
@@ -12,7 +12,7 @@ export interface IFilterSearchProps {
 
 export const FilterSearch = ({ helpKey, onBlur, onSearchChange, value }: IFilterSearchProps) => (
   <div className="FilterSearch sp-margin-s-right">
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/cluster/filter/FilterSearch.tsx
```diff
@@ -20,6 +20,11 @@ export const FilterSearch = ({ helpKey, onBlur, onSearchChange, value }: IFilter
           value={value}
           onBlur={onBlur}
           onChange={onSearchChange}
+          onKeyUp={(event: React.KeyboardEvent<HTMLInputElement>) => {
+            if (event.which == 13) {
```
caseyhebebrand, line 24: Is `event.which` sometimes a `string`? Just wondering if there is a reason for double equals vs triple.

vigneshm: My bad, didn't notice I had used only double equals

PR #8460, opened 2020-08-03
fix(core/pipeline): Fix dropping of characters while filtering

vigneshm:
`ExecutionFilters.tsx` component has a state variable `searchString` to hold the current search/filter term entered by the user, but uses `ExecutionState.filterModel.asFilterModel.sortFilter.filter` to render the search term in the input. This results in a poor experience since `ExecutionState.filterModel.asFilterModel.sortFilter` is updated via debounce.

In this PR, I'm making `searchString` the source of truth for the currently rendered search term in the UI to avoid the above issue.


christopherthielen:
Doe the `searchString` sync back to the filter model somehow?

vigneshm:
> Doe the `searchString` sync back to the filter model somehow?

Yep, it [does](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/filter/ExecutionFilters.tsx#L154-L163), but with a delay since `debounce` is used

PR #8461, opened 2020-08-03
feat(amazon/alb): Support for http-request-method rule conditions on ALBs

caseyhebebrand:
Enable support for listener rules that have conditions of type `http-request-method`. 

These conditions have a different model than `path-pattern` or `host-header` conditions. Prior to this PR, these conditions could only be created in the aws console, and would fail to render when editing an ALB.

Screen Shot 2020-08-03 at 10 24 43 AM

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
@@ -133,7 +133,15 @@ export class ALBListeners extends React.Component<IALBListenersProps, IALBListen
         const missingAuth = !!rule.actions.find(
           a => a.type === 'authenticate-oidc' && !a.authenticateOidcConfig.clientId,
         );
-        const missingValue = !!rule.conditions.find(c => c.values.includes(''));
+        const missingValue = !!rule.conditions.find(c => {
+          if (c.field !== 'http-request-method') {
+            return c.values.includes('');
+          }
+
+          return c.httpRequestMethodConfig && c.httpRequestMethodConfig.values.includes('');
+        });
+        // eslint-disable-next-line
+        console.log(missingValue);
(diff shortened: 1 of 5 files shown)


app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
```diff
@@ -328,6 +328,26 @@ export class ALBListeners extends React.Component<IALBListenersProps, IALBListen
...
+  ): void => {
+    let newValues = condition.values || [];
+
+    if (selected) {
+      newValues.push(newValue);
+    } else {
+      newValues = newValues.filter(v => v !== newValue);
+    }
+
+    condition.values = newValues;
```
christopherthielen, line 344: What's the reason to duplicate the array in `values` and `httpRequestMethodConfig.values`?

caseyhebebrand: For this condition type, `httpRequestMethodConfig` is the source of truth from aws, and then I transform it to `values` to match the model of the other conditions in the UI. We don't really use `httpRequestMethodConfig` in the UI (except in the LB transform), but  I just wanted to keep the values consistent.

christopherthielen: will you add that as a comment in the source code?  The way I see it, this is the kind of tech debt that is impossible to pay off because future me doesn't remember if the fields are used or not.

christopherthielen: The alternative would be to transform both ways (when sending and receiving), but I'm not sure if that's feasible here.

caseyhebebrand: Yes I will add a comment

PR #8467, opened 2020-08-04
fix(core/application): Handle edits to cloud provider in application config links

vigneshm:
There are two issues related to editing cloud providers field in application config links.

1. The `on-select` callback in `<ui-select multiple>` was never invoked due to this [specific override](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/bootstrap/uiSelect.config.ts#L16) that we do in `uiSelect.config`. This override prevented the [default behavior](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectMultipleDirective.js#L170-L185) of `ui-select`.
2. We perform a [string compare](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/application/config/links/applicationLinks.component.js#L94) between the edited value and the original value to toggle 'Save' button. This doesn't take into account two identical empty `cloudProviders` values `[]` and `null`.

This PR fixes the above issues.


app/scripts/modules/core/src/bootstrap/uiSelect.config.ts
```diff
@@ -20,6 +21,17 @@ function uiSelectDecorator($provide: ng.auto.IProvideService) {
...
+              [$select.parserResult.itemName]: item,
+            };
+            $timeout(function() {
+              if ($select.onSelectCallback) {
+                $select.onSelectCallback(scope, {
+                  $item: item,
+                  $model: $select.parserResult.modelMapper(scope, locals),
+                });
+              }
+            });
```
vigneshm, line 34: This change is the only bits missing from the [original default implementation](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectMultipleDirective.js#L170-L185) in `uiSelect` and is responsible for invoking the `on-select` callback.

vigneshm:
Converting this PR to a draft as I'm verifying a few things.

PR #8468, opened 2020-08-05
fix(aws/functions): normalizeFunction is synchronous 

kevinawoo:
https://github.com/spinnaker/deck/commit/72af6cc495cf9f82634f62631bba1f7b6e148eb7 introduced a bug when refactoring that calling `normalizeFunction` always returned a Promise.

When it was removed, some calls to `normalizeFunction` may return plain objects that aren't wrapped in a Promise.

**Fix*:* wrap the return value with a `$q` to satisfy `IPromise`

This needs to get backported to `release-1.21.x`


kevinawoo:
Hey @maggieneterval, thanks for the detailed review!

I took another look and realized the call is only synchronous code. I went ahead and just removed the `$q` wrapping and usage. It's unlikely going to be async because it's just in memory data massaging now. If we need to add back in some async call, we can wrap it again later.

I also got pretty confused reading through what was there, so converted `function.transformer.js` to typescript to help with some types checking.

Also added some tests to make sure it works!

kevinawoo:
@mergifyio backport release-1.21.x

PR #8469, opened 2020-08-05
fix(core/nav): Active nav category for hyperlinks

caseyhebebrand:
Some hyperlinks were not showing the active state on the nav bar because `useSrefActive` was not assigning the `active` class to them. In order to fix this, `useIsActive` can now also add the active class to the `NavRoute`.


christopherthielen:
🤔 seems like useSrefActive should work.  let's investigate why it's not working.

caseyhebebrand:
> 🤔 seems like useSrefActive should work. let's investigate why it's not working.

I spent time yesterday investigating this, but can take a second pass. I am pretty confident that this is an edge case this effect will not fully solve because some hyperlinks are not exact child urls. 

Example: For executions, the sref is `pipelines.executions`, but when you permalink to the details the sref has `pipelines.executionDetails.execution`. This is considered not active by the `useSrefActive` effect. However, we have to generate the `sref` with the dataSource (`pipelines.executions`) in order for the navbar to link to the correct place. 

`isActive` is determined by `dataSource.activeState`. This is more broad and includes this permalink case. We cannot, however, use `dataSource.activeState` to generate the links in for the nav items.

christopherthielen:
> Example: For executions, the sref is pipelines.executions, but when you permalink to the details the sref has pipelines.executionDetails.execution. This is considered not active by the useSrefActive effect. However, we have to generate the sref with the dataSource (pipelines.executions) in order for the navbar to link to the correct place.

ah, ok got it, thanks for the explanation.  Then I approve, but please change from `uiSrefActive` to plain old `uiSref`

PR #8470, opened 2020-08-05
Testing ui multi select functionality

vigneshm:
Testing some deck customizations to `ui-select` component. This is not intended to be reviewed/committed.

### DIFF
--- a/app/scripts/modules/core/src/bootstrap/uiSelect.config.ts
+++ b/app/scripts/modules/core/src/bootstrap/uiSelect.config.ts
@@ -18,6 +18,8 @@ function uiSelectDecorator($provide: ng.auto.IProvideService) {
           if ($select.selected.length >= $select.limit) {
             return;
           }
+          // eslint-disable-next-line no-console
+          console.log('Is defaultPrevented? ', event.defaultPrevented ? 'yes' : 'no');
           if (!event.defaultPrevented) {
             $select.selected.push(item);
             scope.$selectMultiple.updateModel();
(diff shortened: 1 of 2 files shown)


maggieneterval:
Hi @vigneshm, I'm so sorry for the slow followup! Finally getting around to running your branch now and checking out those debug statements.

maggieneterval:
I'm getting a bunch of: `isDefaultPrevented? no`:

Screen Shot 2020-08-13 at 4 55 11 PM

And here's what the UI looks like. To be honest, I'm not sure if it's intended behavior that selecting one firewall seems to result in several being selected?
fe4d6464-a06e-4e62-9896-52ec3c91a9ef

I don't have a ton of context on what problem the `ui-select` customization was trying to solve, but perhaps @danielpeach does since it looks like he added it (many years ago!).

vigneshm:
Thanks a lot @maggieneterval, that confirms my questions related to this customization. I don't think deck's [specific customization](https://github.com/spinnaker/deck/blob/0e1d9ca3d2cc2c0a9a1583b05d748290bcae46f7/app/scripts/modules/core/src/bootstrap/uiSelect.config.ts#L6-L32) that is trying to prevent the _default behavior_ of `uiSelectMulti` does anything. 

Here's why I think this is the case.

1. The `uiSelectController` from `angular-ui/ui-select` library, [`broadcasts`](https://github.com/angular-ui/ui-select/blob/master/src/uiSelectController.js#L445) the `uis:select` event instead of `emitting` it. So only child scopes of `ui-select` directive would be able to listen on those events.
2. In deck, there are a [couple](https://github.com/spinnaker/deck/blob/5b0f15fd9cb7f17fa856eb488e0383192e6439e5/app/scripts/modules/google/src/serverGroup/configure/wizard/securityGroups/tagSelector.component.js#L25-L27) [of](https://github.com/spinnaker/deck/blob/5b0f15fd9cb7f17fa856eb488e0383192e6439e5/app/scripts/modules/google/src/serverGroup/configure/wizard/loadBalancers/elSevenOptions/backendServiceSelector.component.js#L27-L29) places where we try to listen to `uis:select` event and prevent the default behavior. However, these do not receive the event since their scope isn't a child of the `ui-select` directive scope.
3. Given the above, I don't think this [condition check](https://github.com/spinnaker/deck/blob/c956d541429f107a4d5069a7b63e0b0f26dd3590/app/scripts/modules/core/src/bootstrap/uiSelect.config.ts#L22) is going to work and as a result this whole customization in deck seems unnecessary.

Since these customizations are really old, I think something must have changed in the `angular-ui/ui-select` library that must have resulted in this behavior.

I'm leaning towards removing our customization, but I would like to hear from @danielpeach if I'm overlooking something here.

PR #8473, opened 2020-08-05
chore(core): Remove outdated nav components

caseyhebebrand:
After the release of the vertical nav, there are several components that are no longer used in deck. This PR removes those components, which include the old navbar and its subcomponents. 

Please let me know if you feel there is benefit to keeping any of these components.


gcomstock:
Thank you for leaving no trace!

christopherthielen:
What's your confidence level that there is no additional CSS to delete?

caseyhebebrand:
> What's your confidence level that there is no additional CSS to delete?

There is more css. I want to put it in a separate PR because there is some overlap with the projects pages. In certain areas the Projects page classes build on top of classes, so I wanted to do a thorough job of moving/deleting styles.

PR #8474, opened 2020-08-06
fix(core/presentation): Do not render the form input even once ...

christopherthielen:
... (render the spel input instead) if spel is present and freeform is enabled

This fixes a problem encountered where a SelectInput applied a default value using an `useEffect`.  The form value had spel in it and should have been in freeform input mode.  However, because the SelectInput was rendered once, its useEffect kicked in and applied the default value overriding the spel value.

```js
<FormikFormField name="env" input={props => {
  return <SelectInput defaultValue="test" options={["test", "prod"]} {...props} />
}} />
```

Screen Shot 2020-08-05 at 8 07 49 PM

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
+++ b/app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
@@ -1,16 +1,15 @@
-import React from 'react';
-import { toPath, isString } from 'lodash';
-import { FastField, Field, FieldProps, FormikConsumer, FormikContext } from 'formik';
-
 import { firstDefined } from 'core/utils';
+import { FastField, Field, FieldProps, FormikConsumer, FormikContext } from 'formik';
+import { isString, toPath } from 'lodash';
+import React from 'react';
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/presentation/forms/fields/FormikFormField.tsx
```diff
@@ -104,21 +103,17 @@ function FormikFormFieldImpl<T = any>(props: IFormikFormFieldImplProps<T>) {
...
-  const [inputMode, setInputMode] = React.useState(SpelAwareInputMode.DEFAULT);
+  const freeformInputAllowed = firstDefined(props.spelAware, SpelAwareFromContext, false);
 
-  const freeformInputEnabled = firstDefined(props.spelAware, SpelAwareFromContext, false);
-
-  React.useEffect(() => {
-    if (!freeformInputEnabled) {
-      return;
-    }
+  const initialInputMode = React.useMemo(() => {
```
christopherthielen, line 108: compute `initialInputMode` early and use it in the first render

PR #8475, opened 2020-08-06
feat(helm): Add Helm Chart pipeline trigger

jcavanagh:
- Pipeline trigger configuration
- Execution status UI
- Manual trigger UI

Relates to issue: https://github.com/spinnaker/spinnaker/issues/4447 

Related PRs:
https://github.com/spinnaker/echo/pull/988
https://github.com/spinnaker/igor/pull/829
https://github.com/spinnaker/clouddriver/pull/4773

image

image

image


dbyron0:
Let me first say that I'm so excited for this and the other related PRs!  I'm curious what the implications are for https://github.com/spinnaker/deck/pull/7312 though.

jcavanagh:
@dbyron0 Interesting!  I had not seen that particular feature - thanks for pointing it out.

If I'm understanding that PR correctly, I would say these are certainly related, but still distinct use cases.  

The PR you mention is intended to provide an expected Helm artifact in a manual execution, regardless of how the Helm chart is normally delivered.  

This PR does have the effect of providing a Helm artifact, but only via a new Helm trigger, and only when a Helm trigger is configured.

From the #7312 PR message, it appears under the following (pretty narrow) conditions:
```
- The pipeline has exactly one Bake (Manifest) stage
- The pipeline has exactly one expected artifact of type helm/chart
```

A possible solution would be add another condition to activating the functionality from #7312 - only show that if there is also no Helm trigger configured.  How does that sound to you?

dbyron0:
If I'm reading this PR correctly, it's possible to specify the helm chart for pipelines with helm triggers ... and so then yes, it makes sense to disable the https://github.com/spinnaker/deck/pull/7312 functionality.  In the end, it seems like it'll still be possible to specify a helm chart whether there's a trigger or not, which would be lovely.

german-muzquiz:
I'm not that familiar with deck code, perhaps @maggieneterval can you take a look?

app/scripts/modules/core/src/pipeline/config/triggers/helm/HelmTriggerExecutionStatus.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { IExecutionTriggerStatusComponentProps } from 'core/domain';
+import { IHelmTrigger } from 'core/domain/IHelmTrigger';
+
+export const HelmTriggerExecutionStatus = (props: IExecutionTriggerStatusComponentProps) => {
+  const trigger = props.trigger as IHelmTrigger;
+  const name = trigger.artifactName;
+  const version = trigger.version;
+
+  return (
+    <li>
```
maggieneterval, line 12: Curious why this is a list element?

app/scripts/modules/core/src/pipeline/config/triggers/helm/HelmTriggerTemplate.tsx
```diff
@@ -0,0 +1,90 @@
...
+          return a.types.includes('helm/chart');
+        })
+        .map(it => it.name);
+
+      setAccounts(accounts);
+    });
+  }, []);
+
+  const stateChanged = (setter: Function, field: string, value: string) => {
+    setter(value);
```
maggieneterval, line 41: Do we need to store `account`, `chart`, and `version` on state, or could we read them from `props.trigger`?

maggieneterval (approved):
@jcavanagh Looks great; thanks for the awesome feature! 💯 Added a couple of questions inline, but nothing blocking.

I'd be in favor of removing the Bake (Manifest) stage's current [manual execution component](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/ManualExecutionBakeManifest.tsx) to avoid confusion over two different Helm manual execution components, but defer to your judgment.

maggieneterval:
@jcavanagh I'd also recommend making a PR against [spinnaker.github.io](https://github.com/spinnaker/spinnaker.github.io) to add a note to the [curated changelog](https://spinnaker.io/community/releases/next-release-preview/) for next release so that users know that this feature has landed! 🥳

app/scripts/modules/core/src/pipeline/config/triggers/helm/HelmTriggerExecutionStatus.tsx
jcavanagh: The parent `ExecutionStatus` component renders the trigger-specific data under an `ul` element, and each `li` is a row in that section of the execution UI.

app/scripts/modules/core/src/pipeline/config/triggers/helm/HelmTriggerTemplate.tsx
jcavanagh: These fields are editable, and are state for that reason.

app/scripts/modules/core/src/pipeline/config/triggers/helm/HelmTriggerExecutionStatus.tsx
maggieneterval: I see, thanks! 👍

app/scripts/modules/core/src/pipeline/config/triggers/helm/HelmTriggerTemplate.tsx
maggieneterval: Totally understood that these fields are editable; was more wondering if we need to duplicate the source of truth for their current value on both this component's state and whatever wrapper's state we're updating with `onHelmChanged` -- when we call `onHelmChanged` is that not updating the object that we pass back down to this component as `props.trigger`?

dbyron0:
> I'd be in favor of removing the Bake (Manifest) stage's current [manual execution component](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/stages/bakeManifest/ManualExecutionBakeManifest.tsx) to avoid confusion over two different Helm manual execution components, but defer to your judgment.

I can see there being confusion in the code, but not so much for users.  If they have a trigger, they get one way of manual execution, and if they don't they get the other.

It works for me to remove it, but can we wait to do it in a subsequent version of spinnaker?  That way when I upgrade I don't have broken pipelines/user experience until I have a chance to change to this new way.

app/scripts/modules/core/src/pipeline/config/triggers/helm/HelmTriggerTemplate.tsx
jcavanagh: Ah, I see what you mean.  I think the internal component state here is still the "real" source of truth - the `onHelmChanged` just wraps the differing details between which pipeline config fields those pieces of state map back to (it's different for manual UI and for the pipeline trigger UI), and ends up just sort of proxying the data upward.

jcavanagh: Ah, I see what you mean.  The actually-stored fields those pieces of state map back to are different in the manual UI and the pipeline config UI.  In order to update the parent directly, it would need to know each of those actually-stored paths down in the shared component.  The current implementations of `onHelmChanged` abstract the actually-stored paths to a level above, and are most just a thin proxy back upward.

jcavanagh:
@dbyron0 @maggieneterval I've added the aforementioned additional conditional for the old manual Helm UI [here](https://github.com/spinnaker/deck/pull/8475/commits/5ad8d68251e923e72db378276109bcf3d0569e56) - they should hopefully not step on each other now, and the old use case can be preserved.

app/scripts/modules/core/src/pipeline/config/triggers/helm/HelmTriggerTemplate.tsx
maggieneterval: A, got it -- thanks for the explanation! 😄

maggieneterval:
> @dbyron0 @maggieneterval I've added the aforementioned additional conditional for the old manual Helm UI here - they should hopefully not step on each other now, and the old use case can be preserved.

Thanks, LGTM! 👍

dbyron0:
Thank you!

PR #8485, opened 2020-08-11
fix(core/managed): supply references when updating constraints

erikmunson:
Related to https://github.com/spinnaker/keel/pull/1417 — this change always includes a reference when updating the status of a constraint so updates can be uniquely mapped to an artifact version. Once this is live we can make the reference field non-nullable on the API.

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedWriter.ts
+++ b/app/scripts/modules/core/src/managed/ManagedWriter.ts
@@ -21,6 +21,7 @@ export interface IUpdateConstraintStatusRequest {
   application: string;
   environment: string;
   type: string;
+  reference: string;
   version: string;
   status: StatefulConstraintStatus;
 }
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/managed/ManagedWriter.ts
```diff
@@ -88,6 +90,7 @@ export class ManagedWriter {
       .one('constraint')
       .post({
         type,
+        artifactReference: reference,
         artifactVersion: version,
```
erikmunson, line 94: it'd be nice if we could rename these fields to `reference` and `version` to match all the other endpoints in keel — it's easy to spot that this particular feature was implemented by a different human at a different time from all the others.

emjburns: :D totally

PR #8487, opened 2020-08-11
fix(kubernetes): do not self-close element in angular template

maggieneterval:
Closes: https://github.com/spinnaker/spinnaker/issues/5962

Oddly, the `kubernetes-rolling-restart` component was rendering as expected, but each of the remaining elements in the list of actions (Edit and Delete) were not. Self-closing tags are not allowed in Angular templates, so I likely broke this when I added support for rolling restarts.


maggieneterval:
I looked into this a bit more deeply and now suspect that the [jquery update from 3.4 to 3.5](https://github.com/spinnaker/deck/pull/8222) broke custom self-closing tags (see description of security vulnerability fix in the jquery [upgrade guide](https://jquery.com/upgrade-guide/3.5/)). Will submit a separate PR after auditing the other `react2Angular` elements.

PR #8488, opened 2020-08-11
fix(core/CustomDropdown): Add support to dismiss custom dropdown menu when clicked outside the menu

vigneshm:
Custom dropdowns need to explicitly handle 'dismissal on click' scenario as mentioned [here](https://react-bootstrap-v3.netlify.app/components/dropdowns/#btn-dropdowns-custom).


vigneshm:
The core issue is in `CustomMenu` which is only used in one instance. I'm going to remove that occurrence and remove `CustomMenu` from `@spinnaker/core`

PR #8492, opened 2020-08-13
fix(RunJobStage): Make RunJobStage restartable

AbdulRahmanAlHamali:
This is supposed to fix this issue:
https://github.com/spinnaker/spinnaker/issues/5461

Is there a reason why the RunJobStage doesn't have the `restartable` property on it?

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/runJob/runJobStage.js
+++ b/app/scripts/modules/core/src/pipeline/config/stages/runJob/runJobStage.js
@@ -12,5 +12,6 @@ module(CORE_PIPELINE_CONFIG_STAGES_RUNJOB_RUNJOBSTAGE, []).config(function() {
     key: 'runJob',
     label: 'Run Job',
     description: 'Runs a container',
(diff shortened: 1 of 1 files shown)


AbdulRahmanAlHamali:
I confirmed that this is just a UI issue, because I was able to restart it using the API.

Looking at Deck's code, I find a lot of things that should probably be read from the API instead of hard-coded here. There seems to be multiple sources of truth

maggieneterval:
Thanks for the PR, @AbdulRahmanAlHamali! Looping in @ethanfrogers in case there is some reason I'm not recalling that at least the Kubernetes Run Job stage should not be restarted.

christopherthielen:
👍  from me

AbdulRahmanAlHamali:
I get the impression that @ethanfrogers is on vacation. So I don't know if we'll be able to make the release. 

But I've called `restart` using the API a few times, and it worked with no issues. It feels that we just forgot to add the button in the UI

PR #8495, opened 2020-08-13
feat(google): add predictive autoscaling support

maggieneterval:
## Screenshots

<details><summary>Enabling predictive autoscaling</summary>
<p>

71e52cce-4714-45c8-a558-48a68fe51985

</p>
</details>

<details><summary>Server group details for enabled autoscaling</summary>
<p>

0dOMgZnooQV

</p>
</details>

## Enabling

Requires:

- https://github.com/spinnaker/clouddriver/pull/4794 (Will be released with Spinnaker 1.22)
- An `alphaListed` GCE account
- The following added to `settings-local.js`:

```
window.spinnakerSettings.providers.gce.feature.predictiveAutoscaling = true;
```

## Commits

* feat(google): add feature flag for predictive autoscaling 

* refactor(google): move autoscaling policy interfaces and layout components to autoscalingPolicy directory 

  Currently, the only Typescript/React components of the autoscaling policy modal are specific to the scale-in control feature. In the next commit, we will add support for predictive autoscaling, so in order to share these interfaces and layout components, lift them to the top level.

* feat(google): add predictive autoscaling support (add/edit policy modal) 

* feat(google): add predictive autoscaling support (SG details view) 

* fix(google): sp/scaleDownInConfigured/scaleInControlsConfigured 

  I think we missed one spot when we renamed scale-in to scale-down.

### DIFF
--- a/app/scripts/modules/google/src/serverGroup/details/autoscalingPolicy/autoscalingPolicy.directive.html
+++ b/app/scripts/modules/google/src/serverGroup/details/autoscalingPolicy/autoscalingPolicy.directive.html
@@ -6,6 +6,12 @@
   <dd ng-repeat-end>
     <div ng-repeat="target in basis.targets">{{target}}</div>
   </dd>
+  <dt ng-if="$ctrl.predictiveAutoscalingEnabled">
+    Predictive Autoscaling
+  </dt>
+  <dd ng-if="$ctrl.predictiveAutoscalingEnabled">
+    {{$ctrl.policy.cpuUtilization.predictiveMethod}}
+  </dd>
   <dt>
     Min # VMs
     <help-field key="gce.serverGroup.autoscaling.minVMs"></help-field>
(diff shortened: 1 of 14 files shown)


app/scripts/modules/google/src/serverGroup/details/autoscalingPolicy/autoscalingPolicy.directive.html
```diff
@@ -43,7 +43,7 @@
   <dt ng-if="$ctrl.scaleInControlsConfigured">
     Scale-in Time Window
   </dt>
-  <dd ng-if="$ctrl.scaleDownInConfigured">{{$ctrl.timeWindowSecMessage}}</dd>
+  <dd ng-if="$ctrl.scaleInControlsConfigured">{{$ctrl.timeWindowSecMessage}}</dd>
```
plumpy, line 46: whoops, good catch!

PR #8501, opened 2020-08-17
feature(UI): Add support for displaying colors in console outputs

AbdulRahmanAlHamali:
This should allow displaying colours when the console logs contain ANSI colour escape characters.

I will test this on my local installation and will update you if it works

### DIFF
--- a/app/scripts/modules/core/src/manifest/stage/JobManifestPodLogs.tsx
+++ b/app/scripts/modules/core/src/manifest/stage/JobManifestPodLogs.tsx
@@ -5,6 +5,8 @@ import { bindAll } from 'lodash';
 
 import { InstanceReader, IInstanceConsoleOutput, IInstanceMultiOutputLog } from 'core/instance/InstanceReader';
 import { IPodNameProvider } from '../PodNameProvider';
+import DOMPurify from 'dompurify';
+import AnsiUp from 'ansi_up';
 
 // IJobManifestPodLogs is the data needed to get logs
 export interface IJobManifestPodLogsProps {
@@ -61,9 +63,14 @@ export class JobManifestPodLogs extends React.Component<IJobManifestPodLogsProps
     const region = this.resourceRegion();
     InstanceReader.getConsoleOutput(account, region, this.podName(), 'kubernetes')
       .then((response: IInstanceConsoleOutput) => {
+        const containerLogs = response.output as IInstanceMultiOutputLog[];
+        containerLogs.forEach((log: IInstanceMultiOutputLog) => {
+          log.formattedOutput = DOMPurify.sanitize(new AnsiUp().ansi_to_html(log.output));
+        });
+
         this.setState({
-          containerLogs: response.output as IInstanceMultiOutputLog[],
-          selectedContainerLog: response.output[0] as IInstanceMultiOutputLog,
+          containerLogs: containerLogs,
+          selectedContainerLog: containerLogs[0],
         });
         this.open();
       })
@@ -105,7 +112,10 @@ export class JobManifestPodLogs extends React.Component<IJobManifestPodLogsProps
                       </li>
                     ))}
                   </ul>
-                  <pre className="body-small flex-fill">{selectedContainerLog.output}</pre>
+                  <pre
+                    className="body-small fill"
+                    dangerouslySetInnerHTML={{ __html: selectedContainerLog.formattedOutput }}
+                  ></pre>
                 </>
               )}
               {errorMessage && <pre className="body-small">{errorMessage}</pre>}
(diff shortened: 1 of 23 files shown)


AbdulRahmanAlHamali:
image
Tested, works great!

app/scripts/modules/core/src/manifest/stage/JobManifestPodLogs.tsx
```diff
@@ -61,9 +63,14 @@ export class JobManifestPodLogs extends React.Component<IJobManifestPodLogsProps
     const region = this.resourceRegion();
     InstanceReader.getConsoleOutput(account, region, this.podName(), 'kubernetes')
       .then((response: IInstanceConsoleOutput) => {
+        const containerLogs = response.output as IInstanceMultiOutputLog[];
+        containerLogs.forEach((log: IInstanceMultiOutputLog) => {
+          log.formattedOutput = DOMPurify.sanitize(new AnsiUp().ansi_to_html(log.output));
```
maggieneterval, line 68: do we need to create a new `ansiUp` object on every click?

app/scripts/modules/core/src/manifest/stage/JobManifestPodLogs.tsx
```diff
@@ -5,6 +5,8 @@ import { bindAll } from 'lodash';
 
 import { InstanceReader, IInstanceConsoleOutput, IInstanceMultiOutputLog } from 'core/instance/InstanceReader';
 import { IPodNameProvider } from '../PodNameProvider';
+import DOMPurify from 'dompurify';
```
maggieneterval, line 8: can we move these imports up under the other external package imports?

app/scripts/modules/core/src/modal/modals.less
```diff
@@ -46,6 +46,9 @@
     flex-direction: column;
     flex: 1;
   }
+  .fill {
```
maggieneterval, line 49: why do we need this?

package.json
```diff
@@ -44,6 +44,7 @@
     "angular-ui-bootstrap": "^2.5.0",
     "angular-ui-sortable": "^0.17.0",
     "angular2react": "^3.0.2",
+    "ansi_up": "^4.0.4",
```
maggieneterval, line 47: I don't have a strong opinion, but just curious why you chose this library over the [seemingly more popular alternative](https://www.npmjs.com/package/ansi-to-html) (1.6 million weekly downloads vs. 18k)

app/scripts/modules/core/src/manifest/stage/JobManifestPodLogs.tsx
AbdulRahmanAlHamali: nope, we don't. Will fix

package.json
AbdulRahmanAlHamali: No reason actually, weird that this library didn't show up when I searched. I'll try it out

app/scripts/modules/core/src/modal/modals.less
AbdulRahmanAlHamali: I needed to remove `fill-flex` from the `pre`, because the ansi_to_html adds spans, and the flex was sorting all the spans in a column, effectively ruining the formatting.
But I needed to keep `flex: 1` because I wanted to fill the parent, so I had to make this class

package.json
AbdulRahmanAlHamali: I think I'll stay with ansi_up:
* It has double the number of stars on github (which I think is a better popularity measures than number of downloads).
* The other library does not have type support. In fact, type support was introduced 4 months ago, but the author forgot to upload the new version to npm. Not a great sign :sweat_smile:

AbdulRahmanAlHamali:
I addressed the comments :) Any other notes?

app/scripts/modules/core/src/modal/modals.less
maggieneterval: Thanks for the explanation! I'm a little afraid of having this style that only applies to one type of modal in the general modal stylesheet -- feels like it could easily get removed/changed without someone understanding that this is necessary only when we're formatting the modal contents with a particular library. Perhaps we could either add a comment with this explanation, or add a`JobManifestPodLogs`-specific stylesheet.

package.json
maggieneterval: sounds good, thanks for checking it out!

app/scripts/modules/core/src/modal/modals.less
AbdulRahmanAlHamali: Hmmm, what we really want here is for the element to fill the parent withoht being a flex container itself. Maybe we can call it "fill-no-flex"? Or I can put it as an inline style if that's allowed. Not sure if it's worth creating a different file just for this

AbdulRahmanAlHamali: I'll go with `fill-no-flex`. It is general enough. I can imagine wanting an element to fill its parent, even if the element itself is not a flex box.

maggieneterval: sure!

PR #8502, opened 2020-08-17
fix(ecs): update z-index for uibModal

pkandasamy91:
Fixes: https://github.com/spinnaker/spinnaker/issues/5971

modal-z-index-fix

### DIFF
--- a/app/scripts/modules/core/src/cluster/allClusters.controller.js
+++ b/app/scripts/modules/core/src/cluster/allClusters.controller.js
@@ -123,6 +123,7 @@ module(CORE_CLUSTER_ALLCLUSTERS_CONTROLLER, [
               $uibModal.open({
                 templateUrl: providerConfig.cloneServerGroupTemplateUrl,
                 controller: `${providerConfig.cloneServerGroupController} as ctrl`,
+                windowClass: 'modal-z-index-fix',
                 size: 'lg',
                 resolve: {
                   title: () => title,
--- a/app/scripts/modules/core/src/cluster/allClusters.controller.js
+++ b/app/scripts/modules/core/src/cluster/allClusters.controller.js
@@ -123,7 +123,7 @@ module(CORE_CLUSTER_ALLCLUSTERS_CONTROLLER, [
               $uibModal.open({
                 templateUrl: providerConfig.cloneServerGroupTemplateUrl,
                 controller: `${providerConfig.cloneServerGroupController} as ctrl`,
-                windowClass: 'modal-z-index-fix',
+                windowClass: 'modal-z-index',
                 size: 'lg',
                 resolve: {
                   title: () => title,
--- a/app/scripts/modules/core/src/pipeline/config/stages/deploy/deployStage.js
+++ b/app/scripts/modules/core/src/pipeline/config/stages/deploy/deployStage.js
@@ -149,6 +149,7 @@ angular
                   templateUrl: config.cloneServerGroupTemplateUrl,
                   controller: `${config.cloneServerGroupController} as ctrl`,
                   size: 'lg',
+                  windowClass: 'modal-z-index-fix',
                   resolve: {
                     title: () => title,
                     application: () => application,
@@ -186,6 +187,7 @@ angular
                 templateUrl: providerConfig.serverGroup.cloneServerGroupTemplateUrl,
                 controller: `${providerConfig.serverGroup.cloneServerGroupController} as ctrl`,
                 size: 'lg',
+                windowClass: 'modal-z-index-fix',
                 resolve: {
                   title: () => title,
                   application: () => application,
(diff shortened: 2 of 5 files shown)


pkandasamy91:
Ready for review: @maggieneterval @caseyhebebrand @allisaurus

app/scripts/modules/core/src/pipeline/config/stages/deploy/deployStage.js
```diff
@@ -149,6 +149,7 @@ angular
                   templateUrl: config.cloneServerGroupTemplateUrl,
                   controller: `${config.cloneServerGroupController} as ctrl`,
                   size: 'lg',
+                  windowClass: 'modal-z-index-fix',
```
caseyhebebrand, line 152: Does this overwrite the other classes or is it supplementary? 

If this does not overwrite, then I would suggest using our existing styleguide classname `layer-critical` instead of this name. Then you will not need the update to `main.less`

app/scripts/modules/core/src/presentation/main.less
```diff
@@ -641,6 +641,11 @@ body > .container {
...
 }
 
+/* This class is used to override the uibModal default z-index value of 1050 */
+.modal-z-index-fix {
+  z-index: var(--layer-critical) !important;
+}
+
 .modal-pages {
   width: 1200px;
   .modal-page {
```
caseyhebebrand, line 651: Did you try adding `z-index: var(--layer-critical)`  under `modal-page`? I think that may do the trick, then you can remove the rest of the code.

app/scripts/modules/core/src/cluster/allClusters.controller.js
```diff
@@ -123,6 +123,7 @@ module(CORE_CLUSTER_ALLCLUSTERS_CONTROLLER, [
               $uibModal.open({
                 templateUrl: providerConfig.cloneServerGroupTemplateUrl,
                 controller: `${providerConfig.cloneServerGroupController} as ctrl`,
+                windowClass: 'modal-z-index-fix',
```
caseyhebebrand, line 126:
```suggestion
                windowClass: 'modal-z-index-fix',
```
```suggestion
                windowClass: 'layout-critical',
```
```suggestionw
                windowClass: 'modal-z-index-fix',
```

caseyhebebrand, line 126:
```suggestion
                windowClass: 'layout-critical',
```

app/scripts/modules/core/src/pipeline/config/stages/deploy/deployStage.js
pkandasamy91: I think it adds it to the modal window template. When I use the `layer-critical` class the issue continues to exist.

pkandasamy91: Did you mean `layout-critical`?

caseyhebebrand: No `layer-critical`

caseyhebebrand: if that doesn't work, can you just update the class to `modal-z-index` and remove the `fix`?

app/scripts/modules/core/src/cluster/allClusters.controller.js
caseyhebebrand, line 126:
```suggestion
                windowClass: 'layer-critical',
```
```suggestion
                windowClass: 'modal-z-index',
```

pkandasamy91: From Angular's bootstrapping directives, windowClass (Type: string) - Additional CSS class(es) to be added to a modal window template. 

Based on the above description, it seems like it should be supplemental, but it doesn't work when I switch over to using the layout-critical class. I've updated the PR to rename the class to modal-z-index.

app/scripts/modules/core/src/presentation/main.less
pkandasamy91: Yup, tried adding that, but it was still being overridden by the z-index defined in uibModal from Angular's bootstrap UI.

app/scripts/modules/core/src/pipeline/config/stages/deploy/deployStage.js
pkandasamy91: Updated

app/scripts/modules/core/src/cluster/allClusters.controller.js
maggieneterval:
```suggestion
                windowClass: 'modal-z-index',
```
This should now just be `modal-z-index`, right? Otherwise LGTM, thanks for the quick fix!

pkandasamy91: Good catch! Not sure how this file didn't commit with the update everywhere else, but everything should be updated now!

maggieneterval: Great, thanks! Going to merge this now.

PR #8507, opened 2020-08-18
fix(gitRepoArtifact): move subpath value to location field.

edgarulg:
Currently, there are some issues with the `git/repo` artifact because we store `subpath` as metadata, for users that define artifacts with the same URL and branch but different subpath, Orca is not able to match artifacts correctly.

This PR is about move the subpath value of a `git/repo` artifact to be saved in the `location` field.

It will look at `location` field first, then if it is missing will look at `metadata.subPath`

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/gitrepo/GitRepoArtifactEditor.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/gitrepo/GitRepoArtifactEditor.tsx
@@ -1,4 +1,4 @@
-import { cloneDeep, get } from 'lodash';
+import { cloneDeep } from 'lodash';
 import React from 'react';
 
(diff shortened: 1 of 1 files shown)


edgarulg:
Hi @maggieneterval, thanks for the feedback

Clouddriver is the service responsible of use the artifact information to downloader it. I opened this PR: https://github.com/spinnaker/clouddriver/pull/4815 only for `git/repo` artifacts, it will look at the location field for the subPath and fallback to metadata when location is empty.

Additionally, I will open a PR with a Front50 migration that will move the metadata.subPath value to the location field on expected artifacts.

I am going to replace the lodash `get` with Typescript

maggieneterval:
Sorry, I think CI failed because I added `ready to merge` without actually approving!

maggieneterval:
Actually, @edgarulg you might need to edit this PR to allow edits from maintainers so that Mergify can merge master into your branch

PR #8508, opened 2020-08-19
fix(RunJobStage): Add restartability to Kubernetes RunJobStage

AbdulRahmanAlHamali:
I already submitted this to fix the problem:

https://github.com/spinnaker/deck/pull/8492

But it didn't

It seems there are multiple files for `runJobStage`, but the one I change in this PR is the correct one. I have tested it and it works perfectly.

Apologies for the confusion

@maggieneterval FYI


AbdulRahmanAlHamali:
Thank you for merging and for sharing the new criteria. I didn't know about that

PR #8513, opened 2020-08-21
feat(core/managed): add resource kind registry as prep for plugins

erikmunson:
We're getting very close to extending our supported resource kinds via plugins, and as prep for that we need to do some UI plumbing. The first step of that is taking hard-coded configuration for the currently supported resource kinds and moving it all behind a registry contract that can later be used for other kinds outside of the core codebase. A couple things still stand in our way:

- We need a way to show a display name when a resource doesn't conform to the moniker contract (@luispollo will be handling this on the API side)
- We need to wire this registry contract up to the externally facing SDK provided to UI plugins, so that it's possible to add kinds outside of the core codebase itself

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
+++ b/app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
@@ -2,12 +2,13 @@ import React, { memo } from 'react';
 import { useSref } from '@uirouter/react';
 
 import { Application } from 'core/application';
-import { Icon, IconNames } from '../presentation';
+import { Icon } from '../presentation';
 import { IManagedResourceSummary, IManagedEnviromentSummary, IManagedArtifactSummary } from '../domain/IManagedEntity';
 
 import { getKindName } from './ManagedReader';
 import { ObjectRow } from './ObjectRow';
 import { AnimatingPill, Pill } from './Pill';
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/plugins/deck.plugin.ts
```diff
@@ -21,6 +24,7 @@ export interface IDeckPlugin {
 export function registerPluginExtensions(plugin: IDeckPlugin): PromiseLike<any> {
   plugin.stages?.forEach(stage => Registry.pipeline.registerStage(stage));
   plugin.preconfiguredJobStages?.forEach(stage => Registry.pipeline.registerPreconfiguredJobStage(stage));
+  plugin.resourceKinds?.forEach(kind => registerResourceKind(kind));
```
erikmunson, line 27: @vigneshm I _kinda_ think I _maybe_ figured out how to expose this to plugins via the `pluginsdk` contract? Maybe? LMK if you think this will work or if I went down my own special path 😆

vigneshm: Yes, this is the way to open it up!

erikmunson: woohoo! thanks vignesh :D

PR #8516, opened 2020-08-24
fix(appengine): fix rendering of StageArtifactSelector when artifacts are created

zachsmith1:
This is a fix for GAE deploy stage. The stage initially wouldn't correctly show the artifacts that have been added and it also wouldn't allow a user to edit them. This fixes the dropdown artifacts to correctly display and allow a user to modify them. Note: Code in this module will need to be further revisited because there are better ways of doing it (i.e. the way Helm Bake Manifest Stage does). But for now, this solves the issue at hand. I've also added two tests to verify correctness


app/scripts/modules/appengine/src/serverGroup/configure/wizard/configFileArtifactList.spec.tsx
```diff
@@ -0,0 +1,66 @@
+import { ConfigFileArtifactList } from './ConfigFileArtifactList';
+import { mount } from 'enzyme';
+import { API, IArtifactAccount, IArtifactAccountPair, StageArtifactSelector } from '@spinnaker/core';
+import React from 'react';
```
danielpeach, line 4: Usually Deck imports get sorted such that 3rd party dependencies (`react`, `enzyme`) go above internal dependencies (`./ConfigFileArtifactList`) with a space between them.

e.g., https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/stages/deployService/DeployServiceExecutionDetails.tsx#L1-L5

app/scripts/modules/appengine/src/serverGroup/configure/wizard/configFiles.component.ts
```diff
@@ -23,11 +24,13 @@ class ConfigArtifact implements IArtifactAccountPair {
...
   public id: string;
   public account: string;
+  public artifact?: IArtifact;
 
-  constructor($scope: IScope, artifact = { id: '', account: '' }) {
+  constructor($scope: IScope, pair: IArtifactAccountPair = { id: '', account: '' }) {
     const unserializable = { configurable: false, enumerable: false, writable: false };
-    this.id = artifact.id;
-    this.account = artifact.account;
+    this.id = null;
```
kevinawoo, line 31: Why are we dropping the id instead of setting it to `pair.id`?

kevinawoo: Oh... when the json payload gets to orca if the artifact resolver see’s that it has an ID set it will search for it in the “resolvedArtifacts” section, which doesn’t exist. It’s a different pattern that wasn’t adhered to by the stage. Setting the id to null guarantees that it’s resolved directly on the artifact

app/scripts/modules/appengine/src/serverGroup/configure/wizard/configFileArtifactList.spec.tsx
zachsmith1: I did this manually because all my deck webstorm settings are messed up :/

zachsmith1:
@kevinawoo @danielpeach I realized that while this PR worked for inline default artifacts, the ui wouldn't populate correctly for expectedArtifacts from previous stages, etc. It turns out that I was missing a prop in StageArtifactSelector for expectedArtifactId. I now explicitly to a check to see if artifact == null, and only then do I set the id for this prop. This ensures that it will display the correct label for default vs expected artifacts. As far as the id == null in configFiles.component.ts, I now changed this to pair?.id. This ensures that it's only set if the pair already has an id. The only time pair will have an id is if it's an expected artifact. This works as expected and will make sure that both artifacts get properly resolved when they get to orca.

app/scripts/modules/appengine/src/serverGroup/configure/wizard/configFileArtifactList.spec.tsx
dogonthehorizon: @zachsmith1 you could also run `prettier` yourself on the files. There's the `yarn prettier` command that will format the whole project, or you can borrow the script command and run it yourself on a subset:
`bash -c '$(npm bin)/prettier --write \'app/scripts/modules/kubernetes/src/serverGroupManager/*.{js,jsx,ts,tsx,html,css,less}\''`

☝️ just change the path to whatever and make sure you run the command from the root of the project :)

zachsmith1: @dogonthehorizon will def give that a try

PR #8519, opened 2020-08-25
feat(core/managed): use displayName for resources in place of moniker

erikmunson:
Following up #8513, this change switches the way we deal with resource objects to use the [newly added `displayName` field](https://github.com/spinnaker/keel/pull/1456) instead of assuming a resource will have a `moniker` object. That's important for resource plugins/extensions as it doesn't always make sense for a resource to conform to the moniker contract.

I refactored most of this via whack-a-mole-ing the TS compiler errors after making `moniker` optional, so it's possible there's a spot I missed. We'll find out.

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -20,12 +20,12 @@ import { showMarkArtifactAsBadModal } from './MarkArtifactAsBadModal';
 
 import { ConstraintCard } from './constraints/ConstraintCard';
 import { isConstraintSupported } from './constraints/constraintRegistry';
+import { isResourceKindSupported } from './resources/resourceRegistry';
 
 import './ArtifactDetail.less';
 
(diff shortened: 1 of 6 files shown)


app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -20,12 +20,12 @@ import { showMarkArtifactAsBadModal } from './MarkArtifactAsBadModal';
 
 import { ConstraintCard } from './constraints/ConstraintCard';
 import { isConstraintSupported } from './constraints/constraintRegistry';
+import { isResourceKindSupported } from './resources/resourceRegistry';
 
 import './ArtifactDetail.less';
 
 function shouldDisplayResource(reference: string, resource: IManagedResourceSummary) {
-  //TODO: naively filter on presence of moniker but how should we really decide what to display?
```
luispollo, line 27: Haha, love these kinds of comments! 😆

PR #8521, opened 2020-08-26
feat(core/notifier): Make notifier service work with react components

vigneshm:
`notifier.service.ts` currently only accepts string inputs to render as messages. There's a request to open application configuration modal for non-configured applications via the notifications message and it can't be done without making `notifier.service.ts` accept a react component as content. This PR includes this change and updates the existing usages of this service in spinnaker.

NOTE that I'm not removing the old prop `body` from `notifier.service.ts` as I'm not sure if it is used in other customized spinnaker implementations.

### DIFF
--- a/app/scripts/modules/core/src/application/service/InferredApplicationWarningService.tsx
+++ b/app/scripts/modules/core/src/application/service/InferredApplicationWarningService.tsx
@@ -1,3 +1,5 @@
+import React from 'react';
+
 import { Application } from '../application.model';
 import { NotifierService } from 'core/widgets/notifier/notifier.service';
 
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/application/service/InferredApplicationWarningService.tsx
```diff
@@ -1,5 +1,15 @@
+import React from 'react';
+
 import { Application } from '../application.model';
 import { NotifierService } from 'core/widgets/notifier/notifier.service';
+import { Markdown } from 'core/presentation';
+
+const ConfigureApplicationMessage = ({ appName }: { appName: string }) => (
+  <Markdown
+    message={`The application <b>${appName}</b> has not been <a href="#/applications/${appName}/config">configured</a>.`}
+    options={{ ADD_ATTR: ['onclick'] }}
```
erikmunson, line 10: can this `options` prop specifically be removed (i see why the other usages need it)?

vigneshm:
> 👍 as-is this seems totally reasonable, one thought I'll share: just reading the diff I am not seeing any specific need for using the `Markdown` component and sticking with the original HTML strings from the angular days — is it worth quickly moving over to pure JSX and dropping `Markdown` entirely where `content` is used?

Good call! I wasn't sure of the significance of `Markdown` usage, the `options` prop  and left it that way. Let me test it out and remove it.

PR #8524, opened 2020-08-27
fix(amazon/firewall): Ensure we use a valid default vpc instead of none

vigneshm:
For certain accounts, `vpc0` is not a valid VPC but the defaults provided in `AWSProviderSettings.defaults.vpc` points to `vpc0` resulting in no default vpc being selected. When the application has a creation date of earlier than `AWSProviderSettings.classicLaunchLockout`, we end up defaulting to `none` which points to ec2 classic resulting in undesired behavior.

In this PR, I'm attempting to default to the first available vpc for the account if all of the earlier default attempts have failed.

### DIFF
--- a/app/scripts/modules/amazon/src/securityGroup/configure/CreateSecurityGroup.controller.spec.js
+++ b/app/scripts/modules/amazon/src/securityGroup/configure/CreateSecurityGroup.controller.spec.js
@@ -113,8 +113,8 @@ describe('Controller: CreateSecurityGroup', function() {
       this.$scope.securityGroup.regions = ['us-east-1'];
       this.ctrl.accountUpdated();
       this.$scope.$digest();
-      expect(this.$scope.availableSecurityGroups.length).toBe(2);
-      expect(this.$scope.existingSecurityGroupNames).toEqual(['group1', 'group2']);
+      expect(this.$scope.availableSecurityGroups.length).toBe(1);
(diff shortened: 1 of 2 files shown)


jeyrschabu:
I'm wondering if leaving it blank and graying out the submit button would be less surprising to the user.

vigneshm:
@jeyrschabu

> I'm wondering if leaving it blank and graying out the submit button would be less surprising to the user.

I may have misunderstood the desired behavior here. Based on what I see [here](https://github.com/spinnaker/deck/blob/8ef31b5bfd0dde59de70cb18ec2df6f4cf0156e6/app/scripts/modules/amazon/src/securityGroup/configure/configSecurityGroup.mixin.controller.js#L171), we allow choosing 'none' (i.e no VPC/ec2 classic) as an option for security group creation for applications that were created prior to `AWSProviderSettings.classicLaunchLockout`. 

Are you saying that we shouldn't allow the creation of SGs (without VPCs) even for older apps?

vigneshm:
Updated the PR with unit test fixes. Just wanted to clarify the state of SG creation after my PR.

1. For apps created after `AWSProviderSettings.classicLaunchLockout`, we would never default to `null` vpc (even without my fix) and it wouldn't even be an option to choose from the dropdown.
2. For apps created before `AWSProviderSettings.classicLaunchLockout`, we would never default to `null` vpc (due to my fix), but `None (EC2 Classic)` will be an option in the dropdown.

PR #8527, opened 2020-08-28
fix(ecs): set container name to empty when using container inputs

paragbhingre:
This PR fox the issue of container does exist error by setting the value of containerName parameter to empty when you toggle from task definition artifact to input.

fix the issue : [spinnaker/spinnaker/issues/6004](https://github.com/spinnaker/spinnaker/issues/6004)

If using a task definition artifact:
containerNamePresent

if using inputs:
containerMappings

Scenarios for this change:
1. If using a task definition artifact, you will need to set the value of container name in container mappings and target group mappings section.
2. if using inputs, the value of container name for container mappings and target group mappings will automatically set empty. The container name will be auto generated by the spinnaker.


paragbhingre:
@pkandasamy91 , @allisaurus please take a look at it.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -130,7 +130,7 @@ export interface IEcsServerGroupCommand extends IServerGroupCommand {
   loadBalancedContainer: string;
   targetGroupMappings: IEcsTargetGroupMapping[];
   serviceDiscoveryAssociations: IEcsServiceDiscoveryRegistryAssociation[];
-
+  useTaskDefinitionArtifact: boolean;
```
pkandasamy91, line 133: You don't actually need this property.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -59,7 +61,18 @@ export class Container extends React.Component<IContainerProps, IContainerState>
       dockerImages: cmd.backingData && cmd.backingData.filtered ? cmd.backingData.filtered.images : [],
       targetGroupMappings: cmd.targetGroupMappings,
       targetGroupsAvailable: cmd.backingData && cmd.backingData.filtered ? cmd.backingData.filtered.targetGroups : [],
+      containerMappings: cmd.containerMappings,
     };
+
+    if (!cmd.useTaskDefinitionArtifact) {
```
pkandasamy91, line 67: You don't need to check useTaskDefinitionArtifact since Container.tsx is only invoked when it's using containerInputs (and useTaskDefArtifact is false).

pkandasamy91 (commented):
Looks good overall, had a couple comments.

pkandasamy91:
@paragbhingre can you also rebase off of master?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -59,7 +61,16 @@ export class Container extends React.Component<IContainerProps, IContainerState>
...
+      containerMappings: cmd.containerMappings,
     };
+
+      this.state.targetGroupMappings.forEach(targetGroupMapping => {
+        targetGroupMapping.containerName = "";
+      });
+
+      this.state.containerMappings.forEach(containerMapping => {
+        containerMapping.containerName = "";
+      });
```
pkandasamy91, line 73: nit: indentation

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -24,6 +25,7 @@ interface IContainerState {
   dockerImages: IEcsDockerImage[];
   targetGroupsAvailable: string[];
   targetGroupMappings: IEcsTargetGroupMapping[];
+  containerMappings: IEcsContainerMapping[];
```
pkandasamy91, line 28: Now that I think about it, you shouldn't need to set state here. Container mappings is not modified by Container inputs. What you should do is actually set cmd.containerMappings to null or an empty array of ContainerMappings. ContainerMappings are only used when taskDefArtifacts are used.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -59,7 +61,16 @@ export class Container extends React.Component<IContainerProps, IContainerState>
...
+      containerMappings: cmd.containerMappings,
     };
+
+    this.state.targetGroupMappings.forEach(targetGroupMapping => {
+      targetGroupMapping.containerName = "";
+    });
+
+    this.state.containerMappings.forEach(containerMapping => {
+      containerMapping.containerName = "";
+    });
```
pkandasamy91, line 73: You don't actually need to be setting containerMappings here.

What you should do is actually set cmd.containerMappings to null or an empty array of ContainerMappings. ContainerMappings are only used when taskDefArtifacts are used.

pkandasamy91:
Ship it! 💯 @paragbhingre thanks for fixing this issue! Can you confirm that the behaviour/screenshots are still current?

pkandasamy91 (approved):
Ship it! 💯 Looks great @paragbhingre!

PR #8529, opened 2020-09-01
feat(provider/cf): add bake cf manifest stage

zachsmith1:
This is a new stage specifically for the cloudfoundry provider. This stage is similar to the kubernetes bake manifest stage in that its goal is to produce manifest artifacts from templated manifests and overrides/variables files. The CF CLI currently supports this exact feature, but doesn't expose it externally. Our goal with this stage is to closely mimic the functionality in the CF CLI and provide users the ability to use a templated manifest with 1-N variables files. This stage will allow users to create CF manifests on-the-fly, specific to the deployment environment, without having to manage a separate manifest for each.
Screen Shot 2020-09-01 at 1 30 39 PM

https://github.com/spinnaker/rosco/pull/705
https://github.com/spinnaker/orca/pull/3886


kevinawoo (changes requested):
LGTM, would you mind adding some simple tests? 🙏

kevinawoo (approved):

PR #8537, opened 2020-09-01
feat(publish): Add changelogs to package publish

christopherthielen:
- Split each package bump into a separate commit
- Write changelog for each package to its commit
- Write changelog in the PR body
- Attempts to push to `upstream` so spinnakerbot can auto-approve the package bump


christopherthielen:
Example publish PR: https://github.com/spinnaker/deck/pull/8539

PR #8540, opened 2020-09-02
fix(gce/internalhttplb) BackendServices for InternalHTTPLb is not hadled correctly

takaaki7:
spinnaker/spinnaker#6017


maggieneterval (commented):
Thanks for the fix, @takaaki7! From the issue you created:

> BackendService for InternalHTTP must be set at instanceMetadata.region-backend-service-names. But currently it is set at instanceMetadata.backend-service-names

Would you mind elaborating a bit and walking me through this fix? I am not super familiar with this code, and it's tough to feel comfortable merging this change that could potentially break other workflows without a greater understanding given that none of this code is tested or in Typescript.

Thanks!

xike41:
Is there a way to merge this fix?

PR #8548, opened 2020-09-05
fix(aws): Get imageId and imageName from server group info

jeyrschabu:
- updated updateSecurityGroup to get imageId & name from server group info.
- this works for both launch templates and configs


caseyhebebrand (approved):
LGTM

PR #8550, opened 2020-09-08
feat(core/managed): let resource plugins define their own links

erikmunson:
Soon we'll be adding a details drawer (similar to the infra views) which each resource kind can populate with custom information/links. In the immediate term, we'd like to give resource extensions (one of which is now live internally) the ability to provide a details link that is equivalent to the short-term linking we're doing to infra views for native resource kinds.

I'm aiming to reduce the amount of code/systems we need to touch when getting rid of this in the near future, so for now it's taking the form of some client-side hackery. As a plugin, you'd optionally supply something like:

```ts
resourceKinds: [
  {
    kind: 'my/custom-kind@v1',
    iconName: 'config',
    experimentalDisplayLink: resource => `use resource data to create a link href`,
  }
]
```


app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -60,8 +60,14 @@ const getNativeResourceRoutingInfo = (
...
     const { kind, displayName } = resource;
+
     const routingInfo = getNativeResourceRoutingInfo(resource) ?? { state: '', params: {} };
-    const route = useSref(routingInfo.state, routingInfo.params);
+    const routeProps = useSref(routingInfo.state, routingInfo.params);
+
+    const displayLink = getExperimentalDisplayLink(resource);
+    const displayLinkProps = displayLink && { href: displayLink, target: '_blank', rel: 'noopener noreferrer' };
+
+    const linkProps = routeProps.href ? routeProps : displayLinkProps;
```
erikmunson, line 70: Apparently the `route` conditional on L91 below was always hitting the truthy branch, because `useSref` always returns an object. This was leading to resources that didn't have links showing as `<a>` tags without an href. I'm checking for `href` here now so that we show a plain text name (no link) when we have neither a native sref _or_ a display link from the plugin.

luispollo (approved):
LGTM.

PR #8551, opened 2020-09-08
feat(core): Update app icon and canary config icon

caseyhebebrand:
Based on user feedback, the app icons are being updated to include a color indicator of freshness. The canary config icon is also being updated.


caseyhebebrand:
cc: @gcomstock, for some reason I can't add you as a reviewer

gcomstock:
Weird! Not sure why I cant be added. Just a heads up that these 2 new app icons are the only ones with 2 colors declared inline, so the color prop of the icon component will probably be ignored. I don't think this is worth overthinking, but just calling out a departure from other icons in case the group feels it's worth treating this as more of a standalone svg illustration. I'm going to guess if anyone has an opinion on this it'd be @erikmunson  😆

erikmunson:
ah yeah thanks for the ping — no philosophical objection, but Greg is right that we have so far been very disciplined/strict about `<Icon/>` having single-color treatments with predefined options. In the couple of cases where multi-color vectors have come up, we've done a couple of different things:

- [An `<Illustration/>` component](https://github.com/spinnaker/deck/pull/8336) for illustration use cases
- Direct use of SVGs [straight from an `import` statement](https://github.com/spinnaker/deck/pull/8479) (in the linked PR I also noted a conversation Greg and I had about potentially creating a `symbol` concept if this became more widespread)

Up to your judgement whether you'd like to go with either of those two routes, or do something different for this case!

caseyhebebrand:
> ah yeah thanks for the ping — no philosophical objection, but Greg is right that we have so far been very disciplined/strict about `<Icon/>` having single-color treatments with predefined options. In the couple of cases where multi-color vectors have come up, we've done a couple of different things:
> 
> * [An `<Illustration/>` component](https://github.com/spinnaker/deck/pull/8336) for illustration use cases
> * Direct use of SVGs [straight from an `import` statement](https://github.com/spinnaker/deck/pull/8479) (in the linked PR I also noted a conversation Greg and I had about potentially creating a `symbol` concept if this became more widespread)
> 
> Up to your judgement whether you'd like to go with either of those two routes, or do something different for this case!

@erikmunson @gcomstock I will go ahead and switch it to using an `Illustration` to be more consistent.

erikmunson (approved):

app/scripts/modules/core/src/presentation/illustrations/illustrationsByName.ts
```diff
@@ -7,6 +9,8 @@ import { ReactComponent as unpinArtifactVersion } from './vectors/unpinArtifactV
 
 export const illustrationsByName = {
   announcement,
+  appSynced,
+  appUnsynced,
```
erikmunson, line 13: is this illustration referenced via an `<Illustration/>` component somewhere already? I don't see it in the diff.

caseyhebebrand: @erikmunson Not yet, I have an upcoming PR that will integrate it with some of the layout changes: https://github.com/spinnaker/deck/pull/8556/files#diff-e446ec3f6fb430ca891053eadf01aec6R10

erikmunson: 👍 👍

PR #8552, opened 2020-09-08
feat(core/managed): wording + terminology tweaks for resources

erikmunson:
It's been a while since we did a first stab at wording for each resource status and the pause/resume flows, and we've gotten some feedback in the time since launching. This is primarily a pass at refining and tuning the wording based on what we've learned, not necessarily a final destination. Three main changes:

- Switch from `declarative configuration` to `desired state` across all the various UI surfaces. Neither of these is really that intuitive or clear to folks that haven't works with a K8s-esque declarative system before, but desired state seems to be a little more friendly. Either way some folks will still need to read the docs to fully understand
- Switch from passive -> active voice for as much as possible — not always appropriate but 95%+ of the phrasing has been swapped
- Drop 'continuous' from our management vernacular — no one seems to really find this helpful, and it creates extra verbosity in some cases. This means 'continuous management' becomes 'management', 'continuously manage' becomes 'manage', etc.

If folks are happy with the wording, I'll take some updated UI screenshots and update the documentation to match post-merge.

cc @gcomstock


app/scripts/modules/core/src/managed/DeployingIntoManagedClusterWarning.tsx
```diff
@@ -45,9 +45,9 @@ export const DeployingIntoManagedClusterWarning = ({ app, formik }: IDeployingIn
   return (
     <div className="alert alert-danger">
       <p>
-        <b>🌈 Spinnaker is continuously managing this resource.</b>
+        <b>🌈 Spinnaker is managing this resource.</b>
       </p>
-      <p>Any changes you make to this cluster will be stomped in favor of the declarative configuration.</p>
+      <p>Any changes you make to this cluster will be overridden in favor of the desired state.</p>
```
lorin, line 50: While I think "overriden" is better, I will miss "stomped".

app/scripts/modules/core/src/managed/managedResourceStatusConfig.tsx
```diff
@@ -37,11 +37,10 @@ export const viewConfigurationByStatus: { [status in ManagedResourceStatus]: IVi
     popoverContents: (resourceSummary: IManagedResourceSummary) => (
       <>
         <p>
-          <b>Action is being taken to resolve a drift from the declarative configuration.</b>
+          <b>Spinnaker is taking action to resolve a drift from this resource's desired state.</b>
```
lorin, line 40: This is strictly better than what was there before, but I think people still find the "resolve a drift" language confusing. I don't think people think of a new version becoming available as a resource drifting away from desired state.

Maybe "difference" would be less confusing than "drift"?

lorin (approved):
Changes look good.

To re-iterate a comment I made inline: my sense is people find it confusing that we use the term "drift" to describe the situation where a new version of an artifact becomes available for an environment.

"Drift" carries a connotation that the managed resources have changed in an undesirable way.

app/scripts/modules/core/src/managed/managedResourceStatusConfig.tsx
gal-yardeni: +1 for `difference`

gal-yardeni (approved):

app/scripts/modules/core/src/managed/DeployingIntoManagedClusterWarning.tsx
erikmunson: same — may it rest in peace. 🕯️

app/scripts/modules/core/src/managed/managedResourceStatusConfig.tsx
erikmunson: I think you've mostly convinced me, will swap it out. As a side note on your overall comment about deployments, I personally believe that almost _any_ way of communicating deployments or new versions as a change in actual vs. desired state is pretty academic for most people. My overall strategy so far has been to focus on making the environments experience really great at describing the deployment/version view of the world, and keep resource statuses more focused on state/actuation. I think there's probably room in the future to think about creating a dedicated resource status for 'there's a new/different version that needs to be deployed', but unclear how valuable it would be.

erikmunson:
Updated from drift -> difference as suggested — going to merge now and will update the docs shortly. Feel free to comment post-merge with any further suggestions!

PR #8553, opened 2020-09-09
chore(typescript): Upgrade typescript to v4

vigneshm:
Upgraded typescript to `v4` and fixed errors.


app/scripts/modules/core/src/plugins/plugin.registry.spec.ts
```diff
@@ -5,18 +5,14 @@ import { IPluginMetaData, PluginRegistry } from './plugin.registry';
 import { Registry } from 'core/registry';
 import { mock } from 'angular';
 
-type TestPluginRegistry = PluginRegistry & {
-  load(pluginMetaData: IPluginMetaData): Promise<any>;
-  loadModuleFromUrl(url: string): Promise<any>;
-};
```
vigneshm, line 11: This resulted in typescript inferring the type of `TestPluginRegistry` to `never` since `load`, `loadModuleFromUrl` methods are public in this declaration, but private in the actual `PluginRegistry` definition. So removed this custom definition and added `@ts-ignore` in call sites to get unblocked.

vigneshm:
Working on fixing the build issues.

vigneshm:
Needs https://github.com/spinnaker/deck/pull/8555 to be pushed and @spinnaker/eslint-plugin package bumped to fix the build errors.

christopherthielen (approved):

app/scripts/modules/core/src/managed/ManagedReader.ts
```diff
@@ -105,10 +105,10 @@ export class ManagedReader {
       .then((response: IManagedResourceEventHistoryResponse) => {
         response.forEach(event => {
           if (event.delta) {
-            (event as IManagedResourceEvent).delta = transformManagedResourceDiff(event.delta);
+            ((event as unknown) as IManagedResourceEvent).delta = transformManagedResourceDiff(event.delta);
```
caseyhebebrand, line 108: Do you know why the original type wasn't working? (Just curious, not asking you to change this since it is probably more trouble than its worth right now to perfect the type).

vigneshm: `IManagedResourceEventHistoryResponse` has been explicitly typed [here](https://github.com/spinnaker/deck/blob/2f2e850350c7aa1ba75e5b01761710969f81af40/app/scripts/modules/core/src/domain/IManagedEntity.ts#L186-L196) to be different from `IManagedResourceEvent` due to the difference in the type of `delta` property.

I believe the newer version of ts treats `event` to be of that custom type and doesn't allow the type to be changed/modified since it is a function argument? That is my best guess as I didn't spend time debugging it.

The exact error though is

```
Conversion of type 'Pick<IManagedResourceEvent, "type" | "kind" | "id" | "application" | "timestamp" | "plugin" | "tasks" | "message" | "reason" | "exceptionMessage"> & { ...; }' to type 'IManagedResourceEvent' may be a mistake because neither type sufficiently overlaps with the other
```

caseyhebebrand: Thanks for the explanation!

caseyhebebrand (approved):

PR #8556, opened 2020-09-09
feat(core/*): Deck layout optimizations

caseyhebebrand:
Implementing some user feedback after the new deck layout was released:

- Introducing new app sync/unsynced icons
- Move running executions so that it is visible when the navbar is collapsed
- Make the details drawer friendly for widescreen monitors by ensuring it does not block any information. Make content width responsive for infrastructure routes.
- Updating the layout of details sections (you will see a lot of changes to `dl-narrow` class), and other various tweaks to sizing/layout

Before and after
Screen Shot 2020-09-09 at 12 28 48 PM
Screen Shot 2020-09-09 at 12 24 39 PM

### DIFF
--- a/app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
+++ b/app/scripts/modules/amazon/src/function/details/AmazonFunctionDetails.tsx
@@ -98,7 +98,7 @@ export class AmazonFunctionDetails extends React.Component<IAmazonFunctionDetail
     }
 
     const functionDetails = (
-      <dl className="horizontal-when-filters-collapsed dl-horizontal dl-flex">
+      <dl className="horizontal-when-filters-collapsed dl-horizontal dl-narrow">
         <dt>Last Modified </dt>
         <dd>{functionDef.lastModified}</dd>
         <dt>In</dt>
--- a/app/scripts/modules/amazon/src/loadBalancer/details/loadBalancerDetails.html
+++ b/app/scripts/modules/amazon/src/loadBalancer/details/loadBalancerDetails.html
@@ -66,7 +66,7 @@
   </managed-resource-details-indicator>
   <div ng-if="!ctrl.state.loading" class="content">
     <collapsible-section heading="Load Balancer Details" expanded="true">
-      <dl class="dl-horizontal dl-flex">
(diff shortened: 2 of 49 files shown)


gcomstock:
Thanks so much for getting this together, Casey! This isn't a blocker for submission, but ideally the active number that can display next to each menu item will not cause a horizontal offset. The menu item will remain flush with the rest whether the number is visible or not. It looks like in your attached screen shot, the pipelines [2] is getting nudged to the right a bit.

Here's the design with some redlines:

Screen Shot 2020-09-09 at 1 34 16 PM

caseyhebebrand:
No problem, I can fix that. It looks like the badge is a bit wider for
builds since it’s a double-digit number and the div is centered around the
icon.

caseyhebebrand:
> Thanks so much for getting this together, Casey! This isn't a blocker for submission, but ideally the active number that can display next to each menu item will not cause a horizontal offset. The menu item will remain flush with the rest whether the number is visible or not. It looks like in your attached screen shot, the pipelines [2] is getting nudged to the right a bit.

No problem, I can fix that. It looks like the badge is a bit wider for builds since it’s a double-digit number and the div is centered around the icon. Its an easy fix to left-justify the content.

gcomstock:
I couldnt come up with the right words, so I made a little doodle that could be used as a reference for the positioning of the count to keep it stuck to the lower corner of the icon:

https://codepen.io/gcomstock/pen/VwaXKzQ

christopherthielen:
1) I love the revamped left navigation.  Changing the nav collapse widget color to match the nav color really makes the collapse functionality stand out.

2) Also 👍  on the subtle color in the refresh icon

3) Thought I would dislike the offset count badge, but it really works nicely.  Love having all the nav items lined up on the left hand side.

4) WRT to the infrastructure views, I'm finding the breakpoint resize behavior of the content (i.e., server groups and instance chiclets) jarring and unpleasant.

caseyhebebrand:
@christopherthielen I added the side panel behavior we chatted about. I will put the "drag and drop" links in a separate PR, since I may need to update some tests.

christopherthielen (approved):

PR #8561, opened 2020-09-11
feat(stages): Add monitorPipeline stage execution details

marchello2000:
This is a companion PR for https://github.com/spinnaker/orca/pull/3902
I am not adding ability to add the stage from pipeline designer view (making it synthetic) for now.
But I do want ability to visualize the state of the stage in execution since it will be created in pipelines via automated tooling

image


app/scripts/modules/core/src/orchestratedItem/orchestratedItem.transformer.ts
```diff
@@ -165,6 +166,29 @@ export class OrchestratedItemTransformer {
     return null;
   }
 
+  /**
+   * Get exceptions from child pipelines as references by the monitorPipeline stage
+   */
+  private static getMonitoredPipelineExceptions(task: IStage): string {
```
christopherthielen, line 172: I'm questioning if we really need to add another abstraction for "stage error".  Ideally errors would conform to a single model and be sourced from the same place.

marchello2000: I hear you and I struggled with this myself. But I wanted to expose errors to end users (via spel so and attribute to an execution if they are monitoring multiple pipelines).. I supposed i could also aggregate it in orca for deck and have it available in both places... what do you think?

christopherthielen: I think this behavior is probably unique to this particular stage.  I'd prefer to see this logic isolated in the stage's execution details component.

app/scripts/modules/core/src/orchestratedItem/orchestratedItem.transformer.ts
```diff
@@ -86,7 +86,7 @@ export class OrchestratedItemTransformer {
       status: {
         // Returns either SUCCEEDED, RUNNING, FAILED, CANCELED, or NOT_STARTED
         get: (): string => this.normalizeStatus(item),
-        set: status => {
+        set: (status) => {
```
marchello2000, line 89: i guess webstorm did this...? i can undo if needed

app/scripts/modules/core/src/domain/IExecutionStage.ts
```diff
@@ -24,7 +24,6 @@ export interface IExecutionStage extends IOrchestratedItem, IStage {
   after?: IExecutionStage[];
   before?: IExecutionStage[];
   context: IExecutionContext;
-  id: string;
```
christopherthielen, line 27: This is a surprising change, can you explain?

app/scripts/modules/core/src/orchestratedItem/orchestratedItem.transformer.ts
christopherthielen: prettier does what prettier wants

... but this seems to be only contain formatting changes.  might as well revert this file

app/scripts/modules/core/src/pipeline/config/stages/monitorPipeline/MonitorPipelineStageExecutionDetails.tsx
```diff
@@ -0,0 +1,79 @@
...
+    application,
+    execution,
+    stage: { context = {}, outputs = {} },
+    stage,
+    name,
+    current,
+  } = props;
+
+  const statuses = Object.keys(outputs.executionStatuses).map((executionId) => {
+    const status = outputs.executionStatuses[executionId];
```
christopherthielen, line 19: We should avoid mutating executions.


```suggestion
    const status = {...outputs.executionStatuses[executionId]};
```

christopherthielen, line 19: will `executionStatuses` ever be missing?

app/scripts/modules/core/src/pipeline/config/stages/monitorPipeline/MonitorPipelineStageExecutionDetails.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <ExecutionDetailsSection name={name} current={current}>
+      <div className="row">
+        <div className="col-md-12">
+          <h5>Monitor behavior</h5>
+          {context.monitorBehavior}
+
+          <h5>Executions</h5>
+          {_.sortBy(statuses, 'status').map((status) => {
```
christopherthielen, line 44: probably more idiomatic to sort outside of the JSX.  Way up there ☝️

christopherthielen (changes requested):

app/scripts/modules/core/src/pipeline/config/stages/monitorPipeline/MonitorPipelineStageExecutionDetails.tsx
marchello2000: it shouldn't be unless there is a bug in `orca`

app/scripts/modules/core/src/pipeline/config/stages/monitorPipeline/MonitorPipelineStageExecutionDetails.tsx
marchello2000: yes, thank you!

app/scripts/modules/core/src/domain/IExecutionStage.ts
marchello2000: I can not... i bet it was a rebase gone awry, will put back, thank you!

christopherthielen (approved):

PR #8562, opened 2020-09-11
fix(amazon/serverGroup): Do not clone launch template userData

caseyhebebrand:
Moving assignment of launch template userData in a cloned serverGroup to the backend.


jeyrschabu (approved):
LGTM

PR #8563, opened 2020-09-12
fix(amazon/serverGroup): Enable launch templates for rolling push

caseyhebebrand:
- specifies `setLaunchTemplate` when rolling push strategy is selected by user
- adds a settings to conditionally use `launch templates` in help text depending on whether or not launch templates or launch configurations are activated.

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/configure/serverGroupConfiguration.service.ts
+++ b/app/scripts/modules/amazon/src/serverGroup/configure/serverGroupConfiguration.service.ts
@@ -167,7 +167,7 @@ export class AwsServerGroupConfigurationService {
       command.suspendedProcesses.includes(process);
 
     cmd.onStrategyChange = (command: IAmazonServerGroupCommand, strategy: IDeploymentStrategy): void => {
-      if (AWSProviderSettings.serverGroups.enableLaunchTemplates) {
+      if (AWSProviderSettings.serverGroups?.enableLaunchTemplates) {
         command.setLaunchTemplate = strategy.key === 'rollingpush' ? true : undefined;
       }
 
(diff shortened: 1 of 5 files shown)


app/scripts/modules/amazon/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -166,6 +166,8 @@ export class AwsServerGroupConfigurationService {
       command.suspendedProcesses.includes(process);
 
     cmd.onStrategyChange = (command: IAmazonServerGroupCommand, strategy: IDeploymentStrategy): void => {
+      command.setLaunchTemplate = strategy.key === 'rollingpush' ? true : undefined;
```
jeyrschabu, line 169: I am hoping we don't need this. It feels very specific to rollingPush.

jeyrschabu: The tricky thing is that rollingpush is very orca heavy. Let me play around with this branch a bit.

caseyhebebrand: Just let me know. For now its only useful for rollingPush, but we can always take it out if we don't end up needing this.

jeyrschabu: Sounds good!

app/scripts/modules/amazon/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -166,6 +167,10 @@ export class AwsServerGroupConfigurationService {
       command.suspendedProcesses.includes(process);
 
     cmd.onStrategyChange = (command: IAmazonServerGroupCommand, strategy: IDeploymentStrategy): void => {
+      if (AWSProviderSettings.serverGroups.enableLaunchTemplates) {
```
caseyhebebrand, line 170:
```suggestion
      if (AWSProviderSettings.serverGroups?.enableLaunchTemplates) {
```

jeyrschabu (approved):
LGTM

PR #8565, opened 2020-09-15
feat(manual judgment/deck): Added ability to add roles to manual judgment stage

sanopsmx:

Enhanced stage.html to

Display the roles of the application as a list, only if the stage is Manual Judgment.

Enhanced stage.module.js to

If the role exists in the stage roles added in the manual judgment stage,
then we check for role permissions added in the application.
If the stage role has 'READ', then we disable the continue button
If the stage role has 'WRITE,EXECUTE,CREATE', then we enable the continue button.

Enhanced help.contents.ts file

"The current user must have the permission to approve the manual judgment stage. Otherwise, you'll not be able continue to the next pipeline stage."

### DIFF
--- a/app/scripts/modules/core/src/application/service/ApplicationReader.ts
+++ b/app/scripts/modules/core/src/application/service/ApplicationReader.ts
@@ -41,6 +41,17 @@ export class ApplicationReader {
       });
   }
 
+  public static getApplicationPermissions(applicationName: string): IPromise<any> {
+    return API.one('applications', applicationName)
+      .withParams({
+        expand: false,
+      })
+      .get()
+      .then((fromServer: Application) => {
+        return fromServer.attributes.permissions;
+      });
+  }
+
   public static getApplication(name: string, expand = true): IPromise<Application> {
     return API.one('applications', name)
       .withParams({ expand: expand })
--- a/app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
@@ -5,6 +5,8 @@ import { IExecution, IExecutionStage } from 'core/domain';
 import { Application } from 'core/application/application.model';
 import { Markdown } from 'core/presentation/Markdown';
 import { NgReact, ReactInjector } from 'core/reactShims';
+import { ApplicationReader } from 'core/application/service/ApplicationReader';
+import { AuthenticationService } from 'core/authentication';
 
 export interface IManualJudgmentApprovalProps {
   execution: IExecution;
@@ -16,6 +18,8 @@ export interface IManualJudgmentApprovalState {
   submitting: boolean;
   judgmentDecision: string;
   judgmentInput: { value?: string };
+  applicationRoles: { READ?: string[]; WRITE?: string[]; EXECUTE?: string[]; CREATE?: string[] };
+  userRoles: string[];
   error: boolean;
 }
 
@@ -29,8 +33,25 @@ export class ManualJudgmentApproval extends React.Component<
       submitting: false,
       judgmentDecision: null,
       judgmentInput: {},
+      applicationRoles: {},
+      userRoles: [],
       error: false,
     };
+  }
+
+  public componentDidMount() {
+    const applicationName = this.props.execution.application;
+    ApplicationReader.getApplicationPermissions(applicationName).then(result => {
+      if (typeof result !== 'undefined') {
+        this.setState({
+          applicationRoles: result,
+        });
+        this.isManualJudgmentStageNotAuthorized();
+      }
+    });
+    this.setState({
+      userRoles: AuthenticationService.getAuthenticatedUser().roles,
+    });
   }
 
   private provideJudgment(judgmentDecision: string): void {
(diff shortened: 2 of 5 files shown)


app/scripts/modules/core/src/application/service/ApplicationReader.ts
```diff
@@ -41,6 +41,17 @@ export class ApplicationReader {
       });
   }
 
+  public static getApplicationPermissions(applicationName: string): IPromise<any> {
+    return API.one('applications', applicationName)
+      .withParams({
+        expand: false,
+      })
+      .get()
+      .then((fromServer: Application) => {
```
caseyhebebrand, line 50: Could we call this `app` or `application` so it is more clear what is coming from the server?

app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
```diff
@@ -29,17 +33,60 @@ export class ManualJudgmentApproval extends React.Component<
...
+      applicationRoles: {},
+      userRoles: [],
       error: false,
     };
   }
 
+  public componentDidMount() {
+    const applicationName = this.props.execution.application;
+    ApplicationReader.getApplicationPermissions(applicationName).then(result => {
+      if (typeof result !== 'undefined') {
```
caseyhebebrand, line 45: `if (result)` should be fine

app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
```diff
@@ -29,17 +33,60 @@ export class ManualJudgmentApproval extends React.Component<
...
+        this.setState({
+          applicationRoles: result,
+        });
+        this.isManualJudgmentStageNotAuthorized();
```
caseyhebebrand, line 49: I am a bit confused as to why this is invoked here. The function returns a boolean, so I would expect that value to be set to the `state` here. 

I suggest (coupled with my recommendations for this functions, to set the authorizations as part of the state and using that value for your conditionals in the html

`this.setState({ authorizedForManualJudgment: this.isUserAuthorizedForManualJudgment() })`

app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
```diff
@@ -29,17 +33,60 @@ export class ManualJudgmentApproval extends React.Component<
...
+          isStageNotAuthorized = false;
+          return isStageNotAuthorized;
+        } else if (readArray.includes(usrRole)) {
+          isStageNotAuthorized = true;
+        }
+      }
+    }
+    return isStageNotAuthorized;
+  }
+
```
caseyhebebrand, line 89: I think we can optimize the function a bit. It looks like all you care about is if the user can write, execute or create. Please correct me if I am wrong but the logic below considers that case. 

Note, the logic flips since we are checking if the user **is** authorized, since that would be the happy path, and special handling is needed when they are not authorized. 

```js
private isUserAuthorizedForManualJudgment(): boolean {
     const { applicationRoles, userRoles } = this.state;
     const stageRoles = this.props.stage.context.selectedStageRoles || [];

     if (!stageRoles.length) {
        return true;
     }
    
    let isAuthorized = true; 
    userRoles.forEach( userRole => {
         if (stageRoles.includes(userRole)) {
           const { CREATE, EXECUTE, READ, WRITE } = applicationRoles;
           isAuthorized = (WRITE || []).includes(userRole) || (EXECUTE || []).includes(userRole) || (CREATE || []).includes(userRole);
          }
      })

     return isAuthorized;
    }    
}
```

app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
```diff
@@ -132,6 +135,31 @@ module(CORE_PIPELINE_CONFIG_STAGES_STAGE_MODULE, [
...
+      $scope.getApplicationPermissions = function() {
+        ApplicationReader.getApplicationPermissions($scope.application.name).then(result => {
+          appPermissions = result;
+          if (appPermissions) {
+            const readArray = appPermissions.READ || [];
+            const writeArray = appPermissions.WRITE || [];
+            const executeArray = appPermissions.EXECUTE || [];
+            appRoles = readArray.concat(writeArray, executeArray);
+            appRoles = Array.from(new Set(appRoles));
+            $scope.updateAvailableStageRoles();
```
caseyhebebrand, line 147: Lodash's `union` method would be helpful here: https://lodash.com/docs/4.17.15#union

app/scripts/modules/core/src/application/service/ApplicationReader.ts
sanopsmx: done

app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
sanopsmx: done

caseyhebebrand (changes requested):
Authorization logic, as in another comment in the `ManualJudgmentApproval` file

sanopsmx:
Closing this PR...Raised a new PR.

https://github.com/spinnaker/deck/pull/8779

PR #8567, opened 2020-09-15
chore(core): remove CSS module code + build/test scaffolding

erikmunson:
We did some experimental work with CSS modules in the `core` package a while ago, and decide not to move forward with it. This change does a pretty direct conversion from CSS modules -> less stylesheets for the few places we experimented with (modals + the Environments route), and reverses the build/webpack/karma changes from https://github.com/spinnaker/deck/pull/7650 and https://github.com/spinnaker/deck/pull/7674.


christopherthielen (approved):
:taco:

caseyhebebrand (approved):

PR #8570, opened 2020-09-16
fix(kubernetes/serverGroup): use the actual server group manager's kind in name

ncknt:
Prior to this change, the Kubernetes server group manager's name was hardcoded with prefix `deployment`. We're now looking using the kind as it comes from the server (`$kind $name`).


app/scripts/modules/kubernetes/src/serverGroup/serverGroupTransformer.service.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      } as any;
+    };
+
+    const sg = await transformer.normalizeServerGroup(sgAccount, app);
+    expect(sg).not.toBe(null);
+    expect(sg.serverGroupManagers.length).toBe(1);
+    expect(sgMgrRef.name).toBe('daemonSet sgmanager');
+  });
+
+  it('normalized server group different account', async () => {
```
danielpeach, line 38: it's semi-conventional to try to write these such that the test description makes a sentence with `it` e.g. `it('normalizes the server group name')`

app/scripts/modules/kubernetes/src/serverGroup/serverGroupTransformer.service.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+describe('KubernetesV2ServerGroupTransformer', function() {
+  it('normalized server group name', async () => {
+    const transformer = new KubernetesV2ServerGroupTransformer();
+    const ACCOUNT = 'myaccount';
+    const LOCATION = 'myns';
+    const SG_MANAGER = 'sgmanager';
+    const KIND = 'daemonSet';
+
+    const sgMgrRef = { account: ACCOUNT, location: LOCATION, name: SG_MANAGER };
+    const sgAccount: IKubernetesServerGroup = ({
```
danielpeach, line 16: Can drop the left-hand side `IKubernetesServerGroup` typing unless you think it's more readable that way.

danielpeach (approved):
lgtm

app/scripts/modules/kubernetes/src/serverGroup/serverGroupTransformer.service.spec.ts
ncknt: The `d` is too close to the `s` :) Thanks!

PR #8574, opened 2020-09-17
feat(core/managed): point to resource spec in YAML format instead of JSON

erikmunson:
Following up on https://github.com/spinnaker/gate/pull/1342, this sets up deck to point to the YAML-formatted spec. I'll be coming back around to do the same thing for delivery configs once we throw a link to them into Environments.


christopherthielen:
@erikmunson is this mergable?

erikmunson:
closing this for now as it's low-value, old, and some extra sleuthing might be needed to make it work with the backend.

PR #8578, opened 2020-09-18
chore(storybook): Upgrade storybook and react

vigneshm:
Upgrading react, storybook and kayenta versions.

The newer version of storybook provides a lot of features out of the box with some new API for defining configurable stories.


.storybook/main.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
       ...config.resolve,
       alias: deckWebpackConfig.resolve.alias,
       extensions: deckWebpackConfig.resolve.extensions,
-      modules: deckWebpackConfig.resolve.modules,
+      modules: ['node_modules'],
```
vigneshm, line 13: This update is needed to allow various packages use their own version of their dependencies instead of the ones installed directly under `deck/node_modules`. This is particularly needed due to different versions of `core-js` used by storybook and other older dependencies in `deck` such as `react-bootstrap`.

I was attempting to upgrade `react-bootstrap` and other similar packages, but it came clear that this involved hacking the CSS way too much to make latest version of `react-bootstrap` work with older version of `bootstrap`.

In order to unblock myself, I'm putting upgrading `react-bootstrap` on hold and allowing webpack to bundle up the right dependency versions for each package.

caseyhebebrand (approved):

christopherthielen (approved):

PR #8580, opened 2020-09-18
feat(core/managed): surface git metadata on version list

erikmunson:
Now that we've got git metadata, let's show it off at a glance in the list of versions! We'll now be able to show commit messages, branches, pull requests, and authors without even a single click. Screenshots:

Screen Shot 2020-09-18 at 4 15 35 PM
Screen Shot 2020-09-18 at 4 15 47 PM
Screen Shot 2020-09-18 at 4 16 15 PM

Only thing that's a bit gross is the experience of _not_ having git metadata for a version — it's a bit of a step down in layout/UX right now. I feel strongly we should optimize for making the git metadata case great, so I'm shipping this now and will polish the other case next. I've swapped out the relatively meaningless commit sha we were lifting from versions to instead use the `displayName`. Here's what it looks like:

Screen Shot 2020-09-18 at 4 19 54 PM

cc @gcomstock

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactRow.less
+++ b/app/scripts/modules/core/src/managed/ArtifactRow.less
@@ -1,5 +1,5 @@
 .ArtifactRow {
-  height: 56px;
+  height: 94px;
   display: block;
   position: relative;
   cursor: pointer;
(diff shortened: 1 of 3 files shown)


luispollo:
This looks awesome, @erikmunson!

Quick question. In examples like this:
image

...is `#1580` the number of a PR or something? If so, I'd suggest actually using `PR #1580` instead (I don't think it's enough of a convention to skip the prefix).

erikmunson:
> ...is `#1580` the number of a PR? If so, I'd suggest actually using `PR #1580` instead (I don't think it's enough of a convention to skip the prefix).

I was giving this some consideration as well, might try it given someone besides me mentioned it. I expect people will pick it up pretty quickly but the first impression feels slightly cryptic. @gcomstock any opinions?

erikmunson:
Gonna throw that PR tweak in and get this out so we can learn/experiment, here's what it looks like:

Screen Shot 2020-09-21 at 9 03 43 AM

gcomstock:
Good call!

PR #8582, opened 2020-09-20
fix(tencent+pcf): fix imports 'core' -> '@spinnaker/core'

kevinawoo:
changes the import `core` -> `@spinnaker/core`. 

This is required due to how modules are built and consumed for Deck extended projects. The only place an import like `... from 'core'` will work is within the `core` module itself.

cc @zexiplus
cc @zachsmith1


zachsmith1:
LGTM! Thanks

zachsmith1 (approved):

maggieneterval (approved):

zexiplus:
got that, thanks

PR #8584, opened 2020-09-21
feat(validation): Allow app name validators to be overridden

german-muzquiz:
Overwriting the app name validators allows plugins to change the existing app name rules.


kevinawoo (approved):
LGTM, but since it's `core`, I'd like to make sure others are aware of the changes, namely @christopherthielen and @maggieneterval

PR #8585, opened 2020-09-21
refactor(google): clean up GCE-specific feature config

maggieneterval:
* fix(google): default GCE feature config 

  While adding [predictive autoscaling support](https://github.com/spinnaker/deck/pull/8495), I added a `feature` block to the GCE-specific providers config to prevent GCE-specific feature flags from polluting the core `feature` block. I then proceeded to test the behavior with the new autoscaling flag both enabled and disabled, but did _not_ test with the `feature` block missing altogether, which is the default for Halyard and non-Halyard users who do not enable this feature. This manifests as an exception when we try to [read](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/google/src/autoscalingPolicy/components/metricSettings/GcePredictiveAutoscaling.tsx#L18) `GCEProviderSettings.feature.predictiveAutoscaling` in the Autoscaling section of the GCE server group details view. While we could have fixed this with safer lookups every time we read from the `feature` block, let's instead default the `feature` config to an empty object to [mirror](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/config/settings.ts#L133) the defaulting behavior of the top-level `feature` block, and handle both the case where `SETTINGS.providers.gce` is undefined, and `SETTINGS.providers.gce.feature` is undefined, so that we can always safely read from `SETTINGS.providers.gce.feature`.

* refactor(google): move stateful MIG feature flag to GCE-specific features config 

  Stateful MIG support in Spinnaker was an experimental feature that we never documented or promoted in the community and are no longer actively developing. Now that we have a GCE-specific features config block, let's move the stateful MIG feature flag out of the core features block.


maggieneterval:
If anyone is running 1.22 without predictive autoscaling enabled and wishes to prevent the NPE described in the first commit, you can add the following to your `settings-local.js` (and can remove upon upgrading to 1.23):

```
window.spinnakerSettings.providers.gce.feature = {};
```

ezimanyi (approved):
🧹 🚚 🎏 💯 🥳

PR #8586, opened 2020-09-21
fix(appname): encodeURIComponent for app name

german-muzquiz:
Since the application name is a user supplied value, and that its rules may be changed with plugins, let's encode the url value on network requests.


kevinawoo (approved):
LGTM, but since it's `core`, I'd like to make sure others are aware of the changes, namely @christopherthielen and @maggieneterval 

I hope there isn't a function that would try to parse the name from the url, that would be pretty funky if it did. I think it's unlikely since we're pretty good at passing around the app/pipeline contexts around in memory.

christopherthielen:
This seems like something we should do in the API code itself, not at every call site.    I haven't investigated to find out if that is possible, however.

german-muzquiz:
That was my first option, but then I saw some code like [this one](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/slack/SlackReader.ts#L16) calling the API methods with multiple uri components in one string instead of splitting them.
Probably we would need to fix those calls first before doing URI encoding in the API.

german-muzquiz:
@christopherthielen what do you think about fixing calls to API to not use more than one uri component and encode in API, vs the current implementation in the PR?

christopherthielen:
I think we could do something like `url.split('/').map(segment => encodeUriComponent(segment)).join('/')` in the API code and then remove any manual `encodeUriComponent()` calls that are being passed to `API.one()`/`API.all()`

german-muzquiz:
What if an application name includes `/` as part of its name, like `myapp/db`?
I think it's safer to encode in API without the split and fix the code to do `API.one('slack', 'channels')` instead of `API.one('slack/channels')`

christopherthielen:
I'm fairly certain that an application nameed `myapp/db` would break _everything_ in spinnaker _everywhere_ 😆 

That said, I think your solution is good and we should do that instead.

christopherthielen:
We can also write a linter rule in https://github.com/spinnaker/deck/tree/master/packages/eslint-plugin that warns and auto-fixes this if somebody passes a string literal with a `/` char.

german-muzquiz:
@christopherthielen I made the changes for encoding in API service and fixed calls to not use `/` or `encodeURIComponent`. The idea of the linter rule is good but I'm not familiar with them, do you think it could be a separate PR?

app/scripts/modules/core/src/pipeline/config/services/PipelineConfigService.ts
```diff
@@ -82,7 +82,7 @@ export class PipelineConfigService {
     idsToIndices: { [key: string]: number },
     isStrategy = false,
   ): IPromise<void> {
-    return API.one(`actions/${isStrategy ? 'strategies' : 'pipelines'}/reorder`)
+    return API.one('actions', `${isStrategy ? 'strategies' : 'pipelines'}`, 'reorder')
```
christopherthielen, line 85:
```suggestion
    return API.one('actions', isStrategy ? 'strategies' : 'pipelines', 'reorder')
```

app/scripts/modules/core/src/plugins/plugin.registry.ts
```diff
@@ -98,7 +98,7 @@ export class PluginRegistry {
   public loadPluginManifestFromGate() {
     const source = 'gate';
     const uri = '/plugins/deck/plugin-manifest.json';
-    const loadPromise = API.one(uri)
+    const loadPromise = API.one(...uri.split('/').slice(1))
```
christopherthielen, line 101: ```API.one('plugins', 'deck', 'plugin-manifest.json');```
?

christopherthielen (approved):

christopherthielen:
> The idea of the linter rule is good but I'm not familiar with them, do you think it could be a separate PR?
Yes, of course. LGTM, just a couple minor suggestions

PR #8590, opened 2020-09-23
feat(storybook/SpinFormik): Create a story for SpinFormik

vigneshm:
Created a story for `SpinFormik` component in this PR. This story uses the new [`mdx`](https://storybook.js.org/docs/react/writing-docs/mdx) story format that provides a way to write documentation and stories together. See `SpinFormik.stories.mdx` file.

I would imagine us using the typical component story format ([`CSF`](https://storybook.js.org/docs/react/api/csf)) for most of the stories and this mdx format for a few that require additional context through some documentation.

I'm planning to add a couple of more examples under this story.

1. A form with validation
2. A form with dependent field updates
3. A form with custom field input.

Separately, I'm planning to write individual stories for the various form field inputs.

### Canvas View
image

### Docs View
image


.storybook/preview-head.html
```diff
@@ -0,0 +1,5 @@
+<style>
+  body {
+    overflow: scroll;
```
vigneshm, line 3: This change is needed due to [this](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/main.less#L1162-L1164) which prevents scrolling the storybook content

package.json
```diff
@@ -201,7 +204,7 @@
     "postcss-colorfix": "^0.0.4",
     "postcss-loader": "3.0.0",
     "postcss-nested": "^4.2.1",
-    "prettier": "1.19.1",
+    "prettier": "2.1.2",
```
vigneshm, line 207: This upgrade is needed since the older version prettier throws error when dealing with `.mdx` files

caseyhebebrand:
Storybook is a good way for users to see all states of a component, so I suggest showing an invalid form as well. Either in this PR or a follow-up.

vigneshm:
> Storybook is a good way for users to see all states of a component, so I suggest showing an invalid form as well. Either in this PR or a follow-up.

Yep, that is one of the examples I've listed in the to-do list.

caseyhebebrand (approved):
Very excited about this!

caseyhebebrand:
> Yep, that is one of the examples I've listed in the to-do list.

I just re-read the description, glad you are already planning that :)

christopherthielen:
neato

PR #8596, opened 2020-09-25
feat(titus): Add monitored deploy strategy support to titus

aravindmd:
- Add monitored deploy strategy support to titus
I was able to test this out running deck locally and it works as expected. I don't think any additional changes are required.


caseyhebebrand (approved):
LGTM

PR #8599, opened 2020-09-25
feat(provider/appengine): add deploy global configuration stage

zachsmith1:
This new stage was created in order to support various configuration settings for an app engine application. These settings can be found [here](https://cloud.google.com/appengine/docs/standard/python/configuration-files). For example, appengine supports cron configuration. A user could want to update or deploy cron configuration and its done via a similar mechanism to how services are deployed. This stage replicates the functionality offered and allows users to deploy cron, dispatch, index, and queue configuration files to their appengine environments. 

Screen Shot 2020-09-25 at 3 18 16 PM

https://github.com/spinnaker/orca/pull/3929
https://github.com/spinnaker/clouddriver/pull/4944/


app/scripts/modules/appengine/src/pipeline/stages/deployAppengineConfig/deployAppengineConfigStage.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import {
+  validateDeployAppengineConfigurationStage,
+  DeployAppengineConfigurationConfig,
+} from './DeployAppengineConfigurationConfig';
+
+export const DEPLOY_APPENGINE_CONFIG_STAGE_KEY = 'deployAppEngineConfiguration';
+
+Registry.pipeline.registerStage({
+  label: 'Deploy AppEngine Configuration',
```
danielpeach, line 11: AppEngine -> App Engine

danielpeach (approved):

PR #8600, opened 2020-09-25
feat(core/managed): visual re-treatment of statuses, constraints, and more

erikmunson:
This PR incorporates several changes that — while atomic — are closely interrelated enough to justify rolling out together. Namely:

- We've run into issues where, when looking at a specific version, it's not clear that things are not happening because another version is pinned. We'll now warn when looking at a version if that's the case
- Warning-colored visual treatment on version pills in the overview when an artifact is pinned. This will hopefully increase the understanding of the current state of things at a glance
- Major refresh of the color palette across the view for all things status-related, along with changes to the icon colors for statuses to go with that refresh
- Major refresh of the way cards on a given version detail pane are presented, including reserving background colors for the current overall state of a version in an environment vs. using backgrounds on every card. The goal here is to reduce visual noise, increase the ability to scan/glance, and more easily correlate the colored environment bars on the sidebar to the colored cards in the detail pane

These changes add up together to create (IMO) a much more cohesive, scannable experience throughout, and draw attention to the right things at the right time. We'll continue to iterate based on feedback.

<details>
<summary>New pinning treatment</summary>

**Environment overview (note warning-colored pill)**
Screen Shot 2020-09-25 at 3 45 38 PM

**Version details**
Screen Shot 2020-09-25 at 3 46 25 PM

**When a different version is already pinned**
Screen Shot 2020-09-25 at 3 47 55 PM

</details>

<details>
<summary>New card treatments (misc. example states)</summary>

Screen Shot 2020-09-25 at 3 49 29 PM
Screen Shot 2020-09-25 at 3 50 47 PM
Screen Shot 2020-09-25 at 3 51 18 PM
Screen Shot 2020-09-25 at 3 49 50 PM
Screen Shot 2020-09-25 at 3 51 00 PM
Screen Shot 2020-09-25 at 3 49 17 PM

</details>

cc @gcomstock


luispollo:
I've no idea about the code, but the screenshots look marvelous! 😉

erikmunson:
The code is almost certainly excellent and bug-free

PR #8603, opened 2020-09-27
feat(ecs): Enable embedded artifacts for Deploy stage

atyutyunnik:
This PR will allow for AWS ECS task definition artifacts embedded into the pipelines


allisaurus:
Thanks for submitting this, @atyutyunnik ! 
Change LGTM, though I'd like to see info on how this was tested (maybe screenshot of embedded artifact in stage + successful deployment). Also since this is an enhancement to the ECS provider, can you please update the commit msg prefix to "feat(ecs)" so it will show up correctly in future release notes/changelog?

atyutyunnik:
Sometimes we use a heavily parameterized generic pipeline for multiple environments, instead of having specific pipelines:
2-DeployConfig

Then a task definition available as a JSON from cloudformation output:
```
Outputs:
...
  UiTaskDef:
    Value:
      Fn::Base64:
        !Sub
          |
          {
              "networkMode": "bridge",
              "family": "${BaseName}-ui",
              "placementConstraints": [],
              "executionRoleArn": "${InfrastructureStack.Outputs.EcsExecutionRole}",
...
```

can be transmitted as a parameter to such an initial deployment generic pipeline:
3-parameter

and embed it in Deploy ECS's json:
3-passing_TaskDef_as_base64_encoded_parameter

Of course, when editing such a Deploy ECS's task def artifact editor, it won't show anything (can't render the reference to pipeline's parameter; I think it should rather be grayed out and shown as is, if base64-text can't be decoded)

Yet deployment works just fine:

7-SuccessfulDeployment

allisaurus:
@atyutyunnik appreciate the info on testing! For the prefix change, unfortunately the commit message itself (not just the PR title) needs to be updated to reflect the new prefix. If you can do that and update from master, I think this is good to go.

allisaurus (approved):

PR #8604, opened 2020-09-28
fix(core/spelNumberInput): Fix broken html escaping

vigneshm:
Broke this in https://github.com/spinnaker/deck/pull/8553


caseyhebebrand (approved):

christopherthielen:
wait, what?

vigneshm:
> wait, what?

Some weird effect of react interpolation and html escaping I think

PR #8611, opened 2020-10-01
feat(core/managed): show failed, pending bubbles for any constraint type

erikmunson:
We've had some code for a while that will look for `manual-judgement` constraints in a `PENDING` status and throw them into status bubbles on the version sidebar. This change expands that logic in two ways:

1. We'll now surface _any_ constraint types, not just `manual-judgement`
2. We'll now show constraints with a `FAIL` or `OVERRIDE_FAIL` status using an error-colored bubble

As a bonus, I also threw in a one-liner to fix the z-index issue that made for an awkward scrolling experience (before and after below)

**Failed & rejected constraints**
Screen Shot 2020-10-01 at 9 45 52 AM

**Scrolling (before and after)**
Screen Shot 2020-10-01 at 9 45 10 AM
Screen Shot 2020-10-01 at 9 44 30 AM

cc @gcomstock

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactsList.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactsList.tsx
@@ -9,6 +9,8 @@ import {
 } from '../domain/IManagedEntity';
 import { Icon } from '../presentation';
 
+import { isConstraintSupported, getConstraintIcon } from './constraints/constraintRegistry';
+
 import { ISelectedArtifactVersion } from './Environments';
 import { Pill } from './Pill';
 import { IStatusBubbleStackProps, StatusBubbleStack } from './StatusBubbleStack';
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -134,14 +136,29 @@ function getArtifactStatuses({ environments }: IManagedArtifactVersion): Artifac
   // NOTE: The order in which entries are added to `statuses` is important. The highest priority
   // item must be inserted first.
```
erikmunson, line 137: the way i'm stacking these right now is:

1. pending constraints — this feels like a 'things are actively happening and you should know about it' case
2. failed constraints — it feels kind of yucky to potentially not see these if other things are pending/running, but I thought 1 was more important
3. pinned state — important, but hopefully if you don't get a chance to see it here you'll see it in the overview or via our snazzy new warning card on other versions. I'm less worried about people missing this info if there are multiple status bubbles

app/scripts/modules/core/src/managed/ArtifactsList.tsx
```diff
@@ -134,14 +136,29 @@ function getArtifactStatuses({ environments }: IManagedArtifactVersion): Artifac
...
-  if (requiresManualApproval) {
-    statuses.push({ appearance: 'progress', iconName: 'manualJudgement' });
-  }
+
+  pendingConstraintTypes.forEach((type) => {
+    statuses.push({ appearance: 'progress', iconName: getConstraintIcon(type) });
+  });
+  failedConstraintTypes.forEach((type) => {
+    statuses.push({ appearance: 'error', iconName: getConstraintIcon(type) });
+  });
```
erikmunson, line 161: The way I've setup the logic here is that we will reduce all the constraints across all the environments into a list of their unique types – so let's say there are three pending manual judgements across three environments, we'll show _one single pending manual judgement bubble_. If there were any failed manual judgements, we'd show _one single error manual judgement bubble_. So you'll see that some environments have a constraint in that state, but no context about how many. Same is true for pinning right now.

Longer term we may need to think about supporting quantities on these bubbles so we could e.g. have a pending manual judgement bubble with a little "2" to indicate that there are two environments with that constraint status.

erikmunson:
I thought this was going to be dependent on #8610, but I ended up being wrong. 🥳

vigneshm (approved):
looks great!

PR #8620, opened 2020-10-03
feat(core/managed): put constraints in a 'skipped' state on skipped versions

erikmunson:
When a version becomes `skipped` in an environment, constraints are no longer evaluated and version is mostly forgotten, never to be heard from again. The exception would be if you manipulate the normal flow of versions via pinning/marking as bad, which would forcefully switch their state to something else.

Despite this relative irrelevance, constraints on skipped versions continue to appear very meaningful and important because they don't actually change state in any way. [We've discussed adding a `SKIPPED` status on the backend](https://github.com/spinnaker/keel/issues/1267), but there are some pretty big, pretty urgent fish to fry in keel that should get attention first. This change makes a low-effort attempt to patch over the lack of a `SKIPPED` status by checking for the `skipped` state on environments in client side code.

I've elected to make a few design decisions that may or may not be good:
- If a version is skipped in an environment, I'm not considering _any_ pending or failed constraints as worthy of showing up in the sidebar. I think failed constraints are the one I'm unsure about here — do people want to see a big red bubble on the sidebar for something that's skipped?
- When stateful constraints are PENDING or NOT_EVALUATED, I'm switching them to a new skipped state that says something like "manual judgement skipped" and doesn't have buttons/a timestamp. If they are FAIL or PASS, though, I'm retaining that data as I think it's useful, but visually greying it out so it's relative importance is accurately portrayed
- For stateless constraints, I'm adding a `skipped` handler that can check whether the constraint is passing. On the depends-on constraint I'm still showing the 'already deployed to...' message if it's passing, because again I think that's useful information

<details>
<summary>Sidebar</summary>

**Before**
Screen Shot 2020-10-02 at 5 53 38 PM

**After**
Screen Shot 2020-10-02 at 5 53 13 PM

</details>

<details>
<summary>Pending stateful constraint</summary>

**Before**
Screen Shot 2020-10-02 at 5 53 48 PM

**After**
Screen Shot 2020-10-02 at 5 57 15 PM

</details>

<details>
<summary>Pass/fail stateful constraint</summary>

**Before**
Screen Shot 2020-10-02 at 6 09 23 PM

**After**
Screen Shot 2020-10-02 at 5 52 12 PM

</details>

@gcomstock I would love your input on any/all of these choices

### DIFF
--- a/app/scripts/modules/core/src/domain/IManagedEntity.ts
+++ b/app/scripts/modules/core/src/domain/IManagedEntity.ts
@@ -136,6 +136,8 @@ export interface IManagedArtifactVersion {
   };
 }
 
+export type IManagedArtifactVersionEnvironment = IManagedArtifactSummary['versions'][0]['environments'][0];
+
 export interface IManagedArtifactSummary {
   name: string;
   type: string;
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -9,6 +9,7 @@ import {
   IManagedArtifactVersion,
   IManagedEnvironmentSummary,
   IManagedResourceSummary,
+  IManagedArtifactVersionEnvironment,
 } from '../domain';
 import { Application } from '../application';
 import { useEventListener, Markdown } from '../presentation';
@@ -66,14 +67,21 @@ type IEnvironmentCardsProps = Pick<
   IArtifactDetailProps,
   'application' | 'reference' | 'version' | 'allVersions' | 'resourcesByEnvironment'
 > & {
-  environment: IManagedArtifactSummary['versions'][0]['environments'][0];
+  environment: IManagedArtifactVersionEnvironment;
   pinnedVersion: string;
 };
 
 const EnvironmentCards = memo(
   ({
     application,
-    environment: {
+    environment,
+    reference,
+    version: versionDetails,
+    allVersions,
+    pinnedVersion,
+    resourcesByEnvironment,
+  }: IEnvironmentCardsProps) => {
+    const {
       name: environmentName,
       state,
       deployedAt,
@@ -83,13 +91,7 @@ const EnvironmentCards = memo(
       vetoed,
       statefulConstraints,
       statelessConstraints,
-    },
-    reference,
-    version: versionDetails,
-    allVersions,
-    pinnedVersion,
-    resourcesByEnvironment,
-  }: IEnvironmentCardsProps) => {
+    } = environment;
     const {
       stateService: { go },
     } = useRouter();
@@ -156,7 +158,7 @@ const EnvironmentCards = memo(
             <ConstraintCard
               key={constraint.type}
               application={application}
-              environment={environmentName}
+              environment={environment}
               reference={reference}
               version={versionDetails.version}
               constraint={constraint}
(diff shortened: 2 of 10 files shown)


gcomstock:
> If a version is skipped in an environment, I'm not considering any pending or failed constraints as worthy of showing up in the sidebar. I think failed constraints are the one I'm unsure about here — do people want to see a big red bubble on the sidebar for something that's skipped?

Agreed, let's hide them until we get feedback otherwise

> When stateful constraints are PENDING or NOT_EVALUATED, I'm switching them to a new skipped state that says something like "manual judgement skipped" and doesn't have buttons/a timestamp. If they are FAIL or PASS, though, I'm retaining that data as I think it's useful, but visually greying it out so it's relative importance is accurately portrayed

I've been using the outlined version of icons to represent something that hasn't happened. In your first example, where a manual judgement was never given, it makes sense to me. In the second example, where the manual judgement was rejected, by removing the color, I feel we're losing some information unnecessarily. (something happened here but we're conveying that it didn't)

Your inclination to get rid of the red is probably right. Same case if something became pinned then unpinned. It wouldn't feel right to have the same callout as pinned!

As for the skipped constraint, I think I'd also like to know why. See how this feels @erikmunson @emjburns 

Screen Shot 2020-10-05 at 10 30 43 AM

emjburns:
@gcomstock I like that better. Do you think it would be confusing to show some sort of red / green for pass / fail? When I look at that greyed out bubble it's hard to tell what the status was. I guess I'm not sure if I'd need that information, but if it visually matched a bit more to its non-greyed-out self I think I could better associate them in my head and would have an easier time building a mental model.

gcomstock:
> @gcomstock I like that better. Do you think it would be confusing to show some sort of red / green for pass / fail? When I look at that greyed out bubble it's hard to tell what the status was. I guess I'm not sure if I'd need that information, but if it visually matched a bit more to its non-greyed-out self I think I could better associate them in my head and would have an easier time building a mental model.

I'm not sure! I see the value in the color and the distraction in it. I think I'd want to ask more people and see it in action to have a better opinion.

erikmunson:
I like reserving the dotted outline exclusively for things that haven't happened yet/at all — re: the open question about how much color would/would not be useful here I agree it's tough to tell in advance of real world data. I basically just guessed at this first pass. Let me add in the solid color treatment in that mock and we'll see what people think day-to-day. This is the sort of stuff we may also discover new information about via watching people use the UI (we haven't done that yet, but I bet we will eventually).

erikmunson:
Ok I've done a few things, screenshots below:

- Using filled-in bubbles in the cases discussed (vs. the outline ones)
- I started using the approved/rejected manual judgement icons where appropriate. This was a bit tedious and required some refactoring, which is why I haven't done it in the past. I think this change (i.e. not using color) makes the status-specific icons more important
- Modified the status cards to be able to show greyed-out text while showing a filled-in bubble, to match @gcomstock's screenshot in this thread
- Re-worded the skipped depends-on text because it was awkwardly phrased, "skipped" is now the first word
- Note: **I did not add an explanatory detail to skipped depends-on constraints**. This is quite involved to do well, and it's not currently clear to me whether it's worth the technical investment given so much else is signaling 'skipped for a later version' already here in this list. If it's really bothersome I can do it, just some extra tedium

Screen Shot 2020-10-05 at 12 06 41 PM
Screen Shot 2020-10-05 at 12 06 23 PM
Screen Shot 2020-10-05 at 12 07 08 PM
Screen Shot 2020-10-05 at 12 07 17 PM

emjburns:
Using the actual status icons is enough for me right now! Thanks @gcomstock and @erikmunson. As I use things I'll point out if it's hard to tie things together in my head

gcomstock:
Excellent!

PR #8622, opened 2020-10-05
feat(core/managed): apply consistent sorting to resources

erikmunson:
Teeny tiny sorting tweak: let's sort resources by their `kind`, and then their `displayName` within that. This means that resources of the same kind will be grouped together, and within each group they'll be sorted alphabetically.

(cc @gcomstock)


gcomstock:
Easy win!

PR #8623, opened 2020-10-05
feat(core/managed): limit artifact versions to 30

erikmunson:
As part of trying to do some short-term tweaks to improve Environments performance, this change caps the maximum number of artifact versions we fetch to 30. Above 30 I begin to see increased variability in latency that I'm not super comfortable with (i.e. getting >3s latency on a percentage of calls).


app/scripts/modules/core/src/managed/ManagedReader.ts
```diff
@@ -36,7 +36,7 @@ export const getResourceKindForLoadBalancerType = (type: string) => {
 const transformManagedResourceDiff = (diff: IManagedResourceEventHistoryResponse[0]['delta']): IManagedResourceDiff =>
   Object.keys(diff).reduce((transformed, key) => {
     const diffNode = diff[key];
-    const fieldKeys = flatMap<string, string>(key.split('/').filter(Boolean), fieldKey => {
+    const fieldKeys = flatMap<string, string>(key.split('/').filter(Boolean), (fieldKey) => {
```
erikmunson, line 39: prettier blew up this file cause it hasn't been touched in a while 🤦

PR #8626, opened 2020-10-06
Drag n drop nav

caseyhebebrand:
Related to [issue #6009](https://github.com/spinnaker/spinnaker/issues/6009)

Chrome's drag-n-drop for links only works when there is a `span` directly under the anchor tag. This PR refactors `NavItem` so drag-n-drop works in the navigation.


christopherthielen (approved):
nice work 👍

PR #8627, opened 2020-10-06
Revert "fix(appname): encodeURIComponent for app name (#8586)"

caseyhebebrand:
Reverting [the PR](https://github.com/spinnaker/deck/pull/8586).

### DIFF
--- a/app/scripts/modules/amazon/src/image/image.reader.ts
+++ b/app/scripts/modules/amazon/src/image/image.reader.ts
@@ -27,7 +27,7 @@ export class AwsImageReader {
       return $q.when([{ message: 'Please enter at least 3 characters...', disabled: true }]) as any;
     }
 
-    return API.one('images', 'find')
+    return API.one('images/find')
       .withParams(params)
       .get()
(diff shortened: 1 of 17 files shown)


caseyhebebrand:
@german-muzquiz Please take a look. We need to rollout some changes on top of this PR, so we thought the easiest thing would be to revert until the issue could be resolved .The issue was that all of the slashes were replaced with space for any endpoint calls.

german-muzquiz:
@caseyhebebrand I'm trying to uderstand the issue, are you saying that after merging this PR, any and all API calls replace slashes by spaces? I didn't see that in my tests, can you elaborate a bit more for exactly which API calls did you see this?

christopherthielen (approved):

christopherthielen:
@german-muzquiz This lives in proprietary netflix code (think: plugins) that uses `API.one('with/slashes')`.  I'll add the linter rule first to enable folks to update their code, then we can un-revert this PR.

german-muzquiz (approved):

PR #8629, opened 2020-10-06
feat(eslint-plugin): Add a new rule api-no-slashes that forbids using string literals with slashes to API.one() 

christopherthielen:
Pushing this out as an eslint rule to give plugin authors a chance to fix their code before we flip the new `API` encoding on.

This is first pass at this rule and it doesn't detect literals passed in as variables. Also it probably wont pass CI, heh.  I'm going to run this on the OSS repo and see what happens.

### DIFF
--- a/packages/eslint-plugin/rules/api-no-slashes.js
+++ b/packages/eslint-plugin/rules/api-no-slashes.js
@@ -1,0 +1,122 @@
+'use strict';
+const _ = require('lodash');
+
+/**
+ * Recursively grab the callee until an Identifier is found.
+ *
+ * API.all().all().one('foo/bar');
+ *
+ * var calleeOne = ...
+ * getCallingIdentifier(calleeOne).name === 'API'
+ */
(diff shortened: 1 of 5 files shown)


caseyhebebrand (approved):

packages/eslint-plugin/rules/api-no-slashes.js
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * @version 0.1.0
+ * @category
+ */
+const rule = function (context) {
+  return {
+    CallExpression: function (node) {
+      const { callee = {}, arguments: args = [] } = node;
+
+      debugger;
```
christopherthielen, line 38: debugger 🤦

german-muzquiz (approved):

german-muzquiz:
@christopherthielen at which point can we enable encoding in API again? I can create the PR in the meantime to have it ready, ideally it would be merged before release branches for `1.23` are cut next week.

PR #8631, opened 2020-10-06
chore(eslint): eslint --fix api-no-slashes

christopherthielen:
Applies the auto-fixes for new `api-no-slashes` eslint rule from #8629

Duplicates work originally done in #8586

via:
```eslint --fix```

### DIFF
--- a/app/scripts/modules/azure/src/image/image.reader.js
+++ b/app/scripts/modules/azure/src/image/image.reader.js
@@ -6,15 +6,15 @@ import { API } from '@spinnaker/core';
 
 export const AZURE_IMAGE_IMAGE_READER = 'spinnaker.azure.image.reader';
 export const name = AZURE_IMAGE_IMAGE_READER; // for backwards compatibility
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: app/scripts/modules/azure/src/image/image.reader.js [hunk 1]
```diff
@@ -28,10 +28,10 @@ module(AZURE_IMAGE_IMAGE_READER, []).factory('azureImageReader', function() {
       .withParams({ provider: 'azure' })
       .get()
       .then(
-        function(results) {
+        function (results) {
           return results && results.length ? results[0] : null;
         },
-        function() {
+        function () {
```
christopherthielen, line 34: 🤔 Is this because of a new version of `prettier`?

caseyhebebrand: I think so. I have noticed this and some mods to one param arrow functions as well.

caseyhebebrand (approved):

PR #8639, opened 2020-10-09
feat(core/application): sort Environments first for users with it enabled

erikmunson:
We've gotten some feedback that when Environments is turned on, having it sorted higher than pipelines is useful because typically that's the more common landing point for those users.

(fyi @gcomstock)


gcomstock:
I was going to suggest this! it'd be great to have environments be the default home route for MD apps, if not a ton of effort.

erikmunson:
totally agree and would love to! I'm looking into it but it's sadly not straightforward and could require some silliness. We'll see how far i get.

PR #8642, opened 2020-10-09
 refactor(core): Component for instance actions dropdown

caseyhebebrand:
Migrate instance actions dropdown to react. Integrate with aws/titus cloud providers.


app/scripts/modules/amazon/src/instance/details/instance.details.controller.js
```diff
@@ -505,16 +507,69 @@ module(AMAZON_INSTANCE_DETAILS_INSTANCE_DETAILS_CONTROLLER, [
     this.hasHealthState = function hasHealthState(healthProviderType, state) {
       const instance = $scope.instance;
       const healthMetrics = instance.health || [];
-      return healthMetrics.some(function(health) {
+      return healthMetrics.some(function (health) {
         return health.type === healthProviderType && health.state === state;
       });
     };
 
+    const constructInstanceActions = (instance) => {
```
caseyhebebrand, line 515: My plan is to ultimately have this be in the `AmazonInstanceDetails` component which has any custom methods needed, but is composed of the generic components.

app/scripts/modules/core/src/instance/details/InstanceActions.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <Dropdown.Toggle className="btn btn-sm btn-primary dropdown-toggle">
+        <span>{title || 'Instance Actions'}</span>
+      </Dropdown.Toggle>
+      <Dropdown.Menu>
+        {(actionGroups || [])
+          .filter((group) => group.length)
+          .map((group, i) => {
+            return (
+              <>
+                {group.map((action) => (
```
christopherthielen, line 28: this might read slightly better if you `_.flatten()` the actionGroups first

app/scripts/modules/titus/src/instance/details/instance.details.controller.js
```diff
@@ -309,9 +309,33 @@ module(TITUS_INSTANCE_DETAILS_INSTANCE_DETAILS_CONTROLLER, [
...
+        { label: 'Terminate and Shrink Server Gorup', triggerAction: this.terminateInstanceAndShrinkServerGroup },
+      ];
+      const conditionalActions = [];
+
+      if (this.canRegisterWithDiscovery()) {
+        conditionalActions.push({
+          label: 'Enable In Discovery',
+          triggerAction: this.enableInstanceInDiscovery,
+        });
+      }
```
christopherthielen, line 324: I noticed that `instance` is not used in the constructActions block and also that the actions themselves are referenced off the `this` object.  Do you have a plan to refactor this such that the actions can be decoupled from from the instanceDetailsController and managed separately?  i.e., something like:

```
import { enableInstanceInDiscoveryAction }  from 'somewhere'
     if (this.canRegisterWithDiscovery()) {
        conditionalActions.push(instance => enableInstanceInDiscoveryAction(instance));
     }
```

christopherthielen (approved):
Nice refactor, I like the direction you're taking this

app/scripts/modules/titus/src/instance/details/instance.details.controller.js
caseyhebebrand: Yes, that is exactly what I was thinking. Eventually I will make this a functional React component, where there is no `this` object and the component only captures the conditional logic to construct the actions list. The action functions will be kept either in core or cloudProvider specific folders.

I will remove the `instance` from here so its not confusing for now.

PR #8643, opened 2020-10-10
feat(core): make Environments the default route if enabled

erikmunson:
When users migrate to Managed Delivery, we turn on the Environments route for them. Our internal customers have given us feedback that once they migrate it's much more useful to route directly to Environments when hitting `/application/{application}` vs. the standard default (clusters). As a result @gcomstock and I are looking to try out changing the default route _only for apps with Environments enabled_ — any other apps ought to continue to behave the way they already do.

Based on my limited knowledge of UI Router I've taken a stab at doing this, PLEASE give any feedback you may have as I'm more than happy to take a different approach — my primary concern is just getting the end result this PR achieves.

Today we use the static `addRewriteRule` method to redirect to `/clusters`, but now we need to consult the enabled data sources for the chosen app before deciding where to go. I've taken the approach of making the `application` state non-abstract, and adding a `redirectTo` handler for it which grabs the app model already being fetched for the route and checks it to decide where to redirect. AFAICT this is behaviorally equivalent outside of apps with Environments enabled and doesn't cause jank/noticeable changes in loading performance (which I assume is because we were already blocking on fetching the app model to show the route).

In order to make all the existing links in deck work with this new dynamic default (search results, applications list, etc.), I've gone through to all the places which were routing to the `insight.clusters` state as a means to go to the app's default screen and changed them to point to the newly routable `.application`. Any overridden/customized code that's still pointing directly to `insight.clusters` should theoretically continue to work as it always has. I've also tested it with apps that belong to a project and the nested states continue to work as expected.

### DIFF
--- a/app/scripts/modules/core/src/application/application.state.provider.ts
+++ b/app/scripts/modules/core/src/application/application.state.provider.ts
@@ -33,19 +33,6 @@ export class ApplicationStateProvider implements IServiceProvider {
     this.childStates.push(this.insightState);
   }
 
-  private static getDefaultState(app: Application) {
-    const environments = app.getDataSource('environments');
-    const serverGroups = app.getDataSource('serverGroups');
-
-    if (environments && !environments.disabled) {
-      return '.environments';
-    } else if (serverGroups && !serverGroups.disabled) {
-      return '.insight.clusters';
-    } else {
-      return app.dataSources.find((ds) => ds.sref && !ds.disabled).sref;
-    }
-  }
-
   /**
    * Adds a direct child to the application that does not use the Insight (i.e. inspector) views, e.g. tasks
    * @param state
@@ -97,12 +84,13 @@ export class ApplicationStateProvider implements IServiceProvider {
           .injector()
           .getAsync('app')
           .then((app: Application) => {
-            const defaultState = ApplicationStateProvider.getDefaultState(app);
+            const defaultDataSource = app.dataSources.find((ds) => ds.sref && !ds.disabled)?.sref;
 
             const params = transition.params();
-            const options = { relative: transition.to().name };
+            // If there's no data source to route to, we need to use the absolute href 'home.search'
+            const options = { relative: defaultDataSource ? transition.to().name : undefined };
 
-            return transition.router.stateService.target(defaultState, params, options);
+            return transition.router.stateService.target(defaultDataSource || 'home.search', params, options);
           });
       },
       resolve: {
--- a/app/scripts/modules/core/src/application/application.state.provider.ts
+++ b/app/scripts/modules/core/src/application/application.state.provider.ts
@@ -60,9 +60,9 @@ export class ApplicationStateProvider implements IServiceProvider {
    */
   public addInsightDetailState(state: INestedState): void {
     this.detailStates.push(state);
-    this.insightState.children.forEach(c => {
+    this.insightState.children.forEach((c) => {
       c.children = c.children || [];
-      if (!c.children.some(child => child.name === state.name)) {
+      if (!c.children.some((child) => child.name === state.name)) {
         c.children.push(state);
       }
     });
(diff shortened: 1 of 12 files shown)


app/scripts/modules/core/src/application/application.state.provider.ts
```diff
@@ -78,8 +78,21 @@ export class ApplicationStateProvider implements IServiceProvider {
...
+          .injector()
+          .getAsync('app')
+          .then((app: Application) => {
+            const environments = app.getDataSource('environments');
+            const defaultState = environments && !environments.disabled ? '.environments' : '.insight.clusters';
+
+            const params = transition.params();
+            const options = { relative: transition.to().name };
+
+            return transition.router.stateService.target(defaultState, params, options);
```
erikmunson, line 93: I am in no way attached to this specific approach, it's modified [copy-pasta from the serverGroups state](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/serverGroup/serverGroup.states.ts#L28-L43)

christopherthielen: This looks good to me, and accounts for for access to apps via projects and normally

christopherthielen (approved):
> any other apps ought to continue to behave the way they already do.

I think we could consolidate the serverGroup redirect to this centralized location.  They're both fetching the app metadata and checking if the feature is enabled, then choosing a route.

erikmunson:
> I think we could consolidate the serverGroup redirect to this centralized location. They're both fetching the app metadata and checking if the feature is enabled, then choosing a route.

Agreed — will make that change. I was a little hesitant to go that far before getting a sanity check on whether what I was doing was a reasonable idea!

app/scripts/modules/core/src/application/application.state.provider.ts
caseyhebebrand: I like that is it centralized here.

caseyhebebrand (approved):

erikmunson:
@christopherthielen would love a very quick check on [the latest commit](https://github.com/spinnaker/deck/pull/8643/commits/332ebf466166924ef9fc865a67b6aa9fbfcb764e) which consolidates the serverGroups redirect, just to make sure it gets a second pair of eyes. Tested it with all three cases (environments, clusters, neither) and it seems to behave nicely.

app/scripts/modules/core/src/application/application.state.provider.ts
```diff
@@ -33,6 +33,19 @@ export class ApplicationStateProvider implements IServiceProvider {
...
+    const environments = app.getDataSource('environments');
+    const serverGroups = app.getDataSource('serverGroups');
+
+    if (environments && !environments.disabled) {
+      return '.environments';
+    } else if (serverGroups && !serverGroups.disabled) {
+      return '.insight.clusters';
+    } else {
+      return app.dataSources.find((ds) => ds.sref && !ds.disabled).sref;
+    }
```
christopherthielen, line 46: I was going to suggest adding 'pipelines' to this list, then I found this little guy: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/application/service/ApplicationDataSourceRegistry.ts#L6-L13

What would you think of 
```
const defaultDataSource = ApplicationDataSourceRegistry.getDataSources()
  .find(x => !x.disabled && x.visible && x.sref);
return defaultDataSource?.sref || 'home.search';
```
?

christopherthielen: Note this would have the effect of making 'executions' higher priority than 'clusters', so we might want to swap those in the `defaultDataSourceOrder` if we took this approach?

christopherthielen: One more comment: the appealing thing about this approach to me is that it seems like a reasonable source of truth for default data source.

erikmunson: YES this is much nicer, seems obvious in hindsight (especially since I just touched that default data source order thing the other day). I've updated the PR. 

The only thing I adjusted is that I'm still using `app.dataSources` instead of `ApplicationDataSourceRegistry.getDataSources()`, because `app.dataSources` [respects the enabled/disabled flags](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/application/service/ApplicationReader.ts#L71-L101) that come back from front50 whereas the data source configs from the registry are totally static. They both use the same ordering logic from the registry.

christopherthielen: > app.dataSources respects the enabled/disabled flags that come back from front50 whereas the data source configs from the registry are totally static.

😕

PR #8644, opened 2020-10-12
chore(prettier): Just Update Prettier™

christopherthielen:
prettier was updated to 2.x, but wasn't run against the source code

To update prettier in the future, run `yarn prettier` as a second commit after bumping the version of prettier


erikmunson (approved):
i bet this is gonna work great

vigneshm (approved):

vigneshm:
Thanks for doing it. It was distracting to see prettier noise in other PRs.

PR #8646, opened 2020-10-12
fix(API): encodeURIComponent for each path element

christopherthielen:
Re-submittal of reverted PR #8586

I chose to make the first `.one()`/`.all()` exempt from encoding so legacy code like `API.one('/foo/bar').one(param)` will still work.


caseyhebebrand (approved):

sch-spinnaker:
@christopherthielen This PR breaks fetching of builds from (some?) CI systems, at least Travis but probably also Jenkins. To reproduce, start a manual execution of a pipeline with a build trigger set.
image
This is the call that fails: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/ci/igor.service.ts#L36
`job` here can contain slashes that should not be urlEncoded.

jervi:
@christopherthielen This PR breaks fetching of builds from (some?) CI systems, at least Travis but probably also Jenkins. To reproduce, start a manual execution of a pipeline with a build trigger set.
image
This is the call that fails: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/ci/igor.service.ts#L36
`job` here can contain slashes that should not be urlEncoded.

PR #8648, opened 2020-10-12
@anupmpatil Fix (OCI) : Bugfix 5911 Spinnaker - Oracle Load Balancer Regional Subnets don't work

anupmpatil:
Due to this bug, while creating OCI public load balancer using Spinnaker, UI forced users to select 2 regional subnets. 
But the actual logic should be, if chosen subnet is regional, allow load balancer creation using 1 subnet, else chose 2 ad-specific subnets.

Before fix:

image

After fix:

image


christopherthielen:
@anupmpatil I will squash and merge these commits for you.  Normally, we use a bot to squash commits, which requires the _first commit_ in the PR to match the commit conventions.  I will fix the final commit message to match the conventions.  

In case you're not familiar with rebasing or squashing, here is [an article](https://medium.com/@slamflipstrom/a-beginners-guide-to-squashing-commits-with-git-rebase-8185cf6e62ec) that explains how to squash using the cli.

anupmpatil:
Thank you very much @christopherthielen  for merging. I will read the article.

PR #8649, opened 2020-10-12
fix(eslint-plugin): Cast a wider net in api-no-slashes.js

christopherthielen:
Simplified the detection logic for slashes.

Now detects slashes in the following new cases:

some arg is template string or string concatenation with a slash
```
var prefix = 'proxies/foo/v1';
API.one(`${prefix}/endpoint`);
API.one(prefix + "/endpoint"`);
```

some arg is reference to a template string or concatenated string with a slash
```
var prefix = 'proxies/foo/v1';
var url = prefix + "/endpoint";
var url = `${prefix}/endpoint`;
API.one(url);
```

### DIFF
--- a/packages/eslint-plugin/rules/api-no-slashes.js
+++ b/packages/eslint-plugin/rules/api-no-slashes.js
@@ -31,7 +31,7 @@ const rule = function (context) {
       // Get the source code (think .toString()) of the AST node and find a slash
       // This isn't 100% accurate, but it's good enough.
       function sourceCodeHasSlash(node) {
-        const text = context.getSourceCode().getText(node);
+        const text = node ? context.getSourceCode().getText(node) : '';
         return !!(text && text.includes('/'));
       }
 
--- a/packages/eslint-plugin/rules/api-no-slashes.js
+++ b/packages/eslint-plugin/rules/api-no-slashes.js
@@ -1,6 +1,6 @@
 'use strict';
 const { getCallingIdentifier, getVariableInScope } = require('../utils/utils');
-const _ = require('lodash');
+const { get } = require('lodash');
 
 /**
  * No slashes in string literals passed to API.one() / API.all()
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin/rules/api-no-slashes.js
```diff
@@ -20,36 +20,36 @@ const rule = function (context) {
...
-        if (argument.type !== 'Identifier') {
-          return false;
+      // Get the source code and trigger the lint violation if any forward slashes are found in any of the args
+      // This isn't 100% accurate, but it's good enough.
+      const hasSlash = args.some((arg) => {
+        //  Check if the arg expression has a / anywhere, i.e.:
+        // .one(prevArg, 'foo/bar', nextarg)
+        // .one(prevArg, 'foo/' + barid, nextarg)
+        // .one(prevArg, `foo/${barid}`, nextarg)
+        const text = context.getSourceCode().getText(arg);
```
christopherthielen, line 38: this casts a wider net for detection, but is less accurate

packages/eslint-plugin/rules/api-no-slashes.js
```diff
@@ -20,36 +20,36 @@ const rule = function (context) {
...
+        if (text.includes('/')) {
+          return true;
         }
-        const variable = getVariableInScope(context, argument);
-        const literalValue = _.get(variable, 'defs[0].node.init.raw', '');
-        return literalValue.includes('/');
+
+        // Check if the arg is a variable, and check if that variable was initialized with a slash somewhere
+        const variable = getVariableInScope(context, arg);
+        return variable && context.getSourceCode().getText(variable.init).includes('/');
```
christopherthielen, line 45: Does the same thing as above (like for the argument), but instead looks at variable references and their initializers

christopherthielen:
... gonna tweak something before review

packages/eslint-plugin/rules/api-no-slashes.js
```diff
@@ -20,33 +20,55 @@ const rule = function (context) {
...
+        // console.log('calling identifier not API');
+        return;
+      }
 
-      // var badvar = 'foo/bad'; API.one(badvar);
-      //                                 ^^^^^^
-      const variableArgs = args.filter((argument) => {
-        if (argument.type !== 'Identifier') {
+      // Get the source code (think .toString()) of the AST node and find a slash
+      // This isn't 100% accurate, but it's good enough.
```
christopherthielen, line 32: // This isn't 100% accurate, but it's good enough.

vigneshm (approved):

PR #8652, opened 2020-10-14
chore(build): change Dockerfile.compile base to openjdk:11-slim

dogonthehorizon:
The move to `alpine` caused build failures when attempting to use the
`node` binary provided by the `nebula.node` Gradle plugin. The binary
was built assuming the presence of `glibc` while Alpine is based on
`musl`, so the build would always fail.

Moving to openjdk:11-slim appears to get us around these build issues,
we'll see in the final build in gcloud if that's correct!

### DIFF
--- a/Dockerfile.compile
+++ b/Dockerfile.compile
@@ -1,8 +1,8 @@
-FROM alpine:3.11
-RUN apk add --update \
-    openjdk11 \
-    git \
-    && rm -rf /var/cache/apk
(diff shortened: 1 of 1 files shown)


german-muzquiz (approved):
The change looks good to me

plumpy:
You can test this with `cd deck && ../buildtool/build-service-container.sh Dockerfile.slim`

PR #8656, opened 2020-10-14
fix(core/managed): surface artifact reference on details pane

erikmunson:
We have a pretty nice new way of showing artifact reference names inside version pills, which we've added in both the sidebar and the environments overview. Recently it was noticed that the header for the version details pane still uses the _name_ of the artifact, which isn't unique and can cause confusion (e.g. you could have the same artifact name configured to pull from PR builds and master builds). This change switches to using the reference, and modifies the visual treatment to look the exact same as the other version pills (reference if there are > 1 artifacts and you need them, no reference if you only have 1 artifact).

Before:
Screen Shot 2020-10-14 at 3 23 08 PM

After:
Screen Shot 2020-10-14 at 3 22 53 PM

fyi @gcomstock

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -204,17 +204,18 @@ export interface IArtifactDetailProps {
   version: IManagedArtifactVersion;
   allVersions: IManagedArtifactSummary['versions'];
   allEnvironments: IManagedEnvironmentSummary[];
+  showReferenceNames: boolean;
   resourcesByEnvironment: { [environment: string]: IManagedResourceSummary[] };
   onRequestClose: () => any;
 }
 
 export const ArtifactDetail = ({
   application,
-  name,
   reference,
   version: versionDetails,
   allVersions,
   allEnvironments,
+  showReferenceNames,
   resourcesByEnvironment,
   onRequestClose,
 }: IArtifactDetailProps) => {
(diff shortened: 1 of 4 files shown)


erikmunson:
fyi @luispollo this should help contribute to resolving the issue you ran into the other day where it was non-obvious that you were looking at one artifact reference vs. another (since they both came from the same artifact name)

luispollo:
So the name of the artifact won't be shown anywhere after this change? I think that people will miss that. In fact I've missed even having _more_ info, like the artifact type (unless we have a different icon per type, which I don't think we do?).

erikmunson:
sorry for the delayed follow up here — given the feedback customers have given so far I'm currently taking the approach of only including things I _know_ without question are useful to people outside our team/other central teams. I haven't yet heard that any of the nuances of artifacts (type, name, version) are useful to end customers, though central teams frequently find them useful because they want to dig into/understand what's going on underneath the UX. In many cases end customers don't even know the name of the artifact they're using, they just know there's some CI job that publishes some thing and it ends up getting deployed. 

each individual piece of information we add increases the overall cognitive load of viewing this screen, and there's already _a lot_ of very dense information communicated in lots of different ways here. If end customers find it painful to not see those things, I expect we'll hear about it directly from them at some future date.

luispollo:
Fair enough, @erikmunson. Sometimes I fear we may swing too far to one end of the spectrum when making these UX decisions, and I keep thinking that central teams will also be our customers (although certainly not the majority if things go as planned), but I agree 100% with improving based on user feedback either way.

PR #8660, opened 2020-10-16
refactor(storybook/forms): Rewrite stories for SpinFormik

vigneshm:
There were a few issues with the earlier version of the story I had written for `SpinFormik`. The stories were embedded with the documentation and the source code of the story wasn't formatted.

- In this PR, I'm moved the stories out to a separate file making it quickly accessible and linked them to the documentation. Now the documentation looks exactly like the previous version.
- I've added the `Story` panel which automatically displays the currently selected story's source code.
- Added support to display log messages in `Actions` panel.

image


.storybook/main.js
```diff
@@ -1,8 +1,10 @@
-const deckWebpackConfigurer = require('../webpack.config');
+const path = require('path');
```
christopherthielen, line 1: ?

app/scripts/modules/core/src/presentation/forms/stories/FormikIntro.stories.mdx
```diff
@@ -0,0 +1,117 @@
...
+  render={(formikProps) => <YourFormView {...formikProps} />}
+  onSubmit={(values) => {
+    /* handle form submission */
+  }}
+/>
+```
+
+### Props
+
+`SpinFormik` passes all props to `Formik`, so check out the [documentation](https://formik.org/docs/api/formik#reference) for detailed information on props accepted by `Formik`.
```
christopherthielen, line 43: Fine for now, but I'd like us to determine what props are actually being used in Deck and hide or remove the rest from the "supported" api.

app/scripts/modules/core/src/presentation/forms/stories/FormikIntro.stories.mdx
```diff
@@ -0,0 +1,117 @@
...
+Here's a simple form with all supported input fields
+
+<!-- Simple Form Example -->
+<Canvas>
+  <Story id="forms-spinformik--simple-form" />
+</Canvas>
+
+## Validation
+
+Forms can be validated by providing a `validate` function to `FormikFormField`. The validate function is expected to return an error message when the field value is invalid and `FormikFormField` automatically renders the error message.
```
christopherthielen, line 76: We should document when inline `validate` on the `FormikFormField` is acceptable and when a `validate` function should be provided for the overall form.

Again, good for now but let's come back to this later.

app/scripts/modules/core/src/presentation/forms/stories/FormikIntro.stories.mdx
```diff
@@ -0,0 +1,117 @@
...
+If you wish to use the validation errors outside the scope of a field, then `formikProps` (argument provided via the `render` function) provides `errors` as a map of all the errors keyed by their field name.
```
christopherthielen, line 77: This seems like an edge case and maybe it gets its own story but shouldn't be featured prominently in this section

app/scripts/modules/core/src/presentation/forms/stories/FormikIntro.stories.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+
+<!-- Validation Example -->
+<Canvas>
+  <Story id="forms-spinformik--validation" />
+</Canvas>
+
+## Custom Field Input
+
+Spinnaker provides standard form input fields that cover most common scenarios. If your use case is unique, then you can create a custom input field and use that in the `input` prop to `FormikFormField`.
```
christopherthielen, line 88: I think this also should be less prominently featured somehow.

christopherthielen (approved):
This is cool, I like how it's separated now and in .tsx files.

app/scripts/modules/core/src/presentation/forms/stories/FormikIntro.stories.mdx
vigneshm: I'm going to handle it in a separate PR. I agree it should be less prominent, but I also want to avoid removing it completely from documentation as there is some value in showing how an existing field can be modified to work with `FormikFormField`.

PR #8661, opened 2020-10-16
fix(halyard): Bugfix 6112: settings.js doesn't have Oracle default settings

anupmpatil:
fix(halyard): settings.js doesn't have Oracle default settings 
This settings.js is being used as template by halyard and gets hydrated with default settings. Without these settings, oracle default settings were not getting populated, causing at least one know failure while creating OCI load balancer template.


anupmpatil:
@christopherthielen could you please help this getting merged?

christopherthielen (approved):

PR #8664, opened 2020-10-16
feat(core/managed): run commit messages through Markdown

erikmunson:
We're experimenting a bit with commit messages and how they're displayed, since this is new territory for us/our customers. Some tooling and automation likes to put links in commit messages — we have our own stuff that does this internally and the pattern is fairly common. We're interested in enabling this use case to be a bit more convenient by supporting markdown parsing in messages so you can get actual real links, which Slack will do with the notifications we send by default.

This runs commits through the markdown parser and retains the same whitespace/newline rules, we'll see how it goes. Markdown can often be a bit fickle and unpredictable, so we'll want to keep an eye on whether it's inadvertently mangling commit messages in a way that's not useful.

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -302,7 +302,9 @@ export const ArtifactDetail = ({
                 <VersionMetadataItem
                   label="Message"
                   value={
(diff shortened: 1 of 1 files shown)


luispollo (approved):
LGTM! (Not that I know what I'm doing here... 😆 )

PR #8665, opened 2020-10-19
fix(deck): Bugfix 6112: provide default settings for oracle provider in deck microservice to remove Halyard dependency.

anupmpatil:
fix(deck): settings.js doesn't have Oracle default settings
Since the Halyard is planned to sunset, as suggested by @maggieneterval we are moving the config defaults into the microservices themselves rather than have Halyard supply them.


anupmpatil:
@christopherthielen could you please help us with this PR to review and get merged?

christopherthielen (approved):

PR #8666, opened 2020-10-20
fix(core): Check if manifest event is null from k8s.

jeid64:
Addresses issue https://github.com/spinnaker/spinnaker/issues/6111


maggieneterval (approved):
Thanks! 🎉 🎉

PR #8670, opened 2020-10-21
chore(package): Do not typecheck typecript in webpack (we check via 'tsc' in npm run lib)

christopherthielen:
This is a test to see if avoiding typechecking in webpack module builds can ~improve CI speed~ typecheck only the code in each package when it is built.
We already [typecheck in the package.json scripts](https://github.com/spinnaker/deck/blob/97bd8cb5f1d98b8d8fe4c3b71b5f16c0a160548f/app/scripts/modules/core/package.json#L11) when building packages.


christopherthielen:
For some reason, typechecking in webpack via fork-ts-checker-webpack-plugin was checking code *outside the package being built*.  This made it difficult to change an API in `@spinnaker/core` for example while also updating the code in packages (i.e., `@spinnaker/amazon`) that need to use the API change.  It seems there was sometimes a chicken/egg scenario where `core/lib` wasn't built yet so `amazon` wouldn't compile.  However, we couldn't build `core/lib` because `amazon` wouldn't compile when (erroneously... ?) picked up during the `core` build.

caseyhebebrand (approved):

PR #8672, opened 2020-10-21
feat($q): Adds PromiseLike compatibility to $q and $q promises

christopherthielen:
I'm trying to come up with a plan to migrate us away from our heavy reliance on AngularJS's `IPromise` in our typescript.  Migrating from `IPromise` to `PromiseLike` in our calling code looks like it has merit.  

In order to do this, we need to be able to cast the actual objects (`$q` promises from things like `API` calls) to `PromiseLike`, which is a middle ground interface built into Typescript which has `.then` and `.catch`.  At some point when we replace `$http` in `API` with `fetch` (or whatever), fetch will return a Native promise, which can still be assigned to `PromiseLike` in the downstream code.

----

This PR provides these capabilities:

cast a $q promise as `PromiseLike`:

```
const fetchModels: PromiseLike<Model[]> = $http.get('/models');
```

cast a native promise as `PromiseLike`:
```
const fetchModels: PromiseLike<Model[]> = fetch('/models').then(resp => resp.json());
```

wrap a native promise in a $q promise:
```
const fetchModels: IPromise<Model[]> = $q.when(fetch('/models').then(resp => resp.json()));
const fetchModels2: IPromise<Model[]> = $q.resolve(fetch('/models').then(resp => resp.json()));
```

use `$q.all()` on a mix of native and $q promises:
```
const fetchThings: IPromise<[Thing1, Thing2]> = 
  $q.all([
    fetch('/thing1).then(resp => resp.json()), 
    $http.get('/thing2')
  ]);
```


christopherthielen:
I'm trying to come up with a plan to migrate us away from our heavy reliance on AngularJS's `IPromise` in our typescript.  Migrating from `IPromise` to `PromiseLike` in our calling code looks like it has merit.  

In order to do this, we need to be able to cast the actual objects (`$q` promises from things like `API` calls) to `PromiseLike`, which is a middle ground interface built into Typescript which has `.then` and `.catch`.  At some point when we replace `$http` in `API` with `fetch` (or whatever), fetch will return a Native promise, which can still be assigned to `PromiseLike` in the downstream code.

christopherthielen:
I found that my initial implementation didn't work in `@spinnaker/*` packages, so I moved the typings to `app/types` and added config to `tsconfig.json` with a new `typeRoots` entry.

caseyhebebrand (approved):
looks promising to me (LPTM!)

PR #8681, opened 2020-10-26
feat(core/managed): add DIFF_NOT_ACTIONABLE resource status

erikmunson:
Adding a new resource status, DIFF_NOT_ACTIONABLE, which represents cases where a diff is not resolvable automatically. This currently shows up in health-related cases primarily, but may or may not expand.

Screen Shot 2020-10-26 at 10 41 40 AM

cc @gcomstock


gcomstock:
Thanks for the heads up! Looks good.

PR #8686, opened 2020-10-27
refactor(aws/instance): Reactify instance tags and security groups

caseyhebebrand:
Looking by commit would be easiest.

Converting instance tags and security groups to react components. Eventually we won't need the react2angular comnponents.


app/scripts/modules/core/src/presentation/main.less
```diff
@@ -972,7 +972,7 @@ select:invalid {
     }
   }
   padding: 0;
-  border-bottom: 1px solid var(--color-silver);
+  border-top: 1px solid var(--color-silver);
```
caseyhebebrand, line 975: With a hybrid of the `CollapsibleSection` and `collapsibleSection.directive.html` some borders were on top and some were not. This makes them cohesive.

app/scripts/modules/amazon/src/instance/details/InstanceTags.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import _orderBy from 'lodash/orderBy';
```
christopherthielen, line 2: the rest of the deck does `import { orderBy } from 'lodash'`.  nbd, but nice to be consistent, IMO

app/scripts/modules/amazon/src/instance/details/InstanceSecurityGroups.tsx
```diff
@@ -0,0 +1,41 @@
...
+                to="^.firewallDetails"
+                params={{
+                  name: sg.groupName,
+                  accountId: account,
+                  region,
+                  vpcId,
+                  provider,
+                }}
+              >
+                <a>{`${sg.groupName} (${sg.groupId})`}</a>
```
christopherthielen, line 33:
```suggestion
                <a>{sg.groupName} ({sg.groupId})</a>
```
would this work?

christopherthielen (approved):

vigneshm (approved):

PR #8687, opened 2020-10-27
chore(promiselike): Migrate more code away from angularjs IPromise to PromiseLike

christopherthielen:
it passed ci


christopherthielen:
This is a continuation of https://github.com/spinnaker/deck/pull/8672

PR #8690, opened 2020-10-27
fix(core/cronTrigger): Fix regex for weekly cron triggers

vigneshm:
While specifying a weekly trigger such as `0 0 10 ? * MON,TUE,WED,THU *`, there needs to be a space between `*` and `MON`. We incorrectly allow specifying it without any space which leads to pipelines not being triggered.

It is strange that an explicit support was added for an optional space, but I couldn't find any reasoning behind it from the commit history, so I'm removing it so that it is no longer optional.


christopherthielen:
hmm, lots of 😕 here.  What even is this regex?  ```/^(0 \d+ \d+ \? \*\s?([SUN,MON,TUE, WED, THU, FRI, SAT]*) \*)$/```  This `[SUN,MON,TUE, WED, THU, FRI, SAT]*` notation seems rather not right to me.  Also this regex is only used to set local state (to auto-pick the correct cron type in the dropdown).

app/scripts/modules/core/src/pipeline/config/triggers/cron/CronTrigger.tsx
```diff
@@ -21,7 +21,7 @@ export class CronTrigger extends React.Component<ICronTriggerConfigProps, ICronT
     { label: 'minutes', value: 'minutes', regex: [/^(0 0\/\d+ \* 1\/1 \* \? \*)$/g] },
     { label: 'hourly', value: 'hourly', regex: [/^(0 \d+ 0\/\d+ 1\/1 \* \? \*)$/g] },
     { label: 'daily', value: 'daily', regex: [/^(0 \d+ \d+ 1\/\d+ \* \? \*)$/g, /^(0 \d+ \d+ \? \* MON-FRI \*)$/g] },
-    { label: 'weekly', value: 'weekly', regex: [/^(0 \d+ \d+ \? \*\s?([SUN,MON,TUE, WED, THU, FRI, SAT]*) \*)$/g] },
+    { label: 'weekly', value: 'weekly', regex: [/^(0 \d+ \d+ \? \* ([SUN,MON,TUE, WED, THU, FRI, SAT]*) \*)$/g] },
```
christopherthielen, line 24:
```suggestion
    { label: 'weekly', value: 'weekly', regex: [/^(0 \d+ \d+ \? \* (SUN|MON|TUE|WED|THU|FRI|SAT|\?)+ \*)$/g] },
```

christopherthielen:
from regex101.com:

Screen Shot 2020-10-27 at 4 08 37 PM
😆

vigneshm:
> so I don't understand how this PR could rectify the underlying problem, that pipelines aren't being triggered.

If the user updates the cron expression directly in JSON (like it happens in some pipeline templates), the UI shows a warning if the cron expression doesn't match one of the regex listed in `CronTrigger`. In one of the reported bugs, the UI didn't complain about the invalid cron expression and user was wondering why the pipeline wasn't triggered.

> This [SUN,MON,TUE, WED, THU, FRI, SAT]* notation seems rather not right to me. 

I didn't take a detailed look at how day-of-the-week part was set, it is confusing and I'm trying to figure out how it actually works.

`/^(0 \d+ \d+ \? \* (SUN|MON|TUE|WED|THU|FRI|SAT|\?)+ \*)$/g` wouldn't work when setting multiple days  though (`0 0 10 ? * MON,TUE *`). Did you mean `^(0 \d+ \d+ \? \* (SUN|MON|TUE|WED|THU|FRI|SAT|\,)+ \*)$`?

christopherthielen:
>  figure out how it actually works.

It doesn't, not really.  It will match gibberish such as `0 0 0 ? * S,,,,,WEDTUSAT,AAAIR *`

> UI shows a warning if the cron expression doesn't match one of the regex listed in CronTrigger

Are you sure? It looks to me like we're asking the backend to validate it.

https://github.com/spinnaker/deck/blob/8532bdd4c08d59c38a0adde70ccac4f163c9dd97/app/scripts/modules/core/src/pipeline/config/triggers/cron/cronValidator.service.ts#L11

christopherthielen:
> Did you mean <....>

The `\?` in my regex was explicitly to match `0 0 0 ? * *` which I'm not sure is even a valid cron (but if you choose "Weekly" and then select no days of week you get `0 0 0 ? * *`)

The comma was definitely missing in my suggested edit.

vigneshm:
> Are you sure? It looks to me like we're asking the backend to validate it.

You're right that we're asking the backend for validation, but this only happens when the `activeTab` state is set to `advanced`. If the regex matches, then `activeTab` will be set to `weekly` [here](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/triggers/cron/CronTrigger.tsx#L48) and the error message isn't shown to the user.

image

> It doesn't, not really. It will match gibberish such as 0 0 0 ? * S,,,,,WEDTUSAT,AAAIR *

Oh, wow. Just re-read your earlier comment as well. I'm going to test a few use cases with the new regex and update the PR.

vigneshm:
Ok, I tested a few more scenarios. I've modified the regex to `/^(0 \d+ \d+ \? \* (MON|TUE|WED|THU|FRI|SAT|SUN)(,(MON|TUE|WED|THU|FRI|SAT|SUN))* \*)$/g` in order to avoid capturing `0 0 0 ? * MONTUE *` and `0 0 0 ? * MON, *`.

christopherthielen (approved):
looks good

PR #8691, opened 2020-10-27
feat(promiselike): Add then() and finally() methods to PromiseLike interface

christopherthielen:
1) Copy `.then()` from Typescript's `lib.es5` to `PromiseLike`

Not sure why this is necessary; I expected the declaration to get merged

2) Copy `.finally()` method from `lib.es2018.promise`

Both `IPromise` and `Promise` (as of ES2018) have a `.finally()` method.  Adding this to the `PromiseLike` provides much better interop between the two types.

3) Migrate the remainder of deck off of `IPromise'  🎉 

Hopefully this will be a simple find/replace for any downstream projects, such as deck-customized or plugins.
Or better yet, use `yarn lint --fix` to auto-migrate.


caseyhebebrand (approved):

christopherthielen:
Update: typeRoots was a bad idea because it didn't account for plugins or composite builds like netflix's custom deck build.

I reverted the typeRoots change and moved the types to src/types/*
Also added a kludge to the `@spinnaker/core` build to expose the raw `.d.ts` files to consumers of the library build

caseyhebebrand (approved):

PR #8693, opened 2020-10-28
fix(core/clusterFilter): Use filterModel object to retrieve filter values

vigneshm:
In the reactified `ClusterFilters` component, we end up creating a [new `sortFilter` object](https://github.com/spinnaker/deck/blob/6da861a12aaeda744b253491cb67e2b141fe7362/app/scripts/modules/core/src/cluster/filter/ClusterFilters.tsx#L118) for changes to the filter selection (such as `status`, `search`, `instance`). This makes the [`sortFilter` object stale](https://github.com/spinnaker/deck/blob/a929d3fa4db978aaf7b6d8ada12abc5b03403821/app/scripts/modules/core/src/cluster/allClusters.controller.js#L44) elsewhere.

Fixing the above by referencing `ClusterState.filterModel` directly.


christopherthielen (approved):
lgtm.  

Did you look to see if there are other code that follows the same pattern?

caseyhebebrand (approved):

vigneshm:
> Did you look to see if there are other code that follows the same pattern?

I performed the following to look for similar patterns. Is there any other form that I'm missing here (not super familiar with `*FilterModel` structure)? 

image

caseyhebebrand:
> I performed the following to look for similar patterns. Is there any other form that I'm missing here (not super familiar with `*FilterModel` structure)?

The `LoadBalancerFilters` and `SecurityGroupFilters` probably follow this same pattern.

PR #8699, opened 2020-11-01
feat(ecs): migrate networking server group wizard to react

piradeepk:
This PR migrates the networking section of the create server group wizard for the Amazon ECS provider from angular to react.

Screen Shot 2020-10-31 at 11 28 05 PM
Screen Shot 2020-10-31 at 11 34 56 PM


app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/Networking.tsx
```diff
@@ -0,0 +1,223 @@
...
+  associatePublicIpAddress: boolean;
+  networkMode: string;
+  networkModesAvailable: string[];
+  securityGroups: string[];
+  securityGroupsAvailable: string[];
+  subnetType: string;
+  subnetTypesAvailable: string[];
+}
+
+export class Networking extends React.Component<INetworkingProps, INetworkingState> {
```
allisaurus, line 25: I think the react components share a global namespace in the running app (i.e., there couldn't be a core `NETWORKING_REACT` component if we exported it with this name), so can we continue to preface this with "ECS"? 

I know we don't do this _everywhere_ ATM, though for things like "task definition" it's ECS specific enough for collisions to be less of a worry.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -174,6 +177,7 @@ export class EcsServerGroupConfigurationService {
     command.backingData = {
       // terminationPolicies: clone(this.terminationPolicies)
       launchTypes: clone(this.launchTypes),
+      networkModes: clone(this.networkModes),
```
allisaurus, line 180: is adding this here actually a new thing, or was it always here and we just didn't have to define it before b/c the JS implementation didn't need a type explicitly defined?

piradeepk: In the JS, it was directly being referred to. Rather than having it as backing data, it was just directly referring to a static list in the JS.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/networking/Networking.tsx
piradeepk: All of our other react components are similar to this, but good callout (we should fix them everywhere though, I'll send out a separate PR to fix the rest).

allisaurus (approved):

PR #8702, opened 2020-11-02
Fix new resource versions not showing up in environment view

robfletcher:
We added v1.1 of ALBs (ages ago), and clusters (recently) and they currently do not show up as managed resources in the environments view. This was missed in main because nobody has updated their delivery config to these new versions. However, today I made a query fix (https://github.com/spinnaker/keel/pull/1635) that means the resources will be migrated on-the-fly by the query that backs the API endpoint the environment view uses.

### DIFF
--- a/app/scripts/modules/core/src/managed/resources/registerNativeResourceKinds.ts
+++ b/app/scripts/modules/core/src/managed/resources/registerNativeResourceKinds.ts
@@ -12,6 +12,11 @@ export const registerNativeResourceKinds = () => {
   });
 
   registerResourceKind({
+    kind: 'ec2/cluster@v1.1',
+    iconName: 'cluster',
+  });
+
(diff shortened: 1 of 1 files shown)


lorin (approved):
Going forward, would be nice if we didn't have to edit deck code very time we increment a version on an existing resource.

erikmunson (approved):
👍 funny enough, Luis and I discussed whether the UI should do these registrations via kind group/name (i.e. apply to any/all versions) so that we wouldn't have to do this, but at the time the thought was we'd be leaning toward non-breaking changes and doing schema bumps very rarely so we went this route. If we do go over to registrations applying to all versions, we'll likely have to eventually add some way of still targeting a specific version for cases where the upcoming details drawers for resources are tied to the way a specific version is displayed/parsed.

PR #8710, opened 2020-11-06
fix(amazon/securityGroup): Initialize accounts while cloning security groups

vigneshm:
Following what we already do in [`CreateSecurityGroupCtrl`](https://github.com/spinnaker/deck/blob/35fcebbcdd2c6dea2397a90a48db975293096054/app/scripts/modules/amazon/src/securityGroup/configure/CreateSecurityGroupCtrl.js#L45) and [`EditSecurityGroupCtrl`](https://github.com/spinnaker/deck/blob/35fcebbcdd2c6dea2397a90a48db975293096054/app/scripts/modules/amazon/src/securityGroup/configure/EditSecurityGroupCtrl.js#L113) to initialize all the accounts and make them available in the dropdown selector.


aravindmd (approved):
Thanks for the fix!

PR #8711, opened 2020-11-07
fix(titus/pipeline): Set backingData before accessing it

vigneshm:
This [change](https://github.com/spinnaker/deck/commit/37ff2df776e761dbb286fde994259e288cf0add3) resulted in the broken experience.


alanmquach (approved):
Nice find!

PR #8713, opened 2020-11-09
fix(config): Make running executions limit configurable

emagana:
Adding a new settings configuration to modify the limit of running executions.
Fix issue: https://github.com/spinnaker/spinnaker/issues/6169


app/scripts/modules/core/src/pipeline/service/execution.service.ts
```diff
@@ -22,7 +22,7 @@ export class ExecutionService {
   public get activeStatuses(): string[] {
     return ['RUNNING', 'SUSPENDED', 'PAUSED', 'NOT_STARTED'];
   }
-  private runningLimit = 30;
+  private runningLimit = SETTINGS.defaultExecutionResults || 30;
```
anotherchrisberry, line 25: It would be nice to add this to the `settings.ts` file, also, as an optional field — it helps people see what some of the different configuration options are without having to hunt through the code.

anotherchrisberry (commented):
One very small thing; LGTM otherwise!

app/scripts/modules/core/src/pipeline/service/execution.service.ts
emagana: Done!

settings.js
```diff
@@ -49,6 +49,7 @@ window.spinnakerSettings = {
   },
   checkForUpdates: true,
   debugEnabled: debugEnabled,
+  defaultExecutionResults: 10,
```
emagana, line 52: I am making the default value 10 because in large-scale Spinnaker instances, 30 seems to be too much.

anotherchrisberry: Sorry, this is the settings.**js** file. I was referring to the typescript definition file, settings.**ts**. Including it in this file will change the default value for people running the UI locally, but not on deployed installations.

To that end, let's not change the default. It's been the default for 5+ years and nobody has attempted to change it. Changing it now might be confusing for people running the UI locally, e.g. if they're pointing to a production gate, and look at an app that has 25 running executions, they'll only see 10 locally, but 25 in production.

app/scripts/modules/core/src/pipeline/service/execution.service.ts
anotherchrisberry: A better name for this would be `maxRunningExecutionsToRetrieve` or `backgroundRunningExecutionsLimit` — something to indicate what the setting will effect.

anotherchrisberry (changes requested):

emagana:
@anotherchrisberry  Thanks for your feedback. I have incorporated your suggestions.

app/scripts/modules/core/src/config/settings.ts
```diff
@@ -72,6 +72,7 @@ export interface ISpinnakerSettings {
   bakeryDetailUrl: string;
   checkForUpdates: boolean;
   debugEnabled: boolean;
+  maxRunningExecutionsToRetrieve: 30;
```
christopherthielen, line 75:
```suggestion
  maxRunningExecutionsToRetrieve: number;
```

christopherthielen: If this is typed as the default value `30`, then that's the only value typescript will allow

emagana: got it.. Thanks.. learning. 👍

anotherchrisberry (approved):

anotherchrisberry:
LGTM; defer to @christopherthielen on merging

christopherthielen (approved):

PR #8727, opened 2020-11-11
fix(amazon/targetGroups): Handle empty healthCheckPath

vigneshm:
When `healthCheckProtocol` is `TCP`, the `healthCheckPath` value is empty and `deck` doesn't handle it well.

I'm wondering if we should also remove the [link](https://github.com/spinnaker/deck/blob/c457c3b5d6c5d41e25956294dfb052634c6c1bd0/app/scripts/modules/core/src/instance/loadBalancer/InstanceLoadBalancerHealth.tsx#L44-L46) shown in `InstanceLoadBalancerHealth` since it is linking to `tcp://...`?


robzienert (approved):
🎉

christopherthielen:
> I'm wondering if we should also remove the link shown in InstanceLoadBalancerHealth since it is linking to tcp://...?

Perhaps we should still show the `tcp://` but not render it in an anchor tag

christopherthielen (approved):

PR #8730, opened 2020-11-12
feat(core/managed): support git compare link on versions

erikmunson:
Once https://github.com/spinnaker/keel/pull/1656 is merged, versions in the `PENDING`, `DEPLOYING`, `PREVIOUS`, or `CURRENT` states in an environment will include a link directly to a git compare view (Bitbucket/Stash at Netflix) that shows the commits between that version and what we believe is the appropriate comparison version. For example, a `DEPLOYING` version would compare against the version it's currently replacing in that environment.

This PR adds support in the Environments UI for showing a `See changes` button on each environment where applicable, that links out to the compare URL in a new tab. Also includes some analytics so we can see if it's popular.

Screen Shot 2020-11-11 at 5 53 17 PM

cc @gcomstock

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -1,4 +1,5 @@
 import React, { memo, useMemo } from 'react';
+import ReactGA from 'react-ga';
 import classNames from 'classnames';
 import { useRouter } from '@uirouter/react';
 import { useTransition, animated, UseTransitionProps } from 'react-spring';
@@ -91,6 +92,7 @@ const EnvironmentCards = memo(
       vetoed,
       statefulConstraints,
       statelessConstraints,
+      compareLink,
     } = environment;
     const {
       stateService: { go },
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -147,7 +149,15 @@ const EnvironmentCards = memo(
...
         replacedBy={replacedBy}
         vetoed={vetoed}
+        compareLink={compareLink}
         allVersions={allVersions}
+        logClick={(message) => {
+          ReactGA.event({
+            category: 'Environments - version details',
+            action: message,
+            label: `${application.name}:${environmentName}:${reference}`,
+          });
```
erikmunson, line 159: @gal-yardeni for now, this is how we'll be able to track usage of this button. I'm also logging the app name, environment name, and artifact reference name.

gal-yardeni: @erikmunson how will this information be surfaced?

erikmunson: For us at Netflix it will get pumped into our Google Analytics workspace where we can query it - this is how we do analytics across all of deck today. I am looking into more full-featured product analytics options but for now GA should get us what we need.

PR #8735, opened 2020-11-13
feat(core): Add a container for migration banners

caseyhebebrand:
Introduces a an overridable location for displaying banners about specific migrations. This container is integrated into the executions, clusters, load balancers, and firewalls routes. 

The intent is for the `MigrationsContainer` to be a placeholder to override any custom or business-specific migration components. This is why the implementation of the OSS `MigrationBanners` is very simple. 

Here is an example of a custom migration created by overriding the `MigrationsContainer` with custom logic to render migration a migration banner for certain eligible applications:
Screen Shot 2020-11-20 at 1 19 16 PM


christopherthielen:
@caseyhebebrand  Mind providing a sample screenshot of how this is intended to be utilized?

caseyhebebrand:
> @caseyhebebrand Mind providing a sample screenshot of how this is intended to be utilized?

@christopherthielen I updated the description with a sample image.

christopherthielen:
@caseyhebebrand cool, thanks.  Would it be fair to say these could be used for arbitrary banners, not just migrations?  How does the individual migration banner know if it should display for a given app or not?  Can a user dismiss the banner?

caseyhebebrand:
> @caseyhebebrand cool, thanks. Would it be fair to say these could be used for arbitrary banners, not just migrations? How does the individual migration banner know if it should display for a given app or not? Can a user dismiss the banner?

@christopherthielen  Yes that is fair to say - I will update the naming to reflect a generic banner container, etc. For the last two questions, that is up to the customization of the banner. In the config, the banner routes can be determined, then if there is any additional logic that relates to the application attributes or infra attributes, the custom implementation can handle whether or not to render. The use case I have in mind will not be dismissible and remain until the migration has been completed. 

I mainly needed this to be a placeholder, and expect the banner container to be overridden for custom rendering logic. I can try and make it closer to our internal implementation if that makes more sense.

christopherthielen:
No this is fine, I just wanted a better understanding of the expected usage

christopherthielen (approved):

app/scripts/modules/core/src/banner/BannerContainer.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const { state: currentState } = useCurrentStateAndParams();
+
+  const validBanners = (SETTINGS.banners || [])
+    .filter((m: IBannerSettings) => {
+      const validRoute = m.routes.some((route: string) => currentState.name.includes(route));
+      return m.active && validRoute && Boolean(customBannersByName[m.key]);
+    })
+    .map((m: IBannerSettings) => {
+      const CustomBanner = customBannersByName[m.key];
+      return <CustomBanner key={`migration-${m.key}`} app={app} />;
```
caseyhebebrand, line 35:
```suggestion
      return <CustomBanner key={`banner-${m.key}`} app={app} />;
```

PR #8736, opened 2020-11-16
feat(aws): added the ability to modify CPU credits

pdk27:
## Summary  
  - added a toggle for CPU credits with help fields (shown and related state is set only when applicable)
  - added reusable component for a toggle button group
  - added the ability to display toggle and reset conditionally
  - display CPU credits correctly in 'clone' wizard
  - read `unlimitedCpuCredits` correctly from source server group to prevent invalid configuration

## Issue
The background information for the feature addition is detailed in https://github.com/spinnaker/spinnaker/issues/5989

## Screenshots and Gifs
<details>
<summary>Unlimited CPU Credits Help Field</summary>

</details>

<details>
<summary>ToggleGroupButtonInput</summary>

ToggleButtonGroupInputGif
</details>

<details>
<summary> Unlimited CPU credits is shown only when applicable and reset conditionally (i.e. when selected instance type doesn't support bursting)</summary>

ButtonStateMgmt_TypeAndProfileChange_Reset-part1

ButtonStateMgmt_TypeAndProfileChange_Reset-part2
</details>

<details>
<summary>Clone a SG, and unlimited CPU credits is populated from source SG</summary>

**1. Launch templates enabled**
 - Clone SG with `standard` cpu credits (Note: SG created with cpu credits `undefined` will default to `false` for applicable instance types in [clouddriver](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy#L488))
 - Clone SG with `unlimited` cpu credits

Clone-displaysCreditsCorrectly

**2. Launch templates disabled (Toggle is hidden and unset)**
 - Clone SG 

Clone-displaysCreditsCorrectly

- Create SG

CreateSG_LTdisabled
</details>

## Manual Tests

<details>
<summary>Tests</summary>

1. Create Server Group (SG) with unlimited CPU credits turned on
1. Create SG with unlimited CPU credits turned off
1. Create SG with unlimited CPU credits untouched
**Result**: SG created with unlimited CPU credits turned off, which is the default. 
`unlimitedCpuCredits` is set to `undefined` in Deck, which translates to false in clouddriver. 
See [where default is set](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/handlers/BasicAmazonDeployHandler.groovy#L488).

1. unlimited CPU credits field is reset when instance type is changed in the wizard and the new type doesn't support bursting.
1. Clone SG works as expected 
1. Clone SG + change unlimited CPU credits toggle / instance type works as expected 
1. Unlimited CPU credits toggle is displayed only when launch templates is enabled in Deck

</details>


app/scripts/modules/amazon/src/instance/awsInstanceType.service.js
```diff
@@ -297,12 +298,44 @@ module(AMAZON_INSTANCE_AWSINSTANCETYPE_SERVICE, []).factory('awsInstanceTypeServ
...
+      let familyInCategory = defaultCategories
+        .filter((c) => c.type === category)[0]
+        .families.filter((t) => t.type === family);
+
+      let isTypeInCategory =
+        familyInCategory &&
+        familyInCategory.length > 0 &&
+        familyInCategory[0].instanceTypes.map((t) => t.name).includes(instanceType);
+
+      return isTypeInCategory;
```
caseyhebebrand, line 328: We can use lodash `find` (https://lodash.com/docs/4.17.15#find) here to clean this up. Also, use `const` instead of `let` since there is no reassignment happening. 

```js
const instanceCategory = find(defaultCategories, c => c.type === category);
const familyInCategory = find(instanceCategory?.families, t => t.type === family);

return familyInCategory && familyInCategory.instanceTypes.map((t) => t.name).includes(instanceType);
```

app/scripts/modules/amazon/src/serverGroup/configure/wizard/instanceType/CpuCreditsToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { IAmazonServerGroupCommand } from '../../serverGroupConfiguration.service';
+
+export interface ICpuCreditsToggleProps {
+  command: IAmazonServerGroupCommand;
+  newInstanceType?: string;
+  newProfileType?: string;
+  setUnlimitedCpuCredits: (unlimitedCpuCredits: boolean | undefined) => void;
+}
+
+export class CpuCreditsToggle extends React.Component<ICpuCreditsToggleProps> {
```
caseyhebebrand, line 15: Can we make this a functional component?

app/scripts/modules/amazon/src/instance/awsInstanceType.service.spec.js
```diff
@@ -133,4 +133,61 @@ describe('Service: InstanceType', function () {
       ]);
     });
   });
+
+  describe('isBurstingSupported', function () {
+    it('identifies burstable performance instance types correctly', function () {
+      let types = ['t2.small', 't3.nano', 't3a.medium', 't4g.large', 'm5.small', 'r5.4xlarge'];
+      let service = this.awsInstanceTypeService;
```
caseyhebebrand, line 140: These should be `const` since they are not reassigned. `supportedInstanceTypes` is fine, since you are updating that value in thin the for loop.

Same goes for the rest of this file

app/scripts/modules/amazon/src/serverGroup/configure/wizard/instanceType/CpuCreditsToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+  private handleInstanceTypeChange() {
+    const isBurstingSupported = AwsReactInjector.awsInstanceTypeService.isBurstingSupported(this.props.newInstanceType);
+    if (isBurstingSupported) {
+      return true;
+    }
+    this.props.setUnlimitedCpuCredits(undefined);
+    return false;
```
caseyhebebrand, line 22:
```js
const isBurstingSupported = AwsReactInjector.awsInstanceTypeService.isBurstingSupported(this.props.newInstanceType);
if (!isBurstingSupported) {  this.props.setUnlimitedCpuCredits(undefined) }

return isBurstingSupported;
```

app/scripts/modules/amazon/src/serverGroup/configure/wizard/instanceType/CpuCreditsToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private handleProfileChange() {
+    const instanceType = this.props.command.instanceType;
+    const isTypeInProfile = AwsReactInjector.awsInstanceTypeService.isInstanceTypeInCategory(
+      instanceType,
+      this.props.newProfileType,
```
caseyhebebrand, line 29: What is the difference between the `command.instanceType` and the `newInstanceType`?

app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/ServerGroupInstanceType.tsx
```diff
@@ -26,34 +42,56 @@ export class ServerGroupInstanceType
...
   };
 
   public render() {
     const { values } = this.props.formik;
-    const showTypeSelector = !!(values.viewState.disableImageSelection || values.amiName);
 
     const { InstanceArchetypeSelector, InstanceTypeSelector } = NgReact;
+    const showTypeSelector = !!(values.viewState.disableImageSelection || values.amiName);
+
+    const isLaunchTemplatesEnabled = AWSProviderSettings.serverGroups?.enableLaunchTemplates;
```
caseyhebebrand, line 68: A different setting should be used here. There should be a specific flag to enable CPU credits. By this logic, my users would see this toggle even though we do not want them to.

`AWSProviderSettings.serverGroups?.enableCpuCredits`

app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/ServerGroupInstanceType.tsx
```diff
@@ -26,34 +42,56 @@ export class ServerGroupInstanceType
...
+              onTypeChanged={this.instanceTypeChanged}
+              onProfileChanged={this.instanceProfileChanged}
+            />
+            <div style={{ padding: '0 15px' }}>
+              {values.viewState.instanceProfile && values.viewState.instanceProfile !== 'custom' && (
+                <InstanceTypeSelector command={values} onTypeChanged={this.instanceTypeChanged} />
+              )}
+            </div>
           </div>
+          {isLaunchTemplatesEnabled && (
```
caseyhebebrand, line 85: Nit (with above comment): `isCpuCreditsEnabled`

app/scripts/modules/core/src/presentation/ToggleButtonGroupInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+  btnGroupSize?: string;
+  tooltipPropOnBtn?: string;
+  displayTextPropOnBtn?: string;
+  tooltipPropOffBtn?: string;
+  displayTextPropOffBtn?: string;
+  onClick: (value: boolean) => void;
+  isPropertyActive?: boolean;
+}
+
+export class ToggleButtonGroupInput extends React.Component<IToggleButtonGroupInputProps> {
```
caseyhebebrand, line 18: Could we also make this a functional component?

app/scripts/modules/core/src/presentation/ToggleButtonGroupInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+  public static defaultProps: Partial<IToggleButtonGroupInputProps> = {
+    btnGroupSize: 'btn-group-xs',
```
caseyhebebrand, line 20: I recommend that this is an enumerable `xsmall | small | medium | large` so it is decoupled from the class name.

app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/ServerGroupInstanceType.tsx
pdk27: I didn't add a separate setting thinking this feature should be tied to `enableLaunchTemplates`. 

However, it makes sense to create the separate setting. Could you confirm if the following behaviors match your expectations. 
- When the setting is turned off, by default, t2/t3 instances will be launched with `standard` cpu credits (due to the defaults in clouddriver). The server group details will show cpu credits as `standard` in `launch template` details even when `enableCpuCredits` is turned off.
- Users will have to explicitly turn on the `enableCpuCredits` setting to choose `unlimited`.

app/scripts/modules/amazon/src/serverGroup/configure/wizard/instanceType/CpuCreditsToggle.tsx
pdk27: `newInstanceType` is the state maintained in `CpuCreditsToggle's` parent in order to correctly identify an instance type change and differentiate it from a profile change. Replacing `newInstancetype` with `command.instanceType` will cause the `handleInstanceTypeChange` code path to be executed when instance type didn't change i.e. during profile changes.

app/scripts/modules/amazon/src/serverGroup/configure/wizard/instanceType/CpuCreditsToggle.tsx
```diff
@@ -0,0 +1,68 @@
...
+    props.setUnlimitedCpuCredits(state);
+  };
+
+  let showToggle = false;
+  if (props.newInstanceType) {
+    showToggle = handleInstanceTypeChange();
+  }
+  if (props.newProfileType) {
+    showToggle = handleProfileChange();
+  }
```
caseyhebebrand, line 47:
```js
const isBurstingSupportedForNewInstance = AwsReactInjector.awsInstanceTypeService.isBurstingSupported(newInstanceType);

const { instanceType } = command;
    const isTypeInProfile = AwsReactInjector.awsInstanceTypeService.isInstanceTypeInCategory(
      instanceType,
      newProfileType,
    );
const isBurstingSupportedForInstance = AwsReactInjector.awsInstanceTypeService.isBurstingSupported(instanceType);

  const handleToggleChange = (state: boolean) => {
      setUnlimitedCpuCredits(state);
  };

// I think you are trying to observe if the newInstanceType and newProfileType props have changed, then update the toggle in response. If so, this is a better pattern:
const [showToggle, setShowToggle] = React.useState(false);
React.useEffect(() => {
    if (newInstanceType) {
       setShowToggle(isBurstingSupportedForNewInstance);
    }

   if(newProfileType) {
      setShowToggle(instanceType && isTypeInProfile && isBurstingSupported)
   }
}, [newProfileType, newInstanceType])
```

pdk27: Thanks, Casey. Cleaned up the code.

app/scripts/modules/core/src/presentation/ToggleButtonGroupInput.tsx
caseyhebebrand, line 18: We can make this a functional component.

pdk27: `ToggleButtonGroupInput` shouldn't have been in the commit. I will remove this file. 
I replaced `ToggleButtonGroupInput ` from initial commit with `ToggleButtonGroup` which is a functional component. Sorry about that :#

app/scripts/modules/core/src/presentation/ToggleButtonGroupInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+          <b>{this.props.propLabel}</b>
+          {this.props.propHelpFieldId && <HelpField id={this.props.propHelpFieldId} />}
+        </div>
+        <div className="col-md-8 sm-label-left">
+          <span className={`btn-group ${this.props.btnGroupSize}`}>
+            <button
+              name="prop-off"
+              type="button"
+              className={`btn btn-default ${!this.props.isPropertyActive ? 'active btn-primary' : 'disabled'}`}
+              onClick={() => this.btnClicked(false)}
```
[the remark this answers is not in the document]
pdk27: This unused file is now removed.

caseyhebebrand (approved):
Thanks for incorporating the feedback.

PR #8737, opened 2020-11-16
fix(core): Initialize cache for executions permalink

alanmquach:
The router hook registered against `entering: '**.application.pipelines.executions'` does not get run when someone opens a single execution permalink, so the shared code doesn't get to used values stored in the cache.

Not sure if there is a better way register against both states.


christopherthielen (approved):
seems ok

PR #8738, opened 2020-11-17
feat(core/managed): fill some usage logging gaps in Environments

erikmunson:
Correcting an oversight of mine — I hadn't yet added usage tracking to the meaningful/important behaviors in Environments. Focusing first on the behaviors that are most meaningful to the customer value we seek to provide from MD:

- **Pinning, unpinning, and mark as bad (same events for each flow)**: opening/seeing the modal, dismissing the modal without taking action, taking action successfully (pin/unpin/mark), trying to take action and hitting an error
- **Version metadata**: clicking the link to a version's commit, clicking the link to a version's PR
- **Constraints**: clicking a 'constraint action', which is just what we call the buttons on constraints in the UI like "Approve"/"Reject" for manual judgements. The events we log will include the type of constraint and which action was clicked

The rounds out the tracking we already have on the new "See changes" button to capture most of the value-add behaviors a user could interact with, beyond just viewing the page itself which we already have tracking on.

### DIFF
--- a/app/scripts/modules/core/src/managed/MarkArtifactAsBadModal.tsx
+++ b/app/scripts/modules/core/src/managed/MarkArtifactAsBadModal.tsx
@@ -1,4 +1,4 @@
-import React, { memo, useCallback } from 'react';
+import React, { memo, useCallback, useEffect } from 'react';
 import ReactGA from 'react-ga';
 import { Option } from 'react-select';
 
@@ -28,11 +28,11 @@ import { useEnvironmentTypeFromResources } from './useEnvironmentTypeFromResourc
 
 const MARK_BAD_DOCS_URL = 'https://www.spinnaker.io/guides/user/managed-delivery/marking-as-bad/';
 
-const logClick = (label: string, application: string) =>
+const logEvent = (label: string, application: string, environment?: string, reference?: string) =>
   ReactGA.event({
     category: 'Environments - mark version as bad modal',
-    action: `${label} clicked`,
-    label: application,
+    action: label,
+    label: environment ? `${application}:${environment}:${reference}` : application,
   });
 
 type IEnvironmentOptionProps = Option<string> & { disabledReason: string; allResources: IManagedResourceSummary[] };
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/managed/MarkArtifactAsBadModal.tsx
```diff
@@ -62,7 +62,12 @@ export interface IMarkArtifactAsBadModalProps extends IModalComponentProps {
 }
 
 export const showMarkArtifactAsBadModal = (props: IMarkArtifactAsBadModalProps) =>
-  showModal(MarkArtifactAsBadModal, props, { maxWidth: 750 });
+  showModal(MarkArtifactAsBadModal, props, { maxWidth: 750 }).then((result) => {
+    if (status === 'DISMISSED') {
```
vigneshm, line 66: I can't quite figure out where `status` comes from.

erikmunson: it comes from [the result object returned by `showModal()`](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/modal/showModal.tsx#L54)

vigneshm: Hmmm, I don't see `status` retrieved from `result` object here? I'm not sure why eslint didn't complain here.

erikmunson: oopsie, yeah good catch this is a refactor gone bad. Really confused as to why this built or passed CI at all 😕

erikmunson: [fixed](https://github.com/spinnaker/deck/pull/8738/commits/b7d42bb71460ab3983798d5ac0a7381316628ae0) — apparently there is such a thing as `window.status`, which is a string, and hence TS thought this was totally a-ok 🙄

vigneshm (approved):

PR #8739, opened 2020-11-17
fix(core/api): Fix type safety for API.one() builder

christopherthielen:
Previously, `API.one()` returned `any` so all type safety and auto-complete was lost.  Now, those methods return `IRequestBuilder` interface.

- I've cleaned up the calling code whereever it was relying on `any` type being returned
- I added generic parameter to `.get<ReturnType>()` (and `.post`/`.put`/etc)

### DIFF
--- a/app/scripts/modules/core/src/api/ApiService.spec.ts
+++ b/app/scripts/modules/core/src/api/ApiService.spec.ts
@@ -1,6 +1,7 @@
 import Spy = jasmine.Spy;
 import { mock, noop } from 'angular';
 import { AuthenticationInitializer } from '../authentication/AuthenticationInitializer';
+import { ICache } from '../cache';
 import { API, InvalidAPIResponse } from './ApiService';
 import { SETTINGS } from 'core/config/settings';
 
@@ -293,12 +294,6 @@ describe('API Service', function () {
     });
 
     describe('create config with data', function () {
-      it('should not alter the config if no data object passed', function () {
-        const result = API.one('foo').data();
-        expected.url = `${baseUrl}/foo`;
-        expect(result.config).toEqual(expected);
-      });
-
       it('should add data to the config if data object passed', function () {
         const data = { bar: 'baz' };
         const result = API.one('foo').data(data);
@@ -310,11 +305,6 @@ describe('API Service', function () {
   });
 
   describe('create a config with params', function () {
-    it('when no params are provided do not alter config', function () {
-      const result = API.one('foo').withParams();
-      expect(result.config).toEqual({ method: '', url: `${baseUrl}/foo` });
-    });
-
     it('when params are provided', function () {
       const result = API.one('foo').withParams({ one: 1 });
       expect(result.config).toEqual({ method: '', url: `${baseUrl}/foo`, params: { one: 1 } });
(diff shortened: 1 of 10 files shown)


app/scripts/modules/core/src/api/ApiService.spec.ts
```diff
@@ -293,12 +294,6 @@ describe('API Service', function () {
     });
 
     describe('create config with data', function () {
-      it('should not alter the config if no data object passed', function () {
```
christopherthielen, line 296: I removed these tests because it was testing for an implementation detail.

app/scripts/modules/core/src/api/ApiService.ts
```diff
@@ -5,18 +5,21 @@ import { SETTINGS } from 'core/config/settings';
 import { ICache } from 'core/cache';
 import { isNil } from 'lodash';
 
+type toStringable = { toString(): string };
+type IParams = Record<string, toStringable | toStringable[]>;
```
christopherthielen, line 9: Added an `IParams` type which allows:

```
{
  stringParam: 'bar',
  numberParam: 5,
  booleanParam: true,
  multiParam: [2, 'foo', false]
}
```
to generate URLs like
```
?stringParam=bar&numberParam=5&booleanParam=true&multiParam=2&multiParam=foo&multiParam=false
```

app/scripts/modules/core/src/image/image.reader.ts
```diff
@@ -2,19 +2,19 @@ import { module } from 'angular';
 
 import { ProviderServiceDelegate, PROVIDER_SERVICE_DELEGATE } from 'core/cloudProvider/providerService.delegate';
 
-export interface IFindImageParams {
+export type IFindImageParams = {
```
christopherthielen, line 5: for some reason, `Record<>` types need don't work with `interface`s but do with `type`s
https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFApgZygCgPoA8BciASihHAE4AmAPBuTGAOYA0iYIAtgEYrkB8ASkQBvAL4AocQHopiAO4UA1mnFkwGRAwA2DFABUAngAcUhYYmBw4hdt16JRiALwiLVwgCYHAbnGoMmNq6hiYCvtKyCuTKfsYoiABiViHxLuaW1mycPOQOqggaAIZoKYRJcCnOrhmeef5YxSlhkjKIAJJglCi4iGgwjGCFUCDk8RwwaH1M4gyo5MCFEPHlbQtL1e5ZdrkSakVoq4umiVaH62lumQCMdegNB2sozUA

app/scripts/modules/core/src/managed/ManagedReader.ts
```diff
@@ -89,7 +88,7 @@ export class ManagedReader {
       .then(this.decorateResources);
   }
 
-  public static getEnvironmentsSummary(app: string): PromiseLike<IManagedApplicationEnvironmentSummary> {
```
christopherthielen, line 92: @erikmunson I think this refactor could have exposed a typing problem.  Can you confirm that I did the right thing here?

vigneshm (approved):
lg

app/scripts/modules/core/src/api/ApiService.ts
```diff
@@ -5,18 +5,21 @@ import { SETTINGS } from 'core/config/settings';
 import { ICache } from 'core/cache';
 import { isNil } from 'lodash';
 
+type toStringable = { toString(): string };
```
christopherthielen, line 8: Thinking I'll switch this to a more explicit `type toStringable = boolean | string | number`

PR #8741, opened 2020-11-18
feat(core): Create a wizard and modal for multi-step experiences

caseyhebebrand:
The `StepWizard` is a component that takes a set of steps, tracks the current step to be rendered, and injects props into the rendered components.

The `ModalStepWizard` is custom application of the `StepWizard` to a modal. This ends in a customizable modal that renders one step at a time. Here is a little demo of a non-linear progression through steps. 
Step 0 -> Async Action -> Failure -> Go back to Async Action -> Success -> Complete

multistepmodal_demo


christopherthielen:
One of the design flaws of the existing wizard modal is that it's tightly coupled to modal code.  Can you spend some time considering if the wizard componets could be modal-agnostic (could be used inside any component)?

caseyhebebrand:
> One of the design flaws of the existing wizard modal is that it's tightly coupled to modal code. Can you spend some time considering if the wizard componets could be modal-agnostic (could be used inside any component)?

@christopherthielen  I will think through it. In this case it wouldn't be a `ModalStepWizard` it would just be a `GeneralStepWizard`.  There is still a need to integrate this wizard to a modal (whether its this version or a generic version) since a lot of new design patterns that would make this modal useful.

app/scripts/modules/core/src/presentation/stepWizard/StepWizard.tsx
```diff
@@ -0,0 +1,20 @@
...
+  currentStep: React.ReactNode;
+  goToStep: (i: number) => void;
+}
+
+export interface IStepWizardProps {
+  steps: React.ReactNode[];
+  render: (props: IStepWizardInjectedProps) => React.ReactNode;
+}
+
+export const StepWizard = ({ render, steps }: IStepWizardProps) => {
```
caseyhebebrand, line 14: @christopherthielen I am finally revisiting this. Here is the very simple `StepWizard` we talked about a couple weeks ago. 

Given how simple it is, my preference would be to either make this an HOC (rather than using a render prop) to inject the `currentStep`, `currentIndex`, and `goToStep`, or to just abandon this and go back to my original plan of just creating the `ModalStepWizard`.

[This commit](https://github.com/spinnaker/deck/pull/8741/commits/11a5c7f69ce6bda369fcb8d5c9b4e216252702d5#diff-4c5869ab49ace445f582d1d129efba599b1a09942f5d2bc01260bfaf2414031cR26) has an example of the HOC, but it is less abstracted than this iteration.

PR #8746, opened 2020-11-20
fix(titus): Fallback dimensions for target tracking policy

alanmquach:
Just making `dimensions` an empty array if it's undefined to prevent this reduce from throwing:
https://github.com/spinnaker/deck/blob/976e958b25d44086b59108367ab038199ffcdc7e/app/scripts/modules/amazon/src/serverGroup/details/scalingPolicy/chart/metricAlarmChart.component.js#L42


christopherthielen (approved):
ok

PR #8747, opened 2020-11-20
fix(titus): Disable edits of step scaling dimensions

alanmquach:
Not the cleanest, but the least invasive way to temporarily disable editing of dimensions in step scaling policies for Titus.


christopherthielen (approved):
How much work would it be to migrate this to react?

aravindmd:
Thanks for this fix !

alanmquach:
Just the dimension editor or all of auto-scaling? The dimension editor shouldn't be too bad (famous last words), but the rest of the auto-scaling modal would be a bit more involved. I wouldn't say it's an enormous undertaking though.

PR #8748, opened 2020-11-20
fix(core/pipelines): break and wrap long comments text

gcomstock:


christopherthielen (approved):

christopherthielen:
@gcomstock is OK to merge?

gcomstock:
Hey @christopherthielen ! I think so, but I'm not able to run deck locally any more to verify and I don't think it's worth the effort to troubleshoot since I won't be in there any more. Please give it a quick look over for me. Thank you!

christopherthielen (approved):

PR #8749, opened 2020-11-21
fix(core): add dotted underline visual indicator to hover-able popovers

gcomstock:
Minor edit, but presents itself in a few different places. Thoughts? I'm having issues running locally, so just putting this PR up to get feedback and remind myself to check other usages after Thanksgiving.

Screen Shot 2020-11-20 at 3 10 41 PM
Screen Shot 2020-11-20 at 3 10 44 PM


christopherthielen:
@gcomstock  Blue does stand out a bit, wondering if a subtle grey would be preferable?

gcomstock:
Good idea!

christopherthielen (approved):

PR #8750, opened 2020-11-21
feat(core/managed): surface pre-deployment steps for versions

erikmunson:
This is the UI side of https://github.com/spinnaker/keel/pull/1657.

We're working on getting 'pre-deployment' steps surfaced for artifact versions, the two we currently know of are VM image bakes and builds (not covered here). The data model is thankfully pretty generic, so I've set things up in a way where adding build steps later will be trivial (plugging in an icon + some text for each step status).

The visual presentation is very similar to constraints:
1. On the versions sidebar, running and failed steps will show up as status bubbles
2. On the version details pane, we'll add a "pre-deployment" section to the bottom which will include status cards for each step

In order to make sure that it's easy to spot these steps on versions which are in the pre-deployment phase (not deployed to any environment yet), we're going to collapse/hide the environments at the top in a subsequent change. I'm splitting that work out because while related, the actual code to implement it is totally separate and I want to get the customer value in this PR delivered ASAP.

Screen Shot 2020-11-20 at 5 52 08 PM
Screen Shot 2020-11-20 at 5 51 04 PM
Screen Shot 2020-11-20 at 5 48 33 PM
Screen Shot 2020-11-20 at 5 49 24 PM
Screen Shot 2020-11-20 at 5 49 54 PM
Screen Shot 2020-11-20 at 5 50 30 PM

cc @gcomstock


app/scripts/modules/core/src/managed/PreDeploymentRow.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import './PreDeploymentRow.less';
+
+interface IPreDeploymentRowProps {
+  children?: React.ReactNode;
+}
+
+export function PreDeploymentRow({ children }: IPreDeploymentRowProps) {
```
erikmunson, line 9: i'm aware this is extremely duplicative in relation to `EnvironmentRow`, and I certainly could've been clever about making `EnvironmentRow` accept different props for both use cases. It's unclear to me whether this pre-deployment step presentation is going to work out, or if we're going to totally change it up shortly after shipping, so I want to avoid ending up with a mildly brittle, unhelpful component abstraction for a use case we end up not having for long.

gcomstock:
Looks awesome!!

vigneshm (approved):

PR #8752, opened 2020-11-23
feat(mocks) Added mocks for AmazonServerGroup

pdk27:
Added mocks for AmazonServerGroup

### DIFF
--- a/packages/mocks/package.json
+++ b/packages/mocks/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@spinnaker/mocks",
-  "version": "1.0.6",
+  "version": "1.0.7",
   "module": "dist/index.js",
   "typings": "dist/index.d.ts",
   "license": "Apache-2.0",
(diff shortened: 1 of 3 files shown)


packages/mocks/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@spinnaker/mocks",
-  "version": "1.0.6",
+  "version": "1.0.7",
```
pdk27, line 3: Is version increment done manually? Will be helpful if the reviewer can describe the release / publish process for `@spinnaker/mocks`.

caseyhebebrand: Right now this is manual. After this merges, you can go to the `mocks` directory and run `npm publish`, which will update the registry to match this version.

caseyhebebrand (approved):

packages/mocks/package.json
pdk27: Sounds good. Thanks.

PR #8757, opened 2020-11-25
chore(pluginsdk-peerdeps): Publish 0.0.4

caseyhebebrand:
Bumping to use some newly exported core components in a plugin.


christopherthielen (approved):
Was this updated manually?  I suspect tooling for updating this package isn't working quite right.  Either way, this is fine.

caseyhebebrand:
> Was this updated manually? I suspect tooling for updating this package isn't working quite right. Either way, this is fine.
@christopherthielen Yes I updated it manually

PR #8760, opened 2020-11-30
Deprecate API.one('foo').all('bar').get() in favor of REST('/foo/bar').get()

christopherthielen:
I've been doing some thinking about our `API` rest builders and how it could be improved.

Things I don't like:

1) `.one()` vs `.all()`
  What do these mean?  They are borrowed from the Restangular API but don't actually do anything different.
2) `.get()` vs `.getList()`
  Again, borrowed from Restangular, but they do the same thing.
3) `.data(obj).post()` vs `.post(obj)` ...  `.data(obj).put()` vs `.put(obj)`
  Whats the difference?
4) `.withParams({ param: value }).get()` vs `.get({ param: value })`
  Whats the difference?
5) encodeUriComponent in .one() and .all()
  this PR https://github.com/spinnaker/deck/pull/8646 aims to encode path parameters, such as 'appname'  in `.one('applications', appname)`.  However, this makes it less convenient to append a string like `/proxies/pandora/api/v1`.

With all this in mind, this PR refactors the API to a smaller surface area:

```ts
/**
 * A Builder API for making requests to Gate backend service
 */
export interface IRequestBuilder {
  /**
   * Appends one or more path segments to the URL, separated by slashes.
   * Each path segment is uri encoded.
   */
  path(...pathSegments: IPrimitive[]): this;

  /** Adds query parameters to the URL */
  query(queryParams: IParams): this;

  /** Enables or disables caching of the response */
  useCache(useCache?: boolean): this;

  /** issues a GET request */
  get<T = any>(): PromiseLike<T>;
  /** issues a POST request */
  post<T = any, P = any>(data?: P): PromiseLike<T>;
  /** issues a PUT request */
  put<T = any, P = any>(data?: P): PromiseLike<T>;
  /** issues a DELETE request */
  delete<T = any>(): PromiseLike<T>;
}
```

example:

```
REST('/path/to/resourcetype').path(id).query({ expand: true }).get();
```

---

### Backwards compat:

This refactor introduces a builder factory exported as `REST`.  The previous `API` builder factory still exists and can be used with only minor changes -- this refactor is mostly backwards compatible with the `API` implementation.  One exception is that the old builder impl mutated the http config object, so code like this will no longer work with:

```

let getUsers = API.one('users').one(user);

// OLD WAY
if (expand) {
  // Return value is not used, but getUsers is mutated
  getUsers.withParams({ expand: true });  
}

return getUsers.get();
```

This code should be migrated to:

```
let getUsers = API.one('users').one(user);

// NEW WAY
if (expand) {
  getUsers = getUsers.withParams({ expand: true });  // re-assign the return value
}

return getUsers.get();
```

A linter rule (in a separate PR) should help find and migrate code like this.

---

### HttpClient Implementations

This refactor introduces a simple interface to provide different HTTP clients:

```ts
/**
 * An interface to support pluggable http clients
 * In the future, we should have a TestingHttpBackend and a FetchHttpBackend (or whatever http client we go with)
 */
export interface IHttpClientImplementation {
  get<T = any>(config: IRequestBuilderConfig): PromiseLike<T>;
  post<T = any>(config: IRequestBuilderConfig): PromiseLike<T>;
  put<T = any>(config: IRequestBuilderConfig): PromiseLike<T>;
  delete<T = any>(config: IRequestBuilderConfig): PromiseLike<T>;
}
```

Currently, the AngularJS `$http` client is the only implementation, but we should be able to write a `fetch()` based backend (or whatever) as we migrate away from AngularJS.  We should also be able to leverage this "backend" mechanism to implement a test friendly mock http client backend to replace the AngularJS `$httpBackend` `expectGET()` and similar.

### DIFF
--- a/app/scripts/modules/core/src/api/ApiService.ts
+++ b/app/scripts/modules/core/src/api/ApiService.ts
@@ -1,193 +1,254 @@
-import { IRequestConfig } from 'angular';
-import { $q, $http } from 'ngimport';
-import { AuthenticationInitializer } from '../authentication/AuthenticationInitializer';
-import { SETTINGS } from 'core/config/settings';
-import { ICache } from 'core/cache';
 import { isNil } from 'lodash';
+import { $http } from 'ngimport';
+import { ICache } from '../cache/deckCacheFactory';
+import { SETTINGS } from '../config/settings';
+import { AuthenticationInitializer } from '../authentication/AuthenticationInitializer';
 
 type IPrimitive = string | boolean | number;
 type IParams = Record<string, IPrimitive | IPrimitive[]>;
 
+/**
+ * A Builder API for making requests to Gate backend service
+ */
 export interface IRequestBuilder {
-  config?: IRequestConfig;
-  one?: (...urls: string[]) => IRequestBuilder;
-  all?: (...urls: string[]) => IRequestBuilder;
-  useCache?: (useCache?: boolean | ICache) => IRequestBuilder;
-  withParams?: (params: IParams) => IRequestBuilder;
-  data?: (data: any) => IRequestBuilder;
-  get?: <T = any>(params?: IParams) => PromiseLike<T>;
-  getList?: <T = any>(params?: IParams) => PromiseLike<T>;
-  post?: <T = any>(data?: any) => PromiseLike<T>;
-  remove?: <T = any>(params?: IParams) => PromiseLike<T>;
-  put?: <T = any>(data?: any) => PromiseLike<T>;
+  /**
+   * Appends one or more path segments to the URL, separated by slashes.
+   * Each path segment is uri encoded.
+   */
+  path(...pathSegments: IPrimitive[]): this;
+
+  /** Adds query parameters to the URL */
+  query(queryParams: IParams): this;
+
+  /** Enables or disables caching of the response */
+  useCache(useCache?: boolean): this;
+
+  /** issues a GET request */
+  get<T = any>(): PromiseLike<T>;
+  /** issues a POST request */
+  post<T = any, P = any>(data?: P): PromiseLike<T>;
+  /** issues a PUT request */
+  put<T = any, P = any>(data?: P): PromiseLike<T>;
+  /** issues a DELETE request */
+  delete<T = any>(): PromiseLike<T>;
+}
+
+/**
+ * Internal interface to encapsulate a request
+ * Passed to IHttpClientBackend
+ */
+interface IRequestBuilderConfig {
+  url: string;
+  timeout?: number;
+  headers?: { [headerName: string]: string };
+  /** @deprecated used for AngularJS backwards compat */
+  data?: any;
+  params?: object;
+  cache?: boolean;
+}
+
+/**
+ * The old API interface
+ */
+export interface IDeprecatedRequestBuilder extends IRequestBuilder {
+  useCache(): this;
+  useCache(useCache: boolean): this;
+  useCache(useCache: ICache): this;
+  withParams(queryParams: IParams): this;
+
+  /** @deprecated do not use this config object */
+  config: IRequestBuilderConfig;
+  /** @deprecated use SETTINGS.gateUrl */
+  baseUrl: string;
+  /** @deprecated use path() instead (this is a passthrough to path) */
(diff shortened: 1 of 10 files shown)


app/scripts/modules/core/src/api/ApiService.ts
```diff
@@ -1,193 +1,254 @@
...
+  get<T = any>(params?: IParams): PromiseLike<T>;
+  /** @deprecated use delete() instead (this is a passthrough to delete) */
+  remove(params?: IParams): PromiseLike<any>;
+  /** @deprecated use get() instead (this is a passthrough to get) */
+  getList<T = any>(params?: IParams): PromiseLike<T>;
+}
+
+/**
+ * An interface to support pluggable http clients
+ * In the future, we should have a TestingHttpBackend and a FetchHttpBackend (or whatever http client we go with)
```
alanmquach, line 79: Might just be because I haven't done too much Angular, but I find the term "backend" confusing when we're really just referring to an http client.

Now might not be the best time, but can these (eventually) just be referred to as `HttpClient`s? Or is there a distinction I'm missing?

alanmquach (approved):
Great work doing this progressively with the backwards compatible piece.

Excited to use REST but also thankful it wasn't a gigantic PR with the new implementation AND cutting everything over.

app/scripts/modules/core/src/api/ApiService.ts
christopherthielen: I think there should be a way to distinguish between `REST()` being the HTTP client and `$http` being the specific "implementation".  ideas?

app/scripts/modules/core/src/api/ApiService.ts
```diff
@@ -1,193 +1,254 @@
...
-      config.method = 'get';
-      Object.assign(config, this.defaultParams);
-      if (params) {
-        config.params = params;
-      }
-
-      return $http(config).then((result: any) => this.getData(result));
-    };
+  protected get baseUrl(): string {
+    return (this._baseUrl ?? SETTINGS.gateUrl).replace(/\/+$/, '');
```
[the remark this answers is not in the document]
christopherthielen: How about 
```ts
    return joinPaths(this._baseUrl ?? SETTINGS.gateUrl);
```

app/scripts/modules/core/src/api/ApiService.ts
vigneshm: May be just using the word "implementation"?`TestingHttpImplementation`, `FetchHttpImplementation` sound better to me.

app/scripts/modules/core/src/api/ApiService.ts
```diff
@@ -1,193 +1,254 @@
...
+function joinPaths(...paths: IPrimitive[]) {
+  // coerce paths toString() in case somebody sends in a url object
+  // according to https://github.com/spinnaker/deck/pull/6927
+  return paths
+    .filter((path) => !isNil(path) && path !== '')
+    .map((path) => path.toString())
+    .map((path) => path.replace(/^\/+/, '')) // strip leading slashes
+    .map((path) => path.replace(/\/+$/, '')) // strip trailing slashes
+    .join('/');
+}
```
vigneshm, line 139: A naive question, but where is the encoding happening? I was kind of expecting it to happen here.

app/scripts/modules/core/src/api/ApiService.ts
caseyhebebrand: +1 ^^

app/scripts/modules/core/src/api/ApiService.ts
christopherthielen: Good question.  It's not merged yet, but it will go in the `.path()` method

app/scripts/modules/core/src/api/ApiService.ts
christopherthielen: sounds good to me

app/scripts/modules/core/src/api/ApiService.ts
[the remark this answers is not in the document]
alanmquach: Even better

vigneshm (approved):

PR #8767, opened 2020-12-04
feat(ecs): Add basic capacity provider UI 

allisaurus:
Basic UI changes to support https://github.com/spinnaker/spinnaker/issues/5400

* NOTE: Does not (yet) look up capacity providers per cluster and provide them as a drop down. This work is anticipated as a near term enhancement.

* Related PR: https://github.com/spinnaker/clouddriver/pull/5123

### UI

* With launch type (default)
image

* With capacity provider strategies
image

### Deployment:
* Successful deployment w/ https://github.com/spinnaker/clouddriver/pull/5123 : 

image

image

### DIFF
--- a/app/scripts/modules/ecs/src/serverGroup/configure/wizard/CloneServerGroup.ecs.controller.js
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/wizard/CloneServerGroup.ecs.controller.js
@@ -241,5 +241,19 @@ module(ECS_SERVERGROUP_CONFIGURE_WIZARD_CLONESERVERGROUP_ECS_CONTROLLER, [
     $scope.configureCommand = function (query) {
       return ecsServerGroupConfigurationService.configureCommand($scope.command, query);
     };
+
+    // TODO: Migrate horizontalScaling component to react and move this logic there
+    $scope.capacityProviderState = {
+      useCapacityProviders:
+        $scope.command.capacityProviderStrategies && $scope.command.capacityProviderStrategies.length > 0,
+      updateComputeOption: function (chosenOption) {
+        if (chosenOption == 'launchType') {
+          $scope.command.capacityProviderStrategies = [];
+        } else if (chosenOption == 'capacityProviders') {
+          $scope.command.launchType = '';
+          $scope.command.capacityProviderStrategies = $scope.command.capacityProviderStrategies || [];
+        }
+      },
+    };
   },
 ]);
--- a/app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
@@ -1,5 +1,104 @@
 <div class="container-fluid form-horizontal">
   <div class="form-group">
+    <div class="col-md-4 sm-label-right">
+      <b>Compute options</b>
+      <help-field key="ecs.computeOptions"></help-field>
+    </div>
+    <div class="col-md-2 radio">
+      <label>
+        <input
+          data-test-id="ServerGroup.computeOptionsLaunchType"
+          type="radio"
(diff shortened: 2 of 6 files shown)


piradeepk:
Lgtm! Would it be possible to add some unit tests and a functional test?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/CloneServerGroup.ecs.controller.js
```diff
@@ -241,5 +241,21 @@ module(ECS_SERVERGROUP_CONFIGURE_WIZARD_CLONESERVERGROUP_ECS_CONTROLLER, [
     $scope.configureCommand = function (query) {
       return ecsServerGroupConfigurationService.configureCommand($scope.command, query);
     };
+
+    // TODO: Migrate horizontalScaling component to react and move this logic there
+    $scope.capacityProviderState = {
+      useCapacityProviders:
+        $scope.command.capacityProviderStrategies && $scope.command.capacityProviderStrategies.length > 0
```
paragbhingre, line 248: Looks great to me just a minor suggestion.
I guess the condition itself will return true or false. So we don't need to return true or false specifically.

allisaurus: you're right! i will make that change

app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
```diff
@@ -1,5 +1,86 @@
...
+    <div>
+      <div>{{ cp }}</div>
+      <table class="table table-condensed packed tags">
+        <thead>
+          <th style="width: 50%">Provider name <help-field key="ecs.capacityProviderName"></help-field></th>
+          <th style="width: 25%">Base <help-field key="ecs.capacityProviderBase"></help-field></th>
+          <th style="width: 25%">Weight <help-field key="ecs.capacityProviderWeight"></help-field></th>
+        </thead>
+        <tbody>
+          <tr ng-repeat="cp in $ctrl.command.capacityProviderStrategies">
```
paragbhingre, line 48: I guess we need to have an array of strings here as we are just getting CP names and not the strategies. Also, we need to have a separate parameter for default capacity provider strategy as well. 
Default CP strategy will be coming in the form of 
```
"defaultCapacityProviderStrategy": [ 
            { 
               "base": number,
               "capacityProvider": "string",
               "weight": number
            }
         ]
```

And CPs will be coming in the form of 
 `"capacityProviders": [ "string" ]`

Does that make sense?

allisaurus: let's be careful not to conflate the data we need to _submit to clouddriver_ in order to use capacity providers (the "command" obj) and the data we need to _configure the UI_ for capacity providers. 

right now this logic is only displaying what's present in the stage JSON and providing basic text inputs for users to directly update it (i.e., what should be present in the stage JSON). That data should match what we need to submit to clouddriver and will always look like this:

```
"capacityProviderStrategies": [
        {
          "base": "1",
          "capacityProvider": "FARGATE_SPOT",
          "weight": "1"
        }
]
```

You're right that in the near future, we want to enhance the UI and provide a drop-down for `capacityProviderStrategy.name` so users can select CPs associated with their cluster. 

The API call to retrieve that data per cluster returns a more complex object which includes info like `defaultCapacityProviderStrategy`, as you said. But that data will be loaded by the controller and formatted into a list of strings which will populate the drop-down, it will not be added to the stage JSON directly. The drop-down will only update the `capacityProviderStrategy.name` in the stage JSON with the selected value when a user interacts with it and selects something. Let me know if that doesn't clarify things!

piradeepk (approved):
Ship it! 💯  This is awesome @allisaurus

PR #8768, opened 2020-12-04
feat(core/managed): add pre-deployment events for builds

erikmunson:
We recently added support for ['pre-deployment' events](https://github.com/spinnaker/deck/pull/8750), starting with baking. We're now ready to add the same for builds. A couple notable bits that have evolved since the last PR:

- In order to handle the cases that can show up in the build context there's a new status: `ABORTED`. Unfortunately our internal build system (Jenkins) doesn't seem to differentiate between builds that entered this status because of a timeout (bad/error) or because a user manually canceled it (fine/not an error), so we have to be somewhat hand-wavey about this status and our use of colors. For now I've made the decision to show it the same way we show successful builds (neutral color, no badge in the sidebar) because I have a hunch the manual cancellation use case is way more common than timeouts, but either way we're going to be trading something off (i.e. if we treated it as an error we'd be raising the alarm about intentionally canceled builds)
- I added the `UNKNOWN` status for when we can't determine current state, which I've made a warning color (no badge in the sidebar). Not really sure about this and we have zero customer feedback right now so I'm trying not to overthink it

<details>

<summary>Version sidebar</summary>

Screen Shot 2020-12-03 at 10 05 34 AM
Screen Shot 2020-12-03 at 10 09 58 AM

</details>

<details>

<summary>Version details</summary>

Screen Shot 2020-12-03 at 10 07 52 AM
Screen Shot 2020-12-03 at 10 05 19 AM
Screen Shot 2020-12-03 at 10 09 22 AM
Screen Shot 2020-12-03 at 10 10 39 AM
Screen Shot 2020-12-03 at 10 20 51 AM
Screen Shot 2020-12-03 at 10 12 41 AM
Screen Shot 2020-12-03 at 9 13 11 AM

</details>

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -244,7 +244,9 @@ export const ArtifactDetail = ({
   const isPinnedEverywhere = environments.every(({ pinned }) => pinned);
   const isBadEverywhere = environments.every(({ state }) => state === 'vetoed');
   const createdAtTimestamp = useMemo(() => createdAt && DateTime.fromISO(createdAt), [createdAt]);
-  const preDeploymentSteps = lifecycleSteps?.filter(({ scope }) => scope === 'PRE_DEPLOYMENT');
+
+  // These steps come in with chronological ordering, but we need reverse-chronological orddering for display
+  const preDeploymentSteps = lifecycleSteps?.filter(({ scope }) => scope === 'PRE_DEPLOYMENT').reverse();
 
   return (
     <>
@@ -350,7 +352,7 @@ export const ArtifactDetail = ({
               name={environmentName}
               resources={resourcesByEnvironment[environmentName]}
             >
-              <div className="sp-margin-l-right">
+              <div>
                 <EnvironmentCards
                   application={application}
                   environment={environment}
@@ -389,7 +391,7 @@ export const ArtifactDetail = ({
         })}
         {preDeploymentSteps && preDeploymentSteps.length > 0 && (
           <PreDeploymentRow>
-            {lifecycleSteps.map((step) => (
+            {preDeploymentSteps.map((step) => (
               <PreDeploymentStepCard key={step.id} step={step} application={application} reference={reference} />
             ))}
           </PreDeploymentRow>
--- a/app/scripts/modules/core/src/managed/PreDeploymentStepCard.tsx
+++ b/app/scripts/modules/core/src/managed/PreDeploymentStepCard.tsx
@@ -9,17 +9,42 @@ import { Application } from '../application';
 import { StatusCard } from './StatusCard';
 import { Button } from './Button';
 
-const SUPPORTED_TYPES = ['BAKE'];
+const SUPPORTED_TYPES = ['BUILD', 'BAKE'];
+
+const cardAppearanceByStatus = {
+  NOT_STARTED: 'future',
+  RUNNING: 'info',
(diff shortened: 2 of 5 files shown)


app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -389,7 +391,7 @@ export const ArtifactDetail = ({
         })}
         {preDeploymentSteps && preDeploymentSteps.length > 0 && (
           <PreDeploymentRow>
-            {lifecycleSteps.map((step) => (
+            {preDeploymentSteps.map((step) => (
```
erikmunson, line 394: this was an oversight on my part in the previous PR

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -350,7 +352,7 @@ export const ArtifactDetail = ({
               name={environmentName}
               resources={resourcesByEnvironment[environmentName]}
             >
-              <div className="sp-margin-l-right">
```
erikmunson, line 353: slightly tangential to this PR but wasn't worth splitting out: this margin on the right side of constraint cards was from an earlier design iteration where the cards had rounded corners, and as a result looked better with some spacing on the right. The new designs with squared off corners look better flush with the right side of the container, which is how they are supposed to look per the mockups.

app/scripts/modules/core/src/managed/PreDeploymentStepCard.tsx
```diff
@@ -70,25 +97,27 @@ export const PreDeploymentStepCard = memo(({ step, application, reference }: Pre
...
-      appearance={appearance[status]}
+      appearance={cardAppearanceByStatus[status]}
       active={status !== 'NOT_STARTED'}
       iconName={iconName}
       timestamp={getTimestamp(startedAt, completedAt)}
       title={title(step)}
       actions={
         link && (
-          <Button
+          <a
```
erikmunson, line 111: I did a bad thing on both the `See changes` button and this button — I should've used an `<a/>` but instead used `window.open` because... I'm not sure I was paying enough attention to think through it fully. Corrected both here _and_ for the 'See changes' button.

luispollo:
@erikmunson This looks awesome! The only thing I was a bit confused by was that status of "A build will run before deployment". Given that our unit of deployment is an artifact (which is the output of a build), and that we need to detect the new artifact version to even get to this screen, what's the purpose of this status? In the vast majority of cases, I would expect to see a "build succeeded" status when I get to this screen, with a "build failed" being a corner case for when the build succeeded in publishing the artifact but failed subsequently, but never "a build _will_ start". Thoughts?

erikmunson:
You're technically correct that as of this moment it's highly, highly unlikely for `NOT_STARTED` to show up in the wild. At the same time, `NOT_STARTED` is one of the valid states for this system to be in (which we don't control since it's a downstream 3rd party). Much like how all the possible statuses are handled within Keel in the corresponding API code, this code aims to handle whatever that 3rd party may throw at us in a graceful way vs. the alternative which would be throwing/showing nothing.

If it were very time intensive or taxing to add support for additional statuses I would opt to leave out support as the investment wouldn't be worth it, but since [adding support for additional statuses is effectively free](https://github.com/spinnaker/deck/pull/8768/files#diff-7be41e2e6da8b0c89769fd790d357d68baecc80ce10c6629c3aca1fcd360a3e6R31-R32) I can't currently identify a downside to handling it here.

PR #8770, opened 2020-12-04
fix(core): Redirecting aged out executions to permalink

alanmquach:
This addresses a major annoyance in sharing links to executions that are not the most recent (i.e. Show [2] executions per pipeline) results in losing the linked execution because it was not available to be selected.

### Current behavior
Instead of just navigating up (existing behavior) and dropping the executionId from the route:
1. Lands on `#/applications/myawesomeapp/executions/0425LQQ2020FHQWGHADS`
2. Execution not found in already fetched n most recent executions per pipeline
3. Transition up to parent state (executions): `#/applications/myawesomeapp/executions`

### Proposed behavior
We will instead check if the execution exists at all and if so, forward to the SingleExecutionDetails route:
1. Lands on `#/applications/myawesomeapp/executions/0425LQQ2020FHQWGHADS`
2. Execution not found in already fetched n most recent executions per pipeline
3. Ensure execution does indeed exist and we have not forwarded to it yet.
4. Transition to SingleExecutionDetails: `#/applications/myawesomeapp/executions/details/0425LQQ2020FHQWGHADS`

Just using a simple `Set<string>` to track execution IDs and ensure we only forward once to prevent an endless loop:
* Forward to `SingleExecutionDetails`
* User navigates back
* Oh hello, I do not have the droid you are looking for but you can go here instead (aka forward to `SingleExecutionDetails`)
* User navigates back
* Oh hello, I do not have the droid you are looking for but you can go here instead (aka forward to `SingleExecutionDetails`)
* User navigates back
* Oh hello, I do not have the droid you are looking for but you can go here instead (aka forward to `SingleExecutionDetails`)
* User navigates back
* Oh hello, I do not have the droid you are looking for but you can go here instead (aka forward to `SingleExecutionDetails`)
* User navigates back
* Oh hello, I do not have the droid you are looking for but you can go here instead (aka forward to `SingleExecutionDetails`)
* User navigates back
* Oh hello, I do not have the droid you are looking for but you can go here instead (aka forward to `SingleExecutionDetails`)
* You get the idea

Same intention as #8769, but less gotchas.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/executions/Executions.tsx
+++ b/app/scripts/modules/core/src/pipeline/executions/Executions.tsx
@@ -39,6 +39,9 @@ export interface IExecutionsState {
   reloadingForFilters: boolean;
 }
 
+// This Set ensures we only forward once from .executions to .executionDetails for an aged out execution
+const forwardedExecutions = new Set();
+
 export class Executions extends React.Component<IExecutionsProps, IExecutionsState> {
   private executionsRefreshUnsubscribe: Function;
   private groupsUpdatedSubscription: Subscription;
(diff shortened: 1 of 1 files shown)


vigneshm (approved):
Neat fix!

christopherthielen (approved):
love it

PR #8775, opened 2020-12-08
feat(core/api): Introduce a MockHttpClient.ts as a incremental replacement for AngularJS $httpBackend for unit tests.

christopherthielen:
This provides an API similar to `$httpBackend`:
```
const http = mockHttpClient();
http.expectGET('/rest/path/widgets').respond(200, [{ name: 'widget1' }]);
const widgets = await WidgetReader.getAll();
expect(widgets).toEqual([{ name: 'widget1' }]);
```

When `mockHttpClient()` is run, the http client implementation (for both `REST` and `API`) is swapped out with an instance of the mock.

After every test (an `afterEach()` block is registered at the root level):
- it verifies that no outstanding expected requests are missing and verifies that no unexpected requests were detected.
- the original http client implementation object is restored.

This will be introduced at the unit test `it()` block level, migrating away from `$httpBackend` incrementally.


christopherthielen:
I may be iterating on the implementation details, but I'd like review sooner rather than later to validate my approach.

karma-shim.js
```diff
@@ -10,13 +10,15 @@ global.$ = global.jQuery = require('jquery');
 
 import './settings';
 import './app/scripts/app';
+import './test/helpers/customMatchers';
```
christopherthielen, line 13: this file is really finicky about import order and call order.  Reordering all imports to the top breaks a good 10% of our tests.

app/scripts/modules/core/src/slack/SlackReader.spec.ts
```diff
@@ -24,15 +24,10 @@ const mockSlackChannels: ISlackChannel[] = [
 
 describe('SlackReader', () => {
   it('should fetch a list of public Slack channels', async () => {
-    spyOn(SlackReader, 'getChannels').and.callThrough();
-    spyOn(API, 'one')
-      .and.callThrough()
-      .and.returnValue({ getList: () => Promise.resolve(mockSlackChannels) });
+    const http = mockHttpClient();
```
christopherthielen, line 27: here's a real world example

app/scripts/modules/core/src/slack/SlackReader.spec.ts
```diff
@@ -24,15 +24,10 @@ const mockSlackChannels: ISlackChannel[] = [
...
 
-    await SlackReader.getChannels().then((channels: ISlackChannel[]) => {
-      expect(SlackReader.getChannels).toHaveBeenCalled();
-      expect(channels.length).toEqual(2);
-      expect(API.one).toHaveBeenCalledWith('slack', 'channels');
-    });
+    const channels = await SlackReader.getChannels();
+    expect(channels.length).toEqual(2);
   });
 });
```
christopherthielen, line 33: Here's a real-world example
This one already was async

app/scripts/modules/core/src/api/mock/MockHttpClient.ts
```diff
@@ -0,0 +1,176 @@
...
+      return Promise.reject({ status, data });
+    }
+  }
+
+  get = <T = any>(request: IRequest): PromiseLike<T> => this.request('GET', request.url, request.params);
+  put = <T = any>(request: IRequest): PromiseLike<T> => this.request('PUT', request.url, request.params);
+  post = <T = any>(request: IRequest): PromiseLike<T> => this.request('POST', request.url, request.params);
+  delete = <T = any>(request: IRequest): PromiseLike<T> => this.request('DELETE', request.url, request.params);
+
+  verifyNoOutstanding() {
```
christopherthielen, line 132:
```suggestion
  verifyNoOutstandingExpectation() {
```
Might as well match the existing name to make migration easier /shrug

christopherthielen:
Started migrating some tests and immediately ran into some bumps:

```
    it('and updates isLoadingPackageImages', () => {
      $httpBackend.expectGET(`${baseUrl}/images/find?q=testapp*`).respond(200, []);
      shallowComponent = shallow(<AmazonImageSelectInput {...baseProps} />);
      expect(shallowComponent.state().isLoadingPackageImages).toBe(true);
      $httpBackend.flush();
      expect(shallowComponent.state().isLoadingPackageImages).toBe(false);
    });
```

This kind of test requires `flush()` capability, which enables the test to decide when the pending requests should be resolved.

I was hoping not to re-implement this aspect of the angularjs mocks, but I don't see how to write this test without it.

christopherthielen:
A second bump: eagerly throwing an exception for unexpected requests is not failing the test as expected.  The error is written to the console and the test continues to execute.  Perhaps this is something that `enzyme` is capturing?  Perhaps it's a React error boundary?

Screen Shot 2020-12-08 at 3 35 03 PM

christopherthielen:
PR Updated with the following squashed changes:

```
fix(core/api): Allow regex in .expect argument typing
fix(core/api/mock): run AngularJS digest AFTER resolving the http.flush() promise
fix(core/api/mock): Fix partial parameter matching
fix some bugs in mockHttpClient.ts and expectedRequest.ts
Add .patch() support to mock-http-client
feat(core/api): mock-http-client: Display the outstanding expectations when .flush() times out
feat(core/api): expectedRequest.ts: Ignore baseUrl in expects
refactor(core/api): Move ExpectedRequest, ReceivedRequest, and Jasmine helpers to separate files
refactor(core/api): Move mock http util functions to a separate file
fix(core/api): Add support to MockHttpClient.ts for code that interacts with AngularJS $q and $http
```

app/scripts/modules/core/src/api/mock/mockHttpUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+    deferredObj.reject = (error) => {
+      deferredObj.settled = true;
+      reject(error);
+    };
+  });
+
+  return deferredObj;
+}
+
+export function kickAngularJS() {
```
caseyhebebrand, line 32: Kind of a nit, but I think calling this `flushAngularJS` or `flushHttpBackend` is more cohesive with the naming in the MockHttpClient, where it is also called flushing.

christopherthielen: you don't like my dig on angularjs? 😆

app/scripts/modules/core/src/api/mock/expectedRequest.ts
```diff
@@ -0,0 +1,145 @@
...
+    } else {
+      paramsObj[key] = value;
+    }
+    return paramsObj;
+  }, {} as Record<string, string | string[]>);
+}
+
+/**
+ * This internal class represents an HTTP request that is expected to be made by the code under test.
+ * Instances are created by MockHttpClient.expect -- this class should not be used directly.
```
robzienert, line 24: This is my own naiveté speaking here, but the class is exported, so is it actually an internal class at that point?

app/scripts/modules/core/src/api/mock/expectedRequest.ts
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * This class exposes a promise (this.fulfilledDeferred.promise).
+ * The promise will be resolved when the code under test makes the expected HTTP request.
+ *
+ * ExpectedRequest also stores the (optional) mock response status/data to return to the code under test.
+ */
+export class ExpectedRequest implements IExpectBuilder {
```
robzienert, line 31: I would have expected this class to also support a notion of a non-response from the server. Does Deck make the assumption that the backend is always accessible and returning responses of some kind (as opposed to just timing out)? Do we ever assert the behavior of Deck when the backend services are simply unavailable?

robzienert: ... I suppose a timeout in browser land is exposed as a 408? Maybe there's nothing to do on this use case.

app/scripts/modules/core/src/api/mock/mockHttpUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+  try {
+    $httpBackend.flush();
+    // eslint-disable-next-line no-empty
+  } catch (ignore) {}
+}
+
+export const tick = (ms?: number) => new Promise((resolve) => setTimeout(resolve, ms));
+export const isSuccessStatus = (status: number) => status >= 200 && status <= 299;
```
robzienert, line 40: What about 3xx? Do we generally expect the actual HTTP client implementations to automatically follow redirects? Presumably, we don't make use of 304 (Not Modified)?

app/scripts/modules/core/src/slack/SlackReader.spec.ts
```diff
@@ -24,15 +24,10 @@ const mockSlackChannels: ISlackChannel[] = [
...
```
robzienert, line 33: I quite like the improvements made to the developer experience as a result of this refactor - much more intuitive interaction, as a layperson.

robzienert (commented):
Mostly just comments of ignorance, I like it, though!

app/scripts/modules/core/src/api/mock/expectedRequest.ts
christopherthielen: good catch

app/scripts/modules/core/src/api/mock/expectedRequest.ts
christopherthielen: that's a great question.  deck doesn't handle timeouts at this layer and thus we don't provide testing capability.  There is an interceptor in the AngularJS $http client that will retry timed out requests.  If retries are exhausted, the request is rejected (no http codes involved).  When we move from $http to something else (fetch?) we will likely re-create this capability and write tests at that layer.

app/scripts/modules/core/src/api/mock/mockHttpUtils.ts
christopherthielen: Great question, TBH I'm not sure about how all the codes are handled, but I believe that the most common 302 is handled natively by the browser.   We also don't do any caching for API calls -- deck never handles 304 responses directly, but the browser does natively for JS/etc assets.

app/scripts/modules/core/src/api/mock/expectedRequest.ts
christopherthielen: Javascript/Typescript doesn't have the concept of package protected, but this is what was meant by "internal"

PR #8776, opened 2020-12-09
fix(core/managed): temporarily hide build pre-deployment events

erikmunson:
The API data for build events introduced via #8768 has proved to be a bit wonky in the first iteration, and is going to take a little more implementation time on the Keel side to resolve. Events get stuck in the `RUNNING` status — which looks pretty sketchy/untrustworthy in the Environments UI — so I'm going to temporarily disable them here until we're confident the data will look right.

### DIFF
--- a/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/ArtifactDetail.tsx
@@ -34,6 +34,12 @@ import { isResourceKindSupported } from './resources/resourceRegistry';
 
 import './ArtifactDetail.less';
 
+const SUPPORTED_PRE_DEPLOYMENT_TYPES = [
+  // KLUDGE WARNING: disabling build events temporarily while we get the API in shape
+  // 'BUILD',
+  'BAKE',
+];
+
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -34,6 +34,12 @@ import { isResourceKindSupported } from './resources/resourceRegistry';
 
 import './ArtifactDetail.less';
 
+const SUPPORTED_PRE_DEPLOYMENT_TYPES = [
```
erikmunson, line 37: i moved this from `PreDeploymentStepCard.tsx` because it fixes another bug which was a non-issue before this change, but became one after. when there's unsupported types in the list of steps, we were unable to know from this file whether or not all the steps taken in by `PreDeploymentStepCard` would end up rendering anything. That meant that in versions with _only_ a `BUILD` event, we'd still show the "PRE-DEPLOYMENT" section but the card component wouldn't render anything underneath.

With the filtering on L256, we'll avoid showing an empty section.

PR #8778, opened 2020-12-10
fix(core/pipeline): Fix linking to pipeline execution

vigneshm:
Wrapping the link with a `span` tag results in the following html and breaks the link.

image

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/monitorPipeline/MonitorPipelineStageExecutionDetails.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/monitorPipeline/MonitorPipelineStageExecutionDetails.tsx
@@ -47,7 +47,7 @@ export function MonitorPipelineStageExecutionDetails(props: IExecutionDetailsSec
           {statuses.map((status) => {
             return (
               <div key={status.executionId}>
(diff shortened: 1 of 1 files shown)


christopherthielen (approved):

app/scripts/modules/core/src/pipeline/config/stages/monitorPipeline/MonitorPipelineStageExecutionDetails.tsx
```diff
@@ -58,9 +58,7 @@ export function MonitorPipelineStageExecutionDetails(props: IExecutionDetailsSec
                   }}
                   options={{ inherit: false, reload: 'home.applications.application.pipelines.executionDetails' }}
                 >
-                  <span>
-                    &nbsp;<a target="_self">{status.executionId}</a>
```
christopherthielen, line 62: FYI

> In cases like the following, links are not rewritten; instead, the browser will perform a full page reload to the original link.
>
>Links that contain target element
>Example: <a href="/ext/link?a=b" target="_self">link</a>

from https://docs.angularjs.org/guide/$location#html-link-rewriting

since this isn't angularjs, it doesn't apply, but this is likely why the `_self` was in there.

PR #8779, opened 2020-12-10
feat(manual judgment/deck): Added ability to add roles to manual judgment stage

sanopsmx:
This is part of: spinnaker/spinnaker#4792.

Enhanced stage.html to
Get the roles of the application. Display the roles of the application as a list, only if the stage is Manual Judgment.

Enhanced stage.module.js to
Fetch the permissions of the application from the gate application url. populate the list with only the roles of the application with no duplicates.

Fetch the roles of the application, stage and the user during the execution. Iterate through each of the user role to check if the role exists in the stage and application. If the role exists in the stage roles added in the manual judgment stage, then we check for role permissions added in the application. If the stage role has 'READ', then we disable the continue button If the stage role has 'WRITE,EXECUTE,CREATE', then we enable the continue button.

Enhanced help.contents.ts file
The current user must have the permission to approve the manual judgment stage. Otherwise, you'll not be able continue to the next pipeline stage.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
+++ b/app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
@@ -5,6 +5,8 @@ import { IExecution, IExecutionStage } from 'core/domain';
 import { Application } from 'core/application/application.model';
 import { Markdown } from 'core/presentation/Markdown';
 import { NgReact, ReactInjector } from 'core/reactShims';
+import { ApplicationReader } from 'core/application/service/ApplicationReader';
+import { AuthenticationService } from 'core/authentication';
 
 export interface IManualJudgmentApprovalProps {
   execution: IExecution;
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
```diff
@@ -29,17 +33,66 @@ export class ManualJudgmentApproval extends React.Component<
...
     ReactInjector.manualJudgmentService.provideJudgment(application, execution, stage, judgmentDecision, judgmentInput);
   }
 
+  private isManualJudgmentStageNotAuthorized(): boolean {
+      let isStageNotAuthorized = true;
+      const {
+          applicationRoles,
+          userRoles
+      } = this.state;
+      const stageRoles = this.props.stage.context.selectedStageRoles || [];
```
caseyhebebrand, line 69: Optional chaining since its nested to avoid an error being thrown
```suggestion
      const stageRoles = this.props.stage?.context?.selectedStageRoles || [];
```

caseyhebebrand (approved):
LGTM - left one minor comment

app/scripts/modules/core/src/pipeline/config/stages/manualJudgment/ManualJudgmentApproval.tsx
sanopsmx: done..

sanopsmx:
@caseyhebebrand ..Can you merge this PR.

PR #8782, opened 2020-12-11
chore(core): Allow cloud provider logos in registry

caseyhebebrand:
An `svg` used in for cloud provider logos can now be configured while registering the cloud provider via `cloudProviderLogo`. The current implementation uses css to put an image mask over a span. For plugin-based cloud providers, the plugin will load the svg when it registers the plugin on initilization.


christopherthielen (approved):
LGTM

If you have the time and the desire, this would be much simpler as a functional component.

PR #8797, opened 2020-12-15
feat(aws): Add scheme to load balancer details

alanmquach:
Just exposing a load balancer's scheme in the details pane:

Screen Shot 2020-12-14 at 4 57 03 PM

(options are `internal` or `internet-facing`)


christopherthielen (approved):
Does titus need this too?

alanmquach:
There is no "titus load balancer" and I was never here. 😄

PR #8802, opened 2020-12-16
fix(kubernetes): Fix Details link on pipeline deploy status for global resources

jeffherald:
This PR fixes an issue where the URL is malformed for the details link for global resources such as ClusterRoles and ClusterRoleBindings. This is because a blank instead of passing a `_` for the namespace parameter which the controller expects. An example of a malformed URL:
http://localhost:9000/#/applications/demo-app/clusters/manifest/kubernetes/demo-account//clusterrolebinding%20istio-crb
And a working URL:
http://localhost:9000/#/applications/demo-app/clusters/manifest/kubernetes/demo-account/_/clusterrolebinding%20istio-crb

image


christopherthielen (approved):
recommend adding some comments explaining how these edge cases can occur and what they mean.

This region fixup could also be encapsulated into a function

PR #8808, opened 2020-12-17
feat(titus): Support SpEL for Resources number inputs

alanmquach:
Just like we do for `Capacity`:

image

I wonder if we'll need to do something to prevent the SpEL toggle when thess fields are used outside of a pipeline context, i.e. server group clone modal.

Either way, probably shouldn't hold out on adding this just to figure that out.


vigneshm (approved):

christopherthielen:
> I wonder if we'll need to do something to prevent the SpEL toggle when thess fields are used outside of a pipeline context, i.e. server group clone modal.


There is a spel React context provider that would be handy for this, but the form would have to be converted to FormikFormField

PR #8810, opened 2020-12-18
feat(ecs): Add support for Capacity providers -  enhanced Deck UI

paragbhingre:
This PR adds a new enhanced version of UI to support Capacity Providers.

Currently, Capacity Provider UI section provides input fields to enter details about Capacity Provider Strategy for the server group. 
But new UI will have a drop down list for selecting a Capacity Provider. Plus now users can add default Capacity providers without typing in the details of default Capacity Provider by just clicking on the **use cluster default** option. Please refer the following screenshots for the more details. 

New deck UI 1
New deck UI 2

Addresses [spinnaker/spinnaker#6242](https://github.com/spinnaker/spinnaker/issues/6242)

### DIFF
--- a/app/scripts/modules/ecs/src/ecsCluster/IEcsDescribeCluster.ts
+++ b/app/scripts/modules/ecs/src/ecsCluster/IEcsDescribeCluster.ts
@@ -1,0 +1,58 @@
+export interface IEcsDescribeCluster {
+  activeServicesCount : number,
+  attachments : IDescribeClusterAttachments,
+  attachmentsStatus : string,
+  capacityProviders : string[],
+  clusterArn : string,
+  clusterName : string,
+  defaultCapacityProviderStrategy : IDescribeClusterDefaultCapacityProviderStrategy[],
+  pendingTasksCount : number,
+  registeredContainerInstancesCount : number,
+  runningTasksCount : number,
+  settings : IDescribeClusterSettings[],
+  statistics : IDescribeClusterStatistics[],
+  status : string,
+  tags : IDescribeClusterTags[],
+  failures: IDescribeClusterFailures[]
+
+
+}
+
+export interface IDescribeClusterAttachments {
+  details : IDescribeClusterAttachmentDetails[],
+  id : string,
+  Status : string,
+  type : string
+}
+
+export interface IDescribeClusterAttachmentDetails {
+  name : string,
+  value : string
+}
+
+export interface IDescribeClusterDefaultCapacityProviderStrategy {
+  base : number,
+  capacityProvider : string,
+  weight : number
+}
+
+export interface IDescribeClusterSettings {
+  name : string,
+  value : string
+}
+
+export interface IDescribeClusterStatistics {
+  name : string,
+  value : string
+}
+
+export interface IDescribeClusterTags {
+  key : string,
+  value : string
+}
+
+export interface IDescribeClusterFailures {
+  arn : string,
+  detail : string,
+  reason : string
+}
--- a/app/scripts/modules/ecs/src/serverGroup/configure/wizard/CloneServerGroup.ecs.controller.js
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/wizard/CloneServerGroup.ecs.controller.js
@@ -164,6 +164,7 @@ module(ECS_SERVERGROUP_CONFIGURE_WIZARD_CLONESERVERGROUP_ECS_CONTROLLER, [
         initializeCommand();
         initializeSelectOptions();
         initializeWatches();
+        changeCluster();
       });
     }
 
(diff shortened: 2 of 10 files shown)


app/scripts/modules/ecs/src/serverGroup/configure/wizard/CloneServerGroup.ecs.controller.js
```diff
@@ -184,6 +185,13 @@ module(ECS_SERVERGROUP_CONFIGURE_WIZARD_CLONESERVERGROUP_ECS_CONTROLLER, [
       processCommandUpdateResult($scope.command.regionChanged(serverGroupCommand));
     }
 
+    function changeCluster(){
```
piradeepk, line 188: This controller doesn't actually load the UI. It just copies the properties of the previous server group to the new one. You shouldn't need to change any logic in here.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
```diff
@@ -38,63 +38,102 @@
       <help-field key="ecs.capacityProviderStrategy"></help-field>
     </div>
     <div>
+      <div class="radio">
```
piradeepk, line 41: I thought we discussed moving these components to using React rather than keeping them as Angular components @allisaurus correct me if I'm mistaken

piradeepk (changes requested):
I'm a bit confused as to whether we need all of the data being returned by the describe clusters api, if we're only interested in supporting capacity providers, that api returning everything seems a bit heavy for what we're trying to do.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
piradeepk: @paragbhingre you can use https://github.com/spinnaker/deck/pull/8699 as reference (migrated the component to react in one pr, then implement your feature on top in another PR)

paragbhingre: I had already started working on deck (using angular) when we last decided to push the code before the release. I had a conversation with @allisaurus and she mentioned that first of all we need to change the existing code into react and then on top of it we can make our changes for CP which was too much work before the release. And after release, I had almost completed my work in angular just left with the functional tests. That is why this code is in angular and not in react. But if we think that can be our priority in this release then I am happy to do that.

allisaurus:
Since this addresses https://github.com/spinnaker/spinnaker/issues/6242, please mention it at the top of the PR description

app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
allisaurus: My opinion was/is that migrating the entire `horizontalScaling` component to react was beyond the scope of this PR, but that the _dedicated capacity provider inputs_ should be housed in their own react component which `horizontalScaling` component could then consume. In that case, the contents of the `<div>` on line 35 should look something like:

```
<div class="form-group" ng-if="$ctrl.capacityProviderState.useCapacityProviders">
    <ecs-capacity-provider-selector
        command="$ctrl.command"
        capacity-provider-state="$ctrl.capacityProviderState">
    </ecs-capacity-provider-selector>
</div>
```  
...with the entirety of the selection/data population logic in the new react component.

We can still take that approach, however since we're a ways out from a release boundary now I'm also cool with migrating `horizontalScaling` to react (separate PR) and then rebasing this capacity provider enhancement on top of it. I'm fine with whichever is easiest given the changes needed.

piradeepk: Makes sense @allisaurus. I'm alright with leaving the migration of the whole component to a potential task after the fact (as that's a substantial amount of work). That being said, I agreed with the above comment that any new components/functionality that we add should be using react rather than angular (as that's the approach that we're moving towards)

app/scripts/modules/ecs/src/ecsCluster/IEcsDescribeCluster.ts
```diff
@@ -0,0 +1,58 @@
...
+  capacityProvider : string,
+  weight : number
+}
+
+export interface IDescribeClusterSettings {
+  name : string,
+  value : string
+}
+
+export interface IDescribeClusterStatistics {
```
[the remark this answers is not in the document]
paragbhingre: Hey @piradeepk , do you want me to keep names plural or singular? I am little confused with naming convention here. Here I could see the suggestion is IEcsClusterStatistics(plural) and in the very next comment the name is IEcsTag(singular) and both are the same kinds of objects. 
My suggestion is to assign IEcsClusterStatistic[] to statistics and create an interface for IEcsClusterStatistic as mentioned below - 
` statistics : IEcsClusterStatistic[],`

`export interface IEcsClusterStatistic {`

Please let me know your thoughts.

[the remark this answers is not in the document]
piradeepk: It's not a question of making everything plural or singular, you should try to map the values to what the description object is https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeClusters.html#API_DescribeClusters_ResponseElements

piradeepk:
Can you close the other PR and update this PR? We lose continuity of the comments and it's additional work for the reviewer.

paragbhingre:
Closing this PR in reference to #8858

PR #8813, opened 2020-12-21
feat(aws): Automatically enable IPv6 in test environments when IPv6 f…

caseyhebebrand:
When IPv6 as a feature is enabled, automatically opt server groups into this in the test environment. This is preparation for the future world being mainly reliant on IPv6.


christopherthielen:
@caseyhebebrand Is there any potential that this could cause problems for companies that don't need this enabled, or is it low risk?

caseyhebebrand:
> @caseyhebebrand Is there any potential that this could cause problems for companies that don't need this enabled, or is it low risk?

Great point. I think it is low risk since ipv6 is the future, but I should just remove the risk and move it temporary config setting.

alanmquach (approved):

PR #8816, opened 2020-12-21
fix(amazon): Security group names for standalone apps render incorrectly

caseyhebebrand:
For a standalone security group, the app's name is the security group name. When cloning a standalone security group, the previewer show `app-stack-detail-newstack-detail` instead of `app-newstack-newdetail`. This causes some confusion for users who need to clone their security group.


vigneshm (approved):

christopherthielen:
Weird but ok 😅

christopherthielen (approved):

PR #8818, opened 2020-12-30
feat(core/ManualJudgment): Enhanced ManualJudgment

siddhu-opsmx:
### Goals and Non-Goals

Add a filter for filtering pipelines waiting on manual judgment. Checking this filter will automatically select the running stage filter and disable it(since all manual judgment pipelines are running). On deselecting the manual judgment filter the running filter will be enabled but still selected.

There will be a visual indication on all the stages that are waiting on manual judgment and clicking on the stage will take you to the root pipeline waiting on manual judgment.

## Design

1. Enhanced ExecutionMarker.tsx  to
     -> If the leaf node of a pipeline execution waiting for judgment exist then reusing the UiRef component for redirection 
          to leaf node waiting for manual judgment.
     -> If the stage durations setting is disabled, show a waiting icon.

2. Enhanced executionMarker.less to
     -> Added a visual indicator to the stages waiting on manual judgment.


siddhu-opsmx:
@caseyhebebrand @vigneshm @christopherthielen Could you please review the code.

app/scripts/modules/core/src/filterModel/FilterModelService.ts
```diff
@@ -219,6 +219,7 @@ export class FilterModelService {
             clear() {
               // do not reuse the modelVal variable - it's possible it has been reassigned since the tag was created
               const toClearFrom: ITrueKeyModel = model.sortFilter[key] as ITrueKeyModel;
+              value === 'RUNNING' ? delete model.sortFilter.stages.AWAITING_JUDGMENT : false;
```
caseyhebebrand, line 222: Usually the tertiary is used for variable assignment. In this case, the `false` conditional is a noop.

```suggestion
if (value === 'RUNNING') {
   delete model.sortFilter.stages.AWAITING_JUDGMENT
}             
```

app/scripts/modules/core/src/pipeline/executions/execution/Execution.tsx
```diff
@@ -45,6 +45,7 @@ export interface IExecutionProps {
   cancelHelpText?: string;
   cancelConfirmationText?: string;
   scrollIntoView?: boolean; // should really only be set to ensure scrolling on initial page load deep link
+  manualJudgment: any;
```
caseyhebebrand, line 48: Can we make this type more explicit?

app/scripts/modules/core/src/pipeline/executions/execution/Execution.tsx
```diff
@@ -310,6 +311,11 @@ export class Execution extends React.PureComponent<IExecutionProps, IExecutionSt
         key={stage.refId}
         application={application}
         execution={execution}
+        manualJudgment={
+          this.props.manualJudgment !== undefined && this.props.manualJudgment[this.props.execution.id]
+            ? this.props.manualJudgment
+            : []
+        }
```
caseyhebebrand, line 318: I would pull out manual judgement on line 290-299. Execution is already defined so you don't have to repeat `this.props`.

app/scripts/modules/core/src/projects/dashboard/pipeline/ProjectPipeline.tsx
```diff
@@ -53,6 +53,7 @@ export class ProjectPipeline extends React.Component<IProjectPipelineProps, IPro
     const execution = this.props.execution;
     const stages = execution.stageSummaries.map((stage) => (
       <ExecutionMarker
+        manualJudgment={[]}
```
caseyhebebrand, line 56: Why is this defaulted to an empty array here? Would it make more sense to make this an optional prop or default it to an empty array within the component?

app/scripts/modules/core/src/pipeline/executions/execution/executionMarker.less
```diff
@@ -42,3 +42,11 @@
     position: relative;
   }
 }
+
+.execution-marker-waiting {
+  color: var(--color-black);
+  background-color: var(--color-alert) !important;
+  .fa {
```
caseyhebebrand, line 49: Let's use a more detailed classname here. `fa` is a generic class for our icons, and I wouldn't want this to unintentionally affect other icons.

app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
```diff
@@ -81,15 +84,65 @@ export class ExecutionMarker extends React.Component<IExecutionMarkerProps, IExe
...
+  private fetchLeafNodeParameter = (type: string, leafNodeObject: any) => {
+    let leafNodeVal = '';
+    if (!isEmpty(leafNodeObject)) {
+      if (type === 'application') {
+        leafNodeVal = leafNodeObject.app ?? this.props.application.name;
+      } else {
+        leafNodeVal = leafNodeObject.id;
+      }
+    }
+    return leafNodeVal;
```
caseyhebebrand, line 132:
```suggestion
  const appName = leafNodeObject?.app ?? this.props.application.name;
  return isEmpty(leafNodeObject) ? '' : type === 'application' ? appName : leafNodeObject.id; 
```

app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
```diff
@@ -81,15 +84,65 @@ export class ExecutionMarker extends React.Component<IExecutionMarkerProps, IExe
...
 
+  private manualJudgmentStatus = (stageStatus: string, manualJudgment: any) => {
+    let status = stageStatus;
+    if (manualJudgment !== undefined && manualJudgment.length && stageStatus === 'running') {
+      const existStages = this.props.stage.stages.filter((stage) => stage.status.toLowerCase() === 'running');
+      existStages.forEach(({ context, type }) => {
+        if (this.leafNodeExist(context.executionId, manualJudgment) && type === 'pipeline') {
+          status = 'waiting';
+        }
+      });
```
caseyhebebrand, line 95: You can probably use filter for all of these conditionals since you are a just seeing if any stages match these criteria.
`status === 'running' && type === 'pipeline' && this.leafNodeExist(stage.context.executionId, manualJudgement)`

app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
```diff
@@ -81,15 +84,65 @@ export class ExecutionMarker extends React.Component<IExecutionMarkerProps, IExe
...
+    }
+    return leafNodeVal;
+  };
+
   public render() {
-    const { stage, application, execution, active, previousStageActive, width } = this.props;
+    const { stage, application, execution, active, previousStageActive, width, manualJudgment } = this.props;
     const stageType = (stage.activeStageType || stage.type).toLowerCase(); // support groups
+    const PIPELINE_WAITING =
+      this.manualJudgmentStatus(stage.status.toLowerCase(), manualJudgment[execution.id]) === 'waiting';
```
caseyhebebrand, line 139: Since you call this function a couple times, lets extract into a const.
```
const manualJudgmentStatus = this.manualJudgmentStatus(stage.status.toLowerCase(), manualJudgment[execution.id])
```

app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
```diff
@@ -81,15 +84,65 @@ export class ExecutionMarker extends React.Component<IExecutionMarkerProps, IExe
...
     const markerClassName = [
       stage.type !== 'group' ? 'clickable' : '',
       'stage',
       'execution-marker',
       `stage-type-${stageType}`,
-      `execution-marker-${stage.status.toLowerCase()}`,
+      `execution-marker-${this.manualJudgmentStatus(stage.status.toLowerCase(), manualJudgment[execution.id])}`,
```
caseyhebebrand, line 145: See last comment
```suggestion
      `execution-marker-${manualJudgmentStatus}`,
```

caseyhebebrand (commented):
I left some inline comments, but a couple more general observations that would be helpful:

- Defining a more explicit `type` for manualJudgment. Using `any` is too vague, especially for an entity that has so many helper operations
- Some of the helper functions are pretty complex (`crossAppDataFetch`, `nestedManualJudgment`). I would encourage you to either add supplementary comments, break these into smaller functions,  or see if you can reduce any repeated logic that exists between components.

app/scripts/modules/core/src/pipeline/executions/execution/executionMarker.less
siddhu-opsmx: Since .fa is already scoped inside .execution-marker-waiting, So it will not affect any other icons.

christopherthielen: I agree  with Casey, let's change this from `fa` as that classname already has a very specific meaning in Deck and is globally scoped

siddhu-opsmx: As suggested we have added custom classes to increase the styles specificity.

app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
```diff
@@ -1,5 +1,7 @@
+/*eslint-disable*/
```
caseyhebebrand, line 1: Let's remove this

app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
```diff
@@ -81,15 +84,26 @@ export class ExecutionMarker extends React.Component<IExecutionMarkerProps, IExe
...
+      );
+      if (!isEmpty(temp)) return 'waiting';
+    }
+    return stageStatus;
+  };
+
   public render() {
     const { stage, application, execution, active, previousStageActive, width } = this.props;
     const stageType = (stage.activeStageType || stage.type).toLowerCase(); // support groups
+    const PIPELINE_STATUS = this.stageStatus(stage.status.toLowerCase());
```
caseyhebebrand, line 100: nit: Call this `pipelineStatus` since it isn't a constant. Usually we reserve all caps for constants.

siddhu-opsmx: changed PIPELINE_STATUS to pipelineStatus

caseyhebebrand (approved):
Thanks for responding to all the feedback!

PR #8821, opened 2021-01-03
fix(core/search): Exclude duplicate results in search views

vjda:
Include a filter to get unique results w/ same href in `infrastructureSearchV2.service.ts` while building the result set. This ensures that only one result by href is shown in [GlobalSearchResults.tsx](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/search/global/GlobalSearchResults.tsx) and [SearchResultPods.tsx](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/search/infrastructure/SearchResultPods.tsx)

**Global Search**
global-search-results-fixed

**Infrastructure Search**
infrastructure-search-results-fixed

Solves [spinnaker/spinnaker#6261](https://github.com/spinnaker/spinnaker/issues/6261)


christopherthielen (approved):
I'm not sure why we have duplicates in the first place, but this seems 👍

PR #8822, opened 2021-01-03
fix(core): remove save pipeline error message if success

KeisukeYamashita:
## What

There was a problem where an error would be displayed once, and even if the save was successful a second time or later, the error would still be displayed. If the error message kept appearing, it would mislead the user into thinking that the save was successful but there was still an error.

Fixed the problem so that the error message disappears after a successful save. See the animation (gif) below for example: 

| Before (error message remains) | After (Error message disappears if save success) |
|:----:|:----:|
| deck-before | deck-after  |


KeisukeYamashita:
@christopherthielen @caseyhebebrand   Please take a look, this PR improved the UX on pipeline saves. Thanks in advance.

caseyhebebrand (approved):
LGTM

PR #8823, opened 2021-01-04
feat(cloudfoundry): add create service binding stage

zachsmith1:
This is the completed deck work for the new cf stage that will bind applications to services. This functionality allows users to optionally provide a parameter artifact (valid json) with the service binding request. We've also added a few edge cases where users can pick between restarting an application vs restaging after the binding takes place. We also added the ability to override existing bindings or continue on if they already exist.


app/scripts/modules/cloudfoundry/src/pipeline/stages/createServiceBindings/CloudFoundryCreateServiceBindingsConfig.tsx
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+  return formValidator.validateForm();
+}
+
+export function validateServiceBindingRequests(serviceBindingRequests: ServiceBindingRequests[]): boolean {
+  if (serviceBindingRequests?.length < 1) {
+    return false;
+  }
+  for (let i = 0; i < serviceBindingRequests.length; i++) {
```
rvazquezglez, line 67: I feel this could be simplified with ["some"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) or ["every"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) like:

```
return serviceBindingRequests.every((req) => req.serviceInstanceName);
```

zachsmith1: I tried this and it didn't capture when the fields were empty ("") then false. If you want go ahead and give it a shot. I'm always open to simplifying the readability

zachsmith1: my mistake, your solution also worked. Swapped & Merged :)

kevinawoo: it should work, just ran these in chrome console
```js
> [].every(x=>x)
true

> [""].every(x => x)
false

> [
	true,
	1,
	{},
].every(x => x)
true
```

kevinawoo (approved):
LGTM

PR #8824, opened 2021-01-04
feat(ecs): migrate capacity provider wizard to react

paragbhingre:
This PR migrates the capacity providers section of the create server group wizard for the Amazon ECS provider from angular to react.

Screen Shot 2021-01-04 at 1 23 54 PM

Screen Shot 2021-01-04 at 1 22 49 PM

Screen Shot 2021-01-04 at 1 23 20 PM


paragbhingre:
Please review CC @allisaurus @piradeepk

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+  command: IEcsServerGroupCommand;
+  notifyAngular: (key: string, value: any) => void;
+  configureCommand: (query: string) => PromiseLike<void>;
+}
+
+interface ICapacityProviderState {
+  capacityProviderStrategy: IEcsCapacityProviderStrategy[],
+}
+
+class CapacityProvider extends React.Component<ICapacityProviderProps, ICapacityProviderState>{
```
allisaurus, line 17: can we preface this component name with "Ecs", since we're exporting it into the context of the larger app? (see [this convo](https://github.com/spinnaker/deck/pull/8699#pullrequestreview-521950184))

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -113,6 +113,12 @@ export interface IEcsServiceDiscoveryRegistryAssociation {
   containerName: string;
 }
 
+export interface IEcsCapacityProviderStrategy {
```
allisaurus, line 116: since the "strategy" is the list of providers & their respective values, shouldn't we call this [`IEcsCapacityProviderStrategyItem`](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CapacityProviderStrategyItem.html) ?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -0,0 +1,154 @@
+import React from 'react';
+import { module } from 'angular';
+import { react2angular } from 'react2angular';
+import {IEcsCapacityProviderStrategy, IEcsServerGroupCommand} from '../../serverGroupConfiguration.service';
+import { HelpField, withErrorBoundary } from  '@spinnaker/core';
+
+export interface ICapacityProviderProps {
+  command: IEcsServerGroupCommand;
+  notifyAngular: (key: string, value: any) => void;
+  configureCommand: (query: string) => PromiseLike<void>;
```
allisaurus, line 10: this func doesn't seem to be used in this component?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -0,0 +1,154 @@
...
+  private removeCapacityProviderStrategy = (index: number) => {
+    const capacityProviderStrategy = this.state.capacityProviderStrategy;
+    capacityProviderStrategy.splice(index, 1);
+    this.setState({capacityProviderStrategy : capacityProviderStrategy });
+  }
+
+  private updateCapacityProviderName = (index: number, targetCapacityProviderName: string) => {
+    const currentCapacityProviderStartegy = this.state.capacityProviderStrategy;
+    const targetCapacityProviderStrategy = currentCapacityProviderStartegy[index];
+    targetCapacityProviderStrategy.capacityProvider = targetCapacityProviderName;
```
allisaurus, line 42: did you test adding a new capacity provider strat item with this UI and then deploying it? Any mutation to `capacityProviderStrategy` needs to be saved in the parent (angular) command as well w/ `notifyAngular()` (see [here](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/ecs/src/serverGroup/configure/wizard/taskDefinition/TaskDefinition.tsx#L145)) or else you're only updating the state of the _component_ data, but not the top-level command object (which gets persisted to the pipeline)

allisaurus (changes requested):
Thanks for migrating this! I have a couple requested changes, plus an ask that you provide more 
test output that shows a new capacity provider item being added via the UI (vs. just existing in the pipeline already) and successfully deployed on the new service, since it looks like this won't work ATM. Easiest way would be to remove one of the FARGATE ones, deploy, and then try adding it back w/ the new UI.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/CloneServerGroup.ecs.controller.js
```diff
@@ -81,6 +81,10 @@ module(ECS_SERVERGROUP_CONFIGURE_WIZARD_CLONESERVERGROUP_ECS_CONTROLLER, [
         'ecs.serverGroup.serviceDiscovery',
         require('./serviceDiscovery/serviceDiscovery.html'),
       ),
+      capacityProvider: overrideRegistry.getTemplate(
```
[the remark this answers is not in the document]
paragbhingre: Yes, make sense. I will remove this from the `horizontalScaling` page.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
paragbhingre: Understood the context from the conversation, thanks for the convo link. I will change it to preface with "Ecs".

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
paragbhingre: Resolved.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
paragbhingre: Actually we will need this func as a part of the new enhanced UI, so I thought we will need it anyway. But it makes sense to remove it as it is not a part of this implementation. We should be migrating things as is. I will remove that func.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
paragbhingre: ummm, that was a great catch. As a part of testing I added CP strat item to the pipeline via UI and later changed the code without  notifyAngular() and it was taking CP details from the pipeline. But now I understood the flow between angular and react so I will make the changes accordingly. Thanks for this catch @allisaurus 👍

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
allisaurus: @paragbhingre I saw that this change was made earlier, but it seems to have been reverted. Can you add that back and sync from /master ahead of a final review?

allisaurus (approved):

PR #8828, opened 2021-01-06
fix(amazon/loadBalancer): Fix target group health interface

christopherthielen:
Fix target group health interface to match the shape returned from the back end.
This surfaced a bug in passing instance counts for target groups (in the load balancer icon popover in a server group)

Missing health counts 👇 
Screen Shot 2021-01-05 at 3 36 19 PM


christopherthielen:
@caseyhebebrand for some reason I can't add you as a reviewer 🤔

caseyhebebrand (approved):

vigneshm (approved):

PR #8829, opened 2021-01-06
refactor(presentation): Create a new npm package for core presentation modules in deck

vigneshm:
Kickstarting a new package for presentation modules in deck. This should contain many of the core re-usable UI components (and their dependencies) currently available under `@spinnaker/core`. The goal of this effort is to enable plugin developers build their UI experience with minimal dependence on `@spinnaker/core`.


packages/presentation/src/utils/noop.ts
```diff
@@ -0,0 +1,3 @@
+function noopFn() {}
+const noop = noopFn as any;
+export { noop };
```
caseyhebebrand, line 3: Is this just to avoid repeating code in potential future noops?

caseyhebebrand (approved):

packages/presentation/src/utils/noop.ts
vigneshm: Yep. I copied this from the existing implementation in `@spinnaker/core` as an example.

packages/presentation/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "scripts": {
+    "build": "rollup -c",
+    "clean": "npx shx rm -rf dist",
+    "prepublishOnly": "npm run build",
+    "watch": "rollup -c -w"
+  },
+  "peerDependencies": {
+    "lodash": "^4.17.20",
+    "react": "~16.14.0",
+    "react-dom": "^16.14.0"
```
christopherthielen, line 17: Wondering if we really need `react-dom` in this package?
@vigneshm

vigneshm: Good call. I don't think we need it. I'll remove it for now and add it back when needed.

PR #8830, opened 2021-01-06
feat(amazon): Display updated ALB redirect config

alanmquach:
After configuring a redirect in the ConfigureRedirectConfigModal, the new (or updated) config is not shown in the Edit Load Balancer modal. This makes it difficult to review your changes for mistakes, or even know whether your config actually took effect.

### Before (there is a config, it is just not shown):
image

### After:
image

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
@@ -796,6 +796,7 @@ const Action = (props: {
   configureRedirect: (action: IListenerAction) => void;
 }) => {
   if (props.action.type !== 'authenticate-oidc') {
+    const redirectConfig = props.action.redirectActionConfig || props.action.redirectConfig;
     // TODO: Support redirect
     return (
       <div className="horizontal top">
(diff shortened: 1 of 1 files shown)


app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
```diff
@@ -820,13 +822,27 @@ const Action = (props: {
...
+            <dt>Host</dt>
+            <dd>{(props.action.redirectActionConfig || props.action.redirectConfig).host}</dd>
+            <dt>Path</dt>
+            <dd>{(props.action.redirectActionConfig || props.action.redirectConfig).path}</dd>
+            <dt>Port</dt>
+            <dd>{(props.action.redirectActionConfig || props.action.redirectConfig).port}</dd>
+            <dt>Protocol</dt>
+            <dd>{(props.action.redirectActionConfig || props.action.redirectConfig).protocol}</dd>
+            <dt>Status Code</dt>
+            <dd>{(props.action.redirectActionConfig || props.action.redirectConfig).statusCode}</dd>
```
caseyhebebrand, line 835: nit: Define a `const` to avoid repeating this code?

```js
const redirectConfig = props.action.redirectActionConfig || props.action.redirectConfig;

...

<dd>{redirectConfig.host}</dd>
```

vigneshm (approved):

app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
alanmquach: I was thinking about that too but ended up on the other side of the fence.

app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBListeners.tsx
```diff
@@ -820,13 +823,27 @@ const Action = (props: {
...
         {props.action.type === 'redirect' && (
-          <button
-            className="btn btn-link no-padding"
-            type="button"
-            onClick={() => props.configureRedirect(props.action)}
-          >
-            Configure...
-          </button>
+          <dl className="dl-horizontal dl-narrow">
+            <dt>Host</dt>
```
christopherthielen, line 827: I'm late to the party, but we have `<LabeledValue>` for `<dd/><dt/>` pairs

PR #8834, opened 2021-01-08
Deangularize instance writer

christopherthielen:
I was going to tackle a request to add links to instances, which rabbit holed me into de-angularizing the InstanceWriter.

does it pass CI now?


christopherthielen:
it passes ci now

app/scripts/modules/amazon/src/instance/amazon.instance.write.service.ts
```diff
@@ -64,25 +54,25 @@ export class AmazonInstanceWriter extends InstanceWriter {
...
   }
 
-  public registerInstancesWithTargetGroup(
+  public static registerInstancesWithTargetGroup(
     instanceGroups: IMultiInstanceGroup[],
     application: Application,
     targetGroupNames: string[],
   ) {
-    const jobs = this.buildMultiInstanceJob(
+    const jobs = super.buildMultiInstanceJob(
```
vigneshm, line 62: TIL that typescript allows inheritance on static members

vigneshm (approved):

app/scripts/modules/amazon/src/instance/amazon.instance.write.service.ts
christopherthielen: not just typescript, es6 classes support this!

PR #8835, opened 2021-01-08
fix(aws): Add Connection termination on deregistration flag on NLB's

aravindmd:
Add Connection termination on deregistration flag on NLB's

River PR : https://github.com/spinnaker/clouddriver/pull/5178


caseyhebebrand (approved):
LGTM

PR #8836, opened 2021-01-08
feat(md): waiting status

emjburns:
Adding a new status that denotes a new environment waiting for an artifact to become available.


christopherthielen (approved):

blaquezzz (approved):

app/scripts/modules/core/src/managed/managedResourceStatusConfig.tsx
```diff
@@ -223,4 +223,20 @@ export const viewConfigurationByStatus: { [status in ManagedResourceStatus]: IVi
...
+  WAITING: {
+    appearance: 'info',
+    iconName: 'mdCreated',
+    popoverContents: (resourceSummary: IManagedResourceSummary) => (
+      <>
+        <p>
+          <b>Waiting for information.</b>
+        </p>
+        <p>
+          This resource is part of a brand new environment. We are waiting for an artifact to become available to
```
erikmunson, line 235: Prefer contractions where possible as it reads more conversational vs. formal

```suggestion
          This resource is part of a brand new environment. We're waiting for an artifact to become available to
```

erikmunson (approved):
LGTM, thanks for doing this!

app/scripts/modules/core/src/managed/managedResourceStatusConfig.tsx
erikmunson: Actually, looking back at the content of these message I'd suggest using `Spinnaker is...` to be consistent, we haven't yet deviated from that pattern in MD microcopy. Sorry for the thrash between suggestions.

```suggestion
          This resource is part of a brand new environment. Spinnaker is waiting for an artifact to become available to
```

PR #8838, opened 2021-01-12
fix(core/pipeline): Remove note that pipelines are only paused for 72 hours

ajordens:
More recent implementations of the orchestration engine no longer
automatically cancel paused pipelines.


caseyhebebrand (approved):
LGTM

PR #8840, opened 2021-01-12
refactor(core/deployment): Update redblack fields without force updating

caseyhebebrand:
- Move from force updating redblack fields to utilizing the `onChange` prop that is already passed in. This helps with reusing this component when the command originates from a react component. 
- Refactored to a functional component while I was there
- Remove duplicate code from cloudfoundry, and replace with the refactored additional fields


app/scripts/modules/core/src/deploymentStrategy/strategies/redblack/AdditionalFields.tsx
```diff
@@ -1,95 +1,79 @@
...
 
 export interface IRedBlackStrategyAdditionalFieldsProps extends IDeploymentStrategyAdditionalFieldsProps {
   command: IRedBlackCommand;
 }
 
-export class AdditionalFields extends React.Component<IRedBlackStrategyAdditionalFieldsProps> {
-  private handleChange = (key: string, value: any) => {
-    set(this.props.command, key, value);
-    this.forceUpdate();
-  };
```
caseyhebebrand, line 17: This is the main thing I wanted to get rid of, but it gets lost in the refactor. Now, the field `onChange` just directly uses the `onChange` prop.

vigneshm (approved):

PR #8849, opened 2021-01-14
feat(pipelines): Enable filtering on dynamic categories

alanmquach:
Currently, the only way to filter pipelines is either by name or text search (executions can be filtered by status).

This PR adds the ability to filter by additional dimensions (we'll refer to them as categories):
image

The categories themselves are defined in the individual pipeline configs to allow users to freely define any number of filterable attributes. This way they can group/filter their pipelines in any way they may want.
```
{
  "categories": [
    {
      "name": "Opponent",
      "value": "Borgov"
    },
    {
      "name": "Opening",
      "value": "Queen's Gambit Declined"
  ],
  "keepWaitingPipelines": false,
  "lastModifiedBy": "beth@harmonchessacademy.com",
  "limitConcurrent": true,
...
}
```

Of note, filtering multiple categories at the same time will logically `AND` the filters, similar to the filtering behavior in clusters.
```
    // We use .every() to logically AND the different categories
    return groups.every((group) => {
```

As of this PR, categories will have to defined in JSON. The UI for configuring categories (in the pipeline config) can come  separately.


vigneshm:
Wow, this feature is very similar to the one [here](https://github.com/spinnaker/deck/pull/8819).

alanmquach:
> Wow, this feature is very similar to the one [here](https://github.com/spinnaker/deck/pull/8819).

Oh wow. Let me ping #8819 and see what they think.

alanmquach:
This implementation will now also only fetch executions for filtered pipelines (this functionality already existed for filtering by individual pipeline names, it now also works with filtering by dynamic categories). 

Of note, getRunningExecutions does not seem to be asking for a filtered response: https://github.com/spinnaker/deck/blob/c7a58cd4657054b2b1333057333f6aed0151741f/app/scripts/modules/core/src/pipeline/service/execution.service.ts#L41-L43
But I do not want to overload this PR with that change.

cc: @rumit-opsmx

app/scripts/modules/core/src/pipeline/filter/ExecutionFilters.tsx
```diff
@@ -214,6 +249,13 @@ export class ExecutionFilters extends React.Component<IExecutionFiltersProps, IE
             />
           </div>
           <div className="content">
+            {categories.names.map((name) => (
+              <FilterSection heading={name} expanded={true}>
+                {(categories.values[name] || []).map((value) => (
+                  <CategoryFilter group={name} value={value} refresh={this.refreshExecutions} />
+                ))}
+              </FilterSection>
+            ))}
```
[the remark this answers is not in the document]
alanmquach: Done!

caseyhebebrand (approved):

app/scripts/modules/core/src/domain/IPipeline.ts
```diff
@@ -93,3 +94,8 @@ export interface IPipelineRef {
 export interface ITemplateInheritable {
   inherited?: boolean;
 }
+
+export interface IPipelineCategory {
+  name: string;
+  value: string;
+}
```
christopherthielen, line 101: Can you add some comment describing what this means? I would expect "PipelineCategory" to be a simple string or enum on a pipeline.  The name/value combination in this type is not immediately obvious.

app/scripts/modules/core/src/pipeline/filter/ExecutionFilters.tsx
```diff
@@ -348,6 +384,36 @@ const Pipelines = SortableContainer(
   ),
 );
 
+const CategoryFilters = (props: { categories: IOrderedCategoryFilters; refresh: () => void }): JSX.Element => (
+  <>
+    {props.categories.names.map((name) => (
+      <FilterSection heading={name} expanded={true}>
```
christopherthielen, line 390:
```suggestion
      <FilterSection key={name} heading={name} expanded={true}>
```

app/scripts/modules/core/src/pipeline/filter/ExecutionFilters.tsx
```diff
@@ -348,6 +384,36 @@ const Pipelines = SortableContainer(
...
+        {(props.categories.values[name] || []).map((value) => (
+          <CategoryFilter group={name} value={value} refresh={props.refresh} />
```
christopherthielen, line 392:
```suggestion
          <CategoryFilter key={name} group={name} value={value} refresh={props.refresh} />
```

app/scripts/modules/core/src/pipeline/filter/executionFilter.service.ts
```diff
@@ -82,6 +82,40 @@ export class ExecutionFilterService {
...
+      .map(([name, value]) => ({ name, value }));
+    const grouped = groupBy(decoded, 'name');
+    const groups = Object.keys(grouped);
+    // We use .every() to logically AND the different categories
+    return groups.every((group) => {
+      const checkedValues = grouped[group].map((category) => category.value);
+      const relevantValues = (config.categories || [])
+        .filter((category) => category.name === group)
+        .map((category) => category.value);
+      return checkedValues.some((checkedValue) => relevantValues.includes(checkedValue));
```
christopherthielen, line 100: `.some()`

am I reading this right that we AND the categories, but OR the selected values?

christopherthielen (approved):

app/scripts/modules/core/src/domain/IPipeline.ts
alanmquach: Is the confusion here perhaps because I am not simply allowing users to categorize their pipelines into categories defined as strings.

Rather, I am allowing pipeline authors to define category *names* and give their pipelines values for those categories?

For instance, if users wanted to group/filter by multiple dimensions, instead of them having to work around the system by defining categories:

- `"env=test"`
- `"env=prod"`
- `"region=east"`
- `"region=west"`
- `"user=alan"`
- `"user=chris"`

I tried to tackle this from the start by making multiple dimensions a first class concept here:

- So the category name "env" can have values "test" and "prod", the category name "region" can have values "east" and "west"
- And so on and so forth

Perhaps "category" is not the best term to describe this? See above screenshot again for clarification?

alanmquach: Nonetheless, still good feedback. Is there a better way/term to make this more obvious?

app/scripts/modules/core/src/pipeline/filter/executionFilter.service.ts
alanmquach: Yes.

We OR the values because they are assumed to be mutually exclusive and AND the categories (names) because they're assumed to be different dimensions.

This is analogous to clusters where, for instance, we AND between account and region, but we OR the values of those dimensions.

app/scripts/modules/core/src/domain/IPipeline.ts
christopherthielen: how about something more akin to FilterDimension or Metadata?  Category seems not quite right in this context.

app/scripts/modules/core/src/pipeline/filter/executionFilter.service.ts
christopherthielen: makes sense!

app/scripts/modules/core/src/domain/IPipeline.ts
christopherthielen: PipelineTags?

alanmquach: Tags sound good.

Dimension seems to be the most technically accurate but I'm also thinking of how this is present to a user to configure in the pipeline config UI and Dimension doesn't seem obvious/intuitive enough.

Metadata is a little too broad or vague.

PR #8853, opened 2021-01-15
fix(core): Fix incorrect removal when merging serverGroups response

alanmquach:
Since forever, it appears that we've been incorrectly removing `serverGroup`s when merging the response into the `serverGroups` datasource (to preserve referential equality and save on some, hopefully many, rerenders).

It all boils down these few lines:
```
data.forEach((serverGroup: IServerGroup, idx: number) => {
...
  toRemove.push(idx);
...
toRemove.forEach((idx) => data.splice(idx, 1));
```

Since we are pushing into `toRemove` while `forEach`ing data, the indices will naturally be in ascending order. However, as we iterate through the array of ascending indices to splice (remove) those elements out of `data`, we start from the left causing everything to the right to shift left by 1 every iteration. For example:

```
const toRemove =      [0, 1];
const data =          [0, 1, 2, 3, 4, 5];
data.splice(0, 1); // [1, 2, 3, 4, 5]
data.splice(1, 1); // [1, 3, 4, 5]
```

Removal of more than 1 element is pretty much guaranteed to be incorrect for everything but the very first removal.

This manifested as server groups disappearing on a applicationDataSource refresh interval, remaining absent for the entire refresh interval, and most likely coming back on the next refresh interval, if the user had not panic refreshed the page.

Whew!

### DIFF
--- a/app/scripts/modules/core/src/cluster/cluster.service.spec.ts
+++ b/app/scripts/modules/core/src/cluster/cluster.service.spec.ts
@@ -157,6 +157,138 @@ describe('Service: Cluster', function () {
     });
   });
 
+  describe('addServerGroupsToApplication merging to preserve referential equality', () => {
+    const asgFabricator = (x: string): IServerGroup => ({
+      cluster: `cluster-${x}`,
+      name: `cluster-${x}-v001`,
+      account: 'test',
+      region: 'us-east-1',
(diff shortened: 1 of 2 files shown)


vigneshm (approved):
Wow, great catch!

caseyhebebrand (approved):

PR #8858, opened 2021-01-19
feat(ecs): Add support for Capacity providers - enhanced Deck UI

paragbhingre:
This PR adds a new enhanced version of UI to support Capacity Providers.

Currently, Capacity Provider UI section provides input fields to enter details about Capacity Provider Strategy for the server group.
But new UI will have a drop down list for selecting a Capacity Provider. Plus now users can add default Capacity providers without typing in the details of default Capacity Provider by just clicking on the use cluster default option. Please refer the following screenshots for the more details.

Screen Shot 2021-01-19 at 12 12 27 PM

Screen Shot 2021-01-19 at 12 12 44 PM

Addresses [spinnaker/spinnaker#6242](https://github.com/spinnaker/spinnaker/issues/6242)

Continuation of PR [#8810](https://github.com/spinnaker/deck/pull/8810). Implemented review comments from #8810 PR as well.

### DIFF
--- a/app/scripts/modules/core/package.json
+++ b/app/scripts/modules/core/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@spinnaker/core",
   "license": "Apache-2.0",
-  "version": "0.0.543",
+  "version": "0.0.544",
   "main": "lib/lib.js",
   "typings": "lib/index.d.ts",
   "scripts": {


paragbhingre:
@allisaurus, @piradeepk please review this PR.

test/functional/cypress/integration/ecs/server_group.spec.js
```diff
@@ -163,19 +165,50 @@ describe('amazon ecs: ECSApp Server Group', () => {
...
+    cy.get('[data-test-id="ServerGroup.capacityProvider.base.0"]').type(1);
+    cy.get('[data-test-id="ServerGroup.capacityProvider.weight.0"]').type(2);
 
     cy.get('[data-test-id="ServerGroupWizard.submitButton"]').click();
     cy.get('.glyphicon-edit').click();
 
-    cy.get('[data-test-id="capacityProvider.name.0"]').should('have.value', 'FARGATE');
-    cy.get('[data-test-id="capacityProvider.base.0"]').should('have.value', '1');
-    cy.get('[data-test-id="capacityProvider.weight.0"]').should('have.value', '2');
+    cy.get('[data-test-id="ServerGroup.customCapacityProvider.name.0"]').children().children().children(".Select-multi-value-wrapper").children().children('.Select-value-label').should('contain', 'FARGATE_SPOT');
```
allisaurus, line 209: this is a lot of `.children()` - can we not add a test ID to the element we want to access?

app/scripts/modules/ecs/src/ecsCluster/ecsCluster.read.service.ts
```diff
@@ -2,11 +2,19 @@ import { module } from 'angular';
...
 import { REST } from '@spinnaker/core';
 import { IEcsClusterDescriptor } from './IEcsCluster';
+import {IEcsCapacityProvidersDetails} from "../ecsCluster/IEcsCapacityProvidersDetails";
 
 export class EcsClusterReader {
   public listClusters(): PromiseLike<IEcsClusterDescriptor[]> {
     return REST('/ecs/ecsClusters').get();
   }
+
+  public listDescribeClusters(account: string, region: string): PromiseLike<IEcsCapacityProvidersDetails[]> {
```
allisaurus, line 12: let's simplify this func name to `DescribeClusters`, unless it's not actually doing that?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -19,10 +27,30 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
...
   }
 
+  public componentDidMount() {
+    this.props.configureCommand('1').then(() => {
+      const cmd = this.props.command;
+      const useDefaultCapacityProviders = this.state.useDefaultCapacityProviders;
+      this.setState({
+        availableCapacityProviders: cmd.backingData && cmd.backingData.filtered && cmd.backingData.filtered.availableCapacityProviders ? cmd.backingData.filtered.availableCapacityProviders : [],
+        defaultCapacityProviderStrategy: cmd.backingData && cmd.backingData.filtered && cmd.backingData.filtered.defaultCapacityProviderStrategy ? cmd.backingData.filtered.defaultCapacityProviderStrategy : [],
+        capacityProviderStrategy: useDefaultCapacityProviders && cmd.backingData && cmd.backingData.filtered && cmd.backingData.filtered.defaultCapacityProviderStrategy ? cmd.backingData.filtered.defaultCapacityProviderStrategy : this.state.capacityProviderStrategy,
```
allisaurus, line 46: these lines are soooooo long. if we need this many checks for each value, can we at least break them out into helper function to improve readability? Same goes for similarly long lines on 30-35

piradeepk:
Can you close this PR and update the existing PR (so that we have the continuity on comments)?

paragbhingre:
> Can you close this PR and update the existing PR (so that we have the continuity on comments)?

In order to avoid confusion of the files that have been changed as a part of this PR, I decided to keep this PR separate from the previous one. I am thinking of closing the previous PR as this PR is built on top of [#8824](https://github.com/spinnaker/deck/pull/8824) where previous one is not. And I have also implemented the comments from the previous PR in here. I will keep reference to the previous PR in the description section of this PR for a better understanding. What do you think about it? @piradeepk

piradeepk:
> > Can you close this PR and update the existing PR (so that we have the continuity on comments)?
> 
> In order to avoid confusion of the files that have been changed as a part of this PR, I decided to keep this PR separate from the previous one. I am thinking of closing the previous PR as this PR is built on top of [#8824](https://github.com/spinnaker/deck/pull/8824) where previous one is not. And I have also implemented the comments from the previous PR in here. I will keep reference to the previous PR in the description section of this PR for a better understanding. What do you think about it? @piradeepk

Sounds good, I would just like to maintain the review history as there are a bunch of comments spread across multiple PRs that add additional work on the reviewer on each review.

test/functional/cypress/integration/ecs/server_group.spec.js
paragbhingre: I am completely in favor of adding test ID to the element than using so many children(), but unfortunately we are using TetheredSelect component from @spinnaker/core where we can not assign the ID to the select element directly. That is why I had to use so many children to reach to the required element.

app/scripts/modules/ecs/src/ecsCluster/ecsCluster.read.service.ts
```diff
@@ -2,11 +2,19 @@ import { module } from 'angular';
 
 import { REST } from '@spinnaker/core';
 import { IEcsClusterDescriptor } from './IEcsCluster';
+import {IEcsCapacityProvidersDetails} from "../ecsCluster/IEcsCapacityProvidersDetails";
```
piradeepk, line 5: nit: spacing

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -82,6 +85,7 @@ export interface IEcsServerGroupCommandBackingData extends IServerGroupCommandBa
   filtered: IEcsServerGroupCommandBackingDataFiltered;
   targetGroups: string[];
   ecsClusters: IEcsClusterDescriptor[];
+  capacityProviderDetails: IEcsCapacityProvidersDetails[];
```
piradeepk, line 88: Inconsistency btwn the property name and the type (you should stick to either CapacityProviderDetails (IEcsCapacityProviderDetails) or CapacityProvidersDetails(IEcsCapacityProvidersDetails)

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -280,6 +288,49 @@ export class EcsServerGroupConfigurationService {
...
+  public setCapacityProviderDetails(command: IEcsServerGroupCommand): void {
+    this.$q.all({
+      capacityProviderDetails: this.ecsClusterReader.listDescribeClusters(command.credentials, command.region)
+    }).then((result: Partial<IEcsServerGroupCommandBackingData>) => {
+      command.backingData.capacityProviderDetails = chain(result.capacityProviderDetails)
+        .map((cluster) => this.mapCapacityProviderDetails(cluster))
+        .value();
+    });
+
+    if(command.ecsClusterName == null || command.ecsClusterName.length == 0){
```
piradeepk, line 300: nit: for ease of readability, it's usually easier to put the explicityly valid condition, as it's possible in the future to have more or conditions in which it could be false so you would have a bunch of || statements vs just having the condition that is the true case and anything else falls under else

```
if (command.ecsClusterName != null && command.ecsClusterName.length > 0) {
    this.configureAvailableCapacityProviders(command)
} else {
    command.backingData.filtered.availableCapacityProviders = [];
    command.backingData.filtered.defaultCapacityProviderStrategy = [];
}
```

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -2,15 +2,23 @@ import React from 'react';
 import { module } from 'angular';
 import { react2angular } from 'react2angular';
 import {IEcsCapacityProviderStrategyItem, IEcsServerGroupCommand} from '../../serverGroupConfiguration.service';
-import { HelpField, withErrorBoundary } from  '@spinnaker/core';
+import { HelpField, withErrorBoundary, TetheredSelect } from  '@spinnaker/core';
+import {Option} from "react-select";
+import {Alert} from "react-bootstrap";
```
piradeepk, line 7: nit: spacing

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -37,10 +65,10 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
     this.setState({capacityProviderStrategy : capacityProviderStrategy });
   }
 
-  private updateCapacityProviderName = (index: number, targetCapacityProviderName: string) => {
+  private updateCapacityProviderName = (index: number, targetCapacityProviderName: any) => {
```
piradeepk, line 68: We should explicitly take in the label if that's all we want, instead of taking in any type, define an explicit type of the property(ies) that you need.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -37,10 +65,10 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
...
     const capacityProviderStrategy = this.state.capacityProviderStrategy;
     const targetCapacityProviderStrategy = capacityProviderStrategy[index];
-    targetCapacityProviderStrategy.capacityProvider = targetCapacityProviderName;
+    targetCapacityProviderStrategy.capacityProvider = targetCapacityProviderName.label;
```
piradeepk, line 71: if you pass in the label as the input, you wouldn't have to get the label.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -61,6 +89,24 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
     this.setState({ capacityProviderStrategy: capacityProviderStrategy });
   };
 
+  private updateCapacityProviderStrategy = (targetCapacityProviderType: string) => {
+    const useDefaultCapacityProviders = targetCapacityProviderType == 'defaultCapacityProvider';
```
piradeepk, line 93: Is it possible to pass in a boolean here? I'm hesitant doing string comparison (on the off chance that type changes in the future, or the case changes, etc). If you can pass in true/false since that's really all you care about, that would improve maintainability.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -61,6 +89,24 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
     this.setState({ capacityProviderStrategy: capacityProviderStrategy });
   };
 
+  private updateCapacityProviderStrategy = (targetCapacityProviderType: string) => {
```
piradeepk, line 92: It feels like this method is doing a lot. Can we separate this into an update useDefaultCapacityProviders method and a updateCapacityProviderStrategy method (you can involve updateCapacityProviderStrategy from within useDefaultCapacityProviders, but it feels like this method has too much responsibility

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -61,6 +89,24 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
...
+      this.setState({capacityProviderStrategy : []});
+      this.props.notifyAngular('capacityProviderStrategy', []);
+      if (this.state.defaultCapacityProviderStrategy.length > 0){
+        this.setState({capacityProviderStrategy : this.state.defaultCapacityProviderStrategy});
+        this.props.notifyAngular('capacityProviderStrategy', this.state.defaultCapacityProviderStrategy );
+      }
+    } else {
+      this.setState({capacityProviderStrategy : []});
+      this.props.notifyAngular('capacityProviderStrategy', []);
+    }
```
piradeepk, line 107: There's duplicate logic in the if and else, and it's a bit confusing to follow. what do you think about doing something like this:

```
if (useDefaultCapacityProviders && this.state.defaultCapacityProviderStrategy.length > 0) {
    this.setState({capacityProviderStrategy : this.state.defaultCapacityProviderStrategy});
    this.props.notifyAngular('capacityProviderStrategy', this.state.defaultCapacityProviderStrategy );
} else {
    this.setState({capacityProviderStrategy : []});
    this.props.notifyAngular('capacityProviderStrategy', []);
}
```

app/scripts/modules/ecs/src/serverGroup/configure/wizard/horizontalScaling/horizontalScaling.component.html
```diff
@@ -34,7 +34,7 @@
 
   <div class="form-group" ng-if="$ctrl.capacityProviderState.useCapacityProviders">
     <ecs-capacity-provider-react command="$ctrl.command"
-                             notify-angular="$ctrl.notifyAngular"/>
+                             notify-angular="$ctrl.notifyAngular" configure-command="$ctrl.configureCommand"/>
```
piradeepk, line 37: nit: newline btwn notify-angular and configure-command

test/functional/cypress/integration/ecs/server_group.spec.js
piradeepk: You don't need to use.children.childern (take a look at how we do testing for container images, and other tethered select components: https://github.com/spinnaker/deck/blob/master/test/functional/cypress/integration/ecs/server_group.spec.js#L108)

piradeepk (changes requested):

app/scripts/modules/ecs/src/ecsCluster/ecsCluster.read.service.ts
paragbhingre: Resolved.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
paragbhingre: Aahhh make sense.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
paragbhingre: Resolved.

paragbhingre:
CC @piradeepk  @allisaurus, I have addressed the feedback comments. Please let me know if you still have any concerns.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
paragbhingre: This logic make sense to me. After adding this logic the method becomes very compact and doesn't do much than if and else conditions. 

> It feels like this method is doing a lot. Can we separate this into an update useDefaultCapacityProviders method and a updateCapacityProviderStrategy method (you can involve updateCapacityProviderStrategy from within useDefaultCapacityProviders, but it feels like this method has too much responsibility


So I think keeping this logic in the same method makes better sense. What do you think ?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
paragbhingre: Resolved.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
paragbhingre: Yes, the change has been made.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
paragbhingre: Correct, the change has been made.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
paragbhingre: Resolved.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -280,6 +288,49 @@ export class EcsServerGroupConfigurationService {
...
+  public setCapacityProviderDetails(command: IEcsServerGroupCommand): void {
+    this.$q.all({
+      capacityProviderDetails: this.ecsClusterReader.describeClusters(command.credentials, command.region)
+    }).then((result: Partial<IEcsServerGroupCommandBackingData>) => {
+      command.backingData.capacityProviderDetails = chain(result.capacityProviderDetails)
+        .map((cluster) => this.mapCapacityProviderDetails(cluster))
+        .value();
+    });
+
+    if(command.ecsClusterName != null || command.ecsClusterName.length > 0){
```
piradeepk, line 300: Should this be an && condition? (do we want to configure available capacity if it's an empty string?

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -280,6 +288,49 @@ export class EcsServerGroupConfigurationService {
...
+      })
+      .map(availableCPs => availableCPs.defaultCapacityProviderStrategy)
+      .flattenDeep<IEcsCapacityProviderStrategyItem>()
+      .value();
+  }
+
+  public mapCapacityProviderDetails(describeClusters: IEcsCapacityProviderDetails) : IEcsCapacityProviderDetails {
+    return {
+      capacityProviders : describeClusters.capacityProviders,
+      clusterName :describeClusters.clusterName,
```
piradeepk, line 329: nit spacing

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -609,6 +661,7 @@ export class EcsServerGroupConfigurationService {
         this.configureAvailableSecrets(command);
         this.configureAvailableServiceDiscoveryRegistries(command);
         this.configureAvailableRegions(command);
+        this.setCapacityProviderDetails(command);
```
piradeepk, line 664: Does this actually work?

allisaurus (commented):
I played around with this a bit locally and discovered "base" will highlight in red whenever it's not defined, even though _only one_ capacity provider item can have "base" defined (otherwise ECS will return an error and fail the deployment). 
image

It looks like "base" is never strictly required - it'll default to 0 if missing - so what do you think about removing the validation behavior from this field so we're not implying that folks should always be entering this value?

allisaurus:
We don't show any message if a capacity provider that's not in the drop-down list is currently configured on the server group. 

image

image

Since this field was previously open-ended, there's a non-0 chance this could happen and I think it makes sense to show a msg in this case like [you're doing for target groups](https://github.com/spinnaker/deck/pull/8863). Do you think it's practical to add this functionality here as well, or would it be better to fast-follow with a separate PR? (CC @piradeepk who is a reviewer on both)

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
paragbhingre: Ohhh yes. I will change this.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
paragbhingre: Yes, credentialsChanged does get invoked when we select the account in SG wizard

paragbhingre:
> I played around with this a bit locally and discovered "base" will highlight in red whenever it's not defined, even though _only one_ capacity provider item can have "base" defined (otherwise ECS will return an error and fail the deployment).
> image
> 
> It looks like "base" is [never strictly required](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CapacityProviderStrategyItem.html) - it'll default to 0 if missing - so what do you think about removing the validation behavior from this field so we're not implying that folks should always be entering this value?

@allisaurus, that was a nice catch. I have a few observations around this scenario. So we can let all the CapacityProviderStartegyItems to have a base value as 0 or not defined (left blank). But all the weight values can't have value as 0 or not defined. At least one CP should have weight defined and it should be greater than 1.
1. So we can either keep weight value to be required or let ECS fail the deployment if it's undefined. 
2. Currently, when we click on Add New Capacity Provider the base and weight values will be blank along with capacity provider name. I am planning to set base to 0 and weight to 1 whenever we will click on add new capacity provider similar to what AWS console is doing.

Screen Shot 2021-01-27 at 12 33 31 PM

 What's your thought on this?

allisaurus:
@paragbhingre Not requiring the "base" but defaulting "weight" to 1 sounds good to me!

paragbhingre:
> @paragbhingre Not requiring the "base" but defaulting "weight" to 1 sounds good to me!

Just consolidating the expected behavior here as per the discussion:
1. When user clicks on add new capacity provider all the fields i.e. capacity provider name, base and a weight will be blank.
2. User has to choose capacity provider name and weight (those are required fields) at least avoid ECS failing the deployment.

paragbhingre:
Reopening the PR, closed by mistake.

PR #8859, opened 2021-01-20
feat(amazon/loadBalancer): Enable dualstacking in ALBs/NLBs with constraints

caseyhebebrand:
ALBs and NLBs can have an `ipAddressType` of `ipv4` or `dualstack` if:
- The ALB is external
- The NLB is external, but all targets are of type `ip`.

This option is now an advanced setting the server group configuration. If nothing is selected it will default to `ipv4` (which is the current behavior). 

This also refactors the ALB/NLB advanced settings to functional components.

### DIFF
--- a/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx
+++ b/app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx
@@ -1,25 +1,40 @@
 import React from 'react';
 
+import { FormikProps } from 'formik';
 import { FormikFormField, CheckboxInput, NumberInput, HelpField } from '@spinnaker/core';
+import { IAmazonApplicationLoadBalancerUpsertCommand } from 'amazon/domain';
 
-export class ALBAdvancedSettings extends React.Component {
-  public render() {
-    return (
-      <div>
-        <FormikFormField
-          name="idleTimeout"
-          label="Idle Timeout"
-          help={<HelpField id="loadBalancer.advancedSettings.idleTimeout" />}
-          input={(props) => <NumberInput {...props} min={0} />}
-        />
-
-        <FormikFormField
-          name="deletionProtection"
-          label="Protection"
-          help={<HelpField id="loadBalancer.advancedSettings.deletionProtection" />}
-          input={(props) => <CheckboxInput {...props} text="Enable delete protection" />}
-        />
-      </div>
-    );
-  }
+export interface IALBAdvancedSettingsProps {
+  formik: FormikProps<IAmazonApplicationLoadBalancerUpsertCommand>;
 }
+
+export const ALBAdvancedSettings = React.forwardRef<HTMLDivElement, IALBAdvancedSettingsProps>((props, ref) => (
+  <div ref={ref}>
+    <FormikFormField
+      name="idleTimeout"
+      label="Idle Timeout"
+      help={<HelpField id="loadBalancer.advancedSettings.idleTimeout" />}
+      input={(props) => <NumberInput {...props} min={0} />}
+    />
+
+    <FormikFormField
+      name="deletionProtection"
+      label="Protection"
+      help={<HelpField id="loadBalancer.advancedSettings.deletionProtection" />}
+      input={(props) => <CheckboxInput {...props} text="Enable delete protection" />}
+    />
+
+    <FormikFormField
+      name="dualstack"
+      label="Dualstack"
+      help={<HelpField id="loadBalancer.advancedSettings.albIpAddressType" />}
+      input={(inputProps) => (
+        <CheckboxInput
+          {...inputProps}
+          text="Assign Ipv4 and IPv6"
+          disabled={Boolean(props.formik.values?.isInternal)}
+        />
+      )}
+    />
+  </div>
+));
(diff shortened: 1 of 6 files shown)


app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx
```diff
@@ -1,25 +1,40 @@
...
+  formik: FormikProps<IAmazonApplicationLoadBalancerUpsertCommand>;
+}
+
+export const ALBAdvancedSettings = React.forwardRef<HTMLDivElement, IALBAdvancedSettingsProps>((props, ref) => (
+  <div ref={ref}>
+    <FormikFormField
+      name="idleTimeout"
+      label="Idle Timeout"
+      help={<HelpField id="loadBalancer.advancedSettings.idleTimeout" />}
+      input={(props) => <NumberInput {...props} min={0} />}
```
christopherthielen, line 17: I noticed you used `inputProps` down below to avoid shadowing the `props` variable.  Might be good to do that everywhere.  I thought we had a linter rule enabled that forbids shadowing?

app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx
```diff
@@ -1,25 +1,40 @@
...
-          name="idleTimeout"
-          label="Idle Timeout"
-          help={<HelpField id="loadBalancer.advancedSettings.idleTimeout" />}
-          input={(props) => <NumberInput {...props} min={0} />}
-        />
+export interface IALBAdvancedSettingsProps {
+  formik: FormikProps<IAmazonApplicationLoadBalancerUpsertCommand>;
+}
+
+export const ALBAdvancedSettings = React.forwardRef<HTMLDivElement, IALBAdvancedSettingsProps>((props, ref) => (
```
christopherthielen, line 11: I'm questioning if `forwardRef` is necessary at all, since it's being applied on the root jsx element anyway.

app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx
```diff
@@ -1,25 +1,40 @@
...
+
+    <FormikFormField
+      name="dualstack"
+      label="Dualstack"
+      help={<HelpField id="loadBalancer.advancedSettings.albIpAddressType" />}
+      input={(inputProps) => (
+        <CheckboxInput
+          {...inputProps}
+          text="Assign Ipv4 and IPv6"
+          disabled={Boolean(props.formik.values?.isInternal)}
```
christopherthielen, line 35: I'd rather pass an `isInternal` prop to this component instead of the entire `formik` bag.

app/scripts/modules/amazon/src/loadBalancer/configure/network/NLBAdvancedSettings.tsx
```diff
@@ -1,38 +1,44 @@
...
-          <label>
-            <Field type="checkbox" name="loadBalancingCrossZone" checked={values.loadBalancingCrossZone} />
-            Cross-Zone Load Balancing
-          </label>
-        </div>
-      </div>
-    );
-  }
-}
+export const NLBAdvancedSettings = React.forwardRef<HTMLDivElement, INLBAdvancedSettingsProps>((props, ref) => {
```
christopherthielen, line 13: same comment as above; it doesn't look like `forwardRef` is necessary here

app/scripts/modules/amazon/src/loadBalancer/configure/network/NLBAdvancedSettings.tsx
```diff
@@ -1,38 +1,44 @@
...
+  const { formik } = props;
+  /** NLBs can only be dualstacked if they are external and all targets are IPs. This will remain true until NLBs support security groups. */
+  const allIpTargets = every(formik.values.targetGroups, { targetType: 'ip' });
```
christopherthielen, line 16: maybe refactor similar to the comment above passing in `allIpTargets` and `showDualstack` props instead of the entire `formik` bag

christopherthielen (approved):
Have a couple asks, then lgtm

app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx
caseyhebebrand: It is needed for its use as a `WizardPage` in order for me to pass the ref into the root element.

app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx
caseyhebebrand: sounds good!

app/scripts/modules/amazon/src/loadBalancer/configure/application/ALBAdvancedSettings.tsx
christopherthielen: ah, gotcha!

PR #8861, opened 2021-01-20
refactor(presentation): Move icons, illustrations folders to presentation package

vigneshm:
First PR in a series where modules from `@spinnaker/core/.../presentation/` package will be moved to `@spinnaker/presentation` package.  I'm moving these modules in chunks rather than a single one as it'll give us opportunities to clean up the API (if necessary) of the components.

[Edit]
The PR is now split in to 4 commits

1. First commit adds support to rollup to handle svg files and render them as react components
2. Second commit moves the `icon` and `illustration` related modules from `deck/**/core/**/presentation` to `@spinnaker/presentation`. A single change was made to fix a typescript error though.
3. Third commit re-exports the moved modules(in commit 2) in order to avoid breaking applications/packages that rely on `@spinnaker/core`.
4. Fourth commit updates all the references (of the moved modules) to `@spinnaker/presentation`.

I'll be writing eslint rules and fixers in a separate PR.


vigneshm:
This is just a simple move with manual changes only in `rollup.config.js`, so you can ignore reviewing the other files.

christopherthielen:
1) Can we separate out the simple moves into a separate commit (separate from rollup config and any other non-move changes)?
2) I'd prefer to see the deletes in the same commit, so git can track them as file renames.  This is the only way to retain git history across the move.
2b) We would have to also commit the migration as a third commit otherwise deck won't build

vigneshm:
> 1. Can we separate out the simple moves into a separate commit (separate from rollup config and any other non-move changes)?

Yep, I can split this one.

> 2. I'd prefer to see the deletes in the same commit, so git can track them as file renames.  This is the only way to retain git history across the move.
>    2b) We would have to also commit the migration as a third commit otherwise deck won't build

This one's tricky 'cos we need to publish `@spinnaker/presentation` package (with the new modules) before updating its version in `@spinnaker/core` and complete the migration. One way I could think of achieving it is to reference the new modules (from presentation package) directly in the existing call site like the following and make webpack find these modules from the presentation package. What do you think?

```js

// deck/app/scripts/modules/core/src/presentation/illustrations/index.ts
export {Illustration, IllustrationName} from '../../../../../../../packages/presentation/src/';
```

vigneshm:
> ```js
> // deck/app/scripts/modules/core/src/presentation/illustrations/index.ts
> export {Illustration, IllustrationName} from '../../../../../../../packages/presentation/src/';
> ```

I'm running into issues with typescript while doing the above. The `rootDir` in `tsconfig.json` is set to `./src` under `app/../core`, so typescript doesn't allow module imports from outside the `rootDir`. If the `rootDir` is switched to `deck`, then it affects typescript's emitted output tree.

christopherthielen:
>> I'd prefer to see the deletes in the same commit, so git can track them as file renames. This is the only way to retain git history across the move.
>> 2b) We would have to also commit the migration as a third commit otherwise deck won't build


> This one's tricky 'cos we need to publish @spinnaker/presentation package (with the new modules) before updating its version in @spinnaker/core and complete the migration. One way I could think of achieving it is to reference the new modules (from presentation package) directly in the existing call site like the following and make webpack find these modules from the presentation package. What do you think?

Agreed, this is tricky.  I think we can do it cleanly though.  Here's what I suggest:

- create these commits:

commit 1) rollup.config.js, etc
commit 2) file renames from app/.../modules/presentation to packages/presentation
commit 3) re-export `@spinnaker/presentation` from `@spinnaker/core` (for backwards compat in plugins, deck-nflx, etc)

- publish a prelease `@spinnaker/presentation@1.0.0-beta.1` from your working copy (before the code even is committed)

commit 4) Add dependency on `@spinnaker/presentation@1.0.0-beta.1`.  Update all usages in Deck to import from `@spinnaker/presentation`.

- create PR (it should pass)
- create eslint rule with fixer that migrates imports for plugins and deck-nflx from `core` to `presentation`

packages/presentation/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@spinnaker/presentation",
   "description": "Provides re-usable UI components for deck",
-  "version": "0.0.1",
+  "version": "0.0.1-beta.1",
```
vigneshm, line 4: Not a big deal, but I'm wondering if we should just use `beta` temporarily and update these to proper versions?

christopherthielen: yeah, after this merges maybe bump it again? IDK

christopherthielen (approved):

PR #8863, opened 2021-01-21
fix(ecs): Determine invalid target groups for the ECS provider

paragbhingre:
This PR fixes the issue [#5248](https://github.com/spinnaker/spinnaker/issues/5248)

For the ECS Provider in deck, an Alert have been added when there is a target group(s) that was previously selected and is no longer available.

Dirty Target Groups
Screen Shot 2021-01-26 at 5 52 43 PM

Working Target Groups
Screen Shot 2021-01-28 at 2 02 41 PM

### DIFF
--- a/app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
@@ -44,7 +44,7 @@ import { ServiceDiscoveryReader } from '../../serviceDiscovery/serviceDiscovery.
 import { IServiceDiscoveryRegistryDescriptor } from '../../serviceDiscovery/IServiceDiscovery';
 
 export interface IEcsServerGroupCommandDirty extends IServerGroupCommandDirty {
-  targetGroup?: string;
+  targetGroups?: string[];
 }
 
 export interface IEcsServerGroupCommandResult extends IServerGroupCommandResult {
@@ -64,6 +64,7 @@ export interface IEcsServerGroupCommandViewState extends IServerGroupCommandView
   contextImages: IEcsDockerImage[];
   pipeline: IPipeline;
   currentStage: IStage;
+  dirty: IEcsServerGroupCommandDirty;
 }
 
 export interface IEcsServerGroupCommandBackingDataFiltered extends IServerGroupCommandBackingDataFiltered {
@@ -507,7 +508,7 @@ export class EcsServerGroupConfigurationService {
     const newLoadBalancers = this.getLoadBalancerNames(command);
     const vpcLoadBalancers = this.getVpcLoadBalancerNames(command);
     const allTargetGroups = this.getTargetGroupNames(command);
-
+    const currentTargetGroups = command.targetGroupMappings.map(tg => tg.targetGroup)
     if (currentLoadBalancers && command.loadBalancers) {
       const valid = command.vpcId ? newLoadBalancers : newLoadBalancers.concat(vpcLoadBalancers);
       const matched = intersection(valid, currentLoadBalancers);
@@ -520,6 +521,16 @@ export class EcsServerGroupConfigurationService {
       }
       if (removedLoadBalancers.length) {
         result.dirty.loadBalancers = removedLoadBalancers;
+      }
+    }
+
+    if (currentTargetGroups) {
+      const matched = intersection(allTargetGroups, currentTargetGroups);
+      const removedTargetGroups = xor(matched, currentTargetGroups);
+      if (removedTargetGroups && removedTargetGroups.length > 0) {
+        command.viewState.dirty.targetGroups = removedTargetGroups;
+      } else if(command.viewState.dirty && command.viewState.dirty.targetGroups) {
+        command.viewState.dirty.targetGroups = [];
       }
     }
 
--- a/app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
@@ -135,6 +135,7 @@ export class Container extends React.Component<IContainerProps, IContainerState>
     targetMapping.targetGroup = newTargetGroup.value;
     this.props.notifyAngular('targetGroupMappings', currentMappings);
     this.setState({ targetGroupMappings: currentMappings });
+    this.updateDirtyTargetGroups();
   };
 
   private updateTargetGroupMappingPort = (index: number, targetPort: number) => {
@@ -152,6 +153,10 @@ export class Container extends React.Component<IContainerProps, IContainerState>
     this.setState({ targetGroupMappings: currentMappings });
   };
 
+  private updateDirtyTargetGroups = () => {
+    this.props.command.viewState.dirty.targetGroups = [];
+  };
+
   public render(): React.ReactElement<Container> {
     const removeTargetGroupMapping = this.removeTargetGroupMapping;
     const updateContainerMappingImage = this.updateContainerMappingImage;
@@ -159,6 +164,7 @@ export class Container extends React.Component<IContainerProps, IContainerState>
     const updateTargetGroupMappingPort = this.updateTargetGroupMappingPort;
     const updateComputeUnits = this.updateComputeUnits;
     const updateReservedMemory = this.updateReservedMemory;
+    const dirtyTagetGroups = this.props.command.viewState.dirty && this.props.command.viewState.dirty.targetGroups ? this.props.command.viewState.dirty.targetGroups : [];
 
     const dockerImageOptions = this.state.dockerImages.map(function (image) {
       let msg = '';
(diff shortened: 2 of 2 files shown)


paragbhingre:
CC @piradeepk @allisaurus, Please review this PR.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -24,6 +24,7 @@ interface IContainerState {
   dockerImages: IEcsDockerImage[];
   targetGroupsAvailable: string[];
   targetGroupMappings: IEcsTargetGroupMapping[];
+  dirtyTargetGroups: string [];
```
piradeepk, line 27: This state is not something that is being changed in this component. You shouldn't be storing it in the state as a convenience factor to retrieve/dismiss it.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -168,6 +177,30 @@ export class Container extends React.Component<IContainerProps, IContainerState>
...
+           {dirtyTargetGroupList}
+         </ul>
+         <p className="text-right">
+           <a
+             className="btn btn-sm btn-default dirty-flag-dismiss"
+             onClick={() => this.updateDirtyTargetGroups()}
+           >Okay</a>
+         </p>
+       </div>
+       );
```
piradeepk, line 202: What happens if I select another target group, do I still see the error? Do I have to manually dismiss it (ideally it will dismiss itself if the target group is no longer valid)?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -152,13 +155,19 @@ export class Container extends React.Component<IContainerProps, IContainerState>
     this.setState({ targetGroupMappings: currentMappings });
   };
 
+  private updateDirtyTargetGroups = () => {
+    this.props.command.viewState.dirty.targetGroups = null;
+    this.setState({dirtyTargetGroups: []});
+  };
```
piradeepk, line 161: Not sure I agree with marking the invalidness of a target group as part of the state. this property should be determined using the backing data, and be updated in the configuration when the target group changes.

piradeepk (changes requested):

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
paragbhingre: If you select another target group and saves the panel then the error won't appear again. But if you click **okay** on the error and saves the panel without choosing another target group then next time when you open the SG panel you will see the error.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
paragbhingre: So the reason behind setting this param as a part of the state variables is we need to re-render the react component once user clicks the **okay** button. If we need to hide the error message on **okay** click event then we need to save this variable as a part of the state. Also, we are changing the dirty.targetGroups property in the configurations when target group changes and not in the react component.

piradeepk: We shouldn't be hiding this message. If they have a target group that's no longer available, the only way they should be able to remove the message should be to change the target group. Removing the message just gives a false sense of security that everything is okay, when in reality, keeping an invalid target group will just cause the deployment to fail.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
piradeepk: I think a user should continue to see the alert till they fix the underlying issue (as this is an invalid deployment).

paragbhingre: Do you mean we shouldn't be giving **okay** button to close the error?

piradeepk: Yes. If there's a dirty target group, it means that the target group is no longer available to deploy, so the user NEEDS to update it or the deployment will fail. We shouldn't be giving them an option to clear the field since it's broken at that point.

paragbhingre: I got your point, I was just trying to be consistent with the AWS implementation of dirty target groups. For AWS provider as well **okay** button has been provided. Even when user clicks on the **okay** button dropdown value of the target group will not have the value of the dirty target groups.

piradeepk: IIRC, the AWS provider doesn't save the state, it allows you to press okay, but when you reopen the window, it will show up again. That being said, I think this behaviour is better suited since it's a problem.

paragbhingre: In our case also if user doesn't choose the active target group then the error will re-appear once user re-opens the SG wizard.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
paragbhingre: Considering the current implementation where in the error will disappear only when user will choose the active target group from the dropdown we don't need this param on the state, so removed in the latest commit.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
paragbhingre: @piradeepk As per the latest commit, we are hiding the error message only when user chooses the active target group.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -523,6 +524,16 @@ export class EcsServerGroupConfigurationService {
       }
     }
 
+    if (currentTargetGroups) {
+      const matched = intersection(allTargetGroups, currentTargetGroups);
+      const removedTargetGroups = xor(matched, currentTargetGroups);
+      if (removedTargetGroups && removedTargetGroups.length > 0) {
+        command.viewState.dirty.targetGroups = removedTargetGroups;
+      } else if(command.viewState.dirty && command.viewState.dirty.targetGroups) {
+        delete command.viewState.dirty.targetGroups;
```
piradeepk, line 533: I'm not really sure what the impact of delete is here (are there potential negative repercussions to delete?), would it be better to set command.viewState.dirty.targetGroups = [] instead?

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -72,7 +72,7 @@ export class Container extends React.Component<IContainerProps, IContainerState>
       this.setState({
         dockerImages: this.props.command.backingData.filtered.images,
         targetGroupsAvailable: this.props.command.backingData.filtered.targetGroups,
-      });
+        });
```
piradeepk, line 75: nit spacing

piradeepk (commented):
Looks good overall, just had a couple questions!

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
paragbhingre: So deleting the target group from the viewstate.dirty is the convention that we are following in rest of the code base. But setting it to [] also makes sense in order to just empty the dirty target groups. I will change it to set target groups to [] instead of deleting them all together.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -152,13 +153,18 @@ export class Container extends React.Component<IContainerProps, IContainerState>
     this.setState({ targetGroupMappings: currentMappings });
   };
 
+  private updateDirtyTargetGroups = () => {
+    this.props.command.viewState.dirty.targetGroups = null;
```
piradeepk, line 157: Should this be null or an empty array?

paragbhingre: In my latest commit it's an empty array.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/container/Container.tsx
```diff
@@ -152,13 +153,18 @@ export class Container extends React.Component<IContainerProps, IContainerState>
...
+  };
+
   public render(): React.ReactElement<Container> {
     const removeTargetGroupMapping = this.removeTargetGroupMapping;
     const updateContainerMappingImage = this.updateContainerMappingImage;
     const updateTargetGroupMappingTargetGroup = this.updateTargetGroupMappingTargetGroup;
     const updateTargetGroupMappingPort = this.updateTargetGroupMappingPort;
     const updateComputeUnits = this.updateComputeUnits;
     const updateReservedMemory = this.updateReservedMemory;
+    const dirtyTagetGroups = this.props.command.viewState.dirty && this.props.command.viewState.dirty.targetGroups ? this.props.command.viewState.dirty.targetGroups : [];
```
piradeepk, line 167: Wouldn't dirty.targetGroups be either populated or an empty array? could you do something like this instead:

```
const dirtyTagetGroups = this.props.command.viewState.dirty.targetGroups

paragbhingre: Sure, I will take care of this.

paragbhingre: @piradeepk I tried implementing your suggestion but we have kept `targetGroups` as not required (?:) property on `IEcsServerGroupCommandDirty` which makes it to be undefined when it is not yet discovered. If it is undefined then this code breaks and throws **property of undefined** error in later parts of the code. So it is better to have this check at the very start to avoid errors. 
LMK your thought on this.

piradeepk: Sounds good! :+1:

piradeepk (approved):
Ship it! 💯

piradeepk:
@paragbhingre can you add a screenshot in the description for a working target group?

PR #8866, opened 2021-01-26
fix(core/cluster): Clicking an instance breaks rendering of cluster pods

caseyhebebrand:
The `AutoSizer` was only listening to a window resize, but when an instance or server group is selected, rows also need to be resized. This caused the server group views to render on top of each other once the details panel was opened.

This PR adds a subscription to listen for opening/closing the details panel so that the cluster pods can re-render with appropriate heights.


alanmquach (approved):
Nice catch

PR #8870, opened 2021-01-27
feat(core): allow custom stage summary as React components

anotherchrisberry:
Developers can customize the stage summary for a given stage, but only in AngularJS. This provides a way to customize the stage summary with a React component, instead.


alanmquach (approved):
Nice

PR #8877, opened 2021-01-30
feat(core) Expose raw subnet ID

alanmquach:
Data is already there so, might as well make it accessible:
Screen Shot 2021-01-29 at 4 30 46 PM


lorin (approved):
:shipit:

PR #8878, opened 2021-01-30
feat(ecs): Determine invalid capacity providers

paragbhingre:
This PR fixes [#6317](https://github.com/spinnaker/spinnaker/issues/6317)

For the ECS Provider in deck, an Alert have been added when there is capacity provider(s) that was previously selected and is no longer available.

Dirty Capacity Provider

Screen Shot 2021-01-29 at 8 12 24 PM

Active Capacity Providers

Custom CP

### DIFF
--- a/app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
+++ b/app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
@@ -46,6 +46,7 @@ import { IServiceDiscoveryRegistryDescriptor } from '../../serviceDiscovery/ISer
 
 export interface IEcsServerGroupCommandDirty extends IServerGroupCommandDirty {
   targetGroups?: string[];
+  capacityProviders?: string[];
 }
 
 export interface IEcsServerGroupCommandResult extends IServerGroupCommandResult {
@@ -304,6 +305,7 @@ export class EcsServerGroupConfigurationService {
       command.backingData.filtered.availableCapacityProviders = [];
       command.backingData.filtered.defaultCapacityProviderStrategy = [];
     }
+    this.checkDirtyCapacityProviders(command);
   }
 
   public configureAvailableCapacityProviders(command: IEcsServerGroupCommand): void {
@@ -623,6 +625,21 @@ export class EcsServerGroupConfigurationService {
       command.vpcId = subnet ? subnet.vpcId : null;
     }
     return result;
+  }
+
+  public checkDirtyCapacityProviders(command: IEcsServerGroupCommand): void {
+    if(command.capacityProviderStrategy){
+      const availableCapacityProviders = command.backingData.filtered.availableCapacityProviders;
+      const currentCapacityProviders = command.capacityProviderStrategy.map(cp => cp.capacityProvider);
+      const matched = intersection(availableCapacityProviders, currentCapacityProviders);
+      const removedCapacityProviders = xor(matched, currentCapacityProviders);
+
+      if (removedCapacityProviders && removedCapacityProviders.length > 0) {
+        command.viewState.dirty.capacityProviders = removedCapacityProviders;
+      } else if(command.viewState.dirty && command.viewState.dirty.capacityProviders) {
+        command.viewState.dirty.capacityProviders = [];
+      }
+    }
   }
 
   public attachEventHandlers(cmd: IEcsServerGroupCommand): void {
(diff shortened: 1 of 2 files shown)


paragbhingre:
@piradeepk @allisaurus  Please review this PR.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -625,6 +627,21 @@ export class EcsServerGroupConfigurationService {
     return result;
   }
 
+  public checkDirtyCapacityProviders(command: IEcsServerGroupCommand): void {
+    if(command.capacityProviderStrategy && command.backingData.filtered.availableCapacityProviders){
```
piradeepk, line 631: Shouldn't we also check this, if there are no available capacity providers? (if you delete all capacity providers associated to your account, you can still have previously selected an invalid one)

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -104,6 +105,7 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
   private updateCapacityProviderType = (targetCapacityProviderType: boolean) => {
     this.setState({useDefaultCapacityProviders : targetCapacityProviderType});
     this.props.notifyAngular("useDefaultCapacityProviders", targetCapacityProviderType);
+    this.props.command.viewState.dirty.capacityProviders = [];
```
piradeepk, line 108: What happens if there isn't a default cp defined, but I click the default cp toggle? (will it clear the message, and is that actually accurate?)

paragbhingre: Yes, in any case if you toggle from custom to default this message will disappear because we display default CPs coming from the API response.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
paragbhingre: Actually, `command.backingData.filtered.availableCapacityProviders` also holds true even if length of the availableCapacityProviders is zero which takes care of your concern. But if you want I can remove that condition to avoid the confusion. LMK your preference. @piradeepk

allisaurus (approved):
code LGTM, approved pending @piradeepk 's questions are answered to his satisfaction. Thanks for providing screenshots up front, that is always a big help :D

Can you please rebase from master? Also, for future reference I would consider this type of change a "feat"(ure) b/c its a new UX improvement vs. addressing something that was broken before.

app/scripts/modules/ecs/src/serverGroup/configure/serverGroupConfiguration.service.ts
piradeepk: Understood, my concern was if availableCapacityProviders somehow came back as null, this would not run, but the cp would be invalid.

paragbhingre: Than removing `command.backingData.filtered.availableCapacityProviders` will resolve the issue.

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
piradeepk: Is it possible to delete a defined default capacity provider, and could we run into the same issue there?

paragbhingre: When we choose custom capacity provider and if dirty capacity provider exist then we show an error and ask user to select the different CP from the dropdown. But at this point dirty CP is not a part of this dropdown options. Similarly, in default CP case we do not show already deleted CP (which are dirty in this case) hence there is no need of showing the error. User just have to click the **Done** button and pipeline will have the latest default CPs without any dirty one.

piradeepk: I'm thinking more from the perspective that I'm a user, and I click edit existing server group, I scroll down, and don't see any issues, so I click cancel. As a user, how would I be aware that the default capacity provider was deleted? Also, what happens if I toggle from capacity providers to default and then back to capacity providers (does it clear the error message)?

paragbhingre: When you toggle from default to custom capacity provider then you need to set a new capacity provider strategy. In that case error won't be shown as you are setting a new CP and dirty CP is not a part of dropdown options.
In the case of default CP we need to figure out how do we indicate that something has been deleted and they need to click done instead of cancel.

piradeepk (approved):
Ship it! 💯

PR #8879, opened 2021-02-02
feat(dev): Added webpack hot reloading

ranihorev:
Hot reloading allows us to update styles without refreshing and speeds up the dev process.
This feature is disabled in production.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -39,6 +39,10 @@ function configure(env, webpackOpts) {
     }),
   ];
 
+  if (process.env.NODE_ENV !== 'production') {
+    plugins.push(new webpack.HotModuleReplacementPlugin());
+  }
+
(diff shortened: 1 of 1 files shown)


vigneshm:
lgtm, would like to hear @christopherthielen's opinion.

vigneshm (approved):

erikmunson (approved):
🥳 lgtm, welcome to the team!

christopherthielen (approved):
sure, let's give it a shot!  My understanding is that this won't enable HMR for JS without turning more dials, is that right?

ranihorev:
> sure, let's give it a shot! My understanding is that this won't enable HMR for JS without turning more dials, is that right?

Yes, you're right. We can use react-hot-reload or react-fast-refresh for that

PR #8880, opened 2021-02-02
feat(md): Collasible metadata elements 

ranihorev:
Limit the height of metadata elements on the ENVIRONMENT page. This is useful for long commit messages.

Kapture 2021-02-01 at 11 52 01

This can be moved and used as a generic element if needed.


caseyhebebrand:
Overall, this looks good. I was thinking it would be cool to pull this out into `core/presentation`, since it seems like a good candidate for reuse outside of MD. Then we

ranihorev:
> Overall, this looks good. I was thinking it would be cool to pull this out into `core/presentation`, since it seems like a good candidate for reuse outside of MD. Then we

No problem. I'll have to move the height prop outside of the css file to make it more flexible

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
```diff
@@ -206,9 +206,46 @@ const EnvironmentCards = memo(
...
+  const contentRef = React.useRef<HTMLDivElement>(null);
+
+  const checkIsOverflowing = React.useCallback(() => {
+    if (!contentRef.current) return;
+    setIsOverflowing(contentRef.current.offsetHeight < contentRef.current.scrollHeight);
+  }, []);
+
+  React.useEffect(() => {
+    checkIsOverflowing();
+  }, [children, checkIsOverflowing]);
```
erikmunson, line 221: i haven't personally seen this children-as-hook-dep pattern in the past, but trust that you've tested it and it works as you intend. curious to hear if others have input on this as a pattern.

erikmunson (approved):
lgtm overall! i had a comment that is mostly about my lack of experience with the way you've used `children` here, but don't consider that merge blocking.

only one question re: the visual presentation. In the gif you shared (thanks for the visual!) I noticed that when it was expanded the text content seemed to bump _right_ up against the 'read less' border line. Is that actually what it looks like in practice? if so, my instinct would be to add some whitespace under the text content so it doesn't collide with the border.

nice work!

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
ranihorev: It's not very common, however, in practice, `children` is just another prop. I can call it `text` instead :)

ranihorev:
> only one question re: the visual presentation. In the gif you shared (thanks for the visual!) I noticed that when it was expanded the text content seemed to bump _right_ up against the 'read less' border line. Is that actually what it looks like in practice? if so, my instinct would be to add some whitespace under the text content so it doesn't collide with the border.
> 
> nice work!

The "Read Less" text currently overlays the content when you hover over the text. I can add some padding there.

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
erikmunson: i don't think it's worth changing personally, makes sense as is. was just not 100% sure what to expect from `children` re: reference equality behavior

caseyhebebrand (commented):
LGTM - let me know when you are ready to merge and I can help with that.

app/scripts/modules/core/src/managed/ArtifactDetail.tsx
christopherthielen: I have the same question about this pattern.  I suspect the intent is to re-check if the content is overflowing whenever the contents change.  


Although this code definitely runs the effect whenever `children` changes, `children` object equality actually changes [_every time the parent component renders_](https://stackblitz.com/edit/react-ts-qpr7zb). so this effect gets run every time the parent renders.

ranihorev: That's true. We need to re-run this effect every time `children` changes. 
> children object equality actually changes every time the parent component renders.
It depends on the type of `children`. If it's a simple text input then it won't run every time, however, it'll re-run if `children` is a functional component.

PR #8881, opened 2021-02-03
fix(amazon): Remove conditional for enabling ipv6 by default

caseyhebebrand:
This conditional was causing a bug in updating the value. Also, if the account change is invalid, fallback to `undefined` (which is necessary on the backend to determine if this has been explicitly set or not).


app/scripts/modules/amazon/src/serverGroup/configure/wizard/pages/ServerGroupBasicSettings.tsx
```diff
@@ -103,9 +103,8 @@ export class ServerGroupBasicSettings
       AWSProviderSettings?.serverGroups?.enableIPv6 &&
       AWSProviderSettings?.serverGroups?.setIPv6InTest &&
       accountDetails.environment === 'test';
-    if (enableIPv6InTest) {
-      setFieldValue('associateIPv6Address', enableIPv6InTest);
-    }
+
+    setFieldValue('associateIPv6Address', enableIPv6InTest || undefined);
```
caseyhebebrand, line 107:
```suggestion
    setFieldValue('associateIPv6Address', enableIPv6InTest);
```

PR #8886, opened 2021-02-05
refactor(eslint): Lint rules to prevent presentation import from core

vigneshm:
If a presentation module is migrated to `@spinnaker/presentation` package, they should not be imported from `@spinnaker/core` anymore.


christopherthielen (approved):
nice work

PR #8887, opened 2021-02-05
feat(ecs): Determine invalid default capacity provider strategy

paragbhingre:
This PR fixes [#6330](https://github.com/spinnaker/spinnaker/issues/6330)

For the ECS provider in deck, add an Alert when there is a default capacity provider strategy that was previously set and is no longer available

Dirty default capacity Providers

Screen Shot 2021-02-10 at 9 45 26 AM

Active default capacity providers

Screen Shot 2021-02-05 at 12 38 37 PM


app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -123,19 +123,29 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
...
       return { label: `${capacityProviderNames}`, value: capacityProviderNames };
     }) : [];
 
-    const dirtyCapacityProviderList = dirtyCapacityProviders ? dirtyCapacityProviders.map(function (capacityProvider, index){
+    const defaultCPError = useDefaultCapacityProviders && defaultDirtyCapacityProviders ? (
+      <div className="alert alert-warning">
+        <p className="text-left">
+          <i className="fa fa-exclamation-triangle"></i>
+          Invalid capacity providers are a part of default capacity provider strategy.
+          <br/> Please click Done button to remove dirty capacity providers from the pipeline JSON.
```
allisaurus, line 138: I'd like to make this text explain the end state vs. what's going to be removed. e.g.,:
"Please click 'Done' to use current default capacity provider strategy (shown below) or switch to using a custom strategy"

app/scripts/modules/ecs/src/serverGroup/configure/wizard/capacityProvider/CapacityProvider.tsx
```diff
@@ -123,19 +123,29 @@ class EcsCapacityProvider extends React.Component<IEcsCapacityProviderProps, IEc
...
     const capacityProviderNames = this.state.availableCapacityProviders &&  this.state.availableCapacityProviders.length > 0 ?  this.state.availableCapacityProviders.map((capacityProviderNames) => {
       return { label: `${capacityProviderNames}`, value: capacityProviderNames };
     }) : [];
 
-    const dirtyCapacityProviderList = dirtyCapacityProviders ? dirtyCapacityProviders.map(function (capacityProvider, index){
+    const defaultCPError = useDefaultCapacityProviders && defaultDirtyCapacityProviders ? (
+      <div className="alert alert-warning">
+        <p className="text-left">
+          <i className="fa fa-exclamation-triangle"></i>
+          Invalid capacity providers are a part of default capacity provider strategy.
```
allisaurus, line 137: is it possible to show which capacity providers were invalid?

allisaurus (commented):
Overall LGTM, but I have a few questions:
1. the screenshot of the invalid default CP shows the error msg but doesn't list the invalid CPs which were found, even though it looks like we're building this list/component in the code. Am I mistaken, or would that be reflected in a newer screenshot?
2. what happens if the default CP _and_ the list of custom ones has invalid CPs. Is this possible, and what is the behavior? I imagine it would just show the option which is "active"

Please also rebase from master to ensure we're good there. Thank!

paragbhingre:
> Overall LGTM, but I have a few questions:
> 
> 1. the screenshot of the invalid default CP shows the error msg but doesn't list the invalid CPs which were found, even though it looks like we're building this list/component in the code. Am I mistaken, or would that be reflected in a newer screenshot?
> 2. what happens if the default CP _and_ the list of custom ones has invalid CPs. Is this possible, and what is the behavior? I imagine it would just show the option which is "active"
> 
> Please also rebase from master to ensure we're good there. Thank!

Consolidating what we will be implementing here:
To answer your first question, default capacity provider strategy can be invalid if default capacity provider strategy item has been removed from the cluster or capacity providers that were used in the default capacity provider strategy item are deleted/missing. So instead of showing two different messages to the user we are simply surfacing a generalized message that says they need to click done to use current default capacity provider strategy item(s) or they can use custom one as well.

To answer your second question, we have different lists for both dirty custom and dirty default capacity providers. So in any case if both of them have dirty capacity providers then they will surface an error to the user.

allisaurus:
@paragbhingre update looks good to me! please squash into 1 commit called "feat(ecs): Determine invalid default capacity provider strategy", and i'm good to approve.

allisaurus (approved):

PR #8888, opened 2021-02-07
fix(dev): load external source maps in development

ranihorev:
We used to get a bunch of warnings for missing source maps in development. Adding `source-map-loader` to extract those external source maps properly and reduce the noise.


christopherthielen:
@ranihorev Would you mind checking if there is a perf hit enabling this loader?

ranihorev:
> @ranihorev Would you mind checking if there is a perf hit enabling this loader?

AFAIK, it should not affect perf in any way as it doesn't affect the bundle js. This is the module I added:
https://webpack.js.org/loaders/source-map-loader/

christopherthielen:
>  it doesn't affect the bundle js.

I'm talking about perf for devs while running webpack-dev-server or generating bundles bundles

ranihorev:
> I'm talking about perf for devs while running webpack-dev-server or generating bundles

External packages don't change, therefore, it won't affect the rebuilds. I haven't seen any difference in the initial build time, I can dig into this more if you think that it matters.

christopherthielen:
>  I haven't noticed any difference in the initial build time , I can dig into this more if you think that it matters.

Please do, but nothing more involved than a before/after `yarn build`

ranihorev:
> > I haven't noticed any difference in the initial build time , I can dig into this more if you think that it matters.
> 
> Please do, but nothing more involved than a before/after `yarn build`

It seems to be 5-10 seconds faster with the new changes :)

christopherthielen:
okey dokey

christopherthielen (approved):

christopherthielen:
@ranihorev FWIW, you don't generally need to merge master into your PRs manually.  mergify will do it for you once the merge labels are applied.

PR #8889, opened 2021-02-08
feat(eslint): New eslint rule for ensuring import order

vigneshm:
Adding a new eslint rule to ensure a standardized ordering of imports. The following order will be maintained.

  1. npm package
  2. @spinnaker package
  3. relative path module
  4. css imports

Within each of the above partitions, individual import declarations will be sorted alphabetically based on the module/package name.

### DIFF
--- a/packages/eslint-plugin/rules/import-sort.js
+++ b/packages/eslint-plugin/rules/import-sort.js
@@ -1,0 +1,239 @@
+'use strict';
+// @ts-check
+
+const CSS_IMPORT = /\.(css|less|scss|sass)$/;
+const MODULE_PATH_REGEX = /[./]*(.*)$/;
+/**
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/rules/import-sort.js
```diff
@@ -0,0 +1,244 @@
...
+  if (!importSpecifiers) {
+    return importDeclaration;
+  }
+
+  importSpecifiers.sort((a, b) => {
+    a = a.imported.name;
+    b = b.imported.name;
+
+    return a > b ? 1 : a === b ? 0 : -1;
+  });
```
christopherthielen, line 63:
```suggestion
  importSpecifiers.sort((a, b) => a.imported.name.localeCompare(b.imported.name));
```

christopherthielen (approved):
woot!!

PR #8891, opened 2021-02-08
fix(serverGroup): Update capacity text to be more accurate

caseyhebebrand:
Adjusting some misleading text for the advanced mode of the `CapacitySelector` It is necessary to use the advanced mode if you want to use auto-scaling features, but it is not true that doing so sets up auto-scaling.


christopherthielen (approved):
nice, this was probably my misleading fault

PR #8894, opened 2021-02-09
fix(amazon): Stray 0 showing in component

caseyhebebrand:


christopherthielen (approved):
oh! This is caught me before, but TIL again

PR #8903, opened 2021-02-11
feat(md): Environment verification tasks

ranihorev:
This PR introduces the new verifications feature:
image

In addition, it removes styling a bit by removing some of the lines and polishing styles a bit

### DIFF
--- a/app/scripts/modules/core/src/managed/artifactDetail/ArtifactDetail.tsx
+++ b/app/scripts/modules/core/src/managed/artifactDetail/ArtifactDetail.tsx
@@ -1,8 +1,6 @@
 import React, { memo, useMemo } from 'react';
-import ReactGA from 'react-ga';
 import classNames from 'classnames';
 import { useRouter } from '@uirouter/react';
-import { useTransition, animated, UseTransitionProps } from 'react-spring';
 import { DateTime } from 'luxon';
 
 import {
@@ -31,7 +29,9 @@ import { showMarkArtifactAsBadModal } from './MarkArtifactAsBadModal';
 import { ConstraintCard } from './constraints/ConstraintCard';
 import { isConstraintSupported } from './constraints/constraintRegistry';
 import { isResourceKindSupported } from '../resources/resourceRegistry';
+import { VerificationCard } from './verifications/VerificationCard';
 
+import { useLogEvent } from '../utils/logging';
 import './ArtifactDetail.less';
 
 const SUPPORTED_PRE_DEPLOYMENT_TYPES = ['BUILD', 'BAKE'];
@@ -40,41 +40,6 @@ function shouldDisplayResource(reference: string, resource: IManagedResourceSumm
   return isResourceKindSupported(resource.kind) && reference === resource.artifact?.reference;
 }
 
-const logEvent = (label: string, application: string, environment: string, reference: string) =>
-  ReactGA.event({
-    category: 'Environments - version details',
-    action: label,
-    label: `${application}:${environment}:${reference}`,
-  });
-
-const inStyles = {
-  opacity: 1,
-  transform: 'scale(1.0, 1.0)',
-};
-
-const outStyles = {
-  opacity: 0,
-  transform: 'scale(0.95, 0.95)',
-};
-
-const cardTransitionConfig = {
-  from: outStyles,
-  // KLUDGE: all we're *actually* doing in this scary looking handler
-  // is delaying the start of any enter transitions for a fixed time
-  // so parent transitions for the overall layout have time to start.
-  // Unfortunately today useTransition doesn't support fixed delays
-  // without tapping into this promise-based orchestration feature (ew).
-  // When react-spring v9 is released, this can be changed
-  // to a function that returns { to: inStyles, delay: 180 }
-  enter: () => async (next: (_: React.CSSProperties) => any) => {
-    await new Promise((resolve) => setTimeout(resolve, 180));
-    next(inStyles);
-  },
-  leave: outStyles,
-  trail: 40,
-  config: { mass: 1, tension: 600, friction: 40 },
-} as UseTransitionProps<JSX.Element, React.CSSProperties>;
-
 type IEnvironmentCardsProps = Pick<
   IArtifactDetailProps,
   'application' | 'reference' | 'version' | 'allVersions' | 'resourcesByEnvironment'
@@ -82,6 +47,8 @@ type IEnvironmentCardsProps = Pick<
   environment: IManagedArtifactVersionEnvironment;
   pinnedVersion: string;
 };
+
+const LOG_CATEGORY = 'Environments - artifact details';
 
 const EnvironmentCards = memo(
   ({
@@ -108,6 +75,8 @@ const EnvironmentCards = memo(
     const {
       stateService: { go },
     } = useRouter();
+
+    const logEvent = useLogEvent(LOG_CATEGORY);
 
     const differentVersionPinnedCard = pinnedVersion &&
       pinnedVersion !== versionDetails.version &&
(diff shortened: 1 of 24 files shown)


erikmunson:
🥳 Very cool! I have no reviewed the code yet, but a couple small thoughts on visuals/UX:

- Re: the removal of the divider lines, I will defer to @gcomstock's opinion
- You're using the `3m 3s` timestamp formatting for the duration. If you go with that here, can you please also update the way the duration is formatted on the pre-deployment steps (build, bake) to match? Right now they say things like 'built in 8 minutes'
- I believe we originally mocked up the verification card _above_ the current deployment status (i.e. the decommissioned card in your screenshot), because it runs after deployment and we wanted at least a pseudo-chronological feed of activity. I see here that you've got it below the deployment status. @gcomstock any opinions on that? I think in a perfect world we'd have an actual chronological list of events/activities here but the API data isn't modeled that way as of right now

luispollo:
> * I believe we originally mocked up the verification card _above_ the current deployment status (i.e. the decommissioned card in your screenshot), because it runs after deployment and we wanted at least a pseudo-chronological feed of activity. I see here that you've got it below the deployment status. @gcomstock any opinions on that? I think in a perfect world we'd have an actual chronological list of events/activities here but the API data isn't modeled that way as of right now

@erikmunson but in this particular example, the deployment has been superseded/decommissioned, which would have happened _after_ a successful deployment and possibly (probably?) after a successful verification, so perhaps the ordering there needs to be a bit more nuanced? I agree that this should generally display in reverse chronological order, which is why I'm asking.

ranihorev:
> You're using the 3m 3s timestamp formatting for the duration.

I updated it there as well :)

> I believe we originally mocked up the verification card above the current deployment status

I think that sorting by start time would be great (rows without start time will be at the bottom)

gcomstock:
Hi! Sorry, I took yesterday off.

Try removing the divider lines, sure! They're easy to bring back if needed. I'd have a better opinion if I saw it different contexts.

IMO having verification in progress and verification passed is redundant. The in progress card can be updated with the outcome.

conversely, I think there **is** value in showing when there was a successful deployment, and when there was a decommission, so both of these cards, in their chronological order, would give me a richer story of the artifact. In this case, the successful deployment card would take on the same "past tense" grayed out color scheme, though. 

Let me know y'alls thoughts.

app/scripts/modules/core/src/managed/artifactDetail/ArtifactDetail.tsx
```diff
@@ -152,22 +121,31 @@ const EnvironmentCards = memo(
...
+          replacedBy={replacedBy}
+          vetoed={vetoed}
+          compareLink={compareLink}
+          allVersions={allVersions}
+          logClick={(action) => logEvent({ action, label: `${environmentName}:${reference}` })}
+        />
+        {environment.verifications?.map((verification) => (
+          <VerificationCard
+            key={verification.id}
+            {...verification}
```
christopherthielen, line 143: Since `VerificationCard` renders a Verification, why not just pass a single prop `verification={verification}`?  Is there a benefit to using spread here?

My preference would be either to be explicit about what properties of `verification` are being used (`foo={verification.foo} bar={verification.bar}`) or pass the whole object in as a single prop.

app/scripts/modules/core/src/managed/artifactDetail/verifications/VerificationCard.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+
+import { IVerification } from '../../../domain';
```
christopherthielen, line 3: All the top level modules have aliases in webpack and typescript that should be used.
There are linter rules that I think should have flagged this 🤔 


```suggestion
import { IVerification } from 'domain/IManagedEntity';
```

app/scripts/modules/core/src/utils/timeFormatters.spec.ts
```diff
@@ -73,6 +73,23 @@ describe('Filter: timeFormatters', function () {
       });
     });
 
+    fdescribe('timeDiffToString', function () {
```
christopherthielen, line 76:
```suggestion
    describe('timeDiffToString', function () {
```

packages/presentation/src/icons/iconsByName.ts
```diff
@@ -99,6 +99,7 @@ import { ReactComponent as mdUnknown } from './vectors/mdUnknown.svg';
 import { ReactComponent as mdConstraintGeneric } from './vectors/mdConstraintGeneric.svg';
 import { ReactComponent as mdConstraintDependsOn } from './vectors/mdConstraintDependsOn.svg';
 import { ReactComponent as mdConstraintAllowedTimes } from './vectors/mdConstraintAllowedTimes.svg';
+import { ReactComponent as mdVerification } from './vectors/mdVerification.svg';
```
christopherthielen, line 102: don't forget to remove this stuff in `/packages/presentation` if it's already been committed in the other PR

christopherthielen (commented):
Nice work, I had a couple comments/questions and a request

ranihorev:
> IMO having verification in progress and verification passed is redundant. The in progress card can be updated with the outcome.

@gcomstock, the screenshot shows an example of two verifications. We're working on adding names to the API to be able to identify them.

app/scripts/modules/core/src/managed/artifactDetail/ArtifactDetail.tsx
ranihorev: I don't have a strong preference here, I usually prefer to have flat props when possible. I'll change it

app/scripts/modules/core/src/managed/utils/ApplicationContext.tsx
```diff
@@ -0,0 +1,8 @@
+import { Application } from 'core/application';
+import React from 'react';
+
+export const ApplicationContext = React.createContext<Application | null>(null);
+
+export const ApplicationContextProvider: React.FC<{ app: Application }> = ({ app, children }) => {
```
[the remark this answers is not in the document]
ranihorev: Sure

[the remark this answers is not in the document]
ranihorev: I wonder if it makes sense to put it in hooks (it's not a hook) :)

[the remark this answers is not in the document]
ranihorev: I think that it'd be better to place it next to the ApplicationComponent

app/scripts/modules/core/src/utils/timeFormatters.spec.ts
ranihorev: nice catch :)

app/scripts/modules/core/src/managed/utils/ApplicationContext.tsx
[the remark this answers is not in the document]
christopherthielen: yeah, brain fart, I was going to suggest exposing the context object as a `useApplication()` hook, but then deleted that part of the comment

app/scripts/modules/core/src/managed/artifactDetail/verifications/VerificationCard.tsx
christopherthielen: welp, looks like I spoke untrue words.  in fact, `domain` isn't aliased but `core` is and should be accessed at `core/domain`

```
import { IVerification } from 'core/domain/IManagedEntity';

```

lorin:
> @gcomstock, the screenshot shows an example of two verifications. We're working on adding names to the API to be able to identify them.

I suggest displaying the id field until we have proper name support (for example: `test-container:spkr/ballast:stable`).

christopherthielen (approved):

gcomstock:
Thanks for the clarifications!

PR #8909, opened 2021-02-11
feat(core): Automatically save notifications when closing modal

caseyhebebrand:
When adding a new notification in app config, it is confusing for users to have to open a modal, save the modal, _and_ save the config section to complete the add/edit. This leads to users navigating away from the page before actually persisting this info, and being upset/confused when they don't get a notification in the future. 

This PR updates this user flow so that the new/updated notification rule is automatically saved when the modal is closed.
autoSaveNotifications


christopherthielen:
Couple things: 

1) we should really block the user (and show a confirmation dialog) if they navigate away from the config screen with unsaved changes
2) the dialog should wait for the save operation to complete before closing (like the "edit application attributes" modal)
3) the two save buttons is rather wonky

I think we can punt on 1 & 3 (it's been like this forever) but could you please wire in #2?

caseyhebebrand:
> I think we can punt on 1 & 3 (it's been like this forever) but could you please wire in #2?

Yeah I will work on it. I think I have to re-wire the component to be more promise friendly. Its set up kind of weird, which is why I just put it after the existing promise chain.

app/scripts/modules/core/src/notification/modal/EditNotificationModal.tsx
```diff
@@ -12,13 +12,29 @@ export interface IEditNotificationModalProps extends IModalComponentProps {
...
+  }
+
   private formikRef = React.createRef<Formik<any>>();
 
   private submit = (values: INotification): void => {
-    this.props.closeModal(values);
+    this.setState({ isSubmitting: true });
+    this.props.editNotification(values).then(() => {
+      this.setState({ isSubmitting: false });
+      this.props.closeModal();
```
caseyhebebrand, line 36: fyi @christopherthielen, the modal should now close after the edits are made. The angular component was saving notifications async and the updates would eventually propagate. Now, `editNotification` and `saveNotifications` return promises so we can chain `closeModal` to the end.

christopherthielen (approved):

PR #8911, opened 2021-02-12
feat(deploy/blockDevices): Adds a Volume section in the Deploy wizard

jorgebee65:
This feature adds a Volume section in the Deploy Wizard which allows to add one ore more volumes to an instance and also add Tags which applies to the list of volumes. Dependency on this other [PR](https://github.com/spinnaker/clouddriver/pull/5244)

volumes


app/scripts/modules/amazon/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -120,6 +133,7 @@ export class AwsServerGroupConfigurationService {
     'GroupTotalInstances',
   ];
   private healthCheckTypes = ['EC2', 'ELB'];
+  private volumeTypes = ['gp2', 'st1', 'io1', 'sc1'];
```
iandelahorne, line 136:
```suggestion
  private volumeTypes = ['gp2', 'gp3', 'st1', 'io1', 'io2', 'sc1'];
```

`gp3` is the new general-purpose volume type from AWS. This is only available if you're using LaunchTemplates. 
`io2` is their new provisioned IO volume type, only available in preview currently on r5b instances.
 (see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html)

packages/amazon/src/serverGroup/configure/wizard/pages/volumes/ServerGroupVolumes.tsx
```diff
@@ -0,0 +1,71 @@
...
+            formik={this.props.formik}
+          />
+        </div>
+        <div className="form-group">
+          <div className="sm-label-left">
+            <b>Tags (optional)</b>
+            <HelpField id="aws.serverGroup.blockDevice.tags" />
+          </div>
+          <MapEditor model={blockTags as any} allowEmpty={true} onChange={this.tagsChanged} />
+        </div>
```
christopherthielen, line 67: This should use a `FormikFormField` and a `MapEditorInput` which would also make the `validate` and `tagsChanged functions unnecessary

packages/amazon/src/serverGroup/configure/wizard/pages/volumes/VolumeEditor.tsx
```diff
@@ -0,0 +1,131 @@
...
+  public validate(
+    values: IBlockDevicesCommand,
+  ): FormikErrors<IBlockDevicesCommand> {
+
+    const errors = {} as any;
+    const { blockDevices } = values;
+    console.log(blockDevices)
+    return errors;
+  }
+
```
christopherthielen, line 31: I don't think this function gets called ever, and it is also a noop, so might as well remove it

christopherthielen (changes requested):
thanks for migrating this to Formik.  I have a couple minor requests then I think it's good to go.

PR #8916, opened 2021-02-13
fix(core/serverGroup): Ensure failsafe SpEL expressions are provided

vigneshm:
While creating the deployment payload, it is possible to provide SpEL expressions for stack, detail etc and if these expressions evaluate to invalid values, then that could lead to deployment failures. This commit introduces a warning in the server group create/clone wizard when failsafe SpEL expressions are not provided.

### DIFF
--- a/app/scripts/modules/core/src/serverGroup/configure/common/ServerGroupNamePreview.tsx
+++ b/app/scripts/modules/core/src/serverGroup/configure/common/ServerGroupNamePreview.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { isEmpty } from 'lodash';
 
 export interface IServerGroupNamePreviewProps {
   createsNewCluster: boolean;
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/serverGroup/configure/common/ServerGroupNamePreview.tsx
```diff
@@ -8,13 +9,38 @@ export interface IServerGroupNamePreviewProps {
...
+    return (
+      <div className="form-group">
+        <div className="col-md-12">
+          <div className={`well-compact ${anyInvalidSpel ? 'alert alert-warning' : 'well'}`}>
+            <h5 className="text-center">
+              <p>Your server group cluster will be determined by evaluating the expressions dynamically</p>
+              {anyInvalidSpel && (
+                <div className="text-left">
+                  NOTE: Please wrap your expression with #alphanumerical and provide a default value to make sure only
+                  valid values are used after evaluation. i.e {"${#alphanumerical(yourExpression) ?: 'defaultValue'}"}
```
vigneshm, line 34: Please let me know if this could be worded better.

christopherthielen:
What exactly is this guarding against?  Would it be better to validate server side in the deploy stage that the server group name is valid?

vigneshm:
> What exactly is this guarding against? Would it be better to validate server side in the deploy stage that the server group name is valid?

@cfieber is adding server side validations. This would prompt the user to provide the default value that we can use if the expression evaluates to something invalid.

vigneshm:
It also serves as an FYI to the users providing SpEL expressions on how the actual outcome could be different from the expected outcome.

caseyhebebrand (approved):

PR #8919, opened 2021-02-17
refactor(core): ConsoleOutput link and modal

caseyhebebrand:
The console output on the instance details panel is the last thing that need to be reactified before converting the entire instance details panel. Breaking into commits will probably be easier: 

1. [Refactor console log modal to react](https://github.com/spinnaker/deck/commit/26b0f3fb51de1f3e39e8eec2721e224b8c8a9c55)
2. [Use new component in instance details for all cloud providers](https://github.com/spinnaker/deck/commit/be28a204c9dafc11343822ca4a1fd8a550b93c59)
3. [Remove angular files](https://github.com/spinnaker/deck/commit/4b0edd970898ae320cb5b189b78b991f07ddd247)


app/scripts/modules/azure/src/instance/details/instance.details.controller.js
```diff
@@ -345,19 +344,6 @@ module(AZURE_INSTANCE_DETAILS_INSTANCE_DETAILS_CONTROLLER, [UIROUTER_ANGULARJS,
...
-          controller: 'ConsoleOutputCtrl as ctrl',
-          size: 'lg',
-          resolve: {
-            instance: function () {
-              return $scope.instance;
-            },
-          },
-        });
-      };
-
```
caseyhebebrand, line 360: I noticed that this function was not used anywhere within the html so I removed it. The only reference to `showConsoleOutput` is in the new `ConsoleOutput` component. If you know anyone who works with azure, feel free to tag them.

christopherthielen:
Did you test this? It doesn't seem to work for me.

caseyhebebrand:
> Did you test this? It doesn't seem to work for me.

Yes, of course. I was working with it last night and it was fine, but I made some changes this morning. I need to update the component name though for backward compatibility so I will take another look.

caseyhebebrand:
> Did you test this? It doesn't seem to work for me.

I just realized I did not push the react2angular converter so that is probably why.

app/scripts/modules/core/src/instance/details/console/ConsoleOutputModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+    );
+  };
+
+  const jumpToEnd = () => {
+    logContainerEnd.current.scrollIntoView({ behavior: 'smooth' });
+  };
+
+  React.useEffect(() => {
+    fetchLogs();
+  }, []);
```
christopherthielen, line 42: you could switch to `useData` to avoid manually managing some of that state

christopherthielen (approved):
Nice work

app/scripts/modules/core/src/instance/details/console/ConsoleOutputModal.tsx
caseyhebebrand: great idea thanks!

PR #8920, opened 2021-02-17
fix(lint): Handle non-import declaration nodes in import-sort rule

vigneshm:
Non-import declaration nodes found between the first and last import statements are currently just removed leading to unexpected behavior. In this commit, such nodes are collected and written back to the end of the import declarations.


vigneshm:
This fix is needed before running the fixer across `deck`.

christopherthielen (approved):
sure :)

PR #8921, opened 2021-02-17
feat(lint): Add prettier linter

vigneshm:
Adding prettier eslint plugin to ensure all checked-in code follows prettier formatting.


christopherthielen (approved):
Does this run prettier on the files, or does it just raise a linter error?

vigneshm:
> Does this run prettier on the files, or does it just raise a linter error?

It fixes the issues when --fix is used. Otherwise only reports the errors.

PR #8930, opened 2021-02-17
chore(eslint): Sort all import declarations

vigneshm:
All modules (except the ones under `core`) have been updated to have sort statements in the recommended order. There are some CI failures due to re-ordering import statements in `core`, so it is disabled temporarily as the root cause is being investigated.


christopherthielen (approved):
I'm sure this is fine.
Other PRs will likely need to resolve merge conflicts after this is merged, but there's no great way to avoid it.

vigneshm:
Looking into CI failures.

PR #8932, opened 2021-02-18
chore(core): strict null checking

ranihorev:
Enabling strict null checking globally introduces a lot of error, however, we can gradually enforce it everywhere


ranihorev:
arrgh, too much work

PR #8935, opened 2021-02-19
feat(titus/serverGroup): Enable ipv6 in test environment

caseyhebebrand:
Containers in test should have IPv6 auto-enabled.


app/scripts/modules/titus/src/serverGroup/configure/ServerGroupCommandBuilder.js
```diff
@@ -51,7 +51,9 @@ angular.module(TITUS_SERVERGROUP_CONFIGURE_SERVERGROUPCOMMANDBUILDER, []).factor
         targetHealthyDeployPercentage: 100,
         env: {},
         labels: {},
-        containerAttributes: {},
+        containerAttributes: {
+          'titusParameter.agent.assignIPv6Address': 'true',
```
vigneshm, line 55: Should this be `false` by default?

caseyhebebrand: We default to a test environment when a new command is built, so it should be `true`. If a user switches the account to something in another environment it will flip it to false.

vigneshm (approved):

PR #8936, opened 2021-02-19
chore(lint): Upgrade to latest @spinnaker/eslint-plugin

vigneshm:
Upgrading to latest `@spinnaker/eslint-plugin` package that ended up fixing the order for a number of files in `deck`.


app/scripts/app.ts
```diff
@@ -1,4 +1,5 @@
 ///<reference path="./modules/core/src/types/index.d.ts" />
+/* eslint-disable @spinnaker/import-sort */
```
vigneshm, line 2: Explicitly turning off to ensure existing import order that relies on side effects.

app/scripts/modules/core/src/core.module.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @spinnaker/import-sort */
```
vigneshm, line 1: Explicitly turning off to ensure existing import order that relies on side effects.

app/scripts/modules/core/src/presentation/forms/inputs/JsonEditor.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @spinnaker/import-sort */
```
vigneshm, line 1: Explicitly turning off to ensure existing import order that relies on side effects.

christopherthielen (approved):

PR #8938, opened 2021-02-19
fix(core): avoid fetching security groups twice

ranihorev:
If we're trying to fetch again while fetching the security groups, we now return a promise the will be resolved once the first get request returns. 

Future TODO: This feature is built-in in some packages. Consider using them instead :)

### DIFF
--- a/app/scripts/modules/core/src/securityGroup/securityGroupReader.service.ts
+++ b/app/scripts/modules/core/src/securityGroup/securityGroupReader.service.ts
@@ -89,7 +89,13 @@ export interface ISecurityGroupDetail {
   securityGroupRules: ISecurityGroupRule[];
 }
 
+type AwaitingResponseList = Array<
+  [resolve: (value: ISecurityGroupsByAccountSourceData) => void, reject: (reason: any) => void]
+>;
+
 export class SecurityGroupReader {
+  private isFetchingGroups = false;
+  private awaitingGroupsResponseList: AwaitingResponseList = [];
   private static indexSecurityGroups(securityGroups: IReaderSecurityGroup[]): ISecurityGroupsByAccount {
     const securityGroupIndex: ISecurityGroupsByAccount = {};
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/securityGroup/securityGroupReader.service.ts
```diff
@@ -296,18 +302,37 @@ export class SecurityGroupReader {
...
+        for (const [_, reject] of this.awaitingGroupsResponseList) {
+          reject(e);
+        }
+        throw new Error(e);
+      })
+      .finally(() => {
+        this.awaitingGroupsResponseList = [];
+        this.isFetchingGroups = false;
       });
   }
```
christopherthielen, line 337: I'll propose an alternate solution with less state management:

```suggestion
  private getAllSecurityGroupsPromise: PromiseLike<ISecurityGroupsByAccountSourceData>;

  public getAllSecurityGroups(): PromiseLike<ISecurityGroupsByAccountSourceData> {
    const cache = InfrastructureCaches.get('securityGroups');
    const cached = cache ? cache.get('allGroups') : null;
    if (cached) {
      return this.$q.resolve(this.decompress(cloneDeep(cached)));
    } else if (this.getAllSecurityGroupsPromise) {
      return this.getAllSecurityGroupsPromise;
    }

    this.getAllSecurityGroupsPromise = REST('/securityGroups')
      .get()
      .then((groupsByAccount: ISecurityGroupsByAccountSourceData) => {
        if (cache) {
          cache.put('allGroups', this.compress(groupsByAccount));
        }
        return groupsByAccount;
      })
      .finally(() => {
        delete this.getAllSecurityGroupsPromise;
      });

    return this.getAllSecurityGroupsPromise;
  }
```

ranihorev: That's nice! Thanks

christopherthielen (approved):

PR #8940, opened 2021-02-19
chore(lint): Run lint before commit

vigneshm:
Ensure eslint is run before commit so that eslint fixes are automatically applied.

Since lint-staged takes care of running commands for staged files, I'm removing pretty-quick and using prettier directly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -192,6 +192,7 @@
     "karma-webpack": "4.0.2",
     "less": "^2.7.2",
     "less-loader": "4.1.0",
+    "lint-staged": "^10.5.4",
     "loader-utils": "1.2.3",
     "md5": "^2.2.1",
     "minimist": "^1.2.3",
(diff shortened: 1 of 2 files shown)


christopherthielen:
@vigneshm one thing to consider is that some linter fixes I've written are not 100% guaranteed to be correct.

The worst case scenario is that the code will be committed in a broken state.

vigneshm:
> @vigneshm one thing to consider is that some linter fixes I've written are not 100% guaranteed to be correct.
> 
> The worst case scenario is that the code will be committed in a broken state.

Chatted offline with @christopherthielen. I'm going to work on disabling some of the problematic rules when auto fixer is run during pre-commit.

christopherthielen (approved):
To respond to myself, I believe the fixers that don't guarantee correct code after running are migration fixers which are unlikely to be triggered again (since the migrations are complete)

PR #8945, opened 2021-02-24
feat(core/notifications): Save app notifications when modal is submitted

caseyhebebrand:
**Current Behavior**
1. In application config the user clicks add/edit notification and a modal opens
2. The user makes changes and clicks "Update" in the modal, which updates the pipelin config locally
3. The modal closes
4. The user also has to click "Save" to persist these changes on the BE

Have to click and "Update" and "Save" often leads to confusion. The users will "Update" via teh modal then navigate away from the page before actually persisting the new information. The current user flow makes more sense for the pipeline and stage notifications, so this PR only updates the user flow for application-level notifications.

**New Behavior**
1. In application config the user clicks add/edit notification and a modal opens
2. The user makes changes and clicks "Update" in the modal, which persists the state
3. The modal closes and no other actions are needed

### DIFF
--- a/app/scripts/modules/core/src/notification/NotificationsList.tsx
+++ b/app/scripts/modules/core/src/notification/NotificationsList.tsx
@@ -22,6 +22,7 @@ export interface INotificationsListProps {
 }
 
 export interface INotificationsListState {
+  indexEdited: number;
   isNotificationsDirty: boolean;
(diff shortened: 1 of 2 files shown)


christopherthielen (approved):
Interesting, I didn't know we reused this for pipelines and application config

PR #8947, opened 2021-02-24
chore(lint): Remove lint rule for prettier

vigneshm:
IDEs run eslint eagerly and seeing prettier lint errors pop-up aggressively results in a poor dev experience. So removing the prettier linter here.


christopherthielen (approved):

christopherthielen:
There may be a better way to manage our plugin/default rules configuration and how it's exposed by the pluginsdk.  I'm definitely not married to this setup.

PR #8948, opened 2021-02-25
feat(core): global search box - search on enter

ranihorev:
Initiate search on enter. If an exact app match was found, open the app directly.
Bonus: replaced the deprecated "which" with "key"

### DIFF
--- a/app/scripts/modules/core/src/search/global/GlobalSearch.tsx
+++ b/app/scripts/modules/core/src/search/global/GlobalSearch.tsx
@@ -5,7 +5,6 @@ import React from 'react';
 import ReactGA from 'react-ga';
 import { Observable, Subject } from 'rxjs';
 
-import { IApplicationSearchResult } from 'core/application/applicationSearchResultType';
 import { Tooltip } from 'core/presentation/Tooltip';
 import { ReactInjector } from 'core/reactShims';
 import { ClusterState } from 'core/state';
@@ -17,6 +16,7 @@ import { IChildComponentProps, RecentlyViewedItems } from '../infrastructure/Rec
 import { ISearchResultSet } from '../infrastructure/infrastructureSearch.service';
 import { ISearchResult } from '../search.service';
 import { searchRank } from '../searchRank.filter';
+import { findMatchingApplicationResultToQuery, getSearchQuery as getSearchQueryParams } from './utils';
 
 const SLASH_KEY = '/';
 const MIN_SEARCH_LENGTH = 3;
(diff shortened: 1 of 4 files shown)


caseyhebebrand (approved):

christopherthielen:
I like this, but I have some suggestions about the behavior.  The global search already has keyboard navigation and I think it would be best to integrate with that UX.

1) Add visual cue w.r.t. what the enter key will do.  Pre-select the application using the keyboard navigation code.

2) Add visual cue w.r.t. redirecting to `/search?q=notanapp`.  Pre-select the "show all" button if no application is an exact match.

3) Redirecting to `/search?q=foo` when the search results are loaded is jarring/surprising.  (I suspect using the keyboard nav code will address this this)

4) After navigating to the app, unfocus the search (I suspect using the keyboard nav code will address this this too)

christopherthielen:
I see a couple other UX improvements you could make as long as you're in there (if you want):
1) The keyboard nav highlight color is really low contrast
2) Somebody (it was meeeee) made a poor choice... all the "show all" links do the exact same thing.  Maybe that should be a single link on the top of the results.

christopherthielen:
3) sometimes (it's difficult to reproduce) hitting enter submits the form.  We could add an `onSubmit` to the `<form>` to block submittal

app/scripts/modules/core/src/search/global/GlobalSearch.tsx
```diff
@@ -119,32 +120,41 @@ export class GlobalSearch extends React.Component<{}, IGlobalSearchState> {
...
       }
-    } else if (which === 13) {
-      // enter
-      // do not submit the form and reload the page
+    } else if (key === 'Enter') {
+      const { $state } = ReactInjector;
+      if (this.state.categories) {
+        const matchingQueryResult = findMatchingApplicationResultToQuery(this.state.categories, this.state.query);
+        // eslint-disable-next-line no-console
+        console.log('matching', matchingQueryResult);
```
christopherthielen, line 147:
```suggestion

```

app/scripts/modules/core/src/search/global/GlobalSearchResults.tsx
```diff
@@ -36,11 +33,15 @@ export const GlobalSearchResults = ({
     );
   }
 
+  const matchingApp = findMatchingApplicationResultToQuery(categories, query);
+  // eslint-disable-next-line no-console
+  console.log(matchingApp);
```
christopherthielen, line 38:
```suggestion
```

christopherthielen (approved):

PR #8954, opened 2021-03-01
fix(aws): Simplify security group extraction

pdk27:
Simplify security group extraction by removing [redundant logic ](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonServerGroup.groovy#L111)from Deck.


pdk27:
> Do you know if there are any cases where we would need to fall back to the `networkInterfaces`? If not, this LGTM and is much simpler than before

I can't think of any scenario where anything other than `securityGroups` will be needed in this particular React component.

With the addition of support for mixed instances policies, consolidating such redundant logic in Clouddriver API will prevent future changes from being error-prone, IMO.

PR #8957, opened 2021-03-02
fix(amazon/serverGroup): Move spelLoadBalancers and spelTargetGroups to viewState

christopherthielen:
When we switch between regions or accounts, the ui validates that the requested load balancers/target groups exist in the new account/region.  If the new region/account doesn't have the load balancer, it's removed from the UI and a message is shown.  However, spel expressions are treated differently and are never removed.  

This change moves the UI state from the command object to the viewState object so these values are not sent to the server


caseyhebebrand (approved):
LGTM

PR #8959, opened 2021-03-02
feat(core/ExecutionMarket): Extends component to customize status color by output status code

bbrina:
## What does this PR do?
This PR extends the functionality of the `ExecutionMarker` component to add different css classes based on a stage output status code.

The following example shows the UI changing the execution marker based on the status code 2 which means the stage succeeded but there was a diff.

This should be useful for users who run several pipelines at once so they can identify the status of a finished job with just a glance.

## Demo
Screen Shot 2021-03-02 at 15 30 39

### DIFF
--- a/app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
+++ b/app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
@@ -83,7 +83,7 @@ export class ExecutionMarker extends React.Component<IExecutionMarkerProps, IExe
 
   private getStageOutputStatusCode = (): number => {
     const { stage } = this.props;
-    return stage.masterStage.outputs.status && stage.masterStage.outputs.status.code;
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/pipeline/executions/execution/ExecutionMarker.tsx
```diff
@@ -81,9 +81,24 @@ export class ExecutionMarker extends React.Component<IExecutionMarkerProps, IExe
...
+  };
+
+  private getClassNameByStatusCode = (code: number): string => {
+    switch (code) {
+      case 2:
+        return 'success-with-diff';
+      default:
+        return '';
+    }
+  };
```
ethanfrogers, line 96: @bbrina - This is all very specific to our use case and doesn't belong in this repo. What can we do to extend `ExecutionMarker` in a way that allows us to support this specific use case in our extension of Deck?

PR #8960, opened 2021-03-02
feat(aws): Adding support to view details of an ASG with MixedInstancesPolicy(MIP)

pdk27:
## Summary
  - Add new section for instance diversification details
  - Update launch template details section with multiple instance types conditionally
  - Add data types for MixedInstancesPolicy(MIP) fields
  - Add help fields for MIP fields
  - Add / update tests
  - Add helpField to LabeledValue

Corresponding Clouddriver PR - https://github.com/spinnaker/clouddriver/pull/5268 
## Issue
https://github.com/spinnaker/spinnaker/issues/5989

## Related AWS docs:
[Launching ASG with MIP and it's options](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-purchase-options.html)
[MIP API Ref](https://docs.aws.amazon.com/autoscaling/ec2/APIReference/API_MixedInstancesPolicy.html)

## Screenshots 
<details>
<summary>ASG with MIP details - single instance type</summary>

</details>

<details>
<summary>ASG with MIP details - multiple instance types and instance diversification</summary>

</details>

<details>
<summary>Instance Diversification when SpotMaxPrice is empty / not specified i.e. default is On-Demand price for the configuration </summary>

</details>


app/scripts/modules/amazon/src/serverGroup/aws.less
```diff
@@ -0,0 +1,39 @@
+@import (reference) '~core/presentation/less/imports/commonImports.less';
+
+.collapsibleSubsectionOverride {
```
christopherthielen, line 3: Please scope this css using the name of the component that uses it, i.e.:
```
.InstancesDiversificationDetailsSection {
  .collapsibleSubsectionOverride {
   ... etc
```

christopherthielen (approved):
nice work, and clean coding

christopherthielen:
Once you've fixed CI and updated the CSS as requested, we can merge this

app/scripts/modules/amazon/src/serverGroup/aws.less
pdk27: The plan is to use `collapsibleSubsectionOverride` in more components in upcoming feature additions. That's the reason I kept it generic. 
In the current PR, it is used for multiple instance types. In near future, it will be used for fields with multiple properties - this will happen with [multiple launch templates](https://docs.aws.amazon.com/autoscaling/ec2/userguide/asg-launch-template-overrides.html). 

Do you still think it should be scoped?

caseyhebebrand (approved):

app/scripts/modules/amazon/src/serverGroup/aws.less
christopherthielen: > The plan is to use collapsibleSubsectionOverride in more components in upcoming feature additions.

I see, it makes sense why you write it that way.  I think my preference would be to encapsulate the CSS into a component (like `MyAwsCollapsibleSection` but with a name that explains what it does or why it exists).  If for some reason that is not possible, just rename the CSS class  to something more explicit (a top-level CSS class named `collapsibleSubsectionOverride` doesn't tell me much).

pdk27: Gotcha. I will give that a shot. Thanks.

pdk27: @christopherthielen renamed to `aws-multiple-values-subsection` - looks better now. Thanks.

app/scripts/modules/amazon/src/serverGroup/details/sections/multipleInstanceTypesSubSection.less
```diff
@@ -0,0 +1,50 @@
+@import '~bootstrap/less/mixins/clearfix.less';
```
pdk27, line 1: I couldn't get the build to succeed with importing `commonImports.less ` 🤷‍♀️ 
I am guessing modules other than core should be importing from core and not dependencies directly? Any suggestions to keep this clean?

app/scripts/modules/amazon/src/serverGroup/details/sections/MultipleInstanceTypesSubSection.tsx
```diff
@@ -0,0 +1,40 @@
+import { IAmazonLaunchTemplateOverrides } from 'amazon';
+import React from 'react';
+
+import { CollapsibleSection, HelpField } from '@spinnaker/core';
+import './multipleInstanceTypesSubSection.less';
```
pdk27, line 5: I renamed the css to the specific component name as per suggestions. 
This might need to be renamed in the future as it finds more uses, fyi.

caseyhebebrand (approved):

PR #8961, opened 2021-03-02
fix(appengine/docker): Allow save the Artifact in the Pipeline JSON f…

jorgebee65:
Deploying an App Engine. 

Deploy stage -> When the Docker Container is selected and an image is set the Artifact information it was not been saved on JSON Format. This fix allows to save the Artifact Information as **expectedArtifact** attribute

app engine docker


zachsmith1 (approved):
LGTM

german-muzquiz (approved):

PR #8962, opened 2021-03-03
fix(md): Remove allow-list of events in resource history + redesign

ranihorev:
We moved the list of events to the backend to avoid maintaining them on both sides, as it caused us to drop some of the events that weren't on the frontend. Here's the backend PR: https://github.com/spinnaker/keel/pull/1835

I also removed the "Where" column because it was showing the same value for all rows.

New version: 
- We show the absolute time when hovering over the date.
- The icons are not perfect but I tried to find matching ones from what we have :)
image

Old version:
image


ranihorev:
@gcomstock what do you think?

app/scripts/modules/core/src/managed/resourceHistory/ManagedResourceHistoryModal.tsx
```diff
@@ -31,170 +24,23 @@ export interface IManagedResourceHistoryModalProps extends IModalComponentProps
...
-    () => ManagedReader.getResourceHistory(id).then((events) => mergeNewEvents(events, previousHistoryEvents)),
+    () => ManagedReader.getResourceHistory(id),
     null,
     EVENT_POLLING_INTERVAL,
     [],
   );
-  const previousHistoryEvents: IManagedResourceEvent[] | null = usePrevious(historyEvents);
+
+  // eslint-disable-next-line no-console
+  console.log(historyEvents);
```
christopherthielen, line 43:
```suggestion
```

christopherthielen (approved):
Code LGTM (except for a console.log).  It looks like we've lost some granularity on the icon mapping by mapping against the `eventLevel`. 

If needed, perhaps in the future you could map from `ManagedResourceEventType` to icon and fall back to `eventLevelToIcon` if no icon has been mapped for a type.

ranihorev:
@christopherthielen we removed this granularity on purpose because we weren't maintaining it properly

PR #8964, opened 2021-03-03
fix(core/forms): MapEditor: make trash icon position: relative to avoid scrollbars

christopherthielen:
Before:

Kapture 2021-03-02 at 17 08 02

Note the entire modal is scrolling down all the way to the last `position: absolute` element (which is from the `class="sr-only"` bootstrap screenreader only css class)


christopherthielen:
While I was in there I also added CSS component scoping

alanmquach (approved):
Wow nice find!

PR #8967, opened 2021-03-04
fix(pipelines): Add deps to breadcrumbs useMemo

alanmquach:
The `parentExecutions` was `useMemo`'d with an empty deps array `[]` which means it is never invalidated.

This meant that when navigating from child to child to child, the parentExecutions never got recomputed and certain parts of the lineage would be missing.


christopherthielen (approved):

dreynaud:
Thanks @alanmquach 👏

PR #8970, opened 2021-03-04
feat(core): Allow function components to be overridden

caseyhebebrand:
React function components cannot use decorators, so the dev experience hasn't been ideal using `@Overridable` as we try and move toward function components. (the `OverridableComponent` renders a wrapper class component, which renders a functioning function component). 

This ended up being mostly a Typescript update, but now we can utilize the existing `overrideableComponent` function to produce an HOC, which is a better pattern for function components. 

What is in this PR:
- type updates
- A new function `overridesComponent` that is equivalent to the `@Overrides` decorator, but can be used for function components
- An example of utilizing `overrideableComponent` to create an HOC

### DIFF
--- a/app/scripts/modules/core/src/overrideRegistry/Overrides.tsx
+++ b/app/scripts/modules/core/src/overrideRegistry/Overrides.tsx
@@ -18,7 +18,7 @@ export class RegistrationQueue {
     this.flush();
   }
 
-  public register(component: React.ComponentClass, key: string, cloudProvider?: string) {
+  public register(component: React.ComponentType, key: string, cloudProvider?: string) {
     this.queue.push(() => {
       if (!cloudProvider) {
         this.overrideRegistry.overrideComponent(key, component);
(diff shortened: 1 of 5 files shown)


app/scripts/modules/core/src/overrideRegistry/Overrides.tsx
```diff
@@ -61,3 +61,14 @@ export function Overrides(key: string, cloudProvider?: string) {
     overrideRegistrationQueue.register(targetComponent, key, cloudProvider);
   };
 }
+
+/**
+ * This should be applied to a React Function Component, since they cannot utilize decorators.
```
christopherthielen, line 66:
```suggestion

/**
 * Registers a component as a replacement for some other component.
 *
 * This should be applied to a React Function Component, since they cannot utilize decorators.
```

christopherthielen (approved):
🔥🔥🔥

PR #8971, opened 2021-03-05
fix(core/presentation): Avoid remounting CheckListInput on every render

christopherthielen:
This PR creates a stable function reference for each nested component inside ChecklistInput.
I'm not particularly happy with this solution and wanted to solicit feedback and alternative solutions ya'll may know of.

## The problem

I was working with a component that uses CheckListInput and noticed that the input was being _*remounted*_ on every render.  I checked the component and sure enough this was due to the inner Component function references being re-defined on each render.  React sees that the component function is `!==` and unmounts/remounts.  

The inner Components rely on references to props and local variables from the outer Component closure.  This component is in this buggy state because past me extracted some inline JSX and logic out into separate components (for readability) but didn't fully understand the remount gotcha here.  

```jsx
function Outer(props) {
  const { prop1, prop2, prop3 } = props;
  const var1 = computeThing1();
  const var2 = computeThing2();
  const var3 = computeThing3();

  function InnerA() {
    renderStuffUsing(prop1, prop2, var1, var2);
  }
  function InnerB() {
    renderStuffUsing(prop1, prop3, var1, var3);
  }
  function InnerC() {
    renderStuffUsing(prop2, prop3, var2, var3);
  }
  return <div><InnerA/><InnerB/><InnerC/></div>;
}
```

- Alternative 1)

One potential fix would be to de-componentize the inner components but this seems like a step backwards.

- Alternative 2)

Normally I would just hoist the inner components out to the top level so their function references would become stable.  However, now they don't have access to the closure variables so the values have to be passed in as props.  This caused a lot of extra prop drilling and interface definitions and made the whole thing feel overly complicated.

```jsx
interface IInnerAProps {
  prop1: any;
  prop2: any;
  var1: any;
  var2: any;
}
function InnerA(props: IInnerAProps) {
  const { prop1, prop2, var1, var2 } = props;
  renderStuffUsing(prop1, prop2, var1, var2);
}
interface IInnerBProps {
  prop1: any;
  prop3: any;
  var1: any;
  var3: any;
}
function InnerB(props: IInnerBProps) {
  const { prop1, prop3, var1, var3 } = props;
  renderStuffUsing(prop1, prop3, var1, var3);
}
interface IInnerCProps {
  prop2: any;
  prop3: any;
  var2: any;
  var3: any;
}
function InnerC(props: IInnerCProps) {
  const { prop2, prop3, var2, var3 } = props;
  renderStuffUsing(prop2, prop3, var2, var3);
}

function Outer(props) {
  const { prop1, prop2, prop3 } = props;
  const var1 = computeThing1();
  const var2 = computeThing2();
  const var3 = computeThing3();

  return (
    <div>
      <InnerA prop1={prop1} prop2={prop2} var1={var1} var2={var2} />
      <InnerB prop2={prop2} prop3={prop3} var2={var2} var3={var3} />
      <InnerC prop1={prop1} prop2={prop2} var1={var1} var2={var2} />
    </div>
  );
}
```

- Alternative 3)

The inner Components could be called as functions that return JSX instead of rendered as Components, but this also seems like a step backwards to me. 

```jsx
function Outer(props) {
  const { prop1, prop2, prop3 } = props;
  const var1 = computeThing1();
  const var2 = computeThing2();
  const var3 = computeThing3();

  function InnerA() {
    renderStuffUsing(prop1, prop2, var1, var2);
  }
  function InnerB() {
    renderStuffUsing(prop1, prop3, var1, var3);
  }
  function InnerC() {
    renderStuffUsing(prop2, prop3, var2, var3);
  }
  return <div>{InnerA()}{InnerB()}{InnerC()}</div>;
}
```

- Alternative 4

Just define a bag of "All the things!" to use as a props interface for all the internal components:

```tsx
interface IAllTheThings {
  prop1: any;
  prop2: any;
  prop3: any;
  var1: any;
  var2: any;
  var3: any;
}
function InnerA(props: IAllTheThings) {
  const { prop1, prop2, var1, var2 } = props;
  renderStuffUsing(prop1, prop2, var1, var2);
}
function InnerB(props: IAllTheThings) {
  const { prop1, prop3, var1, var3 } = props;
  renderStuffUsing(prop1, prop3, var1, var3);
}

function InnerC(props: IAllTheThings) {
  const { prop2, prop3, var2, var3 } = props;
  renderStuffUsing(prop2, prop3, var2, var3);
}

function Outer(props) {
  const { prop1, prop2, prop3 } = props;
  const var1 = computeThing1();
  const var2 = computeThing2();
  const var3 = computeThing3();

  const allTheThings = { prop1, prop2, prop3, var1, var2, var3 };

  return (
    <div>
      <InnerA {...allTheThings} />
      <InnerB {...allTheThings} />
      <InnerC {...allTheThings} />
    </div>
  );
}
```

Of all the alternatives, this seems less than bad to me but not great.

- The solution in this PR 

This solution uses the `useLatestCallback` hook which returns a _stable function reference_ similar to `useCallback` so React notices that the function refs are `===` equal.  It also delegates the function call to the function created in the most recent render pass, so it always has access to the latest closure and variables such as `var1`, `var2`, `var3`.   

(side note: this hook was created for callbacks to be invoked outside a react callback, such as an async callback like a promise chain.  The hook happens to satisfy the requirements of this problem as well).  

---

Does anyone know of an elegant approach to extracting small internal nested components that doesn't have the problems above?


caseyhebebrand:
Here is my suggestions for this:

- I am not convinced that `InlineOptions` and `VerticalOptions` need to be separate inner components. Unless I am missing something, it looks like the only difference is being vertically vs horizontally aligned. Since the html is similar, I would render the same html, using the `inline` prop to make the container either a `vertical` or `horizontal` flexbox. So for these, I would de-componentize them, with the `CheckListInput` return being something like this: 

```js
return (
   <div className={`checklist  ${classNameDerivedFromInline}`}>
      {checkListOptions.map(option => < CheckBox />)}
      <SelectAllButton />
   <div>
)
```

- I feel the same about the `SelectAllButton`, there is nothing happening within that inner component that cannot be derived from the parent component. I personally do not see it as a step backward to de-componentize this because its just an anchor tag with a conditional `onClick`. The readability will minorly decrease in the `return`, but I feel that the overall flow of the function component would be improved. 

- (question) For `CheckBox`, is there a reason we cannot use the existing `CheckBoxInput` component with a custom `onChange`?

christopherthielen:
> I am not convinced that InlineOptions and VerticalOptions need to be separate inner components.

@caseyhebebrand thanks for the feedback!  I'll try decomponentizing the code and see how it feels.

Even if this particular component doesn't benefit from child components, I believe the desire to use this pattern is legitimate.  I've seen many render functions that are 300+ lines long and that is difficult for me to navigate.  So, regardless of the outcome of this PR. my question still stands... is there a better way? 

> s there a reason we cannot use the existing CheckBoxInput component with a custom onChange?

great question, I'm not sure, but I will try to refactor it and find out!

caseyhebebrand:
> So, regardless of the outcome of this PR. my question still stands... is there a better way?

@christopherthielen  I think you have chosen the right approach. In a functional component, when a prop changes, the entire function body is executed again. So the only way to avoid/optimize this would be to either (1) make a separate component that takes all the parent-context it needs as props or (2) utilize `useCallback`. 

Where I would differ between choosing (1) or (2) would depend on the complexity and potential reusability of the component. I will typically only lean towards `useCallback` approach if the inner component is readable and not highly reusable.

christopherthielen:
@caseyhebebrand I refactored this to remove all the nested components.  I used css as suggested and I think it turned out much nicer.

caseyhebebrand (commented):
LGTM. I love to see a net -18 lines of code with improvements!

alanmquach:
IMO, while `useCallback` is a pretty clever approach, it does take away from readability as it is not immediately obvious why it was used.

I would personally lean more towards (1) for readability, but would be warmer to (2) if the purpose can be made more obvious via something like another hook with a self-explanatory name and/or documentation added to explain.

Past Alan does have a tendency to stump Future Alan, so I try to leave hints where I can.

alanmquach (approved):

PR #8974, opened 2021-03-05
feat(core/presentation): Add a react hook useFormInputValueMapper 

christopherthielen:
This allows FormInput values to be mapped/transformed/converted if the caller model requires

I noticed that we often try to use spinnaker forms and the set of FormInput components.  However, the components may not always return data in the same way that the underlying model requires.  

For example, the form might contain a multi-select checklist which returns an array of strings. However, the form is building a command object which models the selected items as an object with strings for keys and booleans for values, i.e.: 
```
{
  selectedItem1: true,
  selectedItem2: true,
}
```

Mapping `value` is fairly trivial `onChange` requires a bit of trickery and boilerplate.  This hook is an attempt at a simple APi to wire up the bidirectional mappings.  It is primarily intended to be used in a custom FormInput wrapper that defines the mapping and renders the underlying FormInput.  It takes `IFormInputProps` and two mapper functions and returns a `IFormInputProps` with the mapped `value` and `onChange` handler,  This can then be spread onto the underlying input. 

#### Example

A checkbox will be modeled as "ENABLED" or "DISABLED" depending on state:

```tsx
function EnabledDisabledCheckbox(props: ICheckboxInputProps) {
  const mappedProps = useFormInputValueMapper(
    props,
    (val) => val === 'ENABLED', // model -> input
    (val, event) => event.target.checked ? 'ENABLED' : 'DISABLED',
  )
  return <CheckboxInput {...mappedProps} />;
}
```

#### Alternatives

I toyed with a few other APIs and am not married to the hook.  Some other things I tried: 

- Higher order component: 

I generally don't like HOC these days, but this is **/shrug**

```tsx
const EnabledDisabledCheckboxInput = createValueMappedComponent(
  CheckboxInput, 
  (val) => val === 'ENABLED',
  (val, event) => event.target.checked ? 'ENABLED' : 'DISABLED',),
);
```

- a Component with render props

This one ended up a little verbose inside the `FormikFormField`

```tsx
input={props => (
  <MappedValueInput
    {...props}
    toInput={(val) => val === 'ENABLED'}
    fromInput={(val, event) => event.target.checked ? 'ENABLED' : 'DISABLED'}
    render={mappedProps => <CheckboxInput {...mappedProps} />}
  />
)}
```

- onChange helper

This returns a new onChange callback wired up to map the value.  This was the lowest level attempt, but also the least obvious/intuitive.

```tsx
input={({value, onChange, ...props }) => 
  <CheckboxInput
    {...props}
    value={val === 'ENABLED'}
    onChange={onChangeMapper((val, event) => event.target.checked ? 'ENABLED' : 'DISABLED')}
  />
}
```

I'd love to get some thoughts on this, thanks!


vigneshm (approved):
I like this idea. I could use something like this with the details field that splits the details value into multiple shards.

alanmquach:
The HOC is probably the cleanest but I am also meh about it.

The onChange helper actually seems the most obvious to me.

Kinda feels like a thing that `FormikFormField` _could_ be responsible for. Would adding an optional prop be overloading it too much?
```
<FormikFormField
  ...
  mappers={{
    to: (val) => val === 'ENABLED',
    from: (_val, event) => event.target.checked ? 'ENABLED' : 'DISABLED'
  }}
  input={(inputProps) => <CheckboxInput {...inputProps} />}
/>
```

If so, what about a `FormikMappedFormField`?

christopherthielen:
@alanmquach @vigneshm  thanks for the feedback

> Kinda feels like a thing that FormikFormField could be responsible for. Would adding an optional prop be overloading it too much?

That's actually where I started 😆 and the I was also thinking that this might be overloading `FormikFormField` a bit much, but 🤷.

My thought process was that we could decorate inputs and export them as reusable wrapper components.  For example, the `MapInput` models a list of key/value pairs as an object `{ [key: string]: value }` but we could also model it as a array of pairs: `Array<{ key: string, value: string }>`.  We could reuse the wrapped Input instead of copying the mapping code in multiple `FormikFormField`s across different forms.  

I was also thinking that any work in `FormikFormField` has to be done in the `FormField` too.  

Finally, I was thinking that having this functionality at the Input level just felt like the right abstraction level to me at the time, but that's hard to justify really.

---

> The onChange helper actually seems the most obvious to me.

At the lowest level, the `onChange` helper is the only thing doing actual work. It's the building block that all the other implementations are based on.  However, I prefer having the developer API be slightly higher to demonstrate that this is bi-directional mapping to and from Input values.  I don't think the onChange helper on its own gets that point across, compared to the bidirectional apis.  That said, we could potentially expose both levels and see how it goes.

app/scripts/modules/core/src/presentation/forms/hooks/useFormInputValueMapper.hook.ts
```diff
@@ -0,0 +1,40 @@
...
+) {
+  const value = toInputValue(props.value);
+  const onChange = React.useCallback((e: React.ChangeEvent<any>) => {
+    props.onChange(
+      createFakeReactSyntheticEvent({
+        name: e.target.name,
+        value: fromInputValue(e.target.value, e),
+      }),
+    );
+  }, []);
```
ranihorev, line 37: You're missing the dependencies here - `onChange` and `fromInputValue`. I suggest fixing this or switching to `useLatestCallback`

ranihorev (commented):
I'm assuming that this hook is required (I prefer to use the original event when possible). One comment there

app/scripts/modules/core/src/presentation/forms/hooks/useFormInputValueMapper.hook.ts
christopherthielen: oh, nice  spot! What conclusion did we come to w.r.t. the react hooks eslint rule?

ranihorev: https://github.com/spinnaker/deck/blob/76754d288008ce1ad23fcbb6a8c72fcb0e42fa8b/packages/eslint-plugin/base.config.js#L37

Have we ever turned it on? @vigneshm

christopherthielen:
>  (I prefer to use the original event when possible)

I don't think that's possible in this case (correct me if I'm wrong) because we would have to _assign_ the  mapped value to the target `e.target.value = fromInputValue(e.target.value, e)` which would alter the `value` of the actual dom node.

Or maybe if we just replaced the entire `e.target` reference with a fake `target: { name, mappedValue }` 🤔 ?

ranihorev:
@christopherthielen I don't have enough context on how we use this to answer that. 
> we would have to assign the mapped value to the target

Why do we have to do that? Is that what `formik` expects?

christopherthielen:
@ranihorev yeah exactly... formik expects a `ChangeEvent` and pulls the value off the `event.target.value`

ranihorev:
I guess we'll have to live with it 🤷

PR #8986, opened 2021-03-09
fix(core): Running tasks popover surpassing screen dimensions

caseyhebebrand:
A couple issues that this addresses: 
- The popover was too wide, causing it surpass the browser width.
- For a long list of tasks, the popover would either take up or surpass the screen height. 

**Before**
Screen Shot 2021-03-16 at 4 26 49 PM

**After**
Screen Shot 2021-03-16 at 4 27 14 PM


christopherthielen (approved):
LGTM, can you show a screenshot of the original problems and the end result?

caseyhebebrand:
@christopherthielen I updated the description with some screen shots.

PR #8990, opened 2021-03-11
feat(travis): Remove Property File field from Travis stages and triggers

jervi:
Travis doesn't need this field to fetch properties from builds.
Depends on https://github.com/spinnaker/orca/pull/4089 and https://github.com/spinnaker/echo/pull/1090.
Add info box to Travis trigger and Travis stage to inform about proper usage.

Trigger:
image

Stage:
image


jervi:
Do not merge this before spinnaker/orca#4089 and spinnaker/echo#1090 are merged.

jervi:
@Mergifyio update

PR #8993, opened 2021-03-15
feat(core/presentation): add `mode="CREATABLE"` to the ReactSelectInput

christopherthielen:
This PR allows the ReactSelectInput to use the `react-select` 1.x `Creatable` component which allows you to type an arbitrary value and hit Enter, even if it's not an existing option in the list.

### DIFF
--- a/app/scripts/modules/core/src/presentation/forms/inputs/ReactSelectInput.tsx
+++ b/app/scripts/modules/core/src/presentation/forms/inputs/ReactSelectInput.tsx
@@ -78,9 +78,11 @@ export function ReactSelectInput<T = string>(props: IReactSelectInputProps<T>) {
   const className = orEmptyString(inputClassName);
   const { category } = useValidationData(validation.messageNode, validation.touched);
   const style = category === 'error' ? reactSelectValidationErrorStyle : {};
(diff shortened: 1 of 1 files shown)


alanmquach (approved):
Nice.

app/scripts/modules/core/src/presentation/forms/inputs/ReactSelectInput.tsx
```diff
@@ -92,6 +92,8 @@ export function ReactSelectInput<T = string>(props: IReactSelectInputProps<T>) {
       <TetheredSelect {...commonProps} options={options} />
     ) : mode === 'VIRTUALIZED' ? (
       <VirtualizedSelect {...commonProps} options={options} optionRenderer={null} />
+    ) : mode === 'CREATABLE' ? (
+      <CreatableSelect {...commonProps} options={options} />
```
alanmquach, line 96: So the `CreatableSelect` is neither tethered nor virtualized, right? I wonder if that should be called out anywhere. 🤷‍♂️

christopherthielen: correct it is neither, like the base SelectInput.  I'm wondering: did you have an expectation that it might be tethered or virtualized?

PR #8994, opened 2021-03-16
fix(core): allow stage validation to use both validators and validateFn

anotherchrisberry:
Imagine a stage that uses Formik to perform field-level validation, but also wants to validate the presence of an upstream stage, e.g. a stage that provides an image, via a `stageBeforeTypeValidator`.

It's not easy (or seems clean) to include stage-level validation in the formik validator code, but the PipelineConfigValidator only takes one or the other of `validators` and `validateFn`.

This just lets a stage do both.

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/validation/PipelineConfigValidator.ts
+++ b/app/scripts/modules/core/src/pipeline/config/validation/PipelineConfigValidator.ts
@@ -136,14 +136,16 @@ export class PipelineConfigValidator {
             );
           }
         });
-      } else if (config && config.validateFn) {
+      }
+      if (config && config.validateFn) {
         validations.push(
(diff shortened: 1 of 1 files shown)


christopherthielen:
If you're using `FormikStageConfig`, the validator is passed the pipeline and application as additional properties.  I'd rather move forward with this paradigm than add another edge case that we have to unwind later when the outer code gets converted to React.


https://github.com/spinnaker/deck/blob/5a9768bc6db2f527a73d6b1f5fb3120c101e094b/app/scripts/modules/core/src/pipeline/config/stages/FormikStageConfig.tsx#L36

anotherchrisberry:
@christopherthielen I was going down that route originally, but tried to pull in the StageBeforeTypeValidator, which isn't exported. The logic in that validator is not complicated so I will just replicate it.

PR #9003, opened 2021-03-17
feat(core/ApplicationDataSourceRegistry): Remove data source based on key

bbrina:
## What does this PR do?
Adds function to remove a data source based on the key. We are hoping to customize the vertical navigation with this function


bbrina:
@dogonthehorizon What do you think of this approach?

app/scripts/modules/core/src/application/service/ApplicationDataSourceRegistry.ts
```diff
@@ -21,6 +21,8 @@ export class ApplicationDataSourceRegistry {
   }
 
   private static sortDataSources(): void {
+    // eslint-disable-next-line no-console
+    console.log(this.dataSources);
```
aleEspinosaM, line 25: should we delete this guy here?

bbrina: Done!

christopherthielen (approved):

PR #9009, opened 2021-03-18
fix(core): Trim cloud providers after splitting

alanmquach:
If the cloudProviders attribute contains spaces: `"aws, titus"`, simply splitting them into `["aws", " titus"]` will fail an exact match to the actual cloud provider strings.

While this is not typical (since deck will `.join(',')` without adding spaces), inconsistencies in API clients that create applications can easily result in this scenario.

This simple fix makes deck a bit more flexible to accommodate any "not exactly valid" config that has already been persisted.


caseyhebebrand (approved):
Nice catch!

PR #9012, opened 2021-03-18
fix(managed): fix override label on manual judgement, close artifact …

anotherchrisberry:
…details on escape

These are very minor things I ran into when trying to figure out how the managed delivery code works.

### DIFF
--- a/app/scripts/modules/core/src/managed/artifactDetail/constraints/constraintRegistry.tsx
+++ b/app/scripts/modules/core/src/managed/artifactDetail/constraints/constraintRegistry.tsx
@@ -137,7 +137,7 @@ const constraintsRenderer: { [type in IConstraint['type']]?: IConstraintRenderer
         case 'OVERRIDE_FAIL':
           return (
             <span>
-              <span>Manually {status === 'OVERRIDE_PASS' ? 'approved' : 'rejected'}</span>{' '}
+              <span>Manually {constraint.status === 'OVERRIDE_PASS' ? 'approved' : 'rejected'}</span>{' '}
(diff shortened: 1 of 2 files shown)


app/scripts/modules/core/src/managed/artifactDetail/constraints/constraintRegistry.tsx
```diff
@@ -137,7 +137,7 @@ const constraintsRenderer: { [type in IConstraint['type']]?: IConstraintRenderer
         case 'OVERRIDE_FAIL':
           return (
             <span>
-              <span>Manually {status === 'OVERRIDE_PASS' ? 'approved' : 'rejected'}</span>{' '}
+              <span>Manually {constraint.status === 'OVERRIDE_PASS' ? 'approved' : 'rejected'}</span>{' '}
```
anotherchrisberry, line 140: I guess there's a `window.status` field, which is why Typescript doesn't complain about this?

ranihorev: Yeah :(
I opened a PR for that a couple of hours ago

ranihorev: https://github.com/spinnaker/deck/pull/9010

anotherchrisberry: ha

anotherchrisberry: I'll close this and let you deal with the other thing if you want

PR #9013, opened 2021-03-19
fix(pipeline): Move configure into execution

alanmquach:
With the `SingleExecutionDetails` now rendering ancestors as well, the `Configure` button in the header can be a bit confusing as to which pipeline you will be configuring (the oldest rendered ancestor is the closest, but it will actually configure the deepest rendered child).

This PR removes the top level button and replaces that functionality with a configure button inside each execution instead.
image


alanmquach:
I'm not super happy about adding yet another optional prop, but using something like context to determine whether we are in the `SingleExecutionDetails` view feels equally messy. Thoughts?

christopherthielen (approved):
yeah I think this is good

PR #9014, opened 2021-03-19
feat(md): Plugin API for constraints

ranihorev:
This PR adds support for MD constraint plugins, similarly to the resources plugin. It's a hook for rendering a `StatusCard` with an icon, a text React component, and optionally two buttons to override the constraint. 

The plugin interface was updated to include a new `managedDelivery` key in it that will hold all of them MD plugins. As a result, the old `resourcesKind` was deprecated and will be removed in future versions.

Lastly, to have a consistent experience across different MD plugins, I created a `BasePluginManager` that handles registering new plugins.


app/scripts/modules/core/src/plugins/deck.plugin.ts
```diff
@@ -11,8 +11,10 @@ export interface IDeckPlugin {
   stages?: IStageTypeConfig[];
   /** Custom Preconfigured Job Stage UI (configuration and execution details) */
   preconfiguredJobStages?: IStageTypeConfig[];
-  /** Custom managed resource kinds */
+  /** DEPRECATED - Custom managed resource kinds */
   resourceKinds?: IResourceKindConfig[];
+  /** Managed Delivery hooks */
+  managedDelivery?: IManagedDeliveryPlugin;
```
christopherthielen, line 17: could you make a case for packaging more than one MD plugin into a single `IDeckPlugin`?

app/scripts/modules/core/src/plugins/deck.plugin.ts
```diff
@@ -25,10 +27,14 @@ export interface IDeckPlugin {
...
   plugin.preconfiguredJobStages?.forEach((stage) => Registry.pipeline.registerPreconfiguredJobStage(stage));
-  plugin.resourceKinds?.forEach((kind) => resourceManager.registerResource(kind));
   toPairs(plugin.help ?? {}).forEach(([key, value]) => HelpContentsRegistry.register(key, value));
   plugin.search?.forEach((search) => searchResultTypeRegistry.register(search));
 
+  const managedDeliveryPlugin: IManagedDeliveryPlugin = {
+    ...plugin.managedDelivery,
+    resources: plugin.resourceKinds || plugin.managedDelivery?.resources,
+  };
+  registerPlugin(managedDeliveryPlugin);
```
christopherthielen, line 37:
```suggestion
  if (plugin.managedDelivery || plugin.resourceKinds) {
    const managedDeliveryPlugin: IManagedDeliveryPlugin = {
      ...plugin.managedDelivery,
      resources: plugin.resourceKinds || plugin.managedDelivery?.resources,
    };
    registerPlugin(managedDeliveryPlugin);
  }
```

app/scripts/modules/core/src/managed/plugins/index.ts
```diff
@@ -0,0 +1,12 @@
+import { constraintsManager, IConstraintHandler } from '../constraints/registry';
+import { IResourceKindConfig, resourceManager } from '../resources/resourceRegistry';
+
+export interface IManagedDeliveryPlugin {
+  resources?: IResourceKindConfig[];
+  constraints?: IConstraintHandler[];
+}
+
+export const registerPlugin = (plugin: IManagedDeliveryPlugin) => {
```
christopherthielen, line 9: These end up in the top level export, so `registerPlugin` would be a misleading name
```suggestion
export const registerManagedDeliveryPlugin = (plugin: IManagedDeliveryPlugin) => {
```

app/scripts/modules/core/src/managed/constraints/registry.tsx
```diff
@@ -0,0 +1,175 @@
...
+  /** The type of the constraint - versioning is supported by adding @{version}, e.g. myConstraint@ */
+  kind: K;
+  /** The icon can be a string (from IconNames) or a partial map from statuses to IconNames */
+  iconName: IconNames | { [status in ConstraintStatus | 'DEFAULT']?: IconNames };
+
+  /** Render function of the constraint title */
+  titleRender: (constraint: IBaseConstraint | IConstraint) => React.ReactNode;
+
+  /** Optional render function of the constraint description */
+  descriptionRender?: (constraint: IBaseConstraint | IConstraint) => React.ReactNode;
```
christopherthielen, line 38: Wondering if this API could be typed as a `ComponentType<IConstraint>` instead?  I'm personally more comfortable working with components rather than passing JSX  around.  

Just a thought, not a strong opinion

christopherthielen (changes requested):
Refactor LGTM
I have a couple small requests w.r.t. the plugin API.

app/scripts/modules/core/src/plugins/deck.plugin.ts
ranihorev: You can have multiple `resources` and `constraints` within a single `managedDelivery`. I can't think of a reason to have more the one `managedDelivery` if that's what you mean

app/scripts/modules/core/src/plugins/deck.plugin.ts
ranihorev: Why do we need this check?

app/scripts/modules/core/src/managed/constraints/registry.tsx
ranihorev: Sure, sounds good to me, and then change the render function to this? Is that what you were thinking of?

```
  renderTitle(constraint: IConstraint): React.ReactNode {
    const Component = this.getHandler(constraint.type)?.titleRender;
    if (Component) {
      return <Component constraint={constraint} />
    }
    return `${constraint.type} constraint - ${constraint.status}`;
  }
```

ranihorev: Yeah, it cleans up the registry file so I'm good with it :)

app/scripts/modules/core/src/plugins/deck.plugin.ts
christopherthielen: most plugins won't have this property, and my opinion is that it would improve clarity.  As a reader of this code, I don't know if registerPlugin does something or nothing when my plugin doesn't export a MD plugin.

christopherthielen:
@ranihorev let me know when you want this merged and I'll label it

christopherthielen (approved):

anotherchrisberry (approved):
If @christopherthielen approves, I have to approve

PR #9016, opened 2021-03-19
fix(titus/serverGroup): allow underscores in the first character of environment variables

christopherthielen:


alanmquach (approved):
Sure

caseyhebebrand (approved):

PR #9018, opened 2021-03-19
feat(git): Add a .git-blame-ignore-revs file

christopherthielen:
TIL you can ignore specific revisions in `git blame` output.  These revs can be stored in a file, then you can configure git to read the ignore list from the file.

`git config blame.ignoreRevsFile .git-blame-ignore-revs`

This is pretty great, because I'm regularly seeing sweeping code reformat commits when I do a `git blame`.  I want to determine the intention/history of some code but the reformats get in the way.

---

Before -- notice much of the changes are attributed to "Just Use Prettier" reformatting commit:

Screen Shot 2021-03-19 at 10 46 20 AM

After -- the code is attributed to the previous committer:

Screen Shot 2021-03-19 at 10 47 07 AM

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,37 @@
+# Ignores sweeping changes in `git blame` output
+# See: https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame
+# 
+# INSTRUCTIONS:
+#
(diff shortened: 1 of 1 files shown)


vigneshm (approved):

alanmquach (approved):
Neato!

PR #9020, opened 2021-03-19
feat(pipeline): Change back button to breadcrumbs

alanmquach:
This change is maybe a bit controversial 🤷‍♂️ , so I'd love to have some discussion around it, if appropriate.

image

In the `SingleExecutionDetails` header, there is a back button. The slightly confusing part is that this button does not necessarily always go "back" to where you came from. Rather, it always goes "up" to the `.executions` route.

Most of the time, a user probably reached the SingleExecutionDetails by clicking on the `permalink` link from the `.executions` route so the behavior matches expectation. However, in the following cases, navigating to `.executions` would logically be more of an "up" than a "back":

- If the user was already on an `.executionDetails` route for a different execution and clicked on `View Pipeline Execution`
- If the user is simply navigating around ancestors/descendants now that we render an executions ancestry

The last time I recall this behavior discussed (which was a really long time ago), I think we had settled on using breadcrumbs to clearly convey to the user that they would be going "up" to "all executions". What's nice is that breadcrumbs also allows them to choose between seeing executions of 1) all pipelines or 2) the same pipeline.

Controversially (maybe?), there will be a slight loss of functionality where we will not pre-select the current execution for the user, but I'm not sure how often users are clicking back because they want to:

1.  `See the current execution in the context of all other executions, but only if it has not aged out of the current window of [2?] executions per pipeline`, versus just
2.  `See all executions for this app`


christopherthielen:
- I saw `aqjs > baby shark` and assumed this was a parent/child pipeline relationship.  We already use the right chevron to represent pipeline breadcrumbs so I didn't realize that `aqjs` was actually the _application name_. 
- I like the new feature of linking to the executions view but filtered to the named `pipeline` 
  - I think each parent execution could also have this link, much like the treatment you did for the "configure" button.
- We link to `.executionDetails` with `executionParams` from `ExecutionMarkerInformationModal`.  I'm not sure what this does, but let's make sure we're not breaking something else.
  - "slight loss of functionality" Any reason we can't continue to link to `executions.execution` ?

alanmquach:
Thanks for the feedback!
> I saw aqjs > baby shark and assumed this was a parent/child pipeline relationship. We already use the right chevron to represent pipeline breadcrumbs so I didn't realize that aqjs was actually the application name.

Fair, but do you think this was simply because you did not know aqjs is the application name (and users usually will know their app name)? Also open to suggestions here such as `|` `-` `•` `/`

> I think each parent execution could also have this link

Good idea, I can add this

> We link to .executionDetails with executionParams from ExecutionMarkerInformationModal. I'm not sure what this does, but let's make sure we're not breaking something else.

I think this just allows an inbound link from that modal to deep link directly to a stage/step/detail, which should still work. I've simply removed the params from the outbound link from `.executionDetails` up to `.executions`.

> Any reason we can't continue to link to executions.execution ?

We can... but there's a minor bug from `fix(core): Redirecting aged out executions to permalink` (#8770) that redirects you right back to `.executionDetails` that I was hoping to squash with this change as well:

1. Open the permalink of an aged out execution (aka one that is not the most recent n executions) in a fresh window/tab
2. Go back/up/whatever to `executions.execution`
3. #8770 will kick in and forward you right back to the permalink, but will only do this exactly once
4. Go back/up to `executions.execution` again, this time, the old behavior of seeing that the execution has aged out and redirecting to `.executions` will kick in.

While that certainly can be fixed, I'm not sure if it's worth trying to since:

1. The functionality is only lost if users are looking at an execution that is still part of the most recent [2]. Otherwise, they will get sent to `.executions` which is consistent with the current UX for breadcrumbs. The breadcrumbs simply make it more explicit
2. I don't think breadcrumbs UX usually calls for attempting to focus on where you came from. For example, if I remember from Windows Explorer, if I clicked up from `Program Files` to `C:\`, I don't think it attempts to focus/select the `Program Files` folder while viewing `C:\`.

christopherthielen:
> Fair, but do you think this was simply because you did not know aqjs is the application name (and users usually will know their app name)? Also open to suggestions here such as | - • /

Yes and no, but either way I think re-using the chevron is confusing UI.  I'd rather use a `-` or if we're adding a link _to each parent execution_, maybe the app level `aqjs` link jut stays up at the top?

alanmquach:
Ok I will change it to a dash.

I'm going to skip adding a link to each parent execution as it is too close to the breadcrumbs and the difference in behavior may be a bit confusing to users:

Screen Shot 2021-03-22 at 12 22 57 PM

christopherthielen (approved):

caseyhebebrand (approved):

PR #9022, opened 2021-03-22
feat(core/pipeline): Awaiting Judgement Filter

siddhu-opsmx:
### Goals and Non-Goals

Add a filter for filtering pipelines waiting on manual judgement. Checking this filter will automatically select the running stage filter and disable it(since all manual judgment pipelines are running). On deselecting the manual judgment filter the running filter will be enabled but still selected.

## Design

1. Enhanced FilterModelService.ts to
     -> Added a filter tag for awaiting judgment and enhanced the clear filter functionality to clear awaiting judgment filter.

2. Enhanced IFilterModel to
     -> Added filterStages and stages to ISortFilter.

3. Enhanced ExecutionGroups.tsx to
     -> Added a filter which will filter out running executions contain manual judgement stage in it.

4. Enhanced ExecutionFilterModel.ts to 
     -> Added a filterstages property to sortfilter to extend the functionality and to reuse the same object to add other filters 
           to the stages in the future.

5. Enhanced ExecutionFilters.tsx to 
     -> Created a new component to be used for stage filter for filtering Manual Judgment filter. it can be reused to add more 
          stage filter in the future.

6. Enhanced executionFilter.service.ts to 
     -> Created a function to filter all the running executions that contain manual judgment stage in it.


siddhu-opsmx:
@christopherthielen Could you please review the code.

christopherthielen (changes requested):
This PR doesn't appear to work properly.

When a set of pipelines are checked, they should display like this:

Screen Shot 2021-04-06 at 8 37 46 AM

When I filter by status, the pipelines should still appear in the list:

Screen Shot 2021-04-06 at 8 37 51 AM

This code hides the pipelines even though it's checked.

It also shows pipelines with ANY manual judgement stages, even if that stage is not awaiting judgement.

Screen Shot 2021-04-06 at 8 38 16 AM


After the behavior is fixed, I'll review again.

siddhu-opsmx:
@christopherthielen We have pushed the changes suggested by you, could you please have a look on it.

christopherthielen:
Thanks for fixing the pipeline filtering.  However, even with the new code this still shows running pipelines that are not awaiting judgement.


In this screenshot, the execution should not be visible because it is not awaiting judgement:

Screen Shot 2021-04-12 at 2 05 27 PM

siddhu-opsmx:
@christopherthielen sorry about that, That was a silly mistake. I examined the code closely. I think there are no more bugs but it would be nice if you could take another look.

christopherthielen:
awaiting judgement 2

This is still not working correctly.

christopherthielen:
I have fixed the above problems in a separate branch.  I also changed the approach to better fit the existing FilterModel paradigm.

Because the PR doesn't [allow changes from maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork), I've re-created the PR from my branch and will close this one.

PR #9024, opened 2021-03-23
fix(core/projects): Recent projects: only save the link to project dashboard

christopherthielen:
When following a link to the recently used project, we now always link to `home.project` which will redirect to the project dashboard.


alanmquach (approved):
That was easy

PR #9034, opened 2021-03-24
feat(titus): Finish migrating instance details to react

caseyhebebrand:
This PR completes the Titus instance details react migration. What this PR includes:
- **`TitusInstanceDetails`** - This component captures all common behavior. It fetches, transforms and passes all needed information to the presentational `TitusInstanceDetailsContent` component.  It will also show any error or loading states.
- **`OverridableTitusInstanceDetailsContent`** - This is mainly a presentational component composed of different instance details sections. It is meant to be overridden and augmented with custom behavior or collapsible sections. 
- **`utils`** to help generate common instance information.
  - `extractHealthMetrics` - Gets all displayable metrics, and augments health metrics with additional data
  - `buildTaskActions` - Returns an array of actions based on the health state of the instance
  
What this does not include:
- Deleting angular files. There are still some dependencies that will be removed shortly. At that point, I will create another PR that deletes those files.

### DIFF
--- a/app/scripts/modules/titus/src/instance/details/TitusInstanceDetails.tsx
+++ b/app/scripts/modules/titus/src/instance/details/TitusInstanceDetails.tsx
@@ -1,0 +1,202 @@
+import * as React from 'react';
+
+import { IAmazonHealth, InstanceStatus } from '@spinnaker/amazon';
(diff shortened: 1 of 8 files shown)


app/scripts/modules/titus/src/instance/details/TitusInstanceDetails.tsx
```diff
@@ -0,0 +1,202 @@
...
+          moniker={moniker}
+        />
+      </div>
+    </div>
+  );
+};
+
+const OverridableTitusInstanceDetailsContent = overridableComponent<
+  ITitusInstanceDetailsContentProps & IOverridableProps,
+  typeof TitusInstanceDetailsContent
```
christopherthielen, line 101: `typeof TitusInstanceDetailsContent` oh interesting, I wonder if we can get `overridableComponent` to automatically infer this

christopherthielen (approved):
wow, really great work!  
nice refactors, and it's super clean

PR #9035, opened 2021-03-24
feat(trigger): Add an optional link to configure a stash trigger

caseyhebebrand:


app/scripts/modules/core/src/config/settings.ts
```diff
@@ -125,6 +125,7 @@ export interface ISpinnakerSettings {
   slack?: {
     baseUrl: string;
   };
+  slackTriggerInfo?: string;
```
caseyhebebrand, line 128:
```suggestion
  stashTriggerInfo?: string;
```

vigneshm (approved):

PR #9037, opened 2021-03-25
fix(md): remove depends-on icon from the artifacts list

ranihorev:
The PR removes the depends-on icon from the artifact versions list (the example below is showing MJ). I'm not a fan of remove only this constraint, as it creates inconsistency in user expectations but we will deal with it later. 

image


christopherthielen (approved):
Ok!

caseyhebebrand (approved):

PR #9044, opened 2021-03-27
feat(nav): new overridable section at the bottom of the left navigation bar

ranihorev:
This is intended for adding help menus at the bottom of the bar. 
In addition, the navigation bar is now scrollable.

Note: there aren't as many changes as Github shows. Apply the whitespace filter before reviewing :)


christopherthielen (approved):
looks good!  Did you verify that there is no unwanted changes to the layout?

ranihorev:
Yes, however, I think it might be better to make the bottom section fixed
and the center scrollable. What do you think?

PR #9046, opened 2021-03-29
fix(core/presentation): allow null observable in useObservable.hook also export all the hooks from index.ts

christopherthielen:


ranihorev (approved):
Looks good. What's the reason for fixing it?

christopherthielen:
> Looks good. What's the reason for fixing it?

I want to use it in an upcoming amazon PR

PR #9047, opened 2021-03-29
Migrate AWS scaling policies charts to React

christopherthielen:
_I split this PR into 4 separately reviewable commits_

This replaces the n3-chart charts with a React LineChart wrapper based on Chart.js

deck-kayenta still has a dependency on ~n3-charts~ d3 so we're not yet eliminating all the d3 code from our bundle

Old:

Screen Shot 2021-03-29 at 10 48 49 AM

Screen Shot 2021-03-29 at 10 48 57 AM

New:

Screen Shot 2021-03-29 at 10 49 27 AM

Screen Shot 2021-03-29 at 10 49 33 AM

I'll take a quick look at deck-kayenta to see if we can migrate those charts to Chart.js as well


alanmquach (approved):
Nice. Thanks for doing this.

Do we lose any resolution? The older graph looks more fine grained, unless it's just faking it.

christopherthielen:
> Do we lose any resolution? The older graph looks more fine grained, unless it's just faking it.

No loss in resolution. The screenshot for the new chart has "1 hour" selected but if I had selected "1 minute" it would show the same number of points.

PR #9050, opened 2021-03-29
fix(core/pipeline): After saving a pipeline, use the value from the server as the new "original" This change fixes multiple saves of a pipeline from the same configuration screen

christopherthielen:
This change fixes multiple saves of a pipeline from the same configuration screen

### DIFF
--- a/app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
+++ b/app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
@@ -369,8 +369,11 @@ angular
         PipelineConfigService.savePipeline(toSave)
           .then(() => $scope.application.pipelineConfigs.refresh(true))
           .then(
-            () => {
-              setOriginal(toSave);
+            (pipelines) => {
+              const latestFromServer = pipelines.find((p) => p.id === toSave.id);
+              const latest = latestFromServer || toSave;
+              setOriginal(latest);
+              this.revertPipelineChanges();
               markDirty();
               this.setViewState({
                 saveError: false,


app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -369,8 +369,11 @@ angular
         PipelineConfigService.savePipeline(toSave)
           .then(() => $scope.application.pipelineConfigs.refresh(true))
           .then(
-            () => {
-              setOriginal(toSave);
+            (pipelines) => {
+              const latestFromServer = pipelines.find((p) => p.id === toSave.id);
+              const latest = latestFromServer || toSave;
+              setOriginal(latest);
+              this.revertPipelineChanges();
```
christopherthielen, line 376: this might seem odd, but there is an async callback that reverts $scope.pipeline which makes a simple assignment  here get reverted.

alanmquach: Looks good, but can you drop a comment here referencing that async callback?

alanmquach (approved):

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
christopherthielen: since this already merged, I'll just leave this comment here:

https://github.com/christopherthielen/deck/blob/ceb7a4f2cafb06ccf67cc9ad8ebe9e0be31f884f/app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js#L257-L268

This should be irrelevant if we ever convert these pipeline codes to react

alanmquach: Good enough for me!

mergify was too fast.

PR #9055, opened 2021-03-30
chore: remove semiotic from package.json this is required by @spinnaker/kayenta, but kayenta has its own dependency on the package

christopherthielen:
This package is required by @spinnaker/kayenta, but kayenta has its own dependency on it.

If this passes CI then I'm happy


christopherthielen:
narrator: "It didn't pass CI"

christopherthielen:
this failed CI because when yarn installs `@spinnaker/kayenta` it does not hoist the `semiotic` package up to the top level and webpack is not scanning nested node_modules because ❓ reasons ❓

PR #9056, opened 2021-03-30
feat(pipeline): Pipeline Tags config UI

alanmquach:
image

Added a custom input for editing Pipeline Tags (used for filters). Tags are modeled as an ordered array of objects (currently with name+value attributes), so the `MapEditor` could not be used, not even with the `useFormInputValueMapper` hook. 😞 

Since this tags "list editor" shares similarities with the `MapEditor`, I considered adding a `List` vs `Map` toggle but ultimately decided against it as I felt it would become way too overloaded.

The `Description` field also came along for the ride and got moved into a new `Metadata` section which is now the new home for:

- Description
- Tags


christopherthielen (approved):
This looks great!


>  the MapEditor could not be used, not even with the useFormInputValueMapper hook

Is this because key/value pairs can have duplicate keys but MapEditor cannot?

The `MapEditorInput` models its data as a list of [key/value pairs internally](https://github.com/spinnaker/deck/blob/273966d86d1cefe35378c842a3a334e03f5fd606/app/scripts/modules/core/src/forms/mapEditor/MapPair.tsx#L8-L12) and then maps to/from a object `{ [key: string]: string }` for the Controlled Component contract (this was built before the input mapping hook existed).  

I think it would be pretty easy to flip the direction of composition here, such that the current internal implementation of `MapEditorInput ` (which already uses a list of pairs) becomes the `KeyValueListEditorInput`.   Then, a new `MapEditorInput` component could be a wrapper around `KeyValueListEditorInput` which maps the list of pairs to an object and provides the additional "no duplicate key" validation.

alanmquach:
> Is this because key/value pairs can have duplicate keys but MapEditor cannot?

Yes, and the List preserves order (whether it matters or not). I went with a list of objects in case we wanted to add more attributes in the future.

> I think it would be pretty easy to flip the direction of composition here

This is a really good idea. So good that I really want to try it right now. I don't see why it wouldn't work.🤞

PR #9060, opened 2021-03-30
chore(ci): replace hex colors with closes css variables

christopherthielen:
I did this because it's super annoying when running tests
and still annoying (but somewhat less) when running deck locally


ranihorev (commented):
Looks good. I didn't check that all of the colors are matching :)

christopherthielen:
> Looks good. I didn't check that all of the colors are matching :)

neither did I ;)

The colors did look pretty close in the IDE though, close enough for sure.  I wonder if anyone will notice?

PR #9063, opened 2021-03-30
feat(md): Show resource history link on hover

ranihorev:
This PR adds this new button that links to the history of the resource, per @christopherthielen's request (and my personal need as well). 
image

In addition, it includes some refactoring to improve readability. 
We also removed the weird blue dot on the left when deploying :)


app/scripts/modules/core/src/managed/EnvironmentsList.tsx
```diff
@@ -48,11 +49,14 @@ export function EnvironmentsList({
...
-                    artifactVersionsByState={artifactVersionsByState}
-                    artifactDetails={artifactDetails}
-                    depth={0}
+                    metadata={
+                      <ResourceDeploymentStatus
+                        environment={name}
+                        showReferenceName={allArtifacts.length > 1}
+                        artifactVersionsByState={artifactVersionsByState}
+                        artifactDetails={artifactDetails}
+                      />
```
caseyhebebrand, line 58: Here, I would prefer to either:

- keep the props as they were before, and render `ResourceDeploymentStatus`  within the `ManagedResourceObject` file.
- or just directly render `ResrouceDeploymentStatus` as a child of the component
```js
<ManagedResourceObject {...otherProps}>
  <ResourceDeploymentStatus
    environment={name}
    showReferenceName={allArtifacts.length > 1}
    artifactVersionsByState={artifactVersionsByState}
    artifactDetails={artifactDetails}
  />
</ManagedResourceObject>
```

ranihorev: > keep the props as they were before
I rather not to, because they are only needed in one case.

> or just directly render ResrouceDeploymentStatus as a child of the component
What's the difference between assigning it to `metadata` and `children`? I agree that your implementation looks nice, but I suspect that it could be confusing (to use children for the metadata)

app/scripts/modules/core/src/managed/ManagedResourceObject.tsx
```diff
@@ -115,18 +97,21 @@ export const ManagedResourceObject = memo(
...
-        depth={depth}
-        content={resourceStatus}
-        metadata={
-          <span className="flex-container-h middle">
-            {currentPill}
-            {deployingPill}
-          </span>
-        }
-      />
+      <div className="ObjectRow" style={{ marginLeft: 16 * depth }}>
```
caseyhebebrand, line 100: What is the reason for deleting the `ObjectRow` component? The code here seems the same, but I also don't have much context.

ranihorev: It was in use only here, and it added another layer of abstraction for no clear reason (in addition `ObjectRow` is too vague imo)

app/scripts/modules/core/src/managed/EnvironmentsList.tsx
caseyhebebrand: I think it’s clearer to specify a JSX element’s children using the native `props.children` that JSX has.  The `ResourceDeploymentStatus` component is really just a child component of `ManagedResourceObject`.

caseyhebebrand (approved):

caseyhebebrand (approved):

PR #9069, opened 2021-04-02
fix(md): fix order of environments based on the constraints

ranihorev:
This only fixes the order in the artifacts list view

### DIFF
--- a/app/scripts/modules/core/src/managed/ManagedReader.ts
+++ b/app/scripts/modules/core/src/managed/ManagedReader.ts
@@ -2,13 +2,17 @@ import { flatMap, get, set } from 'lodash';
 
 import { REST } from 'core/api';
 import {
+  IManagedApplicationEnvironmentSummary,
   IManagedApplicationSummary,
+  IManagedArtifactVersionEnvironment,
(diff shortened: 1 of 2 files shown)


luispollo (approved):
LGTM!

app/scripts/modules/core/src/managed/ManagedReader.ts
```diff
@@ -82,16 +100,41 @@ export class ManagedReader {
...
+        artifact.versions.map((version) => {
+          version.environments.sort((a, b) => {
+            if (isEnvDependsOn(a, b)) {
+              return -1;
+            } else if (isEnvDependsOn(b, a)) {
+              return 1;
+            }
+            // Prioritize if has any depends on
+            if (envHasAnyDependsOn(a)) return -1;
+            if (envHasAnyDependsOn(b)) return 1;
```
luispollo, line 116: Aren't these backwards? If `a` has any dependencies, it needs to come later in the list, and so is "greater".

ranihorev: Nope, the ones with dependencies should come first.

PR #9072, opened 2021-04-04
feat(md): allowed times constraints - add times group and fixed timezone

ranihorev:
image

### DIFF
--- a/app/scripts/modules/core/src/managed/constraints/AllowedTimes.less
+++ b/app/scripts/modules/core/src/managed/constraints/AllowedTimes.less
@@ -1,0 +1,4 @@
+.DeploymentWindow {
+  margin-top: 4px;
+  padding-left: 16px;
+}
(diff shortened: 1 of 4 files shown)


app/scripts/modules/core/src/managed/constraints/AllowedTimes.less
```diff
@@ -0,0 +1,4 @@
+.DeploymentWindow {
+  margin-top: 4px;
+  padding-left: 16px;
+}
```
caseyhebebrand, line 4: You can remove this file and use the classnames `sp-margin-xs-top sp-padding-l-left` instead.

See the [styleguide reference](https://github.com/spinnaker/styleguide/blob/master/src/styles/spacing.css) for more info.

ranihorev: I was wondering about the source of these classes! thanks
btw, feels like we could use [tailwindcss](https://tailwindcss.com/). It's really awesome

ranihorev: Fixed, thanks!

caseyhebebrand (approved):
LGTM, thanks for writing tests!

PR #9075, opened 2021-04-06
fix(md): Manual judgement failed text

ranihorev:
Before:
image

After:
image

### DIFF
--- a/app/scripts/modules/core/src/managed/constraints/ManualJudgement.tsx
+++ b/app/scripts/modules/core/src/managed/constraints/ManualJudgement.tsx
@@ -21,7 +21,8 @@ export const ManualJudgementTitle = ({ constraint }: { constraint: IConstraint }
     case 'FAIL':
       return (
         <span>
-          Something went wrong <span className="text-regular">—</span>{' '}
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/src/managed/constraints/ManualJudgement.tsx
```diff
@@ -21,7 +21,8 @@ export const ManualJudgementTitle = ({ constraint }: { constraint: IConstraint }
     case 'FAIL':
       return (
         <span>
-          Something went wrong <span className="text-regular">—</span>{' '}
+          {constraint.comment?.includes('Timed out') ? 'Manual judgement was rejected' : 'Manual judgement failed'}{' '}
+          <span className="text-regular">—</span>{' '}
```
christopherthielen, line 25: This seems pretty brittle, but I'm guessing theres' nothing better to key off of?  Would it be possible to flip the detection logic such that the `Manual judgement was rejected` case is explicitly checked for and everything else falls back to `Manual judgement failed` message?

christopherthielen (approved):

PR #9079, opened 2021-04-08
feat(core): Remove pager page

ranihorev:
This page is not in use anymore


ranihorev:
No need to remove this

PR #9083, opened 2021-04-08
feat(core/taskExecutor): Adds failure message to error

bbrina:
## What does this PR do?
Adds `failureMessage` to task for custom error messaging.

Currently the task message it's being returned as `log` while the `setError` function is looking for a `failureMessage` property so I added that besides the log. 
-- Reference: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/task/monitor/TaskMonitor.ts#L102 

I didn't want to exchange `log` for `failureMessage` because I'm unsure if that's used for something else.

Screen Shot 2021-04-08 at 10 30 03


dogonthehorizon (approved):

bbrina:
@christopherthielen Can we get this merged?

ncknt (approved):

PR #9084, opened 2021-04-09
fix(titus/domain): Add titusUiEndpoint to ITitusInstance

caseyhebebrand:


alanmquach (approved):
LGTM!

app/scripts/modules/titus/src/domain/ITitusInstance.ts
```diff
@@ -26,4 +26,5 @@ export interface ITitusInstance extends IInstance {
   privateIpAddress?: string;
   publicDnsName?: string;
   publicIpAddress?: string;
+  titusUiEndpoint?: string;
```
caseyhebebrand, line 29:
```suggestion
  titusUiEndpoint: string;
```

PR #9088, opened 2021-04-14
fix(core/pipeline): Fixes "stale check" failures when saving pipelines twice, also fixes revert behavior of pipeline tags/description

christopherthielen:


app/scripts/modules/core/src/pipeline/config/pipelineConfigView.html
```diff
@@ -6,7 +6,7 @@
       field-updated="stageFieldUpdated"
       pipeline="isV2TemplatedPipeline ? pipeline : pipelineConfig"
       update-pipeline-config="updatePipelineConfig"
-      view-state="viewState"
+      revert-count="viewState.revertCount"
```
christopherthielen, line 9: pass the primitive, so this react2angular component will re-render

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -371,9 +371,11 @@ angular
           .then(
             (pipelines) => {
               const latestFromServer = pipelines.find((p) => p.id === toSave.id);
-              const latest = latestFromServer || toSave;
-              setOriginal(latest);
-              this.revertPipelineChanges();
+              if (latestFromServer && latestFromServer.updateTs) {
+                toSave.updateTs = latestFromServer.updateTs;
```
christopherthielen, line 375: simplify the updateTs checking

Apparently there is some nasty angular/react glue that is causing updateTs to get clobbered when the `revertPipelineChanges` method was being used.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -450,7 +452,7 @@ angular
               this.navigateTo({ section: 'triggers' });
             }
           }
-          $scope.viewState.revertCount++;
+          $scope.viewState.revertCount = ($scope.viewState.revertCount || 0) + 1;
```
christopherthielen, line 455: if revertCount is undefined, `undefined++` results in `revertCount` being `NaN`

`revertCount` is used to re-mount the react components in Triggers.tsx

app/scripts/modules/core/src/pipeline/config/triggers/MetadataPageContent.tsx
```diff
@@ -104,33 +102,30 @@ export function MetadataPageContent(props: IMetadataPageContentProps) {
   const { pipeline, updatePipelineConfig } = props;
 
   return (
-    <SpinFormik<Partial<IPipeline>>
-      initialValues={pipeline}
-      onSubmit={() => {}}
-      render={(formik) => (
-        <>
-          <WatchValue onChange={updatePipelineConfig} value={formik.values} />
```
christopherthielen, line 112: This SpinFormik made a copy of the pipline as internal state from the `initialValues` prop.  Then, any time the description or tags were edited, it clobbered all pipeline fields with the original values that it initially saved.

I switched this to a controlled FormField to avoid confusion.

app/scripts/modules/core/src/pipeline/config/triggers/Triggers.tsx
```diff
@@ -15,15 +15,15 @@ export interface ITriggersProps {
   pipeline: IPipeline;
   fieldUpdated: () => void;
   updatePipelineConfig: (changes: Partial<IPipeline>) => void;
-  viewState: { revertCount: number };
+  revertCount: number;
```
christopherthielen, line 18: here's where we changed to a primitive value so the react component will get re-rendered by react2angular.

app/scripts/modules/core/src/pipeline/config/triggers/Triggers.tsx
```diff
@@ -54,7 +54,7 @@ export function Triggers(props: ITriggersProps) {
         <NotificationsPageContent {...props} />
       </PageSection>
       <PageSection pageKey="description" label="Metadata" noWrapper={true}>
-        <MetadataPageContent {...props} />
+        <MetadataPageContent {...props} key={revertCountKLUDGE} />
```
christopherthielen, line 57: remount MetadataPageContent when the revertCount changes.

christopherthielen: TBH I am not sure why we don't just remount the entire Triggers component when we revert, but this is all so fragile, I don't want to rock the boat any further in this PR.

christopherthielen (commented):
We should really invest in rewriting the pipelineConfigurer wrapper as React and then get rid of all these kludges.

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
```diff
@@ -363,18 +362,25 @@ angular
...
+          markDirty();
+        });
+      };
+
       this.savePipeline = () => {
         this.setViewState({ saving: true });
         const toSave = _.cloneDeep($scope.pipeline);
         PipelineConfigService.savePipeline(toSave)
-          .then(() => $scope.application.pipelineConfigs.refresh(true))
+          .then(() => this.applyUpdateTs(toSave))
```
alanmquach, line 381: Is this racy?

It looks like `savePipeline` POSTs (but that results in an orchestration task), and we immediately get all pipelineConfigs?

I tried it and it seems fine, but I don't know if that is guaranteed.

app/scripts/modules/core/src/pipeline/config/triggers/MetadataPageContent.tsx
alanmquach: Oh geez, that's right. 🤦

alanmquach: Thanks for fixing this

app/scripts/modules/core/src/pipeline/config/triggers/Triggers.tsx
alanmquach: Remounting everything on revert does seem to be the only reasonable thing to do.

alanmquach (approved):
Overall this looks good, apart from my one question about the race.

Although I'm not too familiar with how the stale check works. Is the updateTs generated by the server? And could you describe what was causing the save twice failure?

Was it simply the fact that the local timestamp was not updating to the new value from the server post-save? I vaguely recall something else being described, but maybe that was a different issue.

christopherthielen:
> Although I'm not too familiar with how the stale check works. Is the updateTs generated by the server? 

When we post the pipeline, we post the entire thing including the updateTs that Deck knows about.  When we use the `staleCheck=true` query parameter, front50 just validates that the payload being submitted matches the updateTs in the database and rejects the update if it doesn't match.

> Was it simply the fact that the local timestamp was not updating to the new value from the server post-save? I vaguely recall something else being described, but maybe that was a different issue.

Yes, or rather it actually WAS updating the local timestamp from the server, but then it later gets "reverted" by some other code, I think (I definitely saw it happen when I updated the pipeline MetaData section).  I tried to solve this the first time using the "revert" logic, but apparently it didn't work correctly.   I tested this code manually and it seems to work correctly

alanmquach:
Got it. Makes sense. Could it have been other `SpinFormik`s causing the same `initialValues` clobber?

app/scripts/modules/core/src/pipeline/config/pipelineConfigurer.js
christopherthielen: the [gate endpoint](https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/PipelineController.groovy#L88) waits for front50 to return success or failure

PR #9091, opened 2021-04-15
feat(core): added graphql framework (no actual functionality)

ranihorev:
The purpose of this PR is to introduce graphql to deck. It's not used anywhere in the code in this PR but only sets the foundations for future PRs.

`graphql-sdk.ts` and `graphql.schema.json` are auto generated by graphql-code-generator. It uses the `graphql` files of the schema and the queries to generate Apollo hooks that will be consumed by the components. 

Ideally the `schema.graphql` won't be commit to this codebase. It can be pulled from the backend server, by replacing the `schema` value in `codegen.yml` or by using a separate node package that will be updated automatically when the backend schema changes. Note: the origin of this file is `keel`. 

TODO: add a hook to refresh the data as part of the refresh loop of `deck`


caseyhebebrand:
I am super excited about this, as I am very bullish on graphql and think deck is a particularly good use case. Do you know if we are going to adopt graphql more broadly in Spinnaker services over time? I would hope yes, but I haven't heard of any concentrated effort outside of MD, unless I am missing some context. 

I think we should set up the framework with that thinking in mind. Right now it looks scoped to `/managed` pathways, but I think anything related to the client or queries for core data sources should be centralized to all of core.

ranihorev:
I'm not aware of any plans at the moment. 

While some the code in the the managed delivery folder, we can easily move it to the core folder and use it for the rest of the app.

vigneshm:
This is great. Thanks for bootstrapping it. As Casey mentioned, could you move the generic ones out of `/managed` folder? I think at least `app/scripts/modules/core/src/managed/graphql/client.ts` could be outside managed.

vigneshm (approved):

vigneshm:
Oops, I added the tag too early. The file can be moved in a separate PR :)

ranihorev:
😂
No problem. I'll move it as part of my next PR

PR #9095, opened 2021-04-20
fix(pipeline): Do not infer user from parent

alanmquach:
This is simply not true.

Especially for executions triggered by a pipeline trigger without a runAs user, they are in actuality executing as anonymous, but we incorrectly display them as being run with the same auth as the parent.

Additionally, we should rely on front50/echo/orca to propagate these from parent to child rather than infer them in deck.


caseyhebebrand (approved):
LGTM

christopherthielen:
How does downstream pipeline authn work in the case of a pipeline stage?

alanmquach:
> How does downstream pipeline authn work in the case of a pipeline stage?

The child pipeline get the auth from the parent, but the propagation should be the responsibility of the backend services. Deck should not need to make these inferences.

FWIW I did verify this to be the case (both authentication.user and trigger.user is propagated to pipeline stages).

PR #9096, opened 2021-04-20
feat(core/pipeline): filter by stages awaiting judgement

christopherthielen:
See #9022 for rationale, this fixes some bugs and reorganizes the code to match the existing paradigm better.
 
Supersedes #9022 
Closes #9022


alanmquach (approved):
Looks good!

This PR (or #9022) does not include recursing into descendant pipelines, right?

For that scenario, I would prefer for there to be a more first class status for running-but-awaiting, which if ever implemented would supersede this, but for a single generation this makes perfect sense.

christopherthielen:
> This PR (or #9022) does not include recursing into descendant pipelines, right?


This does not include that, nope.  I think there is work to be done on the backend for that feature.

PR #9097, opened 2021-04-20
Remove webpack modules + webpack consolidation

ranihorev:
This PR removes the `modules` key from all of the webpack files. 
In addition, it consolidates all of the webpack files into a base one.


vigneshm (approved):
Awesome!! 💯 

Thanks for tackling it. It is much more manageable right now!

PR #9098, opened 2021-04-21
pluginsdk updates

christopherthielen:
1) Fixes the `yarn interactive` command of `pluginsdk-peerdeps`.   Before, it would allow you to update devDependencies inside a docker container, but then it didn't copy the results back out to the correct place.
2) In production mode, terser strips comments from code.  This retains the license comments and also any webpack chunk names, for the case where a plugin has a dynamic import (code split) and wants Deck's webpack chunk to be named something obvious in the network tab.
3) Updates all the rollup tooling for pluginsdk (I need to test this on a plugin or three).


ranihorev (approved):

vigneshm (approved):
LGTM

PR #9103, opened 2021-04-23
feat(core): updated notifications to react-toastify

ranihorev:
API changes: 
- Removed the `body` option as it wasn't in use (verify that)
- Added an `options` field for react toastify option

image

### DIFF
--- a/app/scripts/modules/core/src/widgets/notifier/Notifier.tsx
+++ b/app/scripts/modules/core/src/widgets/notifier/Notifier.tsx
@@ -1,39 +1,9 @@
 import React from 'react';
-import { toast, ToastContainer } from 'react-toastify';
-import { Subscription } from 'rxjs';
-
-import { NotifierService } from './notifier.service';
+import { ToastContainer } from 'react-toastify';
 
 import './notifier.component.less';
 import 'react-toastify/dist/ReactToastify.min.css';
 
-export class Notifier extends React.Component {
-  private subscription: Subscription;
-
-  constructor(props: {}) {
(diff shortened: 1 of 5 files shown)


caseyhebebrand:
Curious what the reasoning is for adding `react-toastify`? For me the benefits seem neutral. We are cleaning up some lines of code, but also adding a new package.

Also, while you are in the `Notifier`, could you make it a functional component? The `useObservable` hook in deck is great for subscription lifecycles.

ranihorev:
The current notifications are hiding the new help menu on the bottom left.
Chris also mentioned that it could be useful for some of his features.

I was trying to minimize the changes. If I need to make more substantial
changes, I can remove this component completely as the subscription service
is not needed anymore.

caseyhebebrand:
> The current notifications are hiding the new help menu on the bottom left. Chris also mentioned that it could be useful for some of his features. 

Ok great. That makes sense, I am happy to hear it will be leveraged more. It sounds like the benefits outweigh the costs 👍🏼 

> I was trying to minimize the changes. If I need to make more substantial changes, I can remove this component completely as the subscription service is not needed anymore.

My opinion is that we should get rid of that if we know that the subscription service isn't needed to remove some tech debt.

ranihorev:
> My opinion is that we should get rid of that if we know that the subscription service isn't needed to remove some tech debt.

Good call :)
I removed it. It's much simpler now

caseyhebebrand (approved):
LGTM - thanks for cleaning up the notifier service!

PR #9104, opened 2021-04-23
fix(titus/serverGroup): Remove empty key/pairs from job attributes

caseyhebebrand:
Server groups with no instances have an empty key/pair `{ "": "" }`. This causes validation issues in the UI when a user wants to clone an empty server group. 

To fix, the empty key/pair is deleted where the other labels are deleted. 

Screen Shot 2021-04-23 at 11 24 09 AM

### DIFF
--- a/app/scripts/modules/titus/src/serverGroup/details/serverGroupDetails.titus.controller.js
+++ b/app/scripts/modules/titus/src/serverGroup/details/serverGroupDetails.titus.controller.js
@@ -108,6 +108,11 @@ angular
           delete labels['source'];
           delete labels['spinnakerAccount'];
 
+          // Server groups with no instances have an empty key/value pair that causes validation issues in the clone modal
+          if (labels[''] === '') {
+            delete labels[''];
+          }
+
           Object.keys(labels).forEach((key) => {
             if (key.startsWith('titus.')) {
               delete labels[key];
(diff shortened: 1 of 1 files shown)


app/scripts/modules/titus/src/serverGroup/details/serverGroupDetails.titus.controller.js
```diff
@@ -108,6 +108,11 @@ angular
           delete labels['source'];
           delete labels['spinnakerAccount'];
 
+          // Server groups with no instances have an empty key/value pair that causes validation issues in the clone modal
+          if (labels[''] === '') {
```
christopherthielen, line 112: is there any reason not to always delete the empty key (i.e., is there a reason to check that the value is also an empty string)?

caseyhebebrand: No, we should always delete this, so I am going to update this code

app/scripts/modules/titus/src/serverGroup/details/serverGroupDetails.titus.controller.js
```diff
@@ -108,6 +108,11 @@ angular
...
+            delete labels[''];
+          }
```
caseyhebebrand, line 114:
```suggestion
          delete labels[''];
```

christopherthielen (approved):

PR #9105, opened 2021-04-23
fix(redblack): Update redblack fields for angular DeploymentStrategyS…

caseyhebebrand:
The `DeploymentStrategySelector` had angular implementations in the `cloneServerGroup` stage. The `onChange` function for additional red-black fields was not being passed to the angular implementation.

This PR includes: 
- Create `onChange` function in the angular controllers and pass to the component
- Force `DeploymentStrategySelector` to re-render. The angular updates are applied to the underlying command, but the UI doesn't update

I really don't like forcing the re-render, but it simple enough for now since I don't have time to refactor the entire stage.

### DIFF
--- a/app/scripts/modules/amazon/src/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
+++ b/app/scripts/modules/amazon/src/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
@@ -132,5 +132,13 @@ module(AMAZON_PIPELINE_STAGES_CLONESERVERGROUP_AWSCLONESERVERGROUPSTAGE, [])
           stage.useAmiBlockDeviceMappings = false;
         }
       };
+
+      this.onRedBlackFieldChange = (key, value) => {
+        if (key === 'rollback.onFailure') {
+          stage.rollback.onFailure = value;
+        } else {
+          stage[key] = value;
+        }
+      };
     },
   ]);
--- a/app/scripts/modules/amazon/src/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
+++ b/app/scripts/modules/amazon/src/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
@@ -134,11 +134,7 @@ module(AMAZON_PIPELINE_STAGES_CLONESERVERGROUP_AWSCLONESERVERGROUPSTAGE, [])
       };
 
       this.onRedBlackFieldChange = (key, value) => {
-        if (key === 'rollback.onFailure') {
-          stage.rollback.onFailure = value;
-        } else {
-          stage[key] = value;
-        }
+        _.set(stage, key, value);
       };
     },
   ]);
--- a/app/scripts/modules/core/src/deploymentStrategy/DeploymentStrategySelector.tsx
+++ b/app/scripts/modules/core/src/deploymentStrategy/DeploymentStrategySelector.tsx
@@ -83,17 +83,11 @@ export class DeploymentStrategySelector extends React.Component<
     this.selectStrategy(this.props.command.strategy, true);
   }
 
-  public onChange = (key: string, value: any) => {
-    this.props.onFieldChange(key, value);
-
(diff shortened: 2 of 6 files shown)


app/scripts/modules/amazon/src/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js
```diff
@@ -132,5 +132,13 @@ module(AMAZON_PIPELINE_STAGES_CLONESERVERGROUP_AWSCLONESERVERGROUPSTAGE, [])
           stage.useAmiBlockDeviceMappings = false;
         }
       };
+
+      this.onRedBlackFieldChange = (key, value) => {
+        if (key === 'rollback.onFailure') {
+          stage.rollback.onFailure = value;
+        } else {
+          stage[key] = value;
+        }
```
christopherthielen, line 141: You can use the lodash set function to eliminate this special casing
```suggestion
        set(stage, key, value);
```

app/scripts/modules/core/src/deploymentStrategy/DeploymentStrategySelector.tsx
```diff
@@ -83,8 +83,15 @@ export class DeploymentStrategySelector extends React.Component<
     this.selectStrategy(this.props.command.strategy, true);
   }
 
+  public onChange = (key: string, value: any) => {
+    this.props.onFieldChange(key, value);
+
+    // This hack is needed to force the component to re-render when the angular deployment-strategy-selector updates the underlying command
+    this.setState({ AdditionalFieldsComponent: this.state.AdditionalFieldsComponent });
```
christopherthielen, line 90: Can we use forceUpdate? https://reactjs.org/docs/react-component.html#forceupdate
```suggestion

    // This hack is needed to force the component to re-render when the angular deployment-strategy-selector updates the underlying command
    this.forceUpdate();
```

app/scripts/modules/titus/src/pipeline/stages/cloneServerGroup/titusCloneServerGroupStage.js
```diff
@@ -93,5 +93,13 @@ module(TITUS_PIPELINE_STAGES_CLONESERVERGROUP_TITUSCLONESERVERGROUPSTAGE, [
       this.processIsSuspended = (process) => {
         return stage.suspendedProcesses && stage.suspendedProcesses.includes(process);
       };
+
+      this.onRedBlackFieldChange = (key, value) => {
+        if (key === 'rollback.onFailure') {
+          stage.rollback.onFailure = value;
+        } else {
+          stage[key] = value;
+        }
```
christopherthielen, line 102: same lodash `set` comment as above

app/scripts/modules/core/src/deploymentStrategy/DeploymentStrategySelector.tsx
```diff
@@ -83,8 +83,15 @@ export class DeploymentStrategySelector extends React.Component<
     this.selectStrategy(this.props.command.strategy, true);
   }
 
+  public onChange = (key: string, value: any) => {
+    this.props.onFieldChange(key, value);
+
+    // This hack is needed to force the component to re-render when the angular deployment-strategy-selector updates the underlying command
```
christopherthielen, line 89: I believe this statement is true, but I don't see how `onChange` gets called when the underlying command is updated?

christopherthielen (approved):

app/scripts/modules/core/src/deploymentStrategy/DeploymentStrategySelector.tsx
caseyhebebrand: `onChange` is called which then changes the underlying command via `onFieldChange`. The data changes but the component doesn't re-render. (`forceUpdate` will be good here).

christopherthielen: My question was asking HOW `onChange` is called by angularjs code when it updates the command object.  
NBD, it's probably not important as long as it works, but I just don't see it.

PR #9108, opened 2021-04-26
fix(titus/serverGroup): Fix height of metrics line chart

christopherthielen:
I missed this when switching from c3 to chart.js.


caseyhebebrand (approved):
Thanks for looking into this!

PR #9110, opened 2021-04-27
fix(core/pipeline): Resolved pipeline header alignment issue by ellip…

siddhu-opsmx:
What is in this PR?

Added a class to Account Tag component to ellipsis long account name , So that it will not affect the header alignment.

Before this PR if we add multiple deploy stages in pipeline consisting long account tag results execution group header slides out of alignment as shown in below screenshot.

alignmentIssue

accountName

After the PR if we add multiple deploy stages in pipeline consisting long account tag results as given in below screenshots

ellipsisAccountTag1

We made ACCOUNT_TAG_OVERFLOW_LIMIT to 1 So that more space will be available to pipeline name and rest of Account Tag is shown as below (by default).

ellipsisAccountTag2


siddhu-opsmx:
@christopherthielen @caseyhebebrand @alanmquach @vigneshm  Could you please review the code.

app/scripts/modules/core/src/pipeline/executions/executionGroup/executionGroup.less
```diff
@@ -127,3 +127,12 @@
   display: block;
   margin-bottom: 5px;
 }
+
+.account-tag-name {
+  max-width: 250px;
+  span {
```
caseyhebebrand, line 133: Can we make a more specific classname on the span? I want to prevent any unintentional side-effects.

siddhu-opsmx:
@caseyhebebrand  We have pushed the changes suggested by you, could you please have a look on it.

caseyhebebrand (approved):

PR #9125, opened 2021-04-29
feature(loadbalancer): Add the ability to configure 'preserve client IP' on NLBs

Bijnagte:


app/scripts/modules/amazon/src/loadBalancer/loadBalancer.transformer.ts
```diff
@@ -470,6 +470,7 @@ export class AwsLoadBalancerTransformer {
               deregistrationDelayConnectionTermination: Boolean(
                 targetGroup.attributes['deregistration_delay.connection_termination.enabled'] === 'true',
               ),
+              preserveClientIp: Boolean(targetGroup.attributes['preserve_client_ip.enabled'] === 'true'),
```
christopherthielen, line 473: lgtm except I'm confused why we use `attributes['preserve_client_ip.enabled']` vs `targetGroup.attributes.preserveClientIp`

christopherthielen (approved):

PR #9127, opened 2021-04-30
fix(titus): Fixing React.Children.only issue

alanmquach:
Hitting a weird edge case in the router.

It looks like `UIRouterContextComponent` uses `React.Children.count` to count children and decide whether or not to wrap it with a `<div>`.

https://github.com/ui-router/react-hybrid/blob/a1da8ce73611a2c7d26499c6415afd0f3af5b937/src/react/UIRouterReactContext.tsx#L69

The gotcha is that when `children` is an array, `React.Children.count` will return the length of the array. So whenever we have an array with `length=1`, `React.Children.only` is used, but that throws on an array.

I wrapped with a fragment to workaround the exception, but I think maybe the router also needs to check `React.isValidElement`?


alanmquach:
Opened https://github.com/ui-router/react-hybrid/pull/526 🤷

christopherthielen (approved):

PR #9130, opened 2021-05-04
fix(style): Ensure that the main div won't overflow the page

ranihorev:
This issue happens if addition divs are added above the main page


vigneshm (approved):
lg

PR #9132, opened 2021-05-04
Add spMenuZuul icon

parshap:
Adding a new navigation menu item for a plugin. Talk to ppourkhomami@netflix.com for context.

* Added icon svg file and exported it from `iconsByName.ts`
* Bumped version in `packages/presentation/package.json`


caseyhebebrand:
@parshap We should move this asset to our internal repo, there is a folder called `assets` where this can live. Also I would love to hear your plans for this so I can help make sure we can set up OSS to support this route if needed.

caseyhebebrand (approved):

PR #9133, opened 2021-05-04
feat(titus): Configure disruption budget in Run Job Stage

caseyhebebrand:
The `disruptionBudget` can already be configured the run job stage JSON, so this PR:
- Updates the existing `JobDisruptionBudget` component so it can be reused in this stage
- Allows user to configure `disruptionBudget` in the UI

runJob_disruptionBudget


app/scripts/modules/titus/src/pipeline/stages/runJob/TitusRunJobStageConfig.tsx
```diff
@@ -169,8 +171,14 @@ export class TitusRunJobStageConfig extends React.Component<IStageConfigProps, I
     this.forceUpdate();
   };
 
+  public disruptionBudgetChanged = (values: IJobDisruptionBudget) => {
+    const { stage, stageFieldUpdated } = this.props;
+    stage.cluster.disruptionBudget = values;
+    stageFieldUpdated();
+  };
```
caseyhebebrand, line 178: This is only needed to sync updates between the parent angular controller and the child react component.

app/scripts/modules/titus/src/serverGroup/configure/wizard/pages/disruptionBudget/JobDisruptionBudget.tsx
```diff
@@ -165,103 +178,106 @@ export class JobDisruptionBudget extends React.Component<IJobDisruptionBudgetPro
...
-            </div>
-            {!isSelfManaged && (
-              <div>
-                <div className={`${budget.rateUnlimited ? '' : 'sp-formGroup'}`}>
-                  <div className={`${budget.rateUnlimited ? '' : 'groupHeader'}`}>
-                    <FormikFormField
-                      name="rates"
-                      label="Rates"
-                      input={(props) => (
+          {showConfig && (
```
caseyhebebrand, line 184: I only added this boolean check.

app/scripts/modules/titus/src/serverGroup/configure/wizard/pages/disruptionBudget/JobDisruptionBudget.tsx
```diff
@@ -99,6 +101,15 @@ export class JobDisruptionBudget extends React.Component<IJobDisruptionBudgetPro
     return options.find((o) => !!disruptionBudget[o.field]);
   }
 
+  componentDidUpdate = (oldProps: IJobDisruptionBudgetProps) => {
+    if (
+      this.props.onStageChange &&
+      !isEqual(oldProps.formik.values.disruptionBudget, this.props.formik.values.disruptionBudget)
+    ) {
+      this.props.onStageChange(this.props.formik.values.disruptionBudget);
```
ranihorev, line 109: Does this happen on every change of the form or only on submit?

app/scripts/modules/titus/src/serverGroup/configure/wizard/pages/disruptionBudget/JobDisruptionBudget.tsx
ranihorev: Why?

christopherthielen (approved):

app/scripts/modules/titus/src/serverGroup/configure/wizard/pages/disruptionBudget/JobDisruptionBudget.tsx
caseyhebebrand: I was just pointing that out because the PR makes it seem like I added everything after line 184

app/scripts/modules/titus/src/serverGroup/configure/wizard/pages/disruptionBudget/JobDisruptionBudget.tsx
caseyhebebrand: Only when the disruption budget is updated. I was having some trouble syncing the underlying model to what was rendered between the angular parent and react child

app/scripts/modules/titus/src/serverGroup/configure/wizard/pages/disruptionBudget/JobDisruptionBudget.tsx
caseyhebebrand: But I added the boolean because it says whether or not to automatically show the extra config or not. (The gif shows an example of that). In the deploy stage, this is always visible but in the run job stage it can be toggled.

app/scripts/modules/titus/src/serverGroup/configure/wizard/pages/disruptionBudget/JobDisruptionBudget.tsx
ranihorev: I was just concerned that it will re-render every time the user types. You can ignore if it's not the case or if it's not a big deal

PR #9135, opened 2021-05-05
chore(aws): Migrates AWS lambda stages from lambda plugin

rvazquezglez:
This PR migrates the lambda stages from [the lambda plugin](https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker), removing references to the plugin framework to load stages directly into Spinnaker.

Includes 5 stages:
image

The stages are not enabled by default, they require this line:
```
window.spinnakerSettings.feature.functions = true; 
```
 on `settings-local.js`


rvazquezglez:
Related: https://github.com/spinnaker/orca/pull/4125

PR #9142, opened 2021-05-07
Remove note for Slack bot channel installation

KeisukeYamashita:
## What

Users will no longer need to install the Slack bot to the destination channel since the `chat.write.public` channel will be able to write to channels that the bot is not a member of.

Therefore, I think it's better to remove this message to avoid no need invitation to the channel.

## Ref

Slacks permission: 
Screen Shot 2021-05-07 at 22 09 53


mattgogerly:
Not overly familiar with Slack permissions, is that setting enabled by default? Or is it opt-in?

PR #9143, opened 2021-05-07
fix(core/pipeline): Warn if concurrency disabled on stage restart

caseyhebebrand:
Users get confused if they restart a stage, but don't see it immediately start. This happens when concurrency is prevented and the restart is queued up.  

Screen Shot 2021-05-07 at 3 03 10 PM


app/scripts/modules/core/src/pipeline/config/stages/stage.module.js
```diff
@@ -357,6 +357,12 @@ module(CORE_PIPELINE_CONFIG_STAGES_STAGE_MODULE, [
           body =
             '<p><strong>This pipeline is currently running - restarting this stage will result in multiple concurrently running pipelines.</strong></p>';
         }
+
+        if ($scope.execution.isRunning && $scope.execution.limitConcurrent) {
```
christopherthielen, line 361: does `scope.execution.isRunning` indicate that ANY execution is running? or that the _current execution_ is running?  Based on the wording I suspect it's the former, and that wouldn't be an accurate test.  It should be testing if any execution is running.

caseyhebebrand: Oh yes that is a good point, this only captures the case where a restart happens when the current execution is still running. I will go fix that to be any execution of this pipeline config :)

vigneshm (approved):

PR #9145, opened 2021-05-07
feat(build): Create build tools package

vigneshm:
Creating a new package named `deck-scripts` for grouping all build-related tooling. This package takes care of providing rollup config, relevant dependencies and exposes scripts that could be utilized directly from deck plugins and modules via npm scripts.

The suggested approach is to add `@spinnaker/pluginsdk` as a dependency to the spinnaker plugin or `@spinnaker/deck-scripts` to a spinnaker module and add the following to `package.json`.

```
{
  ...
  "scripts": {
    "start": "deck-scripts start",
    "build": "deck-scripts build"
  }
}
```

By default `deck-scripts` will look for `rollup.config.js` file in the project directory. If none is found there, then the default rollup config for plugins (defined in `/config/basePluginRollupConfig.js`) will be used. Alternatively, the user can provide the rollup config file via `-f` option.

Plugins can re-use `@spinnaker/pluginsdk/pluginconfig/rollup.confg.js` and customize their bundling. Modules can do that with `@spinnaker/deck-scripts/config/baseModuleRollupConfig.js`.


packages/deck-scripts/config/baseRollupConfig.js
```diff
@@ -0,0 +1,63 @@
+const { nodeResolve } = require('@rollup/plugin-node-resolve');
```
vigneshm, line 1: This file is moved over from `packages/pluginsdk/pluginconfig/rollup.config.js`. But since I made some changes to the other file, git doesn't treat this as a move :(.

packages/deck-scripts/helpers/rollup-plugin-angularjs-template-loader.js
```diff
@@ -0,0 +1,82 @@
+const fs = require('fs');
```
vigneshm, line 1: These changes are copied over from `deck-kayenta`. After publishing these changes, I'll make `deck-kayenta` rely on `deck-scripts` and remove one-off customizations.

packages/deck-scripts/helpers/rollup-plugin-angularjs-template-loader.js
```diff
@@ -0,0 +1,82 @@
...
+// This is a rollup replacement for the ngtemplate-loader webpack loader
+module.exports = function angularJsTemplateLoader(options = {}) {
+  function isSourceMapEnabled() {
+    return options.sourceMap !== false && options.sourcemap !== false;
+  }
+
+  return {
+    transform(originalCode, id) {
+      let code = originalCode;
+      const templateRegex = /require\(['"]([^'"]+\.html)['"]\)/g;
```
vigneshm, line 15: I hope this only catches angular templates. Not really sure if pattern will catch some other library usage.

packages/deck-scripts/config/baseRollupConfig.js
christopherthielen: It's worth splitting this up into two commits.  move the file in one commit, and don't make any other changes.  Then make changes in a successive commit.

packages/deck-scripts/config/baseModuleRollupConfig.js
```diff
@@ -0,0 +1,27 @@
+const baseRollupConfig = require('./baseRollupConfig');
```
christopherthielen, line 1: dope, I like how you split these out

packages/deck-scripts/config/baseModuleRollupConfig.js
```diff
@@ -0,0 +1,27 @@
...
+const angularJsTemplateLoader = require('../helpers/rollup-plugin-angularjs-template-loader');
+
+module.exports = {
+  ...baseRollupConfig,
+  output: [{ dir: 'dist', format: 'es', sourcemap: true }],
+  plugins: [...baseRollupConfig.plugins, angularJsTemplateLoader({ sourceMap: true })],
+  external: [
```
christopherthielen, line 8: I'm curious how you came to this list of externals?

packages/deck-scripts/config/basePluginRollupConfig.js
```diff
@@ -0,0 +1,55 @@
+const baseRollupConfig = require('./baseRollupConfig');
```
christopherthielen, line 1: This looks like it was moved from the pluginsdk.  Did git detect that it was a "rename"?  

I was wondering if git had a mechanism to retain history like this.  It turns out it doesn't.  I'd really like to create one commit where this file gets moved from the pluginsdk and a second commit where the pluginsdk config is re-created with the `module.exports = require('......` so we can retain history on this file.

packages/deck-scripts/helpers/rollup-plugin-angularjs-template-loader.js
christopherthielen: We might be able to tighten up the Regex a bit to look for `templateUrl: require(` and any other permutations that we find in angularjs codee

packages/deck-scripts/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@spinnaker/deck-scripts",
```
christopherthielen, line 2: I'm not loving this name.  all npm pckages in`@spinnaker/` are "deck" and scripts doesn't fully describe what this package does.  Maybe `@spinnaker/build` or `@spinnaker/bundle`.

On second thought, `deck-scripts` does mimic `react-scripts` so maybe `@spinnaker/scripts`?

packages/deck-scripts/helpers/rollup-plugin-angularjs-template-loader.js
christopherthielen: we might be able to tighten this regex up to something like `templateUrl\w*:\w*require(` and add any other patterns besides `templateUrl` that is used in the angularjs code.  Should be pretty easy to find this list by grepping the codebase for `/require\(['"]([^'"]+\.html)['"]\)/g` and filtering out all the known patterns.

```
grep -R  "require.....regex_in_grep_format" deck/app/scripts | grep -v "templateUrl: require"
```

ranihorev:
This is exciting! :)
What's the rationale behind creating another package instead of using it directly in deck? 
so other plugins can use it? if that's the reason, maybe we can expose it via deck directly. 
I'm asking because developing multiple packages could be a bit cumbersome when you need to update multiple at the same time.

packages/deck-scripts/config/baseModuleRollupConfig.js
ranihorev: I usually lean towards putting all the dependencies as externals but I don't know if it's better :)

packages/deck-scripts/config/basePluginRollupConfig.js
```diff
@@ -0,0 +1,55 @@
...
+const externalGlobals = require('rollup-plugin-external-globals');
+
+module.exports = {
+  ...baseRollupConfig,
+  input: 'src/index.ts',
+  output: [{ dir: 'build/dist', format: 'es', sourcemap: true }],
+  plugins: [
+    ...baseRollupConfig.plugins,
+    // map imports from shared libraries (react, etc) to global variables exposed by spinnaker
+    externalGlobals(spinnakerSharedLibraries()),
```
ranihorev, line 11: I'm assuming that we currently use globals. Do you know why by any chance?

packages/deck-scripts/config/basePluginRollupConfig.js
```diff
@@ -0,0 +1,55 @@
...
+
+  // This allows us to share libraries with plugin dependencies that would otherwise be double bundled
+  // @rollup/plugin-commonjs rewrites library names in commonjs imported code
+  // This block finds known permutations and replaces them with the shared global variable
+  return libraries.reduce((globalsMap, libraryName) => {
+    const globalVar = getGlobalVariable(libraryName);
+    const libName2 = libraryName + '?commonjs-proxy';
+    const libName3 = libraryName + '?commonjs-require';
+    const libName4 = '\u0000' + libraryName + '?commonjs-proxy';
+    const libName5 = '\u0000' + libraryName + '?commonjs-require';
```
ranihorev, line 44: Not sure that I understand this (even after reading the comment above). Can you elaborate more?

ranihorev: What's the purpose of `\u0000`?

ranihorev (commented):
I only reviewed this briefly and it looks good.
I'm really excited about getting rid of webpack!

One suggestion: it could be nice to use TypeScript for all of the build scripts. From my experience, it makes them way easier to maintain (and upgrade).
It should be pretty simple with node-ts

packages/deck-scripts/config/baseModuleRollupConfig.js
vigneshm: I should make a comment here explaining that this will be updated as I migrate more spinnaker modules. I came up with this list by looking at the package dependencies in `@spinnaker/titus` and took a guess of what should be provided by `deck`.

packages/deck-scripts/config/basePluginRollupConfig.js
vigneshm: The globals are made available to the plugins. @christopherthielen might know this better, but I believe this is needed since plugins are built and loaded separately from `deck` and these globals are the way to make them available to them. If not, plugin rollup builds will have `import React from 'react'` statements and they won't work.

vigneshm:
> This is exciting! :)
What's the rationale behind creating another package instead of using it directly in deck?
so other plugins can use it? if that's the reason, maybe we can expose it via deck directly.

The scripts will be distributed to plugins via `@spinnaker/pluginsdk` package and will be used directly in all `@spinnaker/*` modules. Here are the reasons that made me create a separate package.

1. We want plugins to rely on `pluginsdk` and not on `@spinnaker/core` in the future. So it makes sense to not include it as part of `@spinnaker/core`. 
2. We could have added it to `pluginsdk`, but then it would make it harder to use the scripts in `@spinnaker/*` modules since there is no dependency on `pluginsdk` and currently `pluginsdk` has a dependency on `spinnaker/core`. 
3. Having a separate package makes it easy to organize and it has a clear purpose.

> I'm asking because developing multiple packages could be a bit cumbersome when you need to update multiple at the same time.

That is true. It really becomes a pain. In the next couple of weeks, I'll be moving deck to use workspaces which should take care of this problem. It should handle automatic linking of modules when developing/testing things locally.

packages/deck-scripts/helpers/rollup-plugin-angularjs-template-loader.js
christopherthielen: > Like is it possible to have a .html require for anything other than an angular template? Definitely wouldn't happen within deck, but I'm wondering if some library package may have that statement for something else (may be web components?) and we may end up breaking that?

Yeah that's my concern.  We could probably mitigate in the rollup plugin by checking the name of the source file and skipping if `node_modules` is in the name

packages/deck-scripts/config/basePluginRollupConfig.js
christopherthielen: the null character is a rollup-ism that marks a module created within a plugin as "synthetic".  In this case, the `@rollup/plugin-commonjs` uses it as a marker for the commonjs compatibility modules it creates under the covers.  

I'm not confident that processing all these permutations is necessary for the plugin's rollup config to work properly, but I think at least some of them are necessary.

packages/scripts/config/baseModuleRollupConfig.js
```diff
@@ -0,0 +1,29 @@
+const baseRollupConfig = require('./baseRollupConfig');
```
christopherthielen, line 1: Style thought (and it doesn't _really matter_ because these files aren't accessed directly or often):

What if we named these three files:

- `rollup.config.base.js`
- `rollup.config.plugin.js`
- `rollup.config.module.js`

It reads a bit better to me.  I was reading where these files were being imported and the specific type (base/plugin/module) wasn't standing out to me in the `require()` statement.

or (I don't prefer this, I like `rollup.config` prefixes)

- `base.rollup.config.js`
- `plugin.rollup.config.js`
- `module.rollup.config.js`

or


- `rollup.config.base.js`
- `rollup.config.base.plugin.js`
- `rollup.config.base.module.js`

thoughts?

packages/scripts/config/baseModuleRollupConfig.js
```diff
@@ -0,0 +1,29 @@
...
+const angularJsTemplateLoader = require('../helpers/rollup-plugin-angularjs-template-loader');
+
+module.exports = {
+  ...baseRollupConfig,
+  output: [{ dir: 'dist', format: 'es', sourcemap: true }],
+  plugins: [...baseRollupConfig.plugins, angularJsTemplateLoader({ sourceMap: true })],
+  // NOTE: The following list of external dependencies will be updated as more spinnaker modules are migrated to
+  // use rollup.
+  external: [
```
christopherthielen, line 10: I think for modules we should make everything external as @ranihorev suggests.  Let webpack figure it out when it bundles everything.

packages/scripts/index.js
```diff
@@ -0,0 +1,117 @@
...
+    })
+    .catch((err) => console.error(err));
+};
+
+const optionsConfigurer = (yargs) =>
+  yargs.option('file', {
+    alias: '-f',
+  });
+
+require('yargs/yargs')(process.argv.slice(2))
```
christopherthielen, line 112: I've seen this before:
```
require('yargs/yargs')(process.argv.slice(2))
```


but I've always simply used:
```
require('yargs')
```
Whats the diff?

packages/scripts/index.js
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env node
+
+const chalk = require('chalk');
+const fs = require('fs');
+const loadConfigFile = require('rollup/dist/loadConfigFile');
+const ora = require('ora');
```
christopherthielen, line 6: check out `listr2` instead of `ora`  

I will message you with an internal repo that shows it in use

packages/scripts/index.js
```diff
@@ -0,0 +1,117 @@
...
+};
+
+const OUTPUT_DIR_REGEX = /\/?(\w+)$/;
+
+const getOutputDir = (output) => {
+  const outputDirMatch = OUTPUT_DIR_REGEX.exec(output);
+  return outputDirMatch.length > 1 ? outputDirMatch[1] : output;
+};
+
+const startHandler = ({ file }) => {
```
christopherthielen, line 32: I don't understand where `file` comes from

packages/scripts/index.js
```diff
@@ -0,0 +1,117 @@
...
+  process.env.ROLLUP_WATCH = true;
+
+  loadConfigFile(getRollupConfigPath(file))
+    .then(async ({ options, warnings }) => {
```
christopherthielen, line 36: in this case I'd prefer to either lean into async/await or use promise chaining, but not mix them

packages/scripts/index.js
```diff
@@ -0,0 +1,117 @@
...
+              delete spinners[key];
+            });
+            break;
+        }
+      });
+
+      watcher.close();
+    })
+    .catch((err) => console.error(err));
+};
```
christopherthielen, line 76: I think this section is worth adding high level comments about what it does (i.e., I think it launches rollup in watch mode, listens for bundler events, and start/stop spinners in the console).

I might be getting too far ahead, but it might be worth thinking about what we will do in the monorepo world. 
 I suspect we'll be watching many modules (core/docker/titus/amazon) at the same time.  Have you given this thought yet?

packages/scripts/index.js
vigneshm: This is a custom file that could be provided by `spinnaker-scripts start -f myRollupConfig.js`. This is defined in `yargs` command.

packages/scripts/index.js
vigneshm: Hmm, I can't remember exactly why, but I had issues with chaining `help` and `demandCommand` when done the other way. But I don't see that happening anymore. 🤷

packages/scripts/config/baseModuleRollupConfig.js
vigneshm: Cool, I'm going to use `rollup-plugin-auto-external` to automatically handle it.

packages/scripts/index.js
vigneshm: > I think this section is worth adding high level comments about what it does (i.e., I think it launches rollup in watch mode, listens for bundler events, and start/stop spinners in the console).

Yep, that's what it is doing here. The code is a little hard to read unfortunately. I'll add some comments here.

> I might be getting too far ahead, but it might be worth thinking about what we will do in the monorepo world.
I suspect we'll be watching many modules (core/docker/titus/amazon) at the same time. Have you given this thought yet?

Yep, that is the idea. We should be able to watch many modules get built in parallel. If we go the `vite` route, I've read that there's some out-of-the-box support for handling serving files from dependent packages in a mono-repo. I'm going to dig deeper into that after migrating all spinnaker packages.

packages/scripts/index.js
vigneshm: I'm going to handle that in a separate PR after playing with `listr2` a bit.

packages/scripts/config/baseModuleRollupConfig.js
vigneshm: I do like the style of `rollup.config.base.js`. Makes it easy to parse that it is a rollup config file. I'm going to do it in a separate PR to make things a little more manageable.

vigneshm: I decided to do it in this PR instead.

packages/pluginsdk/package.json
```diff
@@ -33,15 +33,9 @@
     "yargs": "^16.2.0"
   },
   "devDependencies": {
-    "@rollup/plugin-commonjs": "17.1.0",
-    "@rollup/plugin-node-resolve": "11.1.1",
-    "@rollup/plugin-typescript": "8.1.1",
     "@spinnaker/core": "0.0.549",
+    "@spinnaker/scripts": "file:../scripts",
```
christopherthielen, line 37: this is going to cause problems for plugins, I assume this is temporary?

christopherthielen: oh... maybe it won't cause problems because it's a `devDependency`?

vigneshm: Yep, this is temporary. I'll publish it right after merging this PR.

vigneshm: > but... I think it needs to be an actual dependency so that it gets installed in a plugin when they install the pluginsdk.

Missed this comment. I'm going to test it out real quick.

vigneshm: Yep, that's right. I'm fixing it now. I had tested it locally in a plugin, but I guess I would have moved it over to dev dependency later :(

christopherthielen (approved):

PR #9153, opened 2021-05-11
fix(deck): Attempt to fix `Cannot read property 'length' of undefined`

rvazquezglez:
This is an attempt to fix `Cannot read property 'length' of undefined`

image


rvazquezglez:
The root cause of this error was fixed in https://github.com/spinnaker/deck/pull/9156
But it would be nice to have this small change merged to avoid seeing the "side effect" in case something similar happens again.

caseyhebebrand (approved):

link108:
@mergifyio backport release-1.26.x

PR #9154, opened 2021-05-11
feat(md): application management

ranihorev:
This PR adds a new tab for application management and shows a warning when management is disabled.
@christopherthielen I tried to split the commits better this time, lmk if that helps.

Only needs to review commits newer than 37850e2 (including)

image


caseyhebebrand:
Is there anyway you can just include the relevant files? I feel like all of [these PRs](https://github.com/spinnaker/deck/pulls/ranihorev) are fairly similar and now I am not sure what to review where.

ranihorev:
Unfortunately no, I can't set the base as the other branch, because it
lives on my fork and you won't see it. Sorry...

I can create a diff link for you if it helps

app/scripts/modules/core/src/managed/config/Configuration.tsx
```diff
@@ -0,0 +1,127 @@
+import cx from 'classnames';
```
caseyhebebrand, line 1: nit: For consistency, we usually call import this as: `import classNames from 'classnames'`

app/scripts/modules/core/src/managed/config/ManagementWarning.tsx
```diff
@@ -0,0 +1,46 @@
...
+                    await toggleManagement({ variables: { application: appName, isPaused: false } });
+                    refetch();
+                  },
+                  logCategory: 'App::Management',
+                  onSuccess: refetch,
+                  withComment: false,
+                },
+                { maxWidth: MODAL_MAX_WIDTH },
+              );
+            }}
```
caseyhebebrand, line 37: For readability, can you pull this out into a named function? `onClick={showManagementModal}`

app/scripts/modules/core/src/presentation/horizontalTabs/HorizontalTabs.tsx
```diff
@@ -0,0 +1,53 @@
+import cx from 'classnames';
+import React from 'react';
+import './HorizontalTabs.less';
+
+export const TopNav = () => {};
```
caseyhebebrand, line 5: I am looking per commit, so I may be missing something. Can you explain why this is a noop and where this is used?

app/scripts/modules/core/src/presentation/horizontalTabs/HorizontalTabs.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+type TabsProps = {
+  children: Array<React.ReactElement<TabProps>> | React.ReactElement<TabProps>;
+  className?: string;
+  style?: React.CSSProperties;
+};
```
caseyhebebrand, line 11: For consistency with other components: 
```js
export interface ITabsProps {
  ...allTheProps
}
```

app/scripts/modules/core/src/managed/Loading.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+import { Spinner } from 'core/widgets';
+
+export const Loading = ({ message }: { message: string }) => {
+  return (
+    <div style={{ width: '100%' }}>
+      <Spinner size="medium" message={message} />
+    </div>
+  );
+};
```
caseyhebebrand, line 11: I don't think we need a component for this, it's just a div wrapping an existing component.

christopherthielen:
> @christopherthielen I tried to split the commits better this time, lmk if that helps.

Try squashing all the commits you DONT want reviewed into one commit (it can have a commit message like "DONT REVIEW"), then squash all the commits you DO want reviewed into one or two or three commits.  Then we can review those latter commits separately and ignore the first commit

app/scripts/modules/core/src/presentation/horizontalTabs/HorizontalTabs.tsx
ranihorev: Ooops

caseyhebebrand (approved):
Thanks for addressing the comments!

PR #9157, opened 2021-05-12
fix(ecs): Cannot see instance details for ecs cluster

paragbhingre:
This PR fixes the [#6405](https://github.com/spinnaker/spinnaker/issues/6406)

In 1.25 release, De-angularize InstanceWriter PR got released which broke the existing functionality of the InstanceWriter in ecs module as changes related to ecs got skipped in that particular release. We have fixed the InstanceWriter dependency issue in this PR. 

Screen Shot 2021-05-27 at 10 32 44 AM


test/functional/cypress/integration/ecs/clusters_list.spec.js
```diff
@@ -51,4 +53,39 @@ describe('Amazon ECS: aws-prod-ecsdemo cluster', () => {
...
+      .get('instance-details-header')
+      .get('.header-text')
+      .get('h3:contains("5880e2f5d22e485d834d1d38861147aa")');
+
+    cy.get('.details-panel > .content > .collapsible-section').eq(0)
+      .get('.content-body > dl > dd');
+
+    cy.get('.details-panel > .content > .collapsible-section').eq(1)
+      .get('.content-body > dl > dd').eq(0)
+      .get('span:contains("Up")');
```
allisaurus, line 79: there's a lot of drilling down through DOM elements here in order to get and check the content of this span. This makes the specific UI context we're asserting on hard to pin down and brittle to future changes in the page layout. 

Can we instead add [test IDs](https://github.com/spinnaker/deck/blob/master/test/functional/cypress/integration/ecs/server_group.spec.js#L38-L39) to the relevant spans/DOM elements in the instance detail component and discover them that way?

allisaurus (changes requested):
Code changes for using the new `InstanceWriter` LGTM, I mostly have feedback about the test 
(which I've very happy to see you've added, btw!)

test/functional/cypress/integration/ecs/clusters_list.spec.js
```diff
@@ -51,4 +53,39 @@ describe('Amazon ECS: aws-prod-ecsdemo cluster', () => {
       .get('a:contains("Rollback")')
       .click();
   });
+
+  it('shows stored details view and ECS server group actions1', () => {
```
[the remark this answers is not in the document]
paragbhingre: Oops there was a typo there, I have fixed it.

test/functional/cypress/integration/ecs/clusters_list.spec.js
```diff
@@ -51,4 +53,39 @@ describe('Amazon ECS: aws-prod-ecsdemo cluster', () => {
...
+    cy.visit('#/applications/ecsapp/clusters');
+
+    cy.get('.sub-group:contains("aws-prod-ecsdemo")')
+      .find('.server-group:contains("v000")')
+      .find('.cluster-container')
+      .find('.instances')
+      .find('.instance-group')
+      .get('a[title="f8757e00-184d-4288-b535-4124a739e7be"]')
```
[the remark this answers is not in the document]
paragbhingre: Yes that is true, I had gotten confused between cluster detail panel and instance detail panel. But I have changed this details to show right information.

test/functional/cypress/integration/ecs/clusters_list.spec.js
```diff
@@ -51,4 +53,39 @@ describe('Amazon ECS: aws-prod-ecsdemo cluster', () => {
...
+
+    cy.get('.details-panel > .content > .collapsible-section').eq(1)
+      .get('.content-body > dl > dd').eq(1)
+      .get('span:contains("Up")');
+
+    cy.get('.details-panel > .content > .collapsible-section > .section-heading > .collapsible-heading').eq(2)
+      .click();
+
+    cy.get('.details-panel > .content > .collapsible-section > .content-body > dl > dd').eq(0)
+      .get('vpc-tag > span:contains("SpinnakerVPC (vpc-0a9ccfd1cba7bc715)")');
```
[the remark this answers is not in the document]
paragbhingre: So VPC details are coming from `serverGroup.ecsdemo-v000.json`. So there are few parameters which are getting their values from server group details such as `server group` and `VPC`.

test/functional/cypress/integration/ecs/clusters_list.spec.js
paragbhingre: I have changed functional test to have a minimal level of drilling down through the DOM elements. Please take a look at it and let me know.

test/functional/cypress/integration/ecs/clusters_list.spec.js
```diff
@@ -51,4 +53,35 @@ describe('Amazon ECS: aws-prod-ecsdemo cluster', () => {
...
+      .get('dd:contains("2020-07-20 03:41:03 PDT")');
+
+    cy.get('.details-panel > .content > .collapsible-section').eq(0)
+      .get('dd:contains("aws-final-test-v000")');
+
+    cy.get('.details-panel > .content > .collapsible-section').eq(1)
+      .get('span:contains("Up")');
+
+    cy.get('.details-panel > .content > .collapsible-section > .section-heading > .collapsible-heading').eq(2)
+      .click()
```
allisaurus, line 84: this is still a lot of DOM traveral to the right element. I'd prefer if we did something very explicit, like:
```
cy.get('[data-test-id="SectionHeader.Networking"]').click()
```
...for each element we are checking the contents of. You'll have to add the `data-test-id`s to the actual instance detail page elements, but it will make this much easier to understand and less liable to break b/c of later changes.

test/functional/cypress/integration/ecs/clusters_list.spec.js
```diff
@@ -65,22 +65,25 @@ describe('Amazon ECS: aws-prod-ecsdemo cluster', () => {
...
 
-
     cy.get('.details-panel > .header')
       .get('instance-details-header')
       .get('.header-text')
       .get('h3:contains("f8757e00-184d-4288-b535-4124a739e7be")');
 
-    cy.get('.details-panel > .content > .collapsible-section').eq(0)
+    cy.get('[data-test-id="instanceDetails.content"]')
+      .get('.collapsible-section').eq(0)
```
allisaurus, line 74: is there a reason we can't add test ids directly to the `<collapsible-section>` nodes? if we did that then this `.get()` statement would be able to tell use which section we're clicking into vs. the 0/1/2 section of content.

paragbhingre: So I tried getting a hold of collapsible-section using it's heading but looks like it doesn't work in cypress hence I had to do this 0/1/2.

allisaurus: hmm that's unfortunate. What you have is still definitely an improvement, though, so let's go with it. Thanks very much for this change!

allisaurus (approved):
🎉

allisaurus:
@link108 per https://github.com/spinnaker/spinnaker/issues/6406 it looks like this is a regression present in 1.25 and 1.26. Would we be able to backport this to the next patch releases of these versions?

PR #9168, opened 2021-05-13
feat(md): handle BLOCKED state for constraints

ranihorev:
This mainly affects stateful constraints before they can be acted upon by the user, and therefore we want to hide the actions
image


app/scripts/modules/core/src/managed/overview/artifact/VersionOperation.tsx
```diff
@@ -3,26 +3,37 @@ import React from 'react';
...
-  PENDING: { icon: 'far fa-hourglass', displayName: 'pending' },
-  NOT_EVALUATED: { icon: 'far fa-hourglass', displayName: 'pending' },
+  PENDING: { icon: DEFAULT_ICON, displayName: 'pending' },
+  NOT_EVALUATED: { icon: DEFAULT_ICON, displayName: 'pending' },
+  BLOCKED: { icon: DEFAULT_ICON, displayName: 'pending' },
 };
 
+// This forces the consumer to verify that the value exists (important in the case of a breaking change on the backend)
+export const actionStatusUtils: Partial<ActionStatusUtils> = actionStatusUtilsInternal;
+
```
caseyhebebrand, line 27: Can you explain this a bit more? I don't understand what the reassignment does here.

ranihorev: You can see below where I use the object the return type of this `actionStatusUtils[status]` can be undefined and I have to use `?`. By adding this I'm making sure that adding new options to the backend won't cause the frontend to crash

caseyhebebrand: I don't understand why you can't use `actionStatusUtilsInternal[status]?.icon` If a new status is added to the backend, then `actionStatusUtilsInternal[status]` would also be undefined, and the optional chaining would serve the same purpose.

ranihorev: That's exactly what I'm doing, the purpose of line 26 is to force you to do that :)

caseyhebebrand: I am not seeing a difference between `actionStatusUtilsInternal[status]?.icon` and `actionStatusUtils[status]?.icon`, so I don't think that line 26 is necessary.

ranihorev: The difference is:
- `actionStatusUtilsInternal[status].icon` is valid and won't raise an error
- `actionStatusUtils[status].icon` this is not valid and therefore forces the consume to handle the undefined case properly

caseyhebebrand: But aren't they the same thing? This line just says `actionStatusUtils = actionStatusUtilsInternal`

caseyhebebrand (approved):

PR #9177, opened 2021-05-14
fix(core/task): Render "submitting task" instead of throwing an error in MultiTaskMonitor.tsx

christopherthielen:


christopherthielen:
This code hasn't worked since it was migrated to react.  It seems to work now.

These task monitors are quite a mess :(

Kapture 2021-05-13 at 16 47 29

PR #9178, opened 2021-05-14
refactor(presentation): new collapsible section component

ranihorev:
This PR includes a couple of changes:
- Refactored the component from class component functional component
- Only use the arrow icon and not the glyphicon
- Add an option to use a plus/minus icon instead of the arrow as it's the recommended way (https://www.nngroup.com/articles/accordion-icons/)


app/scripts/modules/core/src/presentation/main.less
```diff
@@ -945,14 +945,53 @@ select:invalid {
...
+    margin-right: 4px;
+  }
+
+  &.right {
+    margin-left: 4px;
+  }
+  display: flex;
 }
 
-.rotated-90 {
```
christopherthielen, line 952: make sure to check other repos to see if this is used by, e.g., deck-kayenta or deck-nflx

app/scripts/modules/core/src/presentation/main.less
```diff
@@ -945,14 +945,53 @@ select:invalid {
...
+
+.rotate-m90 {
+  transition: @rotate-transition;
   transform: rotate(-90deg);
 }
 
+.rotate-p90 {
+  transition: @rotate-transition;
+  transform: rotate(90deg);
+}
```
christopherthielen, line 973: What does `m` and `p` stand for?

christopherthielen:
very like.  much nice.  wow ❤️

christopherthielen (approved):
Looks like there are some build errors.  Otherwise, looks great to me! nice work.

app/scripts/modules/core/src/presentation/main.less
ranihorev: Plus/minus. Is it better to rename to -rotate-90 for the negative classes?

christopherthielen: ah, I see that now.  What about `cw` and ccw`?  Is that more intuitive than `p` and `m`?

@caseyhebebrand @vigneshm @alanmquach

ranihorev:
> Looks like there are some build errors. Otherwise, looks great to me! nice work.

It's related to the presentation package. Do we publish that package separately?

ranihorev:
This was merged as part of another PR. Closing

PR #9185, opened 2021-05-17
refactor(titus): Make titus package independent

vigneshm:
Depends on https://github.com/spinnaker/deck/pull/9145. Cannot land this PR until the other one is merged.

This is the start of a new structure for all spinnaker packages such as `titus`, `core`, `amazon`.

The overall idea is to make each package independent by making them specify their own npm dependencies and use `rollup` for building the package distribution. This helps us in migrating towards a npm-workspace-enabled mono-repo and use tools such as lerna to manage local development and package updates.

We're introducing a new `tsconfig.base.json` file to define the base configuration. Once all the packages are migrated, then we can get rid of the top-level `tsconfig.json` file completely.

## Note
- The build time between `webpack` and `rollup` is comparable. It takes about ~10s for a new build from scratch.
- ~~In order to maintain backward compatibility, `package.json` has additional steps to create a `lib` directory so that existing deck distributions that rely on this directory are not broken. However, I'm not sure if this is needed and I'm inclined towards removing it if possible.~~ This isn't needed.


app/scripts/modules/titus/rollup.config.js
```diff
@@ -0,0 +1,7 @@
+import baseRollupConfig from '@spinnaker/scripts/config/rollup.config.base.module';
+
+export default {
+  ...baseRollupConfig,
+  input: 'src/titus.module.ts',
+  external: ['@spinnaker/amazon', '@spinnaker/core', '@spinnaker/docker'],
```
christopherthielen, line 6: the base config uses a plugin to mark everything as external.  What does this do?  Is it additive? Why is it necessary when (shouldn't the plugin do this for us?)

app/scripts/modules/titus/src/serverGroup/configure/serverGroupConfiguration.service.ts
```diff
@@ -312,8 +312,8 @@ export class TitusServerGroupConfigurationService {
       .map('regions')
       .flattenDeep()
       .filter({ name: this.getRegion(command) })
-      .map<IAmazonLoadBalancer>('loadBalancers')
-      .flattenDeep<IAmazonLoadBalancer>()
+      .map<IAmazonLoadBalancer>((region) => region.loadBalancers)
+      .flattenDeep()
```
christopherthielen, line 316: this looks a bit out of place for this PR, can you explain why this change is here?

christopherthielen:
> However, I'm not sure if this is needed and I'm inclined towards removing it if possible.
Same, I'm not sure that it is necessary to keep lib/ since the entry point is updated to dist/

app/scripts/modules/titus/package.json
```diff
@@ -2,14 +2,43 @@
   "name": "@spinnaker/titus",
   "license": "Apache-2.0",
   "version": "0.0.175",
-  "main": "lib/lib.js",
-  "typings": "lib/index.d.ts",
+  "main": "dist/titus.module.js",
```
christopherthielen, line 5: We should use `"module": "dist/titus.module.js"` to be a bit more explicit that we're shipping ESM.  https://rollupjs.org/guide/en/#publishing-es-modules

It can be in addition to the `main` field, or we could just replace `main` with `module`.

app/scripts/modules/titus/rollup.config.js
vigneshm: Oops, this should be handled by the plugin. I had this in here before the update. Removing it now.

app/scripts/modules/titus/src/serverGroup/configure/serverGroupConfiguration.service.ts
vigneshm: This is fixing a typescript failure. Weirdly, it isn't happening when running `tsc` from `deck`. I didn't spend much time on debugging the cause for the difference though.

app/scripts/modules/titus/package.json
vigneshm: Good call. I missed updating it.

christopherthielen (approved):

PR #9186, opened 2021-05-18
feat(titus/serverGroup): Show custom image ids for deploy configs

caseyhebebrand:
**Background:** Some advanced users have to manually specify the image that they want to use in the deploy config, rather than using the inferred image. To do this, they must update the stage JSON to add the `imageId` attribute. 

**Problem:**  The `imageId` is only visible in the JSON, so when looking at the deploy config an advanced user cannot easily debug which image was used in the deploy. However, we do not want to confuse 99% of users who do not need to configure this value and can use the inferred image during the deploy. 

**Solution:** Add a read-only field if an unexpected `imageId` is present in the deploy config, with help text. 
Screen Shot 2021-05-18 at 10 11 22 AM


vigneshm (approved):
👍

PR #9190, opened 2021-05-19
refactor(build): Fix build scripts to handle independent packages

vigneshm:


vigneshm:
This does generate `yarn.lock` files in the other packages, but I'm hoping that it wouldn't be an issue as they should just be discarded.

app/scripts/modules/build_modules.sh
```diff
@@ -24,6 +24,7 @@ BUILD_ORDER=$(./build_order.sh $*)
 for PACKAGE in $BUILD_ORDER ; do
   beginGroup "$PACKAGE"
   pushd "$PACKAGE" || exit $?
+  yarn || exit 255
```
vigneshm, line 27: Hoping this exit code is ok?

christopherthielen (approved):

PR #9193, opened 2021-05-20
fix(dependency): Fix @types/react version

vigneshm:
`@types/react` was set to `16.8.0`, but many of the dependencies were using `16.0.40` due to how yarn set up the dependencies in the first run. So fixing that to point to `16.8.0` and also fixing a few typescript issues that cropped up.

### DIFF
--- a/app/scripts/modules/amazon/src/serverGroup/configure/AmazonImageSelectInput.tsx
+++ b/app/scripts/modules/amazon/src/serverGroup/configure/AmazonImageSelectInput.tsx
@@ -348,6 +348,7 @@ export class AmazonImageSelectInput extends React.Component<IAmazonImageSelector
 
       return (
         <div className="col-md-9">
+          {/* @ts-ignore */}
           <TetheredSelect
             {...commonReactSelectProps}
             menuRenderer={this.buildImageMenu}
@@ -369,6 +370,7 @@ export class AmazonImageSelectInput extends React.Component<IAmazonImageSelector
       // User can pick an image from the preloaded 'packageImages' using the typeahead
       return (
         <div className="col-md-9">
+          {/* @ts-ignore */}
           <TetheredSelect
             {...commonReactSelectProps}
             menuRenderer={this.buildImageMenu}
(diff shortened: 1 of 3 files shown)


app/scripts/modules/amazon/src/serverGroup/configure/AmazonImageSelectInput.tsx
```diff
@@ -348,6 +348,7 @@ export class AmazonImageSelectInput extends React.Component<IAmazonImageSelector
 
       return (
         <div className="col-md-9">
+          {/* @ts-ignore */}
```
vigneshm, line 351: I didn't have a whole lot of context on why this [customization](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/presentation/TetheredSelect.tsx#L6) was needed in the first place. So making `ts` ignore the failure.

christopherthielen: what was the failure?

christopherthielen (approved):

app/scripts/modules/amazon/src/serverGroup/configure/AmazonImageSelectInput.tsx
vigneshm:
```
Type '{ menuRenderer: (params: MenuRendererProps<OptionValues>) => HandlerRendererResult; isLoading: boolean; placeholder: string; ... 73 more ...; ref?: LegacyRef<...>; }' is not assignable to type 'IntrinsicClassAttributes<TetheredSelect>'.
  Types of property 'ref' are incompatible.
    Type 'LegacyRef<ReactSelectClass<any>>' is not assignable to type 'LegacyRef<TetheredSelect>'.
      Type 'RefObject<ReactSelectClass<any>>' is not assignable to type 'LegacyRef<TetheredSelect>'.
        Type 'RefObject<ReactSelectClass<any>>' is not assignable to type 'RefObject<TetheredSelect>'.
          Property '_renderOuter' is missing in type 'ReactSelectClass<any>' but required in type 'TetheredSelect'.
```

PR #9194, opened 2021-05-20
refactor(amazon): Make amazon package independent

vigneshm:
Continuing with the migration of spinnaker packages to make them independent and use rollup for bundling.

Used this [script](https://gist.github.com/vigneshm/5fdf9c4a9981ec7b75d78f4de2c1fabd) to determine the dependencies and their versions.


ranihorev:
Looks good. How can we verify that this package is working properly?

vigneshm:
> Looks good. How can we verify that this package is working properly?

I use `yalc` to locally publish `@spinnaker/amazon` and pull that it `deck-nflx`. I usually do a quick sanity check to make sure the major flows are working as expected.

PR #9205, opened 2021-05-20
feat(md): show error message if there are no artifacts or resources

ranihorev:
Before: Empty section
Now: Simple error message


app/scripts/modules/core/src/managed/overview/EnvironmentsOverview.tsx
```diff
@@ -63,9 +63,11 @@ const EnvironmentOverview = ({ appName, environment }: IEnvironmentProps) => {
   return (
     <BaseEnvironment title={environment.name}>
       <CollapsibleSection heading="Artifacts" {...sectionProps} defaultExpanded enableCaching={false}>
-        {state.artifacts?.map((artifact) => (
-          <Artifact key={artifact.reference} artifact={artifact} />
-        ))}
+        {state.artifacts ? (
```
caseyhebebrand, line 66: Should this also check `artifacts.length`?

caseyhebebrand (approved):
One minor comment, but LGTM

app/scripts/modules/core/src/managed/overview/EnvironmentsOverview.tsx
ranihorev: Yeah, it won't hurt :)

PR #9224, opened 2021-05-24
feat(md): only show first two pending versions (with expandable section)

ranihorev:
People complained that they have too many versions and it's hard to see the rest of the environments
Kapture 2021-05-23 at 19 50 09


christopherthielen:
@ranihorev I'm wondering if you considered using your other "show more" component?


Screen Shot 2021-05-27 at 1 23 10 PM

ranihorev:
Yes, but I wanted it to feel like part of the time line. Not sure that I
like the end result though...

PR #9243, opened 2021-05-26
fix(cloudfoundry/mapLoadBalancer): Remove validation on Route in order to allow to save Spel Expressions

jorgebee65:
Issue Reference: [issues/6197](https://github.com/spinnaker/spinnaker/issues/6197)
Remove validation in Route in order to allow to save Spel Expressions

cf map load balancer allow spel expression


app/scripts/modules/cloudfoundry/src/pipeline/stages/mapLoadBalancers/cloudfoundryMapLoadBalancersStage.module.ts
```diff
@@ -18,6 +18,6 @@ Registry.pipeline.registerStage({
     { type: 'requiredField', preventSave: true, fieldName: 'credentials', fieldLabel: 'account' },
     { type: 'requiredField', preventSave: true, fieldName: 'region' },
     { type: 'requiredField', preventSave: true, fieldName: 'target' },
-    { type: 'cfRequiredRoutesField', preventSave: true, fieldName: 'loadBalancerNames' },
+    //{ type: 'cfRequiredRoutesField', preventSave: true, fieldName: 'loadBalancerNames' },
```
zachsmith1, line 21: any reason why we want to leave this commented rather than just deleting it?

zachsmith1 (approved):
LGTM pending comment

app/scripts/modules/cloudfoundry/src/pipeline/stages/mapLoadBalancers/cloudfoundryMapLoadBalancersStage.module.ts
jorgebee65: deleted

zachsmith1:
Does this allow you to save an empty loadbalancer name?

jorgebee65:
> Does this allow you to save an empty loadbalancer name?

if is empty the UI shows a required field validation BUT it is possible to save anyways.
cloudfoundry loadbalancer validation

jorgebee65:
The current validation for the domain format (host.some.domain[:port][path]) is still working
An additional validation to allow to save a Spel expression was added
If the field Route is empty then the “Save Changes” button is disabled

zachsmith1 (approved):
LGTM

PR #9245, opened 2021-05-26
feat(md): Added mark as good + refactor actions to mutations and hooks

ranihorev:
image


app/scripts/modules/core/src/managed/overview/artifact/hooks.ts
```diff
@@ -0,0 +1,116 @@
...
+    showModal(
+      UnpinActionModal,
+      {
+        application: payload.application,
+        environment: payload.environment,
+        title: modalTitle,
+        actionName: 'Unpin',
+        onAction: async () => {
+          await onUnpin();
+        },
```
christopherthielen, line 37: just a curiosity: why the async/await? why not just `onAction: () => onUnpin()`?

ranihorev:
The return type of onUnpin doesn't match the return type of onAction

PR #9246, opened 2021-05-26
feat(core): add google analytics to logger

ranihorev:


app/scripts/modules/core/src/reactShims/react.ga.ts
```diff
@@ -1,7 +1,14 @@
 import ReactGA from 'react-ga';
 
 import { SETTINGS } from 'core/config/settings';
+import { logger } from 'core/utils';
 
 if (SETTINGS.analytics.ga) {
   ReactGA.initialize(SETTINGS.analytics.ga, {});
+  logger.subscribe({
```
christopherthielen, line 8: Have you considered flipping the directionality for this?  We use ReactGA.event a lot already.  What if we continued to use ReactGA.event as the primary API for signaling events, and then piped all those events through the logger?

I guess it depends on what sort of things you intend to log using the logger... can you elaborate on your intention?

christopherthielen (approved):

PR #9247, opened 2021-05-26
feat(scripts): Improve build tooling

vigneshm:
Handling a couple of improvements to build tooling.

- Angular template URLs across different packages share similar names (such as `./findAmiStage.html`) between `titus` and `amazon`, and this leads to issues when both these packages are imported into deck. This commit fixes it by expanding the identifier for the template to include the spinnaker package name.
- In the new setup with individual spinnaker packages, it becomes harder to manage linking packages. This adds a new capability where `yalc` is used internally to automatically push updated builds to packages that link spinnaker packages.

### DIFF
--- a/packages/scripts/helpers/rollup-plugin-angularjs-template-loader.js
+++ b/packages/scripts/helpers/rollup-plugin-angularjs-template-loader.js
@@ -15,7 +15,7 @@ module.exports = function angularJsTemplateLoader(options = {}) {
       const templateRegex = /require\(['"]([^'"]+\.html)['"]\)/g;
 
       // look for things like require('./template.html')
-      if (!code.includes("require('") || code.includes('node_modules')) {
+      if (!code.includes("require('") || id.includes('node_modules')) {
         return;
       }
 
@@ -25,6 +25,8 @@ module.exports = function angularJsTemplateLoader(options = {}) {
 
       // Find the directory the JS source file is in
       const baseDir = fs.lstatSync(id).isDirectory() ? id : path.dirname(id);
+      const moduleRootDir = path.resolve('.');
+      const moduleDirName = path.basename(moduleRootDir);
 
       let match = templateRegex.exec(code);
       if (!match) {
@@ -42,6 +44,8 @@ module.exports = function angularJsTemplateLoader(options = {}) {
           throw new Error('Unable to load AngularJS template; could not find template file ' + contentPath);
         }
 
+        const templatePathFromModuleRoot = `${moduleDirName}${path.sep}${path.relative(moduleRootDir, contentPath)}`;
+
         const startIdx = match.index;
         const endIdx = startIdx + match[0].length;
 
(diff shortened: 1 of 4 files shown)


packages/scripts/helpers/rollup-plugin-angularjs-template-loader.js
```diff
@@ -15,7 +15,7 @@ module.exports = function angularJsTemplateLoader(options = {}) {
       const templateRegex = /require\(['"]([^'"]+\.html)['"]\)/g;
 
       // look for things like require('./template.html')
-      if (!code.includes("require('") || code.includes('node_modules')) {
+      if (!code.includes("require('") || id.includes('node_modules')) {
```
vigneshm, line 18: 🤦

packages/scripts/helpers/rollup-plugin-angularjs-template-loader.js
```diff
@@ -25,6 +25,8 @@ module.exports = function angularJsTemplateLoader(options = {}) {
 
       // Find the directory the JS source file is in
       const baseDir = fs.lstatSync(id).isDirectory() ? id : path.dirname(id);
+      const moduleRootDir = path.resolve('.');
```
christopherthielen, line 28: wondering how this will behave if you run e.g. `yarn build` from a subdirectory like `src/core/presentation` and not from the root

christopherthielen:
```suggestion
      const deckRootDir = path.resolve(__dirname, '..', '..', '..');
      const modulesRootDir = path.resolve(deckRoot, 'app', 'scripts', 'modules');
```

packages/scripts/index.js
```diff
@@ -44,6 +45,14 @@ const startHandler = async ({ file }) => {
     // Rollup will be emitting lifecycle events for the bundling process such as start, end and error. These events are
     // used to control a spinner in the terminal for each `input-output` bundle.
     switch (event.code) {
+      case 'END':
+        if (push) {
+          console.log();
+          console.log(chalk.gray.bold('yalc push'));
+          const output = child_process.execSync('yalc push');
```
christopherthielen, line 52: Does `yalc` build the package first before pushing?

christopherthielen, line 52:
```suggestion
          const yalcBin = path.resolve(__dirname, 'node_modules', '.bin', 'yalc');
          const output = child_process.execSync(`${yalcBin} push`);
```

packages/scripts/helpers/rollup-plugin-angularjs-template-loader.js
vigneshm: Good call. I haven't run `yarn` from a subdirectory and TIL that it works!.

packages/scripts/index.js
vigneshm: `yalc push` does not. It uses the existing build, publishes and then pushes it. Since this would be called right after rollup finishes the build, `push` seemed the better option here.

packages/scripts/helpers/rollup-plugin-angularjs-template-loader.js
vigneshm: There's an issue with `path.resolve(__dirname, '..', '..', '..');` though. `__dirname` would point to the one within `node_modules`. For ex, it could be `deck/app/scripts/modules/amazon/node_modules/@spinnaker/scripts/helpers`.

vigneshm: I ran a quick test. `path.resolve('.');` ends up resolving to the directory that has `package.json`. Wondering if `yarn` is doing some magic there.

packages/scripts/index.js
```diff
@@ -44,6 +45,15 @@ const startHandler = async ({ file }) => {
     // Rollup will be emitting lifecycle events for the bundling process such as start, end and error. These events are
     // used to control a spinner in the terminal for each `input-output` bundle.
     switch (event.code) {
+      case 'END':
+        if (push) {
+          console.log();
+          console.log(chalk.gray.bold('yalc push'));
+          const yalcBin = path.resolve(__dirname, 'node_modules', '.bin', 'yalc');
```
vigneshm, line 52: @christopherthielen, I've been thinking about this change. I'm wondering if there's a chance for `@spinnaker/scripts/node_modules/.bin/yalc` not to exist if `yalc` package is hoisted to the parent package's `node_modules`. I tried testing it locally and it didn't fail, but wondering when that hoisting might happen.

packages/scripts/helpers/rollup-plugin-angularjs-template-loader.js
christopherthielen: Cool as long as works 👍

packages/scripts/index.js
christopherthielen: It should still get a link in .bin even if it's hoisted, AFAIK

christopherthielen (approved):

PR #9255, opened 2021-05-27
feat(core/spinner): Consolidate spinners

caseyhebebrand:
The first in a series of PRs to consolidate Spinners. This PR: 
- Updates `Spinner` with a `vertical` and `circular` mode. The vertical mode is the existing loading bars. The circular mode uses a loading circle.
- Updates the `Spinners`'s angular converter with new props
- Replaces `ButtonBusyIndicator` and `LegacySpinner` with a temporary component for backwards compatibility
- Replaces angular `button-busy-indicator` in core module

Upcoming, once this merges and core is bumped:
- In other packages, replace both Angular and React occurrences of `ButtonBusyIndicator` and `LegacySpinner` with circular `Spinner`
- Remove legacy components from `NgReact` once they are no longer used
- Remove controllers and html for the legacy components

### DIFF
--- a/app/scripts/modules/core/src/reactShims/ngReact.ts
+++ b/app/scripts/modules/core/src/reactShims/ngReact.ts
@@ -25,13 +25,11 @@ import { ITaskMonitorProps } from 'core/task/monitor/TaskMonitorWrapper';
 import { TaskMonitorWrapper } from 'core/task/monitor/TaskMonitorWrapper';
 import { IAccountRegionClusterSelectorProps } from 'core/widgets/AccountRegionClusterSelector';
 import { accountRegionClusterSelectorWrapperComponent } from 'core/widgets/accountRegionClusterSelectorWrapper.component';
+import { ButtonBusyIndicator as ButtonBusyIndicatorComponent } from 'core/widgets/spinners/Spinner';
 
-import { IButtonBusyIndicatorProps } from '../forms/buttonBusyIndicator/ButtonBusyIndicator';
-import { buttonBusyIndicatorComponent } from '../forms/buttonBusyIndicator/buttonBusyIndicator.component';
 import { IHelpFieldProps } from '../help/HelpField';
 import { helpFieldWrapperComponent } from '../help/helpField.component';
 import { ReactInject } from './react.injector';
-import { ILegacySpinnerProps, spinnerWrapperComponent } from '../widgets/Spinner';
 
 import IInjectorService = angular.auto.IInjectorService;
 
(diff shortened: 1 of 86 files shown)


app/scripts/modules/core/src/reactShims/ngReact.ts
```diff
@@ -41,12 +39,12 @@ export class NgReactInjector extends ReactInject {
 
   // Reactified components
   public AccountRegionClusterSelector: React.ComponentClass<IAccountRegionClusterSelectorProps> = angular2react('accountRegionClusterSelectorWrapper', accountRegionClusterSelectorWrapperComponent, this.$injectorProxy) as any;
-  public ButtonBusyIndicator: React.ComponentClass<IButtonBusyIndicatorProps>                   = angular2react('buttonBusyIndicator', buttonBusyIndicatorComponent, this.$injectorProxy) as any;
+  public ButtonBusyIndicator: React.FunctionComponent<{}>                               = ButtonBusyIndicatorComponent;
```
christopherthielen, line 42:
```suggestion
  public ButtonBusyIndicator: React.FunctionComponent<{}>                                       = ButtonBusyIndicatorComponent;
```

christopherthielen (approved):
❤️

app/scripts/modules/core/src/widgets/spinners/Spinner.tsx
```diff
@@ -1,15 +1,35 @@
...
+  mode?: 'circular' | 'horizontal';
   fullWidth?: boolean;
+  size?: 'nano' | 'small' | 'medium' | 'large';
 }
 
-export class Spinner extends React.Component<ISpinnerProps> {
-  public getBarRows(): React.ReactNode[] {
-    const { size } = this.props;
+export const Spinner = ({
+  color = '#ffffff',
```
ranihorev, line 15: suggest using `var(--color-white)` instead

app/scripts/modules/core/src/widgets/spinners/Spinner.tsx
```diff
@@ -1,15 +1,35 @@
...
+  mode = 'horizontal',
+  size = 'small',
+}: ISpinnerProps) => {
+  if (mode === 'circular') {
+    const sizeToHeight = {
+      nano: 12,
+      small: 16,
+      medium: 24,
+      large: 32,
+    };
```
ranihorev, line 27: I usually prefer to pull maps like this outside of the component and type them

app/scripts/modules/core/src/widgets/spinners/Spinner.tsx
```diff
@@ -26,27 +46,25 @@ export class Spinner extends React.Component<ISpinnerProps> {
...
-
-    return (
-      <div className={classnames('load', size || 'small', { 'full-width': fullWidth })}>
-        {messageNode}
-        {bars}
-      </div>
-    );
-  }
-}
+  const messageClassNames = `message color-text-accent ${size === 'medium' ? 'heading-4' : 'heading-2'}`;
```
ranihorev, line 51: I know that it was here before but it's kind of odd that only `medium` is using a smaller heading

app/scripts/modules/core/src/widgets/spinners/Spinner.tsx
ranihorev: btw, should we use the color for the bars as well?

PR #9258, opened 2021-05-27
refactor(core/help): Point angular and ngReact to HelpField

caseyhebebrand:
There is both an angular and React version of the `HelpFIeld` used in deck. This PR makes the React component the source of truth for everything.

**In This PR:**
- Removes `HelpField` dependencies on `NgReact`
- Adds a `react2angular` version of `HelpField`
- Repurposes the angular component into a thin wrapper that maps `key` to `id`. This keeps the React component as the source of truth.


vigneshm:
Looks good to me. I wish there was a way to prevent further use of `help-field` over `help-field-react`, but can't think of any.

vigneshm (approved):

PR #9260, opened 2021-05-28
feat(md): improve styles on mobile

ranihorev:
Per @luispollo's request, some improvements for the mobile view + fixes for safari (arrrgggh)

### DIFF
--- a/app/scripts/modules/core/src/managed/Environments2.less
+++ b/app/scripts/modules/core/src/managed/Environments2.less
@@ -3,6 +3,11 @@
   padding: 0 var(--l-spacing);
   --base-horizontal-padding: var(--xl-spacing);
 
+  @media (max-width: 768px) {
+    padding: 0;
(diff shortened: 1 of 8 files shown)


luispollo:
Unfortunately I can't review the CSS changes, but thanks so much for doing this!

app/scripts/modules/core/src/managed/Environments2.less
```diff
@@ -30,4 +35,8 @@
   border: 1px solid var(--color-alto);
   bottom: 4px;
   right: 30px;
+
+  @media (max-width: 768px) {
```
christopherthielen, line 39: I'm wondering if we have these breakpoints as css variables?

christopherthielen (approved):
i'm no css master, but lgtm

ranihorev:
I'll check 👍

ranihorev:
We don't have these breakpoints

PR #9263, opened 2021-05-28
refactor(core): Convert alias imports to relative imports in less

vigneshm:
This is needed to migrate the build tooling to rollup as it doesn't support alias for less imports.


app/scripts/modules/canary/canary/canaryScore.component.less
```diff
@@ -1,4 +1,4 @@
-@import (reference) '~core/presentation/less/imports/commonImports.less';
+@import (reference) '../../core/src/presentation/less/imports/commonImports.less';
```
christopherthielen, line 1: we'll probably want to figure out a better way to manage these cross-package imports.

PR #9264, opened 2021-05-31
feat(scripts): Add support for svgr

vigneshm:


packages/scripts/config/rollup.config.base.js
```diff
@@ -37,6 +38,7 @@ const plugins = [
     // In watch mode, always emit javascript even with errors (otherwise rollup will terminate)
     noEmitOnError: !ROLLUP_WATCH,
   }),
+  svgr(),
```
christopherthielen, line 41: What becomes of the `url` plugin, which is also processing **/*.svg?

```
  url({    include: ['**/*.html', '**/*.svg', '**/*.png', '**/*.jp(e)?g', '**/*.gif', '**/*.webp'],    fileName: '[dirname][hash][extname]',  }),
```

PR #9266, opened 2021-06-01
fix(core): Remove string interpolation token from ng template for rollup

vigneshm:
Rollup has some trouble bundling ng templates that have string interpolation token `${` in them. So fixing it by removing some occurrences in `core`.


christopherthielen:
the rollup plugin wraps the template in string template backticks.   I bet we can quote all the dollar signs `$` -> `\$` to avoid running into this anywhere else

PR #9269, opened 2021-06-01
feat(md): New option to display environments side by side

ranihorev:
To better utilize large screens, I'm trying to display environments side by side. We might want to make it the default but, for now, users can pick their preferred view. 

image


app/scripts/modules/core/src/presentation/hooks/useDimensions.hook.ts
```diff
@@ -0,0 +1,57 @@
...
+    } else {
+      return () => {};
+    }
+  }, [delay, isActive]);
+
+  React.useLayoutEffect(() => {
+    if (isActive) {
+      setDimension({ width: window.innerWidth, height: window.innerHeight });
+    }
+  }, [isActive]);
```
vigneshm, line 23: It would be helpful to add comments describing what is going on here. Why are there two uses of `useLayoutEffect` and what does each one of them do? Also describe the purpose of `delay` and `isActive`. I can't quite figure out what they do within the context of this file.

vigneshm: Just confirmed offline that this is NOT used at all. So going to ignore it and look at the rest

app/scripts/modules/core/src/presentation/hooks/useDimensions.hook.ts
```diff
@@ -0,0 +1,57 @@
...
+}) => {
+  const [dimension, setDimension] = React.useState(getElementDimensions(ref));
+
+  React.useLayoutEffect(() => {
+    const debouncedResizeHandler = debounce(() => {
+      setDimension(getElementDimensions(ref));
+    }, delay);
+
+    if (isActive && ref.current) {
+      const observer = new ResizeObserver(debouncedResizeHandler);
```
vigneshm, line 48: TIL `ResizeObserver`.

vigneshm (approved):
The state management and resize handler bits look good to me.

app/scripts/modules/core/src/managed/environmentBaseElements/EnvironmentsRender.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+const environmentsDirectionState = atom<Direction>({
+  key: 'environmentsDisplay',
+  default: isDirection(storedDirection) ? storedDirection : 'list',
+});
+
+// The goal of this hook is to store the value in an atom to be shared across the app but also update the local storage
+const useEnvironmentDirection = () => {
+  const [direction, setDirection] = useRecoilState(environmentsDirectionState);
+  React.useLayoutEffect(() => {
```
christopherthielen, line 28: I can't think of a reason  why useLayoutEffect would be appropriate here (as opposed to regular old useEffect).  Can you explain?

app/scripts/modules/core/src/presentation/hooks/useDimensions.hook.ts
```diff
@@ -0,0 +1,33 @@
+import { debounce } from 'lodash';
+import React from 'react';
+
+const getElementDimensions = (ref: React.RefObject<HTMLElement>) =>
+  ref.current ? { width: ref.current.offsetWidth, height: ref.current.offsetHeight } : { width: 0, height: 0 };
+
+export const useElementDimensions = ({
```
christopherthielen, line 7: the filename should probably match the name of the hook

app/scripts/modules/core/src/presentation/hooks/useDimensions.hook.ts
```diff
@@ -0,0 +1,33 @@
...
+  ref,
```
christopherthielen, line 8: this is just a style preference comment, not feedback per se

When an arg is "required" I prefer to pass it as a parameter, not as part of an object.  i.e., I would have probably written this as:

```
export const useElementDimension = (ref: ..., options: ...) => 
```

christopherthielen (commented):
nice work

app/scripts/modules/core/src/managed/environmentBaseElements/EnvironmentsRender.tsx
ranihorev: You can't think of a reason because there is no reason :)

PR #9285, opened 2021-06-03
fix(nav): fix unsubscribe on unmount

ranihorev:
The previous code was re-subscribing again instead of unsubscribing

### DIFF
--- a/app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
+++ b/app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
@@ -51,11 +51,8 @@ export const ApplicationNavigation = ({ app }: IApplicationNavigationProps) => {
   });
 
   React.useEffect(() => {
-    appRefreshSubscription();
-
-    return () => {
-      appRefreshSubscription();
-    };
+    const unsubscribe = appRefreshSubscription();
+    return unsubscribe;
   }, []);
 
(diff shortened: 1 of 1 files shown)


caseyhebebrand (approved):

app/scripts/modules/core/src/application/nav/ApplicationNavigation.tsx
```diff
@@ -51,11 +51,8 @@ export const ApplicationNavigation = ({ app }: IApplicationNavigationProps) => {
   });
 
   React.useEffect(() => {
-    appRefreshSubscription();
-
-    return () => {
-      appRefreshSubscription();
-    };
+    const unsubscribe = appRefreshSubscription();
+    return unsubscribe;
```
christopherthielen, line 55: I assume you separated these into two statements so it's more explicit what happens here?

It might be even more better to show an arrow doing an action:

```suggestion
    const unsubscribe = appRefreshSubscription();
    return () =>  unsubscribe();
```

no big deal if you don't like 

FWIW I am pretty sure we have code elsewhere in deck written like this: 
```
React.useEffect(() => appRefreshSubscription(), []);
```
but that is definitely less apparent

christopherthielen (approved):

ranihorev:
Yes, exactly.
I'm fine with your suggestion, but I think that this is explicit enough :)

PR #9291, opened 2021-06-04
feat(md): change the deploying state blinking green instead of teal

ranihorev:


app/scripts/modules/core/src/managed/versionMetadata/VersionMetadata.less
```diff
@@ -54,8 +54,8 @@
   }
 
   .version-deploying {
-    background-color: var(--color-accent);
-    color: var(--color-white);
+    animation: deployingAnimation 2s infinite;
+    color: var(--color-mineshaft);
```
caseyhebebrand, line 58: Loving the name of this color.

caseyhebebrand (approved):

PR #9293, opened 2021-06-04
feat(md): improve error handling for non-managed applications

ranihorev:
I originally wanted to consolidate this error handling to the `Environments` component, but I had some issues with removing the `UIView`, and this solution gives us more flexibility anyway. We might want to revisit it again in the future

### DIFF
--- a/app/scripts/modules/core/src/managed/UnmanagedMessage.tsx
+++ b/app/scripts/modules/core/src/managed/UnmanagedMessage.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+
+import { SETTINGS } from 'core/config';
+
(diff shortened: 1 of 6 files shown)


app/scripts/modules/core/src/managed/UnmanagedMessage.tsx
```diff
@@ -0,0 +1,37 @@
...
+    return <UnmanagedMessage />;
+  }
+  console.warn(error);
+  return (
+    <div style={{ width: '100%' }}>
+      Failed to load environments data, please refresh and try again.
+      <p>{error.message}</p>
+    </div>
+  );
+};
```
caseyhebebrand, line 37: I think this component should be in a separate file since its a unique error state that is unrelated to being unmanaged (or it seems like its unrelated to being unmanaged from my context :) )

PR #9295, opened 2021-06-04
feat(build): Add custom rollup config for core

vigneshm:
The `core` package has some unique characteristics that need a custom rollup config file rather than using the generic one from `@spinnaker/scripts`. This includes

- copying over typescript definition files
- defining an alias mapping for `root`
- a custom external configurer function to accurately mark external modules since `auto-external`
  rollup plugin can't handle imports with paths in them.
- a plugin to fix typescript's path auto-rewrites.


christopherthielen:
I'd like to chat about this, mostly so I understand why this is needed

christopherthielen (approved):

PR #9297, opened 2021-06-04
feat(build): Add custom rollup config for core

vigneshm:
Depends on https://github.com/spinnaker/deck/pull/9295

The `core` package has some unique characteristics that need a custom rollup config file rather than using the generic one from `@spinnaker/scripts`. This includes

- copying over typescript definition files
- defining an alias mapping for `root`
- a plugin to fix typescript's path auto-rewrites.

### DIFF
--- a/app/scripts/modules/core/rollup.config.js
+++ b/app/scripts/modules/core/rollup.config.js
@@ -1,0 +1,48 @@
+import alias from '@rollup/plugin-alias';
+import fs from 'fs';
+import path from 'path';
+import copy from 'rollup-plugin-copy';
+
+import baseRollupConfig from '@spinnaker/scripts/config/rollup.config.base.module';
+import externalConfigurer from '@spinnaker/scripts/helpers/rollup-node-auto-external-configurer';
+
+const packageJSON = JSON.parse(fs.readFileSync('./package.json', 'utf8'));
+const externals = {
+  ...packageJSON.dependencies,
+  'root/version': true,
+};
+
+// In index.ts, the triple slash reference to global namespace types get re-written to match the 'core' alias
+// This kludge re-re-writes them back to a relative path
+// Likely related to https://github.com/microsoft/TypeScript/issues/36763
+const fixTSPathRewrite = () => {
+  return {
+    name: 'fixTSPathRewrite',
+    writeBundle: () => {
+      const dts = path.join(__dirname, 'dist', 'index.d.ts');
+      const fixed = fs
+        .readFileSync(dts)
+        .toString()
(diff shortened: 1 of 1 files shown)


app/scripts/modules/core/rollup.config.js
```diff
@@ -0,0 +1,42 @@
...
+export default {
+  ...baseRollupConfig,
+  plugins: [
+    ...baseRollupConfig.plugins,
+    alias({
+      entries: [{ find: 'root', replacement: path.join(__dirname, '../../../../') }],
+    }),
+    copy({
+      targets: [{ src: 'src/types/**/*', dest: 'dist/types' }],
+    }),
```
christopherthielen, line 38: why does core need types but nothing else does?  For example titus depends on both the amazon and docker packages.

christopherthielen (approved):
couple questions but otherwise lgtm

app/scripts/modules/core/rollup.config.js
vigneshm: I'm going to add some comments in this file.

Rollup uses the typescript plugin to automatically generate the type definitions of modules, however `core` contains some additional typescript declarations such as [`promise.d.ts`](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/types/promise.d.ts) and [`svg.d.ts`](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/types/svg.d.ts) that are needed when linking them to `deck-nflx`.

app/scripts/modules/core/rollup.config.js
```diff
@@ -0,0 +1,42 @@
...
+import fs from 'fs';
+import path from 'path';
+import copy from 'rollup-plugin-copy';
+
+import baseRollupConfig from '@spinnaker/scripts/config/rollup.config.base.module';
+import externalConfigurer from '@spinnaker/scripts/helpers/rollup-node-auto-external-configurer';
+
+const packageJSON = JSON.parse(fs.readFileSync('./package.json', 'utf8'));
+const externals = {
+  ...packageJSON.dependencies,
```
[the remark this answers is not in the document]
vigneshm: As discussed offline, `auto-externals` plugin doesn't work as expected when handling imports with paths in them. There's an [issue](https://github.com/stevenbenisek/rollup-plugin-auto-external/issues/16) that has been open for a while and it hasn't been addressed yet.

There's another plugin mentioned there (https://github.com/Septh/rollup-plugin-node-externals) that appears to solve it, but when I tried using it I ended up getting OOM error, so came up with this tiny little utility function.

PR #9301, opened 2021-06-06
feat(build): Add utility for package bumps

vigneshm:
Adding a temporary utility script that shall be used to bump packages to the latest version across all spinnaker packages until lerna migration is complete.

### DIFF
--- a/app/scripts/modules/bumpPackage.js
+++ b/app/scripts/modules/bumpPackage.js
@@ -1,0 +1,69 @@
+#!/usr/bin/env node
+
+/**
+ * A temporary utility script that shall be used to bump packages to the latest version across all spinnaker packages
+ * until lerna migration is complete.
+ */
+
+/* eslint-disable no-console */
+const { execSync } = require('child_process');
+const { readFileSync } = require('fs');
+const path = require('path');
+
+if (process.argv.length < 3) {
+  console.log('Error: package name must be provided like `./bumpPackage.js package` ');
+  process.exit(1);
+}
+
+const packageToUpgrade = process.argv[2];
+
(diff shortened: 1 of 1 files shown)


christopherthielen (approved):
How will this be utilized?  Would it be useful to bump more than one package to avoid multiple `yarn install`?  There is an existing script in `packages/pluginsdk-peerdeps/sync-versions-from-deck.js` that does version syncing.  If the package exists in top level package.json and also in the target package.json, it updates the target package.json with the semver from the top level one.  Would that be useful here?  How long does this take to run `yarn install`?  If it takes a long time, I wonder if running `yarn install` in parallel would help.

How long you think it will be until we are on lerna?

vigneshm:
> There is an existing script in packages/pluginsdk-peerdeps/sync-versions-from-deck.js that does version syncing. If the package exists in top level package.json and also in the target package.json, it updates the target package.json with the semver from the top level one. Would that be useful here?

Didn't realize the existence of this script. However, I'm going to prune the dependencies in the top-level package.json and also move that into an `app` package to make it easier to move to workspaces.

> How will this be utilized?

I'm primarily looking at updating spinnaker dependencies such as `@spinnaker/scripts`, `@spinnaker/core` in other spinnaker packages. I've had to do it a few times already and it was getting painful, so I wrote this little utility.

> How long does this take to run yarn install? If it takes a long time, I wonder if running yarn install in parallel would help. How long you think it will be until we are on lerna?

It is taking a while actually and haven't explored parallel runs. I'm hoping that I wouldn't need it for more than a week or two as lerna should be ready by then.

PR #9302, opened 2021-06-07
feat(md): Preview environments v0

ranihorev:
This PR adds support for preview environments on the overview page. 
To do that, I refactored the code a bit and added `useOrderedEnvironment` to handle the sorting better. 
I also renamed the directions to `DIRECTIONS = ['listView', 'gridView']`. This means that users will have to set their setting again, but it's not a big deal imo...

### DIFF
--- a/app/scripts/modules/core/src/presentation/hooks/useDimensions.hook.ts
+++ b/app/scripts/modules/core/src/presentation/hooks/useDimensions.hook.ts
@@ -7,7 +7,7 @@ const getElementDimensions = (ref: React.RefObject<HTMLElement>) =>
 export const useDimensions = (
   ref: React.RefObject<HTMLElement>,
   {
-    delay = 200,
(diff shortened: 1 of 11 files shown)


app/scripts/modules/core/src/presentation/hooks/useDimensions.hook.ts
```diff
@@ -17,9 +17,13 @@ export const useDimensions = (
...
   React.useLayoutEffect(() => {
-    const debouncedResizeHandler = debounce(() => {
-      setDimension(getElementDimensions(ref));
-    }, delay);
+    const debouncedResizeHandler = debounce(
+      () => {
+        setDimension(getElementDimensions(ref));
+      },
+      delay,
+      { leading: true },
```
ranihorev, line 25: This ensures that we take an action immediately on first update

PR #9304, opened 2021-06-07
refactor(packages): Migrate packages to make them independent

vigneshm:
This is the start of a new structure for all spinnaker packages under `app/scripts/modules/`.

The overall idea is to make each package independent by making them specify their own npm dependencies and use rollup for building the package distribution. This helps us in migrating towards a npm-workspace-enabled mono-repo and use tools such as lerna to manage local development and package updates.

We're introducing a new `tsconfig.base.json` file to define the base configuration. Once all the packages are migrated, then we can get rid of the top-level `tsconfig.json` file completely.

The changes are split into 2 commits and it makes it easier to review them separately.

## NOTE
The build times are mostly comparable between Rollup and Webpack, but seems Rollup seems faster for `core`. Here's my test run results

| Package  | Rollup | Webpack | 
| ------------- | ------------- | ------------- |
| core  | 56s  | 68s  | 
| amazon  | 20s  | 20s  | 
| docker  | 8s  | 8s  | 
| titus  | 11s  | 12s  |


vigneshm:
Going to fix the build errors.

app/scripts/modules/core/package.json
```diff
@@ -2,14 +2,117 @@
...
-    "prepublishOnly": "npm run lib",
-    "generate-types": "../../../../node_modules/.bin/tsc --emitDeclarationOnly",
-    "webpack": "node ../../../../node_modules/.bin/webpack",
-    "build": "npm run clean && npm run generate-types && npm run webpack"
+    "clean": "shx rm -rf dist",
+    "prepublishOnly": "npm run build",
+    "build": "npm run clean && spinnaker-scripts build",
+    "dev": "spinnaker-scripts start",
+    "dev:push": "spinnaker-scripts start --push",
+    "lib": "npm run build"
```
ranihorev, line 13: Can you describe the purpose of each script in the README?

app/scripts/modules/core/postcss.config.js
```diff
@@ -1,11 +0,0 @@
...
-
-module.exports = {
-  plugins: {
-    autoprefixer: {},
-    'postcss-colorfix': {
-      colors: colorMap,
-    },
-    'postcss-nested': {},
-  },
-};
```
ranihorev, line 11: Do we have it somewhere else?

app/scripts/modules/docker/tsconfig.json
```diff
@@ -1,18 +1,13 @@
...
-    "declarationDir": "lib",
-    "outDir": "lib",
-    "rootDir": "./src",
-    "typeRoots": ["../../../../node_modules/@types"],
+    "outDir": "dist",
     "paths": {
-      "@spinnaker/core": ["../../core/lib"],
-      "core/*": ["core/src/*"]
-    }
+      "core/*": ["../node_modules/@spinnaker/core/lib/*"]
```
ranihorev, line 7: Are we planning to remove these imports as well?

ranihorev: I also thought `core/lib` will be removed

app/scripts/modules/amazon/tsconfig.json
```diff
@@ -1,19 +1,15 @@
...
-    "rootDir": "./src",
-    "typeRoots": ["../../../../node_modules/@types"],
+    "outDir": "dist",
     "paths": {
-      "@spinnaker/core": ["../../core/lib"],
-      "core/*": ["../../core/lib/*"],
-      "amazon/*": ["*"],
-      "amazon": ["."]
-    }
+      "core/*": ["../node_modules/@spinnaker/core/lib/*"],
```
ranihorev, line 7: `lib` -> `dist`?

ranihorev (changes requested):
Looks good, only one issue with some tsconfigs that are still using `lib`.

Do you have any thoughts on managing versions of dependencies across packages? for example, do we want to use the same version of a given dependency in `core` and `amazon`?

vigneshm:
> Looks good, only one issue with some tsconfigs that are still using lib.

We will start using `dist` for these packages only after their dependencies are published. For example, `titus` depends on the prior version of `core` which doesn't bundle the output to `dist` yet. However, `azure` depends on core via path requires so it needs to use `dist`.

app/scripts/modules/docker/tsconfig.json
vigneshm: > Are we planning to remove these imports as well?

It is not planned at the moment. But can be handled via a codemod at the some point after the current migration is done.

app/scripts/modules/core/postcss.config.js
vigneshm: Yep, this is handled in [scripts](https://github.com/spinnaker/deck/blob/master/packages/scripts/config/rollup.config.base.js#L46-L55)

vigneshm:
> Do you have any thoughts on managing versions of dependencies across packages? for example, do we want to use the same version of a given dependency in core and amazon?

Once we migrate to workspaces, we can use the workspace to manage the dependencies. Until then, I have a small utility that I have [here](https://github.com/spinnaker/deck/pull/9301) (that might need some improvements) to do package bumps.

ranihorev (approved):
I haven't tested it locally but looks good, thanks for addressing my questions.

app/scripts/modules/amazon/package.json
```diff
@@ -2,14 +2,57 @@
...
+    "lodash": "4.17.21",
+    "ngimport": "0.6.1",
+    "react": "16.14.0",
+    "react-bootstrap": "0.32.1",
+    "react-ga": "2.4.1",
+    "react-select": "1.2.1",
+    "react-sortable-hoc": "0.6.8",
+    "react-virtualized-select": "3.1.3",
+    "react2angular": "3.2.1",
+    "rxjs": "6.6.7"
```
christopherthielen, line 39: What do you think of making these deps be transitive from `@spinnaker/core`?  Two thoughts: 
- you shouldn't directly reference a package (like import react from React) if it's not in the package.json
- this will be a lot of duplicated dependencies in all the packages.  this can be managed via tooling.

app/scripts/modules/amazon/src/aws.module.ts
```diff
@@ -71,11 +71,13 @@ import { VPC_MODULE } from './vpc/vpc.module';
 
 import './logo/aws.logo.less';
 
+/* Start - Rollup Remove */
```
christopherthielen, line 74: sneaky

christopherthielen (approved):
🤞

app/scripts/modules/amazon/package.json
vigneshm: Not sure I follow. The packages listed here are directly referenced in `@spinnaker/amazon`. (expect `react-ga` which I think was removed recently).

PR #9313, opened 2021-06-09
feat(pluginsdk): Add https support to dev-proxy

parshap:
Enables the `dev-proxy` script used for local Deck plugin development to serve https traffic on port 9443 in addition to http on port 9000. This helps resolve SameSite cookie issues when sharing cookies with an https domain.

Additional context: https://netflix.slack.com/archives/C015HHZMR28/p1623185931034000


parshap:
Is there any documentation that would make sense to update?

vigneshm (approved):
👍  Thanks for fixing it.

PR #9326, opened 2021-06-09
refactor(packages): Migrate packages to make them independent

vigneshm:
Following the changes in https://github.com/spinnaker/deck/commit/9da3751a3b7420eb83ee6b589c1f73b12faed572, the rest of the packages are migrated to have their own dependencies, rollup builds etc.

Users of these packages should make the following changes while migrating to the newer version.

1. The output dir of these packages is moved from `lib` to `dist`. So please update your build tooling to reflect that.
2. Libraries like formik, recoil etc need the same instance of the library to be used. If there's more than one version of these libraries, please use `package.resolutions` in your project to choose a single version of that library.


christopherthielen (approved):
makes sense lgtm 🤞

PR #9340, opened 2021-06-11
refactor(core/entityTag): Convert entity source components to React

caseyhebebrand:
- Convert `EntitySource` to react with an angular converter
- Set up `ngReact` to reference this component

Future:
- Directly use `EntitySource` in amazon module and remove dependency in `ngReact`


app/scripts/modules/core/src/entityTag/EntitySource.tsx
```diff
@@ -1,6 +1,76 @@
...
+        execution={execution as IExecution}
+        executionType={executionType}
+        metadata={metadata.value}
+      />
+    ),
+    [executionId],
+  );
+
+  // eslint-disable-next-line
+  console.log(executionType, status, isLoading, metadata);
```
vigneshm, line 54: You probably forgot to remove it.

app/scripts/modules/core/src/entityTag/EntitySource.tsx
```diff
@@ -1,6 +1,76 @@
...
+    () => (
+      <EntitySourcePopover
+        comments={comments}
+        execution={execution as IExecution}
+        executionType={executionType}
+        metadata={metadata.value}
+      />
+    ),
+    [executionId],
+  );
```
vigneshm, line 51: Curious on the use of `useCallback` here. I'm wondering if we just define `PopoverContent` as a react element, React may do the right thing.

Also, is `executionId` a good proxy for determining a change in the props supplied to `EntitySourcePopover`?

app/scripts/modules/core/src/entityTag/EntitySource.tsx
caseyhebebrand, line 54:
```suggestion
```

app/scripts/modules/core/src/entityTag/EntitySource.tsx
```diff
@@ -1,6 +1,71 @@
...
+    executionType === 'pipeline' ? { application, executionId, stageId } : { application, taskId: executionId };
+
+  const PopoverContent = () => (
+    <EntitySourcePopover
+      comments={comments}
+      execution={execution as IExecution}
+      executionType={executionType}
+      metadata={metadata.value}
+    />
+  );
```
vigneshm, line 48: I was previously referring to the following form. Does this work?

```
const PopoverContent = (
  <EntitySourcePopover
      comments={comments}
      execution={execution as IExecution}
      executionType={executionType}
      metadata={metadata.value}
  />
);
```

vigneshm (approved):

app/scripts/modules/core/src/entityTag/EntitySource.tsx
caseyhebebrand: The component prop will not accept an element, so I had to make this into a functional component.

PR #9353, opened 2021-06-14
fix(imports): Convert requires to imports

vigneshm:
The build tooling is now changed to rollup and we now prefer using imports over requires for modules. So updating the existing call sites that still use requires.


danielpeach (approved):
Thank you!

PR #9358, opened 2021-06-15
feat(md): ignore allowed-time window

emjburns:
Since the `allowed-times` constraint is now stateful, you can override the judgment. This PR adds a little button to the constraint to let you ignore the window. 

Wording options:
`Ignore time window`
`Skip time window`
`Override time window`
`Skip constraint`

Here is what it looks like when it's visible, and when it was overridden.

Screen Shot 2021-06-15 at 1 37 41 PM
Screen Shot 2021-06-15 at 1 39 13 PM

### DIFF
--- a/app/scripts/modules/core/src/managed/constraints/registry.tsx
+++ b/app/scripts/modules/core/src/managed/constraints/registry.tsx
@@ -99,7 +99,7 @@ const baseHandlers: Array<IConstraintHandler<IConstraint['type']>> = [
     overrideActions: {
       FAIL: [
         {
-          title: 'Ignore time window',
+          title: 'Skip constraint',
           pass: true,
         },
       ],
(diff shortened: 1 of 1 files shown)


ranihorev (approved):
lgtm :)
Suggestion: 'Skip constraint"

emjburns:
User feedback 1:

> Rahul Somasunderam  [1:59 PM]
I think Ignore time window and Skip constraint make enough sense.
To me:
Skip time window suggests that it wouldn’t deploy in spite of an open window.
Override time window suggests that we’re changing the window inline during a deployment.

emjburns:
User feedback 2
> Snehal Chennuru  [2:27 PM]
If I were to rank them I would do:
Override time window
Skip constraint
Ignore time window
Skip time window

1 being the most clear.

Override asserts that the operator is overriding the set constraint.

PR #9359, opened 2021-06-16
refactor(core): Reactify ViewChangesLink, ChangesModal, CommitHistory and JarDiff

caseyhebebrand:
Refactor several components to view changes so that we can remove it from `ngReact`. 

**Link**
Screen Shot 2021-06-16 at 1 40 41 PM

**Modal with subcomponents**
Screen Shot 2021-06-16 at 1 40 35 PM


app/scripts/modules/core/src/diffs/ViewChangesLink.tsx
```diff
@@ -1,8 +1,78 @@
...
+export const ViewChangesLink = ({ changeConfig, linkText, nameItem, viewType }: IViewChangesLinkProps) => {
+  const changeConfigValue = changeConfig.metadata.value;
+  const isExecution = changeConfigValue.executionType === 'pipeline';
+
+  const fetchExecution = () => {
+    if (isExecution) {
+      return ReactInjector.executionService.getExecution(changeConfigValue.executionId);
+    }
+    /** A noop promise so `useData` can be utilized */
+    return (new Promise(() => {}) as unknown) as PromiseLike<IExecution>;
```
christopherthielen, line 32: this will return a pending promise that never resolves.  Would it be better to use `Promise.resolve({})` instead?
```suggestion
    return Promise.resolve({}) as PromiseLike<IExecution>;
```

app/scripts/modules/core/src/diffs/ViewChangesLink.tsx
```diff
@@ -1,8 +1,78 @@
...
+  };
+
+  const { result: executionDetails, status } = useData(fetchExecution, {} as IExecution, [
+    changeConfigValue.executionId,
+    changeConfigValue.stageId,
+  ]);
```
christopherthielen, line 38: rule-of-hooks fixes

```suggestion
  const fetchExecution = () => {
    const isExecution = changeConfigValue.executionType === 'pipeline';
    if (isExecution) {
      return ReactInjector.executionService.getExecution(changeConfigValue.executionId);
    }
    /** A noop promise so `useData` can be utilized */
    return (new Promise(() => {}) as unknown) as PromiseLike<IExecution>;
  };

  const { result: executionDetails, status } = useData(fetchExecution, {} as IExecution, [
    changeConfigValue.executionId,
    changeConfigValue.executionType,
  ]);
```

app/scripts/modules/core/src/diffs/ViewChangesLink.tsx
```diff
@@ -1,8 +1,78 @@
...
+
+  const stage = (executionDetails.stages || []).find((s: IExecutionStage) => s.id === changeConfigValue.stageId);
+  const commits = stage?.context?.commits || changeConfig.commits || [];
+  const jarDiffs = stage ? stage.context.jarDiffs : changeConfig.jarDiffs;
```
christopherthielen, line 42: these 2 lines are subtly different in behavior.  can we use the same style though?
```suggestion
  const commits = stage?.context?.commits || changeConfig.commits || [];
  const jarDiffs = stage?.context?.jarDiffs || changeConfig.jarDiffs;
```

app/scripts/modules/core/src/diffs/ViewChangesLink.tsx
```diff
@@ -1,8 +1,78 @@
...
+  const buildInfo = stage
+    ? {
+        ...changeConfig.buildInfo,
+        ...stage.context.buildInfo,
+      }
+    : changeConfig.buildInfo;
+
+  const isLoading = status !== 'RESOLVED';
```
christopherthielen, line 50: IMO, isLoading is a misleading name here because it's true when `status === 'REJECTED'`.  I'd prefer to see this named `isLoaded` to be precise and then invert the logic below

christopherthielen (approved):
nice work, clean components

PR #9360, opened 2021-06-17
fix(core): expose user menu

aleEspinosaM:
fix regression to expose UserMenu so it can be imported from ```@spinakker/core```


dogonthehorizon (approved):

dogonthehorizon:
Without this exported we're unable to successfully override the header in our own install

caseyhebebrand (approved):

PR #9364, opened 2021-06-22
refactor(app): Move app to its own directory

vigneshm:
Moving all "app" related code into its own directory to make it easier to move on to workspaces. All the npm dependencies needed to run the deck app locally are handled now directly in `app/scripts/modules/app/package.json`.

This commit mostly involves just moving all related code in to `app/scripts/modules/app/` along with some updates to references/paths to make the tests work. The ideal folder structure would have all the test related modules for both karma and cypress tests moved into the new app folder, but we are not quite there yet and will get to it after moving to workspaces.

NOTE: `protractor` config file and npm script to run protractor tests are removed since they were not used.


app/scripts/modules/assert_package_bumps_standalone.sh
```diff
@@ -32,8 +32,9 @@ HAS_PURE_PKG_BUMP=false
 for PKGJSON in */package.json ; do
   MODULE=$(basename "$(dirname "$PKGJSON")")
 
+  IS_PRIVATE_PKG=$(jq .private $PKGJSON)
   HAS_PKG_BUMP=$(git diff -U0 "$TARGET_BRANCH" -- "$PKGJSON" | grep -c '"version":')
-  if [ "$HAS_PKG_BUMP" -ne 0 ] ; then
+  if [ $IS_PRIVATE_PKG != "true" ] &&  [ "$HAS_PKG_BUMP" -ne 0 ] ; then
```
vigneshm, line 37: Adding a new fix here to ignore changes in private package json files. Otherwise the build will fail for this PR.

app/scripts/modules/app/src/app.ts
```diff
@@ -17,25 +16,17 @@ import { KAYENTA_MODULE } from '@spinnaker/kayenta';
 import { TITUS_MODULE } from '@spinnaker/titus';
 import { ECS_MODULE } from '@spinnaker/ecs';
 import '@spinnaker/cloudfoundry';
-import { AZURE_MODULE } from '@spinnaker/azure';
-import { HUAWEICLOUD_MODULE } from '@spinnaker/huaweicloud';
-import { TENCENTCLOUD_MODULE } from '@spinnaker/tencentcloud';
```
vigneshm, line 22: These modules are not published to npm, so removing these requires.

caseyhebebrand (approved):
LGTM - if there are any changes other than moving the filepaths and the ones you already pointed out, let me know and I can review more closely.

christopherthielen (approved):

package.json
```diff
@@ -16,13 +16,13 @@
     "yarn": ">=1.21.1"
   },
   "scripts": {
-    "clean": "npx rimraf .awcache .cache-loader build node_modules/.cache transpiled",
-    "build": "NODE_OPTIONS=--max_old_space_size=8192 webpack --bail --mode=production",
+    "clean": "npx rimraf .awcache .cache-loader build node_modules/.cache transpiled && cd app/scripts/modules/app/ && npm run clean",
+    "installDeps": "yarn --cwd app/scripts/modules/app",
+    "build": "npm run installDeps && NODE_OPTIONS=--max_old_space_size=8192 webpack --bail --mode=production --config app/scripts/modules/app/webpack.config.js",
```
christopherthielen, line 21: wait, where do the webpack bundles end up?

vigneshm: They end up in `app/scripts/modules/app/`. I'm going to fix up `clean` after completing the migration.

app/scripts/modules/app/src/app.ts
vigneshm: I'm not sure why these packages weren't published to begin with. We can probably add them back as `file:` dependencies if folks rely on it.

package.json
christopherthielen: this will be a problem for folks using Docker containers.  We should either move, or symlink the output to `/build/webpack`, or update any builds and tooling that expects the output to be off the root directory

vigneshm: Fixing it now.

PR #9366, opened 2021-06-22
feat(md): backend messages 

ranihorev:
In order to replace the delivery config import pipeline, we need a way to show users messages in the UI when the import fails. 
This PR introduces a new UI component for messages, including the management warning, and populates them with messages from the backend. 

image


app/scripts/modules/core/src/managed/graphql/graphql-sdk.ts
```diff
@@ -197,6 +205,20 @@ export interface MdMoniker {
   detail?: Maybe<Scalars['String']>;
 }
 
+export interface MdNotification {
```
luispollo, line 208: Just for my own learning, are these types auto-generated from the GraphQL schema?

app/scripts/modules/core/src/managed/messages/MessageBox.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+const typeToClassName: { [key in Required<IMessageBoxProps>['type']]?: string } = {
+  ERROR: 'fas fa-times',
+  WARNING: 'fas fa-exclamation',
+};
+
+export const MessageBox: React.FC<IMessageBoxProps> = ({ children, type }) => {
+  return (
+    <div className={classnames('MessageBox sp-padding-m', type?.toLowerCase())}>
```
luispollo, line 16: It doesn't look like `type` can be null, can it?

app/scripts/modules/core/src/managed/messages/Messages.tsx
```diff
@@ -0,0 +1,52 @@
...
+  if (!notifications.length) return null;
+
+  return (
+    <MessagesSection>
+      {notifications.map((notification) => (
+        <MessageBox key={notification.id} type={notification.level}>
+          {notification.message}{' '}
+          {notification.triggeredAt && (
+            <>
+              (<RelativeTimestamp timestamp={notification.triggeredAt} />)
```
luispollo, line 33: Nice!

graphql.schema.json
```diff
@@ -294,6 +294,18 @@
             "isDeprecated": false,
             "deprecationReason": null
           },
+          {
```
luispollo, line 297: Again just for my education, it looks like this file is auto-generated from the schema. Is there a reason to keep it in source control? Could it be just generated at build time? 

BTW, could we also retrieve the latest schema from keel as part of the build to avoid the issue with discrepancies? It would be another nudge for us to keep keel OSS up to date.

luispollo (approved):
I can't review the CSS, unfortunately. Everything else looked good to me. Love the new design of the notifications!

app/scripts/modules/core/src/managed/graphql/graphql-sdk.ts
ranihorev: Yes

app/scripts/modules/core/src/managed/messages/MessageBox.tsx
ranihorev: Look at line 5. type is optional

graphql.schema.json
ranihorev: It's complicated :)
We don't use this file at the moment so we can also ignore it, but let's put this aside. 

We do need the sdk file in the source code foe the types. However, the schema isn't necessary and we can generate the sdk file directly by querying the backend, in theory. To do so, we'll need to replace the path of the schema in the type generation config with the endpoint of the backend and somehow deal with authentication. It's kind of a PITA so I decided that it's easier to keep it in the source code...

luispollo: I'm still a bit confused. My thought was that, if the only requirement is for the file to be present in source, it could be generated during deck build time from [the latest schema in the keel OSS repo](https://raw.githubusercontent.com/spinnaker/keel/master/keel-web/src/main/resources/schema/schema.graphql), not at runtime. No authentication required.

ranihorev:
This file is not in use at all at the moment, so it doesn't really matter.

PR #9368, opened 2021-06-22
fix(core/modal): Make SubmitButton compatible with flex display

caseyhebebrand:
Fixing this issue: 
Screen Shot 2021-06-22 at 12 21 10 PM

Adding `flex-container-h` directly to the `button` really messes up all the angular modal footers. Instead I made a new container within the button that allows the button to be flex and the footer to be compatible with the rest of the modal.


vigneshm (approved):

ranihorev:
Thank you, thank you, thank you! :)
It was driving me crazy

cc: @luispollo

PR #9372, opened 2021-06-24
fix(build): Fix gradle builds

vigneshm:
- Webpack build output is now shifted back to the top level from `app/scripts/modules/app` to make sure other tooling that relies on this path is not broken.
- Updated gradle build script to run npm commands rather than directly invoking node scripts. This makes it easier to make changes in package json file and have them working in gradle automatically.


caseyhebebrand (approved):

christopherthielen:
@danielpeach PTAL, let us know if this broke any builds or containers

danielpeach:
Seems OK, I don't think we use any of this stuff.

PR #9379, opened 2021-06-25
chore(core): Remove deps on ngReact for EntitySource and ViewChangesLink

caseyhebebrand:


vigneshm (approved):

vigneshm:
re: failing tests

You might need to upgrade the version of `@spinnaker/core` in `@spinnaker/amazon`. This would not be required once I move everything to workspaces which I'm hoping to complete early next week.

PR #9382, opened 2021-06-29
chore(build): Integrate with yarn workspaces

vigneshm:
Making all packages under `app/scripts/modules/` be part of yarn workspace. This should remove the need to manually handle dependency linking during development. With this setup, just running `yarn dev` under the package (e.g core) would be enough to test it along with the local running version of deck (using `yarn start`).

The other packages under `/packages` will be included into the workspaces in subsequent PRs. Eventually we will have all spinnaker packages listed under `/packages` along with build scripts moved to `/scripts`.


caseyhebebrand:
If there are any updates to the `README` will you include them in this PR as well?

vigneshm:
> If there are any updates to the `README` will you include them in this PR as well?

Great point. I'll update the README once I finish a few more things with the migration.

app/scripts/modules/core/src/widgets/spelText/SpelText.tsx
```diff
@@ -33,7 +33,7 @@ export class SpelText extends React.Component<ISpelTextProps, ISpelTextState> {
   private autocompleteService: SpelAutocompleteService;
   private readonly spelInputRef: any;
   private destroy$ = new Subject();
-  private $input: JQLite;
```
vigneshm, line 36: This change was needed unfortunately. `JQLite` from angular is only defined with `@types/angular@1.6.43` and due to some yarn magic it ended up being made available to `core`. With the migration to workspace, that version can no longer be used since the majority of `core` relies on `@types/angular@1.6.26` and I had to explicitly use that version using `resolutions`.

vigneshm:
Need to fix "production build" and "functional test" actions. Both of them need the individual packages to be built before they can be imported in `app/scripts/modules/app/src/index.ts`.

app/scripts/modules/core/src/managed/Environments2.tsx
```diff
@@ -56,7 +56,7 @@ export const UISwitcher = () => {
 };
 
 const tabsInternal: { [key in Routes]: string } = {
-  overview: 'Overview',
+  overview: 'Test',
```
ranihorev, line 59: Is it temporary? :)

vigneshm: Ah no :( Thanks for catching it. I was testing local development and forgot to remove it.

PR #9401, opened 2021-07-03
feat(md): MD debug mode

ranihorev:
There is some functionality that we don't want to expose to users by default, but could be useful for us for debugging. This PR includes:
- Store to the debug mode to localStorage by setting `md_enabled` as `enabled` and drop it by setting it to `disabled`
- Show the processed delivery config when enabled
- Show "Copy artifact version" when enabled
- More to come


caseyhebebrand (approved):
LGTM

lorin (approved):
Excited to see this!

PR #9407, opened 2021-07-06
feat(cf/versioning): Feat cloudFoundry service versioning

jorgebee65:
The **Deploy Service stage** needs a new checkbox, (updatable already exists)

1. - versioned

if versioned is checked, updatable will get unchecked and disabled (so user can't update it)
if versioned is unchecked, updatable will get enabled.

cf-versioned


zachsmith1 (approved):
LGTM

PR #9413, opened 2021-07-08
feat(managed-delivery): Add resource deleting status and icon

luispollo:
Adds a new `MdResourceActuationStatus` and corresponding icon to represent a resource currently being deleted.

### DIFF
--- a/app/scripts/modules/core/src/managed/overview/Resource.tsx
+++ b/app/scripts/modules/core/src/managed/overview/Resource.tsx
@@ -25,6 +25,7 @@ const statusUtils: {
   NOT_MANAGED: { color: 'var(--color-status-warning)', icon: 'fas fa-pause', defaultReason: 'Resource is not managed' },
   WAITING: { icon: 'far fa-hourglass', defaultReason: 'Resource is currently locked and can not be updated' },
   PROCESSING: { icon: 'far fa-hourglass', defaultReason: 'Resource is being updated' },
+  DELETING: { icon: 'fas fa-trash-o', defaultReason: 'Resource is being deleted' },
 };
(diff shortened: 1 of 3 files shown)


app/scripts/modules/core/src/managed/overview/Resource.tsx
```diff
@@ -25,6 +25,7 @@ const statusUtils: {
   NOT_MANAGED: { color: 'var(--color-status-warning)', icon: 'fas fa-pause', defaultReason: 'Resource is not managed' },
   WAITING: { icon: 'far fa-hourglass', defaultReason: 'Resource is currently locked and can not be updated' },
   PROCESSING: { icon: 'far fa-hourglass', defaultReason: 'Resource is being updated' },
+  DELETING: { icon: 'fas fa-trash-o', defaultReason: 'Resource is being deleted' },
```
luispollo, line 28: @ranihorev I wasn't sure what the difference between `fas` and `far` was here, so I just picked one at random...

ranihorev: Check on fontawesome.com (make sure to filter for free icons)

luispollo: I did, but in fontawesome they all just had `fa` as the prefix. 🤷‍♂️

luispollo: Changed it to `far fa-trash-can` (assuming the `r` is for `regular`): https://fontawesome.com/v6.0/icons?d=gallery&p=1&q=trash&s=regular&m=free

PR #9415, opened 2021-07-09
refactor(build): Use esbuild transpiler with rollup

vigneshm:
The `typescript` rollup plugin does type checking, transpiling and emitting declaration files together in a single process when used with rollup. This is particularly painful during the dev mode as it takes above one minute to finish bundling for `core`.

This commit introduces esbuild plugin to handle transpiling while type checking and declaration file emits are handled via a different process (in parallel) using typescript. The results are encouraging and shared below for `core`.

New
`yarn build`: 33s
`yarn dev`: 12.5s for dist output (exluding type checking and declaration file emits)
`yarn dev` with changes: 4s for dist output (exluding type checking and declaration file emits)

Old
`yarn build`: 55s
`yarn dev`: 55s
`yarn dev` with changes: 20s

This is the first attempt in improving the build times and dev workflow. There are other ideas planned to further reduce the build/refresh times to make it as close to instant refresh.


packages/scripts/config/rollup.config.base.js
```diff
@@ -4,14 +4,14 @@ const commonjs = require('@rollup/plugin-commonjs');
...
-const { terser } = require('rollup-plugin-terser');
-const typescript = require('@rollup/plugin-typescript');
 const url = require('@rollup/plugin-url');
 const svgr = require('@svgr/rollup').default;
 const autoPrefixer = require('autoprefixer');
 const postCssColorFix = require('postcss-colorfix');
 const postCssNested = require('postcss-nested');
 const postCssUrl = require('postcss-url');
+const esbuild = require('rollup-plugin-esbuild');
+const { terser } = require('rollup-plugin-terser');
```
christopherthielen, line 14: `rollup-plugin-esbuild` claims to also replace `rollup-plugin-terser` (for prod builds) but I don't know if it's actually faster

packages/scripts/index.js
```diff
@@ -30,21 +33,50 @@ const getOutputDir = (output) => {
...
+      if (stderr) {
+        console.error(stderr);
+      }
+    })
+    .finally(() => {
+      options.forEach((option) => {
+        const fixTSPathRewritePlugin = option.plugins.find((plugin) => plugin.name == 'fixTSPathRewrite');
+        if (fixTSPathRewritePlugin) {
+          fixTSPathRewritePlugin.writeBundle();
+        }
```
christopherthielen, line 56: What does this do?

packages/scripts/index.js
```diff
@@ -30,21 +33,50 @@ const getOutputDir = (output) => {
   return outputDirMatch.length > 1 ? outputDirMatch[1] : output;
 };
 
+// Runs typescript compiler for type checking and emitting declarations in a separate process to run in parallel with
+// the bundler.
+const runTsc = (options) => {
+  const tscBin = path.resolve(`./node_modules/.bin/tsc`);
+  return exec(`${tscBin} --emitDeclarationOnly`)
+    .then(() => ora().succeed(chalk.green.bold('type checking done')))
+    .catch(({ stderr, stdout }) => {
```
christopherthielen, line 42: If I'm reading this right, it's impossible to fail due to typescript compilation errors?  That's obviously OK for dev mode, but we want to fail when building the prod bundle.

vigneshm: Good call. Going to fix that.

packages/scripts/index.js
vigneshm: It [re-writes](https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/rollup.config.js#L14-L29) typescript's auto path aliasing with references. 

This needs to be run after typescript finishes generating `.d.ts` files and that is why it is run here. This shall be removed after removing aliases from spinnaker modules.

packages/scripts/config/rollup.config.base.js
vigneshm: I ran a quick test and the total runtime is comparable (with `rollup-plugin-terser` running slightly faster by a second). There's some custom logic in our rollup terser plugin for preserving comments which I didn't want to change. I'll investigate esbuild's minify functionality more broadly when experimenting with it for end-to-end builds.

christopherthielen (approved):

PR #9417, opened 2021-07-09
refactor(scripts): Prune build scripts

vigneshm:
- Moving all build scripts defined under `app/scripts/modules/` to `scripts/` folder along with updates to references to these scripts.
- Removing some unused or old scripts and replaced them with newer ones


vigneshm:
I tested the updated build scripts locally (by turning off github action checks within the files) and they all seem to work reasonably well. Will do a fast follow up if I see any failures.

scripts/github_actions_publish.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
 MODULE=$1
 
-# Run this script from the 'deck/app/scripts/modules' directory
+# Run this script from the 'deck/scripts' directory
```
christopherthielen, line 4: `# Run this script from the 'deck/' directory`
?

scripts/publish.js
```diff
@@ -15,10 +17,20 @@ yargs.scriptName('publish.js').usage('$0 [args] [explicitPackage1...] [explicitP
 
 const explicitPackages = yargs.argv._;
 
-process.cwd(__dirname);
 const p = (str) => process.stdout.write(str);
-const files = fs.readdirSync(process.cwd());
-const pkgs = files.filter((file) => fs.statSync(file).isDirectory() && fs.existsSync(path.join(file, 'package.json')));
+const getPackagePath = (pkg) => path.resolve(`${MODULES_DIR}/${pkg}`);
```
christopherthielen, line 21: NBD but I generally try to use path.resolve or path.join everywhere in case windows devs exist

PR #9419, opened 2021-07-09
refactor(all): Remove aliases from all modules

vigneshm:
Ran the following automated script using `jscodeshift` to remove all aliases across all modules (The following script only handles `core`).

```
const path = require("path");
const MODULE_REGEX = /^core\/?(.*)$/;

const DIR = "app/scripts/modules/core/src";

export default function transformer(file, api) {
  const j = api.jscodeshift;
  const ast = j(file.source);

  return ast
    .find(j.ImportDeclaration, {})
    .forEach((nodePath) => {
      const match = MODULE_REGEX.exec(nodePath.value.source.value);
      if (match && match.length > 1) {
        const module = match[1] != "" ? match[1] : "index";

        let newImportSource = path.relative(path.resolve(path.dirname(file.path)), path.resolve(`${DIR}/${module}`));
        if (!newImportSource.startsWith(".")) {
          newImportSource = `./${newImportSource}`;
        }

        nodePath.value.source.value = newImportSource;
      }
    })
    .toSource();
}
```

Tested the changes by running typescript against all the modules.


christopherthielen (approved):
nice work!

christopherthielen:
My browser is freezing due to the number of files.  Does this also remove the aliases from all the configs, i.e., webpack and ts configs?

vigneshm:
> My browser is freezing due to the number of files. Does this also remove the aliases from all the configs, i.e., webpack and ts configs?

Yep, it removes it from all `tsconfig.json` files. We didn't directly use them in webpack though.

PR #9420, opened 2021-07-10
refactor(build): Move packages from app/scripts/modules/ to packages/

vigneshm:
One of the last changes needed for folder restructuring. This is split into 2 changes.

1. The first one simply moves all packages to `/packages` folder and nothing else.
2. The second one updates the references in build scripts, configs etc to make the above move work.

I've temporarily disabled `assert_package_bumps_standalone` to always return 'no pure bump' to have this PR landed. I'll update it after merging this PR.


christopherthielen:
this is a big change but lgtm.

PR #9424, opened 2021-07-12
fix(core): move all graphql code generator components to core and fix the build script

ranihorev:


vigneshm (approved):
Thanks for handling it.

PR #9426, opened 2021-07-13
fix(amazon/securityGroup): Resolve angular/react modal-body conflict

caseyhebebrand:
The `body.modal-open` class was not being applied to the angular Edit Security Group modal because it was competing with the React MD confirmation modal. This caused the z-index of everything to be thrown off after the user pauses management. 

The fix is a hack that forces a 500ms wait so the react modal actually closes before the angular one opens. We use this hack for both MD managed Titus and AWS server groups. 

**Problem**
Screen Shot 2021-07-13 at 2 34 28 PM


christopherthielen (approved):
I love it!
(I secretly hate it)

caseyhebebrand:
> I love it!
> (I secretly hate it)

Same! Always a love-hate relationship when it comes to our two favorite frameworks

PR #9493, opened 2021-07-20
fix(titus): Display entryPointList if not empty

alanmquach:
`entryPointList` takes precedence over `entryPoint` if not null or empty


caseyhebebrand (approved):
LGTM

PR #9495, opened 2021-07-21
feat(build): Exploring vitejs as dev server and build tool

vigneshm:
Adding support to experimentally use `vitejs` as a dev server and build tool for deck. Vite doesn't bundle source code during development and instead transpiles them on the fly when requested by the browser which results in instant server start-up. However, it writes transpiled JS directly to `<script type="module">` which doesn't support old/deprecated JS practices. So this is going to result in runtime failures if either deck or its dependencies rely on such old JS practices.

The rollout of vite is going to be gradual as we fix these runtime issues while testing various user flows. Until then we will continue to use webpack during development and bundling while vite can be tested via `npm run dev:experimental`.

RE: `index.html`
`vite` relies on `index.html` as the starting point unlike other build tools and explicitly needs us to add the JS/TS files in a `<script type="module">` tag. This doesn't work well with our existing webpack html template, so a separate `index.html` file is created.

RE: jquery
Importing `jquery` directly in `app.ts` doesn't set `window.$` or `window.jQuery` which some of the other dependencies rely on, so adding it directly in `index.html`.

RE: `VITE_` prefix on env variables
vite only passes env variables prefixed with `VITE_` into app files and makes it available through `import.meta.env`. It also uses `dotenv` to load values for the env variables. So it is handly to define a `.env.local` file in `app/scripts/modules/app` folder and specify the values typically needed to run deck. On a related note, since webpack doesn't work with `import.meta.env`, a custom webpack loader has been created to replace that with `process.env`.


ranihorev:
This is exciting!
Any reason not to use vote foe development? (I understand why not for production)

vigneshm:
> Any reason not to use vote foe development? (I understand why not for production)

You can use it by `npm run dev:experimental` for development if you like (although there will be some runtime errors you might have to deal with), but it is not quite ready to replace webpack.

ranihorev:
Dogfooding ftw :)
I can spend the time I save on waiting for it to rebuild on fixing those
bugs ;)

vigneshm:
Going to look into the test failures.

app/scripts/modules/app/index.html
```diff
@@ -0,0 +1,268 @@
...
+          })(window, document, 'script', 'https://www.google-analytics.com/analytics.js', 'ga');
+        }
+      });
+    </script>
+
+    <!-- Actually injects our application -->
+    <spinnaker ng-if="!authenticating"></spinnaker>
+
+    <script src="https://code.jquery.com/jquery-3.5.1.min.js"></script>
+    <script type="module" src="./src/settings.js"></script>
```
ranihorev, line 265: Looking at the webpack config - What about `settings-local`? Do we need it?

app/scripts/modules/app/vite.config.js
```diff
@@ -0,0 +1,46 @@
...
+
+const envLocalFilePath = path.resolve(`${__dirname}/.env.local`);
+if (fs.existsSync(envLocalFilePath)) {
+  dotenv.config({
+    path: envLocalFilePath,
+  });
+}
+
+export default defineConfig({
+  clearScreen: false,
```
ranihorev, line 20: Do we want to keep this in dev? (I'm not sure)

app/scripts/modules/app/vite.config.js
```diff
@@ -0,0 +1,46 @@
...
+      ),
+    }),
+    angularTemplateLoader({ sourceMap: true }),
+  ],
+  resolve: {
+    alias: [
+      { find: 'root', replacement: DECK_ROOT },
+      {
+        find: 'coreImports',
+        replacement: `${NODE_MODULE_PATH}/@spinnaker/core/src/presentation/less/imports/commonImports.less`,
```
ranihorev, line 36: What's the deal with this file btw? How's it different from the rest of the css files? :)

app/scripts/modules/app/vite.config.js
```diff
@@ -0,0 +1,46 @@
...
+      },
+    ],
+    mainFields: ['module', 'jsnext:main', 'jsnext', 'main:esnext'],
+  },
+  server: {
+    host: process.env.DECK_HOST,
```
ranihorev, line 42: It could be nice to add the `open: true` option imo

app/scripts/modules/app/vite.config.js
```diff
@@ -0,0 +1,46 @@
...
+  ],
+  resolve: {
+    alias: [
+      { find: 'root', replacement: DECK_ROOT },
+      {
+        find: 'coreImports',
+        replacement: `${NODE_MODULE_PATH}/@spinnaker/core/src/presentation/less/imports/commonImports.less`,
+      },
+    ],
+    mainFields: ['module', 'jsnext:main', 'jsnext', 'main:esnext'],
```
ranihorev, line 39: Do we use all of them?

app/scripts/modules/app/vite.config.js
```diff
@@ -0,0 +1,46 @@
...
+  },
+  server: {
+    host: process.env.DECK_HOST,
+    https: process.env.DECK_HTTPS === 'true' ? { maxSessionMemory: 100, peerMaxConcurrentStreams: 300 } : false,
```
ranihorev, line 43: Worth adding a comment with a link to this issue for future reference:
https://github.com/vitejs/vite/pull/3895

ranihorev (commented):
Looks great, thanks for doing this!

app/scripts/modules/app/index.html
```diff
@@ -0,0 +1,268 @@
+<!DOCTYPE html>
+<html class="no-js" ng-app="netflix.spinnaker" ng-strict-di="true">
+  <head>
+    <title>Spinnaker</title>
+    <meta charset="utf-8" />
+    <meta name="description" content="" />
+    <meta name="viewport" content="width=device-width" />
+
+    <!-- Sample loading styles -->
```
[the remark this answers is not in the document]
vigneshm: No idea if it is being used. I wanted to have minimum changes outside of what was required to get this working.

app/scripts/modules/app/index.html
vigneshm: With vitejs, `.env.local` file will be used to provide local overrides.

app/scripts/modules/app/vite.config.js
vigneshm: I definitely want to keep it in dev 'cos it clears some of the errors which is not ideal.

app/scripts/modules/app/vite.config.js
vigneshm: This is an alias that directly relies on a dependency's module (https://github.com/spinnaker/deck/blob/abf4c7590ee4bf84adfb1624c5aafcaccf829be5/app/scripts/modules/app/src/canary/canary/canaryScore.component.less).

app/scripts/modules/app/vite.config.js
vigneshm: `main:esnext` was needed by one of the libraries we use while the rest of are recommended by vite.

PR #9531, opened 2021-07-28
fix(vite): Add vite fixes

vigneshm:
Some relatively minor vite fixes.


app/scripts/modules/app/vite.config.js
```diff
@@ -42,7 +42,6 @@ export default defineConfig({
     host: process.env.DECK_HOST,
     // See https://github.com/vitejs/vite/pull/3895 for details on the config.
     https: process.env.DECK_HTTPS === 'true' ? { maxSessionMemory: 100, peerMaxConcurrentStreams: 300 } : false,
-    open: true,
```
vigneshm, line 45: Removing this 'cos it opens up a new tab every time instead of reloading the existing tab.

PR #9532, opened 2021-07-28
refactor(build): Handle jquery-ui in webpack and vite

vigneshm:
`jquery-ui` relies on `amd` style module definition to manage module dependencies within its package. This worked with `webpack` since it automatically supports `amd` style module definitions, but this doesn't work well with `vite`.

In this commit, `jquery-ui` is removed from ts/js files and instead added to html/template file. This way it doesn't rely on `amd` style definitions anymore and directly adds features to global `jQuery` object which would be used by angularjs based libraries..


app/scripts/modules/app/webpack.config.js
```diff
@@ -162,10 +162,6 @@ function configure(env, webpackOpts) {
           test: require.resolve('jquery'),
           use: [{ loader: 'expose-loader?$' }, { loader: 'expose-loader?jQuery' }],
         },
-        {
-          test: /ui-sortable/,
-          use: ['imports-loader?$UI=jquery-ui/ui/widgets/sortable'],
-        },
```
christopherthielen, line 168: did you check that this angularjs drag-sort feature still works?  I think it's only used in one or two places anymore.

christopherthielen (approved):

app/scripts/modules/app/webpack.config.js
vigneshm: Yep, it is used in re-ordering links in config page and re-ordering deploy stages in pipeline config. I confirmed that both are working.

PR #9534, opened 2021-07-28
fix(all): Fix lodash global usage

vigneshm:
Remove direct usages of lodash without imports.


christopherthielen:
> Many modules across different spinnaker app packages directly use _ without importing lodash

This is surprising to me...  is it in Javascript code (not TS)?  I would love if deck code didn't rely on globals.

vigneshm:
> > Many modules across different spinnaker app packages directly use _ without importing lodash
> 
> This is surprising to me... is it in Javascript code (not TS)? I would love if deck code didn't rely on globals.

You're right, it is only js files. I was using an incorrect ast for searching it earlier but now it looks like only a handful of files. Let me update them directly.

PR #9538, opened 2021-07-30
feat(core/pipeline): Check if deploy stage has a trigger that provide…

caseyhebebrand:
**Problem**
For deploys to cloud providers like Titus, the image metadata is expected to be inferred from the pipeline trigger. If not, the user must manually specify an image Id. This is not clear to the user when the pipeline is set up. So, they will create and run their pipeline and have an image-related failure. Currently, there is no way to warn the user that their pipeline is in a bad state. 

**Solution**
Add a pipeline validator that checks for triggers that provide images or a custom image id. If none are present, warn the user who is configuring the pipeline.
Screen Shot 2021-07-30 at 11 15 28 AM


packages/core/src/pipeline/config/stages/deploy/deployStage.js
```diff
@@ -39,6 +39,16 @@ angular
                   clusterServiceProvider.$get().isDeployingArtifact(cluster),
               ),
           },
+          {
+            type: 'imageProviderBeforeType',
+            message: 'You must have a trigger that provides the image metadata  before any deploy stage.',
```
ranihorev, line 44:
```suggestion
            message: 'You must have a trigger that provides the image metadata before any deploy stage.',
```

vigneshm (approved):
lgtm

packages/core/src/pipeline/config/validation/imageProviderBeforeType.validator.ts
```diff
@@ -0,0 +1,30 @@
...
+    _config: IStageOrTriggerTypeConfig,
+  ): string {
+
+    const triggersToTest = pipeline.triggers || [];
+    const hasImageProvider = Boolean(triggersToTest.length) && triggersToTest.every((trigger) => (validator.triggerTypes || []).includes(trigger.type));
+
+    const hasCustomImage = (stage.clusters || []).every((cluster: ICluster) => cluster.imageId && cluster.imageId !== '${trigger.properties.imageName}');
+
+    if (!hasImageProvider && !hasCustomImage) {
+      return `${validator.message} Add a trigger or manually input the imageId. Suggested triggers: ${(validator.triggerTypes || []).join(', ')}`;
```
vigneshm, line 23: I'm not familiar with the pipeline validator design but curious to know why the validation and the message is split across two places?

packages/core/src/pipeline/config/stages/deploy/deployStage.js
```diff
@@ -39,6 +39,16 @@ angular
...
+            triggerTypes: ['docker', 'jenkins'],
+            skipValidation: (pipeline, stage) => 
+              (stage.clusters || []).every(
+                (cluster) =>
+                  !CloudProviderRegistry.getValue(cluster.provider, 'serverGroup.checkForImageProviders')
+              ),
```
ranihorev, line 50: What's the purpose of this? (I'm not familiar with this codebase enough, sorry)
btw, you can use `some` instead of `every` and `!`

packages/core/src/pipeline/config/validation/imageProviderBeforeType.validator.ts
```diff
@@ -0,0 +1,30 @@
+import { IStageOrTriggerValidator, IValidatorConfig, PipelineConfigValidator } from './PipelineConfigValidator';
+import { ICluster, IPipeline, IStage, IStageOrTriggerTypeConfig } from '../../../domain';
+
+export interface IImageProviderBeforeTypeValidationConfig extends IValidatorConfig {
+  triggerTypes?: string[];
+  message: string;
+}
+
+export class ImageProviderBeforeTypeValidator implements IStageOrTriggerValidator {
```
ranihorev, line 9: Worth documenting the purpose of this validator (the text in the PR for example)

packages/core/src/pipeline/config/validation/imageProviderBeforeType.validator.ts
```diff
@@ -0,0 +1,30 @@
...
+    pipeline: IPipeline,
+    stage: IStage,
+    validator: IImageProviderBeforeTypeValidationConfig,
+    _config: IStageOrTriggerTypeConfig,
+  ): string {
+
+    const triggersToTest = pipeline.triggers || [];
+    const hasImageProvider = Boolean(triggersToTest.length) && triggersToTest.every((trigger) => (validator.triggerTypes || []).includes(trigger.type));
+
+    const hasCustomImage = (stage.clusters || []).every((cluster: ICluster) => cluster.imageId && cluster.imageId !== '${trigger.properties.imageName}');
```
ranihorev, line 20: 1. `cluster` is not a property of `stage`, but `Istage` also has `[k: string]: any;` in its definition. I'm just wondering why :)
2. You can do `stage.clusters?.every...` instead

packages/core/src/pipeline/config/stages/deploy/deployStage.js
caseyhebebrand: This is to control which cloud providers use this validation. For now this validation is Titus-specific and doesn't apply to anyone else. That is why I only added `checkForImageProviders` flag to the Titus module.

packages/core/src/pipeline/config/validation/imageProviderBeforeType.validator.ts
caseyhebebrand: They are split up so that the message can be configured based on the stage that is used. A good example is the `stageBeforeType` validator. Both the canary and deploy stage use this validator, but have a different warning message if the validation fails.

The messages are mostly similar. If there was ever a validator refactor, then I think the message should be an optional override.

ranihorev:
Worth migrating `deployStage.js` to TS at some point. Don't know if now is the time :)

packages/core/src/pipeline/config/validation/imageProviderBeforeType.validator.ts
caseyhebebrand: 1. This is because all the stage types have such different metadata. A deploy stage has `cluster`, which falls under `[k: string]: any`, but most other stages do not. The `IStage` types could definitely use an overhaul for some more explicit typing (i.e. `IDeployStage extends IStage`). 
2. 👍🏼

caseyhebebrand:
> Worth migrating `deployStage.js` to TS at some point. Don't know if now is the time :)

I agree. The entire pipeline configuration code needs a good migration. It will be a huge effort, but we are definitely approaching the point where it would be beneficial to do this work.

PR #9546, opened 2021-07-30
Don't add browser history entries when changing infrastructure filter values

christopherthielen:
- refactor(core/cluster): simplify FilterSearch.tsx
- feat(core/cluster): Remove unused react2angular <filter-search> component
- fix(core/filterModel): Do not add browser histroy events when changing infrastructure filter queries


packages/core/src/filterModel/FilterModelService.ts
```diff
@@ -33,7 +33,7 @@ export class FilterModelService {
     // Apply any mutations to the current sortFilter values as ui-router state params
     filterModel.applyParamsToUrl = () => {
       const toParams = FilterModelService.mapSortFilterToRouterParams(filterModel);
-      ReactInjector.$state.go('.', toParams);
+      ReactInjector.$state.go('.', toParams, { location: 'replace' });
```
christopherthielen, line 36: This is the juicy bit

vigneshm (approved):

PR #9555, opened 2021-08-04
fix(core): fix layout, click-to-scroll on diff view

anotherchrisberry:
I wanted to use the DiffView in a plugin and noticed some things broke, probably when it was converted to React from Angular.

Currently, it looks like this:
Screen Shot 2021-08-03 at 10 05 04 PM

Issues:
* the scrolling nav (the part with clickable colored blocks mapping to changes) is pushed almost entirely out of the modal, and is not clickable, and is also pushed up vertically to the top of the modal body instead of aligned with the diff itself.
* the pre tag wrapping the diff contents does not fit within the modal body.
* the modal itself is too big and the bottom of it is cut off.

With this PR, it looks like this:
Screen Shot 2021-08-03 at 10 05 16 PM

It's possible I'm the only person who ever used the click-to-scroll thing so nobody else noticed it was broken. Also, the CSS for this thing is way too finicky and could probably be fixed with a better modal solution than the Bootstrap ones, but that's outside the scope of what I'm trying to do here, which make it a little less janky.


christopherthielen (approved):
successfully lured you back to spinnaker code

PR #9558, opened 2021-08-05
feat(md): an option to enable the auto import delivery config from git 

ranihorev:
image


caseyhebebrand (approved):

packages/core/src/managed/config/GitIntegration.tsx
```diff
@@ -0,0 +1,41 @@
...
+export const GitIntegration = ({ isEnabled, branch, repository }: IGitIntegrationProps) => {
+  const appName = useApplicationContextSafe().name;
+  const [updateIntegration, { loading }] = useUpdateGitIntegrationMutation({
+    refetchQueries: [{ query: FetchApplicationManagementDataDocument, variables: { appName } }],
+  });
+
+  return (
+    <div className="sp-margin-xl-top">
+      <h4>Git integration</h4>
+      <CheckboxInput
```
luispollo, line 24: nit: Any chance we could make this into a little switch toggle like in the mockup?

ranihorev: We can but :)
1. We don't have switches at all in deck, so it seems a bit out of patterns imo
2. We don't have a component, so we'll need to create one or use an external package like this: https://react-switch.netlify.app/

I don't think it worth the effort at the moment

luispollo: Fair enough. The only reason I asked is that I always have trouble trusting that checking a box in the UI without some sort of "save" button actually did what it was meant to do. I believe this is why we this little thingie in other places in the UI:
image

You might want to add that to this page as well if the goal is to stick with the pattern.

ranihorev: I agree and I also like the switches. If we'll need more, I can add a new component :)
btw, when you select the option, it will show a loading indicator and the checkmark will appear only when it's done.

luispollo: Oh, nice. In that case, I think we're good! 👍

PR #9562, opened 2021-08-09
feat(core/pipeline): pipeline graph minimize based on number of stages

CalvinTse:
### Feature

Pipeline graphs with many stages take a very long time to load. To remediate this situation, pipeline graphs can now be minimized on the initial mount of the pipeline graph based on a configurable number of stages. When minimized, users can click **Expand** to see the full pipeline graph rendering.

The threshold for the number of stages before triggering a minimization is configurable and is defined [here](https://github.com/spinnaker/deck/commit/1ddd179b999d900e77d20bb012f18a3cfdd4dd4e#diff-44166998715ad434d66ec82ab5833ddf9ce1d48656738dbc3aaeed49b5991a46R51). The default value is 500 stages, which is a large enough number so that most users will not see this change. 

This minimization feature only affects the pipeline graph from the **Execution Details** section the pipeline execution. This feature will not minimize the graph in the **Configure** page. 

Link to this issue https://github.com/spinnaker/spinnaker/issues/6504

## Series of Pics to Show Behaviour

**1. Minimized View If Pipeline has More than 500 stages**
Minimized View If Pipeline has More than 500 stages

**2. After Clicking Expand -> Loading Screen**
Loading Screen

**3. Regular Graph View after Processing**
After Clicking the Expand and Loading

Any feedback would be appreciated! I would like to get this merged in to contribute to the community. Thanks 😄


CalvinTse:
Hello @christopherthielen! 😄  Is it possible to give this a review? Or do you know anyone that can? Thanks in advance!

packages/core/src/pipeline/config/graph/PipelineGraph.tsx
```diff
@@ -359,24 +368,40 @@ export class PipelineGraph extends React.Component<IPipelineGraphProps, IPipelin
     newState.graphHeight = Math.max(sum(newState.rowHeights) + this.graphVerticalPadding, this.minExecutionGraphHeight);
   }
 
-  private updateGraph(props: IPipelineGraphProps, statesOnly = false): void {
+  private updateGraph(props: IPipelineGraphProps, expand = false, statesOnly = false): void {
     let newState: IPipelineGraphState;
-    if (!statesOnly) {
+
```
mattgogerly, line 373: Indentation + extra new line - might be worth running lint and/or spotless:apply.

packages/core/src/pipeline/config/graph/PipelineGraph.tsx
```diff
@@ -451,54 +589,9 @@ export class PipelineGraph extends React.Component<IPipelineGraphProps, IPipelin
...
-                    nodeClicked={this.props.onNodeClick}
-                    highlight={this.highlight}
-                    nodeRadius={nodeRadius}
-                    node={node}
-                  />
-                </g>
-              ),
-          )}
-        </svg>
+        {this.getPipelineGraphRender()}
```
mattgogerly, line 594: Is there a reason for moving everything out of render into its own method? If this was one part of what was being returned it might make sense, but given this is the only thing returned by this component it's an extra layer of abstraction that I don't think is needed?

mattgogerly (changes requested):

mattgogerly:
This is a great improvement for UX. Thanks for putting it together. I've left a comment but I'm by no means an expert on Deck's code conventions so feel free to ignore me :)

packages/core/src/pipeline/config/graph/PipelineGraph.tsx
CalvinTse: Yeah, I agree with you! I don't think we need to make this abstraction (Putting it in the render method would make it more readable too!). Good catch! Ill make the change 👍

dbyron-sf:
@mergifyio update

PR #9590, opened 2021-08-17
chore(core): show "Deleting" when environment gets deleted

ranihorev:
Kapture 2021-08-17 at 12 29 02

### DIFF
--- a/packages/core/src/managed/environmentBaseElements/BaseEnvironment.less
+++ b/packages/core/src/managed/environmentBaseElements/BaseEnvironment.less
@@ -7,6 +7,7 @@
   background-color: var(--color-white);
 
   .EnvironmentTitle {
+    position: relative;
     width: 100%;
     min-height: 40px;
     display: flex;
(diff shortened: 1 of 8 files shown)


packages/core/src/managed/environmentBaseElements/BaseEnvironment.less
```diff
@@ -38,6 +39,18 @@
       font-weight: 700;
     }
 
+    .env-deleting {
+      position: absolute;
+      right: 8px;
+      font-size: 13px;
+      color: var(--color-white);
+      animation: deletingAnimation 2s infinite;
+      padding: var(~'--3xs-spacing') var(~'--xs-spacing');
```
luispollo, line 48: Just for my education, what's the `~` here?

packages/core/src/managed/overview/baseStyles.less
```diff
@@ -118,3 +118,15 @@ button.md-btn-danger:hover {
     background-color: var(--color-status-success);
   }
 }
+
+@keyframes deletingAnimation {
```
luispollo, line 122: Cool! Thanks for adding the animation like I suggested! ❤️

luispollo (approved):
Not sure why the build is failing, but looks good to me.

packages/core/src/managed/environmentBaseElements/BaseEnvironment.less
ranihorev: LESS doesn't like the format `--3xs-spacing`, the `~` tells it to treat it as a string. It'll work without it but the UI will complain

PR #9596, opened 2021-08-17
Move app to packages dir

christopherthielen:
let's ee what breaks in CI


christopherthielen:
@vigneshm is there anything else I might have missed?

vigneshm (approved):
Looks like karma and cypress tests have succeeded, we should be good to go!

PR #9604, opened 2021-08-20
fix(amazon): Pass in entire step policy object

caseyhebebrand:
Next up: 
- add to titus module
-  delete extra prop


packages/amazon/src/serverGroup/details/scalingPolicy/upsert/step/StepPolicyAction.tsx
```diff
@@ -64,6 +65,10 @@ export const StepPolicyAction = ({
     stepsChanged(newSteps);
   };
 
+  React.useEffect(() => {
+    setSteps(step.stepAdjustments);
+  }, [step?.stepAdjustments]);
```
caseyhebebrand, line 70: Detecting changes from other components.

PR #9609, opened 2021-08-24
refactor(amazon): Scaling policy additional settings

caseyhebebrand:
Creates a component with the additional settings section of the step scaling policy modal. This component will be used for both AWS and Titus.

### DIFF
--- a/packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyAdditionalSettings.tsx
+++ b/packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyAdditionalSettings.tsx
@@ -1,0 +1,123 @@
+import { module } from 'angular';
+import { set } from 'lodash';
+import * as React from 'react';
+import { react2angular } from 'react2angular';
+
+import { HelpField, NumberInput, withErrorBoundary } from '@spinnaker/core';
+
+import { IUpsertScalingPolicyCommand } from '../ScalingPolicyWriter';
+
+export interface IScalingPolicyAdditionalSettingsProps {
+  command: IUpsertScalingPolicyCommand;
+  isInstanceType: boolean;
+  isNew: boolean;
+  operator: string;
+  updateCommand: (command: IUpsertScalingPolicyCommand) => void;
+}
+
+export const ScalingPolicyAdditionalSettings = ({
+  command,
+  isInstanceType,
+  isNew,
+  operator,
+  updateCommand,
+}: IScalingPolicyAdditionalSettingsProps) => {
+  const [commandView, setCommandView] = React.useState<IUpsertScalingPolicyCommand>(command);
+  const setCommandField = (path: string, value: number) => {
+    const newCommand = { ...command };
+    set(newCommand, path, value);
+    setCommandView(newCommand);
+    updateCommand(newCommand);
+  };
+
+  return (
+    <div>
+      <h4 className="section-heading">Additional Settings</h4>
+      <div className="section-body section-additional-settings">
+        {!isNew && (
+          <div className="row">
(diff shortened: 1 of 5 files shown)


alanmquach (approved):
Looks good to me!

Might be slightly more readable if the repeated label/content divs were pulled out instead of repeated, but that's splitting hairs, really.

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyAdditionalSettings.tsx
```diff
@@ -0,0 +1,123 @@
...
+              <span className="form-control-static select-placeholder">{commandView.name}</span>
+            </div>
+          </div>
+        )}
+        {!isInstanceType && (
+          <div className="row">
+            <div className="col-md-2 sm-label-right">Adjustment Step</div>
+            <div className="col-md-10 content-fields">
+              <span className="form-control-static select-placeholder">
+                {`${operator} instannces in increments of at least `}
```
christopherthielen, line 50:
```suggestion
                {`${operator} instances in increments of at least `}
```

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/upsertScalingPolicy.controller.js
```diff
@@ -122,6 +124,10 @@ module(AMAZON_SERVERGROUP_DETAILS_SCALINGPOLICY_UPSERT_UPSERTSCALINGPOLICY_CONTR
       };
     }
 
+    this.commandChanged = (updatedCommand) => {
+      this.command = updatedCommand;
+    };
```
christopherthielen, line 129: I think this needs to be wrapped in applyAsync for it to take effect in ng

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyAdditionalSettings.tsx
```diff
@@ -0,0 +1,123 @@
...
+  operator,
+  updateCommand,
+}: IScalingPolicyAdditionalSettingsProps) => {
+  const [commandView, setCommandView] = React.useState<IUpsertScalingPolicyCommand>(command);
+  const setCommandField = (path: string, value: number) => {
+    const newCommand = { ...command };
+    set(newCommand, path, value);
+    setCommandView(newCommand);
+    updateCommand(newCommand);
+  };
```
christopherthielen, line 31: Assume all this is necessary because this component is being used by an AngularJS component?  Did you try embedding a Formik + FormFields then updating the parent whenever any field changes?

christopherthielen (approved):
as a straight ng->react migration, seems ok.  However, I would like this better if we went all the way to a formik migration if possible.

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyAdditionalSettings.tsx
caseyhebebrand: Yes, its part of the mid-migration dual-updates I had to do otherwise the react children were not re-rendering for 3-5 seconds after the command was updated and propagated through the angular controller. 

I haven't tried using Formik yet, but plan to at the high level of the modal. For your reference, my next PRs will be the parent modal for AWS and then Titus. I intend to make the modal the centralized place for the command and change handler. Once this happens, I can go remove some of this mid-migration tech debt. Here are my thoughts on what it would look like: 

1. Refactor parent modal to centalize command building and updating
2. Remove dual-updates/clones from children components. Remove references to a legacy observables that helped manage updates between angular and react. 
3. Incorporate `SpinFormik` and `FormikFormField`s to the modal sections.

PR #9618, opened 2021-08-26
refactor(amazon): Create TargetMetricFields component for aws and titus

caseyhebebrand:
Componentize the `Target Metric` section of the Target Tracking Policy modal. This will be usable for both AWS and titus policies.


packages/amazon/src/serverGroup/details/scalingPolicy/targetTracking/upsertTargetTracking.controller.ts
```diff
@@ -85,6 +89,10 @@ export class UpsertTargetTrackingController implements IComponentController {
     this.alarmUpdated.next();
   };
 
+  public commandChanged = (updatedCommand: ITargetTrackingPolicyCommand) => {
+    this.command = updatedCommand;
+  };
```
christopherthielen, line 94: I think this should be wrapped in `applyAsync`

christopherthielen (approved):
similar comment to the related pr: I don't like all the cloning and state management that is in this, but 👍 to this step in ng->react migration

PR #9620, opened 2021-08-26
feat(provider/azure): Allow empty load balancer for Azure scale sets

kmccarp:
Azure health settings can be placed directly on scale sets. If client-
side load balancing (e.g. through Eureka) is used, a load balancer
is not necessary.

This change is dependent on https://github.com/spinnaker/clouddriver/pull/5494.

**Testing done**:

1. Modified an existing pipeline to remove load balancer
2. Modified pipeline from 1) to re-add load balancer
3. Created a new pipeline without a load balancer
4. Created a new pipeline with a load balancer


jkschneider (approved):
LGTM

PR #9623, opened 2021-08-27
refactor(amazon): Create command builder for scaling policies

caseyhebebrand:


packages/amazon/src/serverGroup/details/scalingPolicy/ScalingPolicyWriter.ts
```diff
@@ -51,6 +51,7 @@ export interface IUpsertAlarmDescription extends IConfigurableMetric {
   threshold: number;
   statistic: AlarmStatisticType;
   unit: StandardUnitType;
+  actionsEnabled?: boolean;
```
caseyhebebrand, line 54:
```suggestion
  actionsEnabled?: boolean;
  disableEditingDimensions?: boolean;
```

alanmquach (approved):

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyCommandBuilderService.ts
```diff
@@ -0,0 +1,127 @@
...
+  IUpsertAlarmDescription,
+  IUpsertScalingPolicyCommand,
+} from '../ScalingPolicyWriter';
+import { IAmazonServerGroup, IScalingPolicy, IStepAdjustment, ITargetTrackingPolicy } from '../../../../domain';
+
+type PolicyType = 'Step' | 'TargetTracking';
+
+export const ScalingPolicyCommandBuilder = {
+  buildAlarm: (policy: IScalingPolicy, region: string, asgName: string): IUpsertAlarmDescription => {
+    const alarm = policy?.alarms[0];
```
christopherthielen, line 15: This will blow up if `alarm` is undefined; should we account for it?

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyCommandBuilderService.ts
```diff
@@ -0,0 +1,127 @@
...
+      const step = {
+        scalingAdjustment: Math.abs(adjustment.scalingAdjustment),
+      } as IStepAdjustment;
+
+      if (adjustment.metricIntervalUpperBound !== undefined) {
+        step.metricIntervalUpperBound = adjustment.metricIntervalUpperBound + threshold;
+      }
+      if (adjustment.metricIntervalLowerBound !== undefined) {
+        step.metricIntervalLowerBound = adjustment.metricIntervalLowerBound + threshold;
+      }
```
christopherthielen, line 48: I don't know how to interpret these numbers, but just wanted to check if this was supposed to be `metricIntervalLowerBound - threshold`

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyCommandBuilderService.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      return step;
+    });
+
+    return {
+      estimatedInstanceWarmup: policy.estimatedInstanceWarmup || cooldown || 600,
```
christopherthielen, line 54:
```suggestion
      estimatedInstanceWarmup: policy.estimatedInstanceWarmup ?? cooldown ?? 600,
```
Should generally prefer `??` over `||` especially when working with numbers so explicitly set `0` values don't get overwritten by a default value.

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyCommandBuilderService.ts
caseyhebebrand: The policy command builder (new) or api response (for existing) always sets it to an empty array, but it never hurts to do an additional safety check so I will add that.

packages/amazon/src/serverGroup/details/scalingPolicy/upsert/ScalingPolicyCommandBuilderService.ts
caseyhebebrand: No, according to the existing controller it is added: https://github.com/spinnaker/deck/blob/master/packages/amazon/src/serverGroup/details/scalingPolicy/upsert/upsertScalingPolicy.controller.js#L108

There is some subtraction that happens when the command is prepared for submit.

PR #9626, opened 2021-08-30
fix(aws): Clarify compatible target groups

alanmquach:
EC2 can only associate with `instance` type target groups and Titus can only associate with `ip` type target groups.

Just clarifying this in the message so that it's more well known that the UI helpfully hides the incompatible ones.

image


caseyhebebrand (approved):
LGTM

PR #9632, opened 2021-09-01
feat(provider/azure): Add image settings to server group

kmccarp:
Allow users to select a custom image for Azure server group deployments.

This functionality already exists in clouddriver and users can specify custom pipeline json to pass image information. The only required fields for Azure are `region`, `imageName`, `ostype`, and `uri`, so I have added those to the UI.

**Screenshots**:

Default (don't use custom image)
image

Use custom image
image

**Testing done**:

1. Select `Use custom image?`, fill out fields, and deploy
2. Deselect `Use custom image?` with no bake stage; see that it fails to deploy
3. Deselect `Use custom image?` with a bake stage and see that it deploys correctly


jkschneider (approved):
LGTM

PR #9633, opened 2021-09-01
feat(core): Allow notification email SpEL

alanmquach:
Letting validation slide when notification emails are SpEL.

This feels like the best way to do it as the form-wide validators don't really work well with dynamic input fields and I don't think we had `spelAware()` for field-local validators.


packages/core/src/presentation/forms/validation/validators.ts
```diff
@@ -85,6 +85,12 @@ const skipIfUndefined = (actualValidator: IValidator): IValidator => {
   };
 };
 
+const skipIfSpel = (actualValidator: IValidator): IValidator => {
+  return function skipIfSpel(val: any, label = THIS_FIELD) {
+    return typeof val === 'string' && val.indexOf('${') !== 0 && actualValidator(val, label);
```
mattgogerly, line 90: Sorry if I'm being dense, what's the reason for checking the value doesn't start with ${?

mattgogerly (approved):
LGTM

alanmquach:
Nice catch, I had forgotten that expressions could be anywhere in the value and doesn't have to be the entire value.

vigneshm (approved):

caseyhebebrand (approved):

PR #9634, opened 2021-09-02
feat(md): added "import now" to delivery config

ranihorev:
image

### DIFF
--- a/packages/core/src/managed/config/DeliveryConfig.tsx
+++ b/packages/core/src/managed/config/DeliveryConfig.tsx
@@ -1,6 +1,14 @@
 import React from 'react';
+
 import { RelativeTimestamp } from '../RelativeTimestamp';
+import {
+  FetchApplicationManagementDataDocument,
+  FetchApplicationManagementDataQueryVariables,
+  useImportDeliveryConfigMutation,
+} from '../graphql/graphql-sdk';
+import { useApplicationContextSafe } from '../../presentation/hooks/useApplicationContext.hook';
 import { YamlViewer } from '../utils/YamlViewer';
+import { NotifierService, Spinner } from '../../widgets';
 
 interface IDeliveryConfigProps {
   config?: string;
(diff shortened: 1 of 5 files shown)


packages/core/src/managed/config/DeliveryConfig.tsx
```diff
@@ -1,24 +1,65 @@
...
+      key: 'import-error',
+      content: `Failed to import delivery config - ${error.message}`,
+      options: { type: 'error' },
+    });
+  }, [error]);
+
+  return (
+    <>
+      <small>
+        (
```
luispollo, line 39: The parenthesis outside of `<button />` are adding spaces around the text, which looks a bit weird IMHO.

packages/core/src/managed/config/DeliveryConfig.tsx
```diff
@@ -1,24 +1,65 @@
...
+  const appName = useApplicationContextSafe().name;
+  const refetchVariables: FetchApplicationManagementDataQueryVariables = { appName };
+  const [importDeliveryConfig, { error, loading }] = useImportDeliveryConfigMutation({
+    variables: { application: appName },
+    refetchQueries: [{ query: FetchApplicationManagementDataDocument, variables: refetchVariables }],
+  });
+
+  React.useEffect(() => {
+    if (!error) return;
+    NotifierService.publish({
```
luispollo, line 29: I'm assuming this has an effect on the UI? Can you please post a sample of what that looks like?

packages/core/src/managed/graphql/schema.graphql
```diff
@@ -108,16 +113,16 @@ enum MdLifecycleEventScope {
...
   BUILD
 }
 
 enum MdLifecycleEventStatus {
-  NOT_STARTED,
-  RUNNING,
-  SUCCEEDED,
-  FAILED,
-  ABORTED,
+  NOT_STARTED
```
luispollo, line 121: nit: It's kind of annoying that every time you push changes to the schema it gets reformatted and looks different from what's in Keel...

luispollo (approved):
Minor style nit and a question, otherwise LGTM.

packages/core/src/managed/config/DeliveryConfig.tsx
ranihorev: This is the button right notifications you see when an error happens

packages/core/src/managed/graphql/schema.graphql
ranihorev: Fixed it

packages/core/src/managed/config/DeliveryConfig.tsx
ranihorev: image

PR #9640, opened 2021-09-03
feat(md): extended and simplified constraint title rendering

ranihorev:
Instead of provide a free format for the title, it's preferred to render the title in the following format:
"<Friendly name> - <Friendly status>"

image

We still support the old format though. 

cc @anotherchrisberry - I'd love to get your feedback. You can update the canary plugin to this format if you want :)


packages/core/src/managed/constraints/registry.tsx
```diff
@@ -4,9 +4,9 @@ import React from 'react';
 
 import { IconNames } from '@spinnaker/presentation';
 
-import { AllowedTimesDescription, AllowedTimesTitle } from './AllowedTimes';
-import { DependsOnTitle } from './DependsOn';
-import { ManualJudgementTitle } from './ManualJudgement';
+import { AllowedTimesDescription, getAllowedTimesStatus } from './AllowedTimes';
+import { getDependsOnState as getDependsOnStatus } from './DependsOn';
```
anotherchrisberry, line 8: just curious, why the alias?

anotherchrisberry (approved):
Thanks for the heads up — let me know when it's out in the wild and we'll update our stuff

packages/core/src/managed/constraints/registry.tsx
ranihorev: Bah, silliness?

PR #9641, opened 2021-09-04
fix(titus/subnet): Support default titus subnetType in SubnetSelectInput

christopherthielen:


christopherthielen:
Allows a titus default subnet type to be configured.  Previously, the SubnetSelectInput always chose the aws default subnet.

PR #9643, opened 2021-09-07
fix(amazon/serverGroup): SubnetSelectField: Use default subnets from settings if no defaultSubnetTypes prop is passed

christopherthielen:


christopherthielen:
fixing the case where no `defaultSubnetTypes` prop is supplied

PR #9647, opened 2021-09-07
feat(core): Expand Image Provider validation to find image stages

caseyhebebrand:
Expanding the `imageProviderBeforeType` validator to check for any find image stages.


caseyhebebrand:
@cfieber Tagging you for visibility.

PR #9648, opened 2021-09-08
refactor(aws): Migrate step policy modal to react

caseyhebebrand:
Migrate AWS `UpsertScalingPolicyModal` to react. This modal creates/edits `step` policies. 

Broken into 3 commits: 
1. Modal component w/ functionality
2. Integrates modal to the create and edit flows. Updates some css.
3. Removes angular files that are not needed.  Some of these components are also used in Titus, so we have to leave the adapters until the Titus upsert modal is refactored.


packages/amazon/src/serverGroup/details/scalingPolicy/upsert/step/StepPolicyAction.tsx
```diff
@@ -150,7 +150,7 @@ export const StepPolicyAction = ({
       ))}
       <div className="row sp-margin-s">
         <div className="col-md-10 col-md-offset-1">
-          <button className="btn btn-block btn-sm add-new" onClick={addStep}>
+          <button type="button" className="btn btn-block btn-sm add-new" onClick={addStep}>
```
caseyhebebrand, line 153: The `TaskMonitorModal` was treating this a submit button by default so I had to be explicit with the type.

PR #9651, opened 2021-09-10
fix(style): keep submit button status indicator inline with label

anotherchrisberry:
This only affects old-timey Angular submit buttons. The React ones look fine.

Before:
Screen Shot 2021-09-10 at 1 02 26 PM

After:
Screen Shot 2021-09-10 at 1 01 57 PM


caseyhebebrand (approved):
LGTM

PR #9657, opened 2021-09-14
feat(md): display manifest path and make it configurable

ranihorev:
This allows users to specify the location of the app's config file

image

image


luispollo:
I recommend displaying the directory hosting the file, otherwise I think this will be confusing to users. We do that in the pipeline stage:

image

ranihorev:
Yup, that's why I added the PR to keel. But I don't have that info in the
UI yet :)

luispollo:
Yes you do. It's a setting in Deck:
https://github.com/spinnaker/deck/blob/b96078e304aad10f147bfeb65b91299c9033a6d1/packages/core/src/pipeline/config/stages/managed/ImportDeliveryConfigStageConfig.tsx#L22

ranihorev:
Doh, I had no idea. I can probably revert my PR on keel then

luispollo:
Yeah, sorry I missed that in the Keel PR. I also didn't think the reason for that was to consume in Deck when reviewing.

PR #9659, opened 2021-09-15
fix(md): Added DELETING state to resource states. 

ranihorev:
The UI currently breaks when deleting a preview env. This should fix it.
It's not ideal, but it should be enough until we'll upgrade these components


christopherthielen (approved):

packages/core/src/managed/externals/managedResourceStatusConfig.tsx
```diff
@@ -31,6 +31,15 @@ const LearnMoreLink = ({ resourceSummary }: { resourceSummary: IManagedResourceS
 );
 
 export const viewConfigurationByStatus: { [status in ManagedResourceStatus]: IViewConfiguration } = {
+  DELETING: {
+    appearance: 'warning',
+    iconName: 'mdActuating',
+    popoverContents: () => (
+      <p>
+        <b>Spinnaker is deleting the resource</b>
```
luispollo, line 39: I wonder if we should include something to the effect that we only do this for preview environments...

ranihorev: I don't think that it's a big deal because most people don't look at it (and it's transient), but happy to change if you have something in mind
Anything in mind? Happy to change

luispollo: How about "Spinnaker is deleting this temporary managed resource"?

erikmunson: I would def change `the` to `this` (i.e. "this resource") but don't feel strongly about clarifying beyond that. If this is going to show up in the context of a preview env in the UI then re-stating it may be redundant

luispollo: This will show in the legacy views...

PR #9673, opened 2021-09-18
chore(deps): switch to Source Sans 3

mattgogerly:
source-sans-pro is [deprecated](https://www.npmjs.com/package/source-sans-pro), so switching to source-sans.


mattgogerly:
@caseyhebebrand @christopherthielen mind checking this? I'm unable to install the core package because the dependency is missing.

caseyhebebrand:
@mattgogerly I checked out the PR and saw the error you mentioned. After running `yarn fixup` it cleaned up `dist` and now there are no missing deps.

PR #9674, opened 2021-09-20
fix(md): render resource status as markdown

ranihorev:
This renders the `event` coming from the backend as markdown and prevents it from overflowing. It only happens for specific errors


packages/core/src/managed/overview/Resource.tsx
```diff
@@ -63,7 +63,7 @@ const Status = ({
         />
         <div>
           <div>{state.reason || statusUtils[state.status].defaultReason}</div>
-          {state.event && state.event !== state.reason && <div>{state.event}</div>}
+          {state.event && state.event !== state.reason && <Markdown className="event" message={state.ev} />}
```
caseyhebebrand, line 66: This line is causing the build failures: 

```js
src/managed/overview/Resource.tsx:66:102 - error TS2339: Property 'ev' does not exist on type '{ __typename?: "MdResourceActuationState"; } & Pick<MdResourceActuationState, "status" | "reason" | "event"> & { tasks?: ({ __typename?: "MdResourceTask"; } & Pick<...>)[]; }'.

66           {state.event && state.event !== state.reason && <Markdown className="event" message={state.ev} />}
```

ranihorev: Thanks. I probably deleted part of it accidentally

PR #9690, opened 2021-09-23
fix(md): allow users to enable resource management

ranihorev:
The UX is not ideal, but I just wanted to put it out there.
What's not ideal?
- The UI is not very elegant imo
- When you enable, sometimes MD won't update the staus immediately and the new status will appear only on the next refresh. Not sure why.

Feedback is welcomed.
image

image


luispollo:
Thanks a lot for picking this up so quickly. While you're at it, just a minor suggestion: could we please change the text to say "Resource management is paused"? I think that's more self-explanatory and less confusing than the current text ("why is it not managed if I'm on Managed Deliver?")

ranihorev:
Sure

luispollo (approved):

PR #9698, opened 2021-09-24
lil cloudprovider refactors

christopherthielen:
I happened to be in these files and noticed that they could be simplified

- refactor(core/cloudProvider): use lodash get/set path functions
- refactor(core/cloudProvider): replace the no-longer-necessary Providers class with a plain old Map
- fix(core/cloudProvider): remove unused ICloudProviderLogoState interface

### DIFF
--- a/packages/core/src/cloudProvider/CloudProviderRegistry.ts
+++ b/packages/core/src/cloudProvider/CloudProviderRegistry.ts
@@ -1,5 +1,5 @@
 /* tslint:disable: no-console */
-import { cloneDeep, get, isNil, set, uniq, without } from 'lodash';
+import { cloneDeep, get, isNil, set } from 'lodash';
 
 import { SETTINGS } from '../config/settings';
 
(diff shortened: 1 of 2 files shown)


packages/core/src/cloudProvider/CloudProviderRegistry.ts
```diff
@@ -13,37 +13,11 @@ export interface ICloudProviderConfig {
   [attribute: string]: any;
 }
 
-class Providers {
-  private providers: Array<{ cloudProvider: string; config: ICloudProviderConfig }> = [];
-
-  public set(cloudProvider: string, config: ICloudProviderConfig): void {
-    // The original implementation used a Map, so calling #set could overwrite a config.
-    // The tests depend on this behavior, but maybe something else does as well.
```
christopherthielen, line 21: There was some logic that was added for kubernetes provider v1 and v2 that made some of this Providers logic useful, but those k8s providers have since been collapsed into a single version.

caseyhebebrand (approved):
LGTM

PR #9702, opened 2021-09-27
fix(Kubernetes): Support instance links in k8s

Shadi:
This is related to the issue reported here https://github.com/spinnaker/spinnaker/issues/6532, currently custom links on Kubernetes instances don't show any custom links, this PR will add that.

This will work for absolute urls, for custom urls the base url will be undefined since clouddriver-kubernetes does not support dns or ip addresses for instances, this can serve as a starting point for that


Shadi:
@christopherthielen it seems that the checks only run on PRs from the Spinnaker organisation: https://github.com/spinnaker/deck/blob/master/.github/workflows/ci.yml#L17

is there anything I need to do before this PR is reviewed ?

christopherthielen (approved):

PR #9707, opened 2021-09-28
fix(kubernetes): hide ad-hoc infrastructure action buttons when spinnaker has multiple accounts

CalvinTse:
This PR will attempt to fix some issues that were observed when using the `kubernetesAdHocInfraWritesEnabled` flag (https://github.com/spinnaker/deck/pull/7822)

### Current Behaviour:
If the `kubernetesAdHocInfraWritesEnabled` flag is not set, or if the `window.spinnakerSettings.kubernetesAdHocInfraWritesEnabled` flag is set to false. We are still able to see the Create Infrastructure Buttons (ie. **Create Server Group**, **Create Load Balancer**, **Create Firewall**) if we have more than 1 kubernetes account in spinnaker:

To Reproduce the issue:
1. Application with only Kubernetes as its cloud provider & Spinnaker has more than one account
AppConfig
2. The Create Server Group Button Still Exists 
Screen Shot 2021-09-27 at 3 28 08 PM
3. When Clicked, the Modal that the UI provides tries to create an AWS Server Group. Even though, the application is only configured to have Kubernetes as a cloud provider. 
Screen Shot 2021-09-27 at 3 28 21 PM

In addition, there are still some existing Kubernetes ad-hoc infrastructure action buttons that are still showing up, this results in a UI inconsistency, because buttons are hidden for some infrastructure but not others:

Pod Actions are still showing up, even when `window.spinnakerSettings.kubernetesAdHocInfraWritesEnabled: false`
Screen Shot 2021-09-14 at 2 28 41 PM

Server Group Actions are still showing up, even when `window.spinnakerSettings.kubernetesAdHocInfraWritesEnabled: false`
Screen Shot 2021-09-14 at 2 28 14 PM

Under the Kubernetes Raw Resources Tab:
Screen Shot 2021-09-21 at 9 42 05 AM

The current behaviour seems like a bug and this PR addresses these issues. 

This PR:
1. Fixes the behaviour of the `isDisable` function to hide buttons for applications that have Kubernetes as its only cloud provider, when more than one account exists, and `window.spinnakerSettings.kubernetesAdHocInfraWritesEnabled: false`. 
2. Hides (if `window.spinnakerSettings.kubernetesAdHocInfraWritesEnabled: false`) any remaining buttons that the Kubernetes cloud provider uses to perform ad-hoc infrastructure writes/updates. 
These buttons include the Pod Actions, Server Group Actions (Multiple Server Groups), Multiple Instances Actions, and Kubernetes Raw Resources Actions (Config Map Actions, Job Actions, etc) 
3. If `kubernetesAdHocInfraWritesEnabled` is not set we will **show** the ad-hoc infrastructure action buttons for applications with Kubernetes as a provider.  The reasoning for this can is because we want to show all buttons if the config flag is not set. 

### Testing
This change was tested in our own environment to ensure that no regression has occurred. 
Unit tests were written for the `isDisabled` function to ensure that the create infrastructure buttons get displayed properly.


dbyron-sf:
@mergifyio update

dbyron-sf (approved):

PR #9708, opened 2021-09-28
refactor(amazon): Step Policy popover and details

caseyhebebrand:
- `StepPolicyContent` and `StepPolicySummary` components
- Utilize components in `@spinnaker/amazon`
- Register upsert modals with cloud provider so `StepPolicySummary` can be used across providers
- temporary `react2angular` converter until parent is converted to react

### DIFF
--- a/packages/amazon/src/serverGroup/details/scalingPolicy/StepPolicySummary.tsx
+++ b/packages/amazon/src/serverGroup/details/scalingPolicy/StepPolicySummary.tsx
@@ -1,0 +1,89 @@
+import * as React from 'react';
+
+import {
+  Application,
+  CloudProviderRegistry,
+  ConfirmationModalService,
+  HoverablePopover,
+  IServerGroup,
+  ReactModal,
+  robotToHuman,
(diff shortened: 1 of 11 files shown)


packages/amazon/src/serverGroup/details/scalingPolicy/StepPolicySummary.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const modalProps = { dialogClassName: 'wizard-modal modal-lg' };
+    ReactModal.show<typeof UpsertModalComponent>(UpsertModalComponent, upsertProps, modalProps);
+  };
+  const deletePolicy = () => {
+    const taskMonitor = {
+      application,
+      title: `Deleting scaling policy ${policy.policyName}`,
+    };
+
+    const submitMethod = () => ScalingPolicyWriter.deleteScalingPolicy(application, serverGroup, policy);
```
caseyhebebrand, line 48: Fix this so it is compatible with Titus

PR #9712, opened 2021-09-29
feat(aws/infrastructure): Hide aws ad hoc infrastructure action buttons

CalvinTse:
This PR will introduce a new flag `awsAdHocInfraWritesEnabled` which performs the same action as the `kubernetesAdHocInfraWritesEnabled` flag. This new `awsAdHocInfraWritesEnabled` flag will hide all buttons that attempt to create/delete/update AWS infrastructure from the Functions, Clusters, Load Balancers, and Firewalls tabs, if it is set to false. 

Some buttons that will be hidden if `window.spinnakerSettings.kubernetesAdHocInfraWritesEnabled: false` are shown below:
Screen Shot 2021-09-28 at 8 17 17 PM
Screen Shot 2021-09-28 at 8 17 34 PM

Core action buttons like Create Server Group, Create Firewall, Create Load Balancer will only be hidden if all cloud providers for that application has its infrastructure writes enabled turned off (Currently, only works for kubernetes (`kubernetesAdHocInfraWritesEnabled`) and AWS (`awsAdHocInfraWritesEnabled`) cloud providers). 

By default, AWS infrastructure Buttons are not hidden. Users will need to opt into this feature.  

This PR has been tested in our environments, and unit tests were written to show the behaviour of the core infrastructure buttons.


dbyron-sf:
@mergifyio update

dbyron-sf (approved):

caseyhebebrand:
@CalvinTse A couple retroactive comments: 

1.  I would recommend defaulting settings like this to `false/disabled` so that those who are aware of the feature can enable it on their own custom settings. Luckily, I saw this in staging otherwise it would have caused an outage for our aws users.  
2. Move these flags to provider specific settings `SETTINGS.providers.aws.awsAdHocInfraWritesEnabled`

CalvinTse:
Thanks for the feedback @caseyhebebrand! 😄  
1. We decided to set the initial settings to `true` so that users can see all of the infrastructure buttons by default, so that they would not have seen a behaviour change. I am curious to why this would cause an outage to aws users (for me to keep in mind for next time to make a change)? 
2. I completely agree I can move settings like these to provider specific settings. 
Thanks again for the comments!

caseyhebebrand:
@CalvinTse No problem!

Specifically with the scaling policy section (which is what I noticed), `SETTINGS.awsAdHocInfraWritesEnabled` was undefined since it wasn't in our settings we were seeing the buttons be hidden. 

```js
      {SETTINGS.awsAdHocInfraWritesEnabled ? (
        <CreateScalingPolicyButton serverGroup={serverGroup} application={app} />
      ) : (
        <p>Can not create scaling policy, because ad-hoc operations are disabled for AWS providers.</p>
      )}
```

CalvinTse:
@caseyhebebrand Thanks for the reply!
We tested the UI without these fields in our settings and we were still able to see the **Create new scaling policy** link. 
Screen Shot 2021-09-30 at 5 46 31 PM

Can I see what your UI looks like after the change?
Are we setting the defaults in the correct place? In the `settings.ts` [here](https://github.com/spinnaker/deck/blob/master/packages/core/src/config/settings.ts#L162-L163). 
I want to make sure that I am doing things correctly for next time. Thanks! 😄

caseyhebebrand:
@CalvinTse Hmm...now when I am running my composite build I see the expected behavior. I recently ran `yarn fixup` so maybe this morning my local env was in a weird state.

CalvinTse:
@caseyhebebrand Thanks for doing the check! :) I still agree with you that these flags should be moved to provider specific settings `SETTINGS.providers.aws.awsAdHocInfraWritesEnabled`. I will make a PR to make that change (with be a refactor)

PR #9713, opened 2021-09-29
feat(md): added a default component if preview environments configured but don't exist

ranihorev:
image

The link to the docs is optional and will appear only if configured

### DIFF
--- a/packages/core/src/managed/overview/PreviewEnvironments.tsx
+++ b/packages/core/src/managed/overview/PreviewEnvironments.tsx
@@ -15,14 +15,16 @@ export const PreviewEnvironments = ({ orderedEnvironments, isConfigured }: IPrev
   const { environments: previewEnvironments, ...previewEnvironmentsProps } = orderedEnvironments;
   const logClick = useLogEvent('PreviewEnvironments');
   const docsLink = getDocsUrl('previewEnvironments');
+  const hasActivePreviewEnvironments = previewEnvironments.length > 0;
+
   return (
     <>
-      {(Boolean(previewEnvironments.length) || isConfigured) && (
+      {(isConfigured || hasActivePreviewEnvironments) && (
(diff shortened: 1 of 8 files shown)


packages/core/src/managed/overview/PreviewEnvironments.tsx
```diff
@@ -0,0 +1,45 @@
...
+  isConfigured?: boolean;
+}
+
+export const PreviewEnvironments = ({ orderedEnvironments, isConfigured }: IPreviewEnvironmentsProps) => {
+  const { environments: previewEnvironments, ...previewEnvironmentsProps } = orderedEnvironments;
+  const logClick = useLogEvent('PreviewEnvironments');
+  const docsLink = getDocsUrl('previewEnvironments');
+  return (
+    <>
+      {(Boolean(previewEnvironments.length) || isConfigured) && (
```
luispollo, line 20: Just out of curiosity, what's the advantage of writing it this way versus `(isConfigured && previewEnvironments.length > 0)` (which is more readable IMHO)?

packages/core/src/managed/overview/PreviewEnvironments.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <h4 className="sp-margin-2xl-top sp-margin-m-bottom self-left">
+          <b>Preview Environments</b>
+        </h4>
+      )}
+      {!Boolean(previewEnvironments.length) && isConfigured && (
```
luispollo, line 25: Same comment/question here. I find this much harder to read than a simple greater-than comparison.

luispollo (approved):
Minor question inline, otherwise LGTM!

packages/core/src/managed/overview/PreviewEnvironments.tsx
ranihorev: No advantage, just a personal perference. I don't mind changing though :)

PR #9717, opened 2021-09-29
Consistent type imports

christopherthielen:
This rule encourages use of `import type { IFoo } from './foo';` to distinguish type-only imports from imports used as code, such as `import { REST }   from '@spinnaker/core'; REST('/foo').get()`


christopherthielen:
I'm generally 👍 for this rule it but am open to discussion.

caseyhebebrand (approved):
I am also a huge fan of this rule

PR #9718, opened 2021-09-30
feat(core): Prepopulate name for app creation

alanmquach:
This change enables support for deep linking into deck for app creation with a prepopulated name: `#/applications?create=supercalifragilisticexpialidocious`.

image

- If the app already exists, we'll just redirect to the app
- If the app sorta kinda maybe exists (inferred), we redirect to the config page for the inferred app to be fully configured
- Otherwise if the app doesn't exist at all, we open the Create Application modal with the name prepopulated

The existing inferred app flow is not great (it simply drops you into a hydrated-ish config page without opening the edit modal for you), but I will improve that in a separate PR.


packages/core/src/application/modal/createApplication.modal.controller.js
```diff
@@ -71,11 +76,12 @@ module(CORE_APPLICATION_MODAL_CREATEAPPLICATION_MODAL_CONTROLLER, [
...
-        $state.go('home.applications.application', {
-          application: this.application.name,
-        });
-      }, 1000);
+      $uibModalInstance.close(this.application);
+      // navigateTimeout = $timeout(() => {
+      //   $state.go('home.applications.application', {
+      //     application: this.application.name,
+      //   });
+      // }, 1000);
```
caseyhebebrand, line 84: Can this be deleted?

alanmquach: Ah yes, thanks for catching. I was testing moving it out to the modal result handlers but forgot to remove here.

mattgogerly:
@alanmquach this still good to go?

alanmquach:
@mattgogerly Yup

mattgogerly (approved):

PR #9719, opened 2021-09-30
feat(scripts): make yarn clean delegate to all lerna packages

christopherthielen:


caseyhebebrand (approved):

scripts/monorepo_fixup.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+# This script should be run when updating a developer's working copy to the monorepo codebase
+
+rm -rf node_modules build packages/*/node_modules packages/*/dist packages/*/lib
+yarn
+yarn modules
```
christopherthielen, line 6: @vigneshm do we still need `yarn fixup` or `yarn cleanModules`?

vigneshm: This sounds good to me. They aren't used as often and I'm fine with moving it into a script.

vigneshm (approved):

PR #9748, opened 2021-10-06
Remove warning about pipeline triggers and quiet period

dreynaud:
Pipeline triggers now go through echo like every other triggers, they are no longer special


caseyhebebrand (approved):
LGTM

PR #9754, opened 2021-10-08
fix(md): short preview env intro text

ranihorev:


ranihorev:
Almost (see my response on Slack) :)
The first one is shorter and provides the same value

luispollo:
Ok.

PR #9755, opened 2021-10-09
feat(md): allow users to reset a constraint if it failed

ranihorev:
image

Users can click on reset if a constraint failed. Disclaimers:
1. It only appears if the constraint state is FAILED (I don't see another reason to reset)
2. It will be available for any constraint. Should we allow constraints to disable this option? For example, it doesn't make sense for allowed times constraints.
UPDATE: It can be controlled now, I removed it for allowed times and manual judgement
3. It takes about 30 seconds to reset its state. We can improve it on the backend.


anotherchrisberry:
This looks great!

PR #9763, opened 2021-10-27
chore(deps):  bump select2 to npm versions

mattgogerly:
Bumping select2 dependencies to versions that are in NPM. I couldn't get an answer on why these are currently fixed at Git tags, so please let me know if you know why.


mattgogerly:
cc @christopherthielen @dogonthehorizon mind reviewing this? These dependencies are causing us issues in procurement and I can't find anybody who knows why they're pinned.

christopherthielen:
I also couldn't find a reason these were pinned.  I don't think we're even including select2 in deck except for CSS

mattgogerly:
@Mergifyio backport release-1.27.x

PR #9771, opened 2021-11-15
fix(core): conditionally hide expression evaluation warning messages

mattgogerly:
Filtering out expression evaluation failure messages if there are actual stage failure message(s) or if skipExpressionEvaluation is set to true in the context (since Orca still evaluates these, it just doesn't fail the stage).


packages/core/src/pipeline/details/StageFailureMessage.tsx
```diff
@@ -71,18 +71,30 @@ export class StageFailureMessage extends React.Component<IStageFailureMessagePro
...
-        );
+
+      let filteredMessages: string[] = [];
+      if (isFailed || stage.context?.skipExpressionEvaluation) {
+        // filter out expression evaluation failure messages if:
+        // - there was an actual failure as these can get *really* long and hide the failure message
+        // - expression evaluation was explicitly disabled (as Orca still processes these and populates errors
+        //   even when disabled)
+        filteredMessages = stageMessages.filter((m) => !m.startsWith('Failed to evaluate'));
+      }
```
christopherthielen, line 88: Can you explain this further?  Why would we filter "failed to evaluate" messages when "isFailed" is true and "skipExpressionEvaluation" is false?

mattgogerly: The failed to evaluate messages can get extremely long and are displayed even if expression evaluation is explicitly skipped in the stage config. In contrast the failure message is usually quite small. So if there's an actual failure message, OR if skipExpressionEvaluation is true, filtering them out removes noise and lets you see the actual error.

An example of this would be where an entire configmap is being printed because Orca failed to evaluate expressions in it, *even though* expression evaluation was explicitly disabled in the stage.

christopherthielen (approved):

mattgogerly:
@Mergifyio update

link108:
@mattgogerly I'm seeing a warning with "No reason provided" in successful stages and I think this PR is what introduced the change, any chance you could take a look?

mattgogerly:
@link108 any chance you could paste the execution?

martin-armory:
@mattgogerly This can be reproduced on any pipeline, see screenshot for a simple 5-second Wait stage.
image
I'll point out the suspecting line shortly.

dogonthehorizon:
I believe this should resolve the behavior we've been seeing:
https://github.com/spinnaker/deck/pull/9794

PR #9776, opened 2021-11-19
fix(bake/oci): Bake stage not loading configured OCI account.

hegdevenky:
Issue - https://github.com/spinnaker/spinnaker/issues/6545
Bake stage not loading configured OCI account.

`$scope.account` was not updating if base OS was not configured as `$scope.stage.baseOs = $scope.baseOsOptions.baseImages[0].id;` resulting in error, included this in the previous if check.


hegdevenky:
@christopherthielen could you please help us with this PR to review and get merged?

christopherthielen (approved):

PR #9780, opened 2021-12-01
update README for necessary `yarn modules` step before `yarn start`

cdll:
It's necessary to run `yarn modules` before running `yarn start` for some reasons without pre-run `yarn modules` and then any other `@spinnaker/*` packages will be losed after started, and it's also important to be noted in the starter README.
image


mattgogerly:
Thanks for this! Can you change the commit to be chore(docs)?

cdll:
@mattgogerly commit msg updated. thank you for the suggestion 🤝

mattgogerly (approved):

PR #9781, opened 2021-12-01
chore(build): sync and include pluginsdk-peerdeps in the package graph

mattgogerly:
The publish GHA currently fails for pluginsdk-peerdeps. Lerna is building a graph of packages but there is currently no entry for pluginsdk-peerdeps, which means it returns undefined and subsequently fails.

https://github.com/spinnaker/deck/runs/4374820011


abannachZen (approved):
LGTM

PR #9786, opened 2021-12-07
chore(build): single Mergify queue

mattgogerly:
Turns out multiple queues is a premium feature.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,11 +1,6 @@
 queue_rules:
   - name: default
     conditions:
-      - base=master
-      - status-success=Deck CI
-  - name: backport
-    conditions:
-      - base~=^release-
       - status-success=Deck CI
 
 pull_request_rules:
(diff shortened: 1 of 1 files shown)


dbyron-sf (approved):

.mergify.yml
```diff
@@ -1,18 +1,15 @@
...
-      - base~=^release-
       - status-success=Deck CI
 
 pull_request_rules:
   - name: Automatically merge on CI success and review
     conditions:
       - 'label=ready to merge'
       - 'approved-reviews-by=@oss-approvers'
+      - base=master
+      - status-success=Deck CI
```
christopherthielen, line 12: since this is already in the `queue_rules`, does it need to be repeated in  each of the `pull_request_rules`?

mattgogerly: I'm not sure. The example in the blog has it repeated. I asked in the Mergify repo and haven't had a response yet. I guess this would prevent it from bouncing back and forth between the queue and out if the CI is failing?

mattgogerly: Got a response, apparently it's necessary (*waves hands mysteriously*)

PR #9788, opened 2021-12-13
feat(core/pipeline): Load quiet period configuration from Gate, not from SETTINGS

christopherthielen:
Previously,  quiet periods start/end ranges were configured in Echo and in Deck settings.js.  Now, deck fetches the current quiet period date range from Echo via Gate.  The Deck feature flag is still used to toggle "respectQuietPeriod" on and off in a pipeline config.


caseyhebebrand (approved):
LGTM

PR #9794, opened 2022-01-12
fix(core): ensure that users are not warned of their success

dogonthehorizon:
The ternary as written would allow a singleton list containing a single
`null` to propagate throughout the render method of this warning
message. This would result in a menacing "warning: no reason provided"
message appearing on all stage executions.

The assignment now states that a singleton list will only be provided
iff `message` is non null and the `messages` array is empty.


cristhian-castaneda (approved):
LGTM

link108 (approved):
Thank you!

link108 (approved):

mattgogerly:
Thanks and sorry for not catching this case!

PR #9802, opened 2022-01-20
feat(core): HTML5 Routing

jcavanagh:
This adds configuration to enable HTML5 (aka path-based, not hash-based) routing for Deck.

The default remains unchanged, and is still hash-based.

To enable this feature:
- Add `window.spinnakerSettings.feature.html5Routing = true;` to your `settings-local.js` file.
- Or, set `HTML5_ROUTING=true` in your Deck container, in whichever way you prefer.

### DIFF
--- a/docker/spinnaker.conf.gen
+++ b/docker/spinnaker.conf.gen
@@ -1,5 +1,6 @@
 <VirtualHost {%DECK_HOST%}:{%DECK_PORT%}>
   DocumentRoot /opt/deck/html
+  FallbackResource /index.html
 
   ProxyPass "/gate" "{%API_HOST%}" retry=0
   ProxyPassReverse "/gate" "{%API_HOST%}"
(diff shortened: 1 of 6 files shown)


mattgogerly (approved):
LGTM

mattgogerly:
@Mergify-io update

mattgogerly:
@Mergifyio update

mat-613:
Hi @jcavanagh, after upgrading my spinnaker to the 1.28.1 it stops working.

I have a spinnaker exposed behind a reverse proxy, so the url of ui is something like https://zz.com/spinnaker/ and the url of api is https://zz.com/spinnaker-api/. 
With the ` <base href="/">` that you have added in packages/app/index.deck all the fetch for the static resources are going to https://zz.com/ without the second part.

Why you have added this base href index.deck? Maybe it's more appropriate to insert something like that:
`<base href="./">`

Let me know, i can't upgrade my ten instances of spinnaker without a resolution!

Thank you

PR #9812, opened 2022-03-01
fix(cf): return credentials in an array

mattgogerly:
Account extractor is [supposed to return an array](https://github.com/spinnaker/deck/blob/554367925d9111c88e04610adb1097236ce48646/packages/core/src/domain/IStageTypeConfig.ts#L10) but this stage returns a string, which breaks the execution history parsing in Deck.

### DIFF
--- a/packages/cloudfoundry/src/pipeline/stages/runJob/cloudFoundryRunJob.module.ts
+++ b/packages/cloudfoundry/src/pipeline/stages/runJob/cloudFoundryRunJob.module.ts
@@ -5,7 +5,7 @@ import { CloudFoundryRunJobStageConfig } from './CloudFoundryRunJobStageConfig';
 import { RunJobExecutionDetails } from './RunJobExecutionDetails';
 
 Registry.pipeline.registerStage({
-  accountExtractor: (stage: IStage) => stage.context.credentials,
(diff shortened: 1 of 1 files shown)


packages/cloudfoundry/src/pipeline/stages/runJob/cloudFoundryRunJob.module.ts
```diff
@@ -5,7 +5,7 @@ import { CloudFoundryRunJobStageConfig } from './CloudFoundryRunJobStageConfig';
 import { RunJobExecutionDetails } from './RunJobExecutionDetails';
 
 Registry.pipeline.registerStage({
-  accountExtractor: (stage: IStage) => stage.context.credentials,
+  accountExtractor: (stage: IStage) => [stage.context.credentials],
```
dbyron-sf, line 8: Thanks for this.  Any chance for an automated test to demonstrate the bug/fix?

mattgogerly: Not sure I can to be honest - I'm surprised this even compiled in the first place.

dbyron-sf (approved):

dbyron-sf:
@mergifyio update

PR #9831, opened 2022-03-22
fix(ci): Build yarn modules before building eslint

kskewes-sf:
Noticed in `release-1.27.x` branch PR CI that eslint was unable to be
built.
Building eslint's dependencies enabled it to run and PASS/FAIL tests.

```
Run cd packages/eslint-plugin
yarn run v1.22.17
$ /home/runner/work/deck/deck/node_modules/.bin/tsc
../../app/scripts/modules/app/src/app.ts:6:29 - error TS2307: Cannot find module '@spinnaker/core' or its corresponding type declarations.

6 import { CORE_MODULE } from '@spinnaker/core';
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,8 @@ jobs:
             ${{ runner.os }}-yarn-
       - name: Install Dependencies
         run: yarn --frozen-lockfile
+      - name: Build Modules
+        run: yarn modules
       - name: Eslint Plugin Build
         run: |
           cd packages/eslint-plugin
@@ -44,8 +46,6 @@ jobs:
         run: yarn lint
       - name: Assert - code is formatted
         run: yarn prettier:check
-      - name: Build Modules
-        run: yarn modules
       - name: Unit Tests
         run: yarn test --single-run
       - name: Functional Tests
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -28,6 +28,8 @@ jobs:
             ${{ runner.os }}-yarn-
       - name: Install Dependencies
         run: yarn --frozen-lockfile
+      - name: Build Modules
+        run: yarn modules
       - name: Eslint Plugin Build
         run: |
           cd packages/eslint-plugin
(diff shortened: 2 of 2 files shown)


mattgogerly:
Hm, looks like we're running a couple of things twice with this - yarn, yarn modules and yarn test are all run by ./gradlew build and previously in the action too.

kskewes-sf:
> Hm, looks like we're running things twice with this - yarn, yarn modules, yarn test and yarn build are all run by ./gradlew build and previously in the action too.

Yes, we can see it in the `master` CI run [here](https://github.com/spinnaker/deck/runs/5651398866?check_suite_focus=true).
I've been running `gradlew` locally without the `yarn` commands and so haven't validated if `gradle` would de-dupe in the GHA, but it doesn't seem so.

We need `gradle` for debian package publishing.

I guess we could move eslint and anything else into `gradle` and just do `./gradlew build` in GHA?
Alternative?

kskewes-sf:
With respect to this PR, thankfully `master` is still passing eslint tests after building dependent modules first. :relieved:

mattgogerly:
> > Hm, looks like we're running things twice with this - yarn, yarn modules, yarn test and yarn build are all run by ./gradlew build and previously in the action too.
> 
> Yes, we can see it in the `master` CI run [here](https://github.com/spinnaker/deck/runs/5651398866?check_suite_focus=true). I've been running `gradlew` locally without the `yarn` commands and so haven't validated if `gradle` would de-dupe in the GHA, but it doesn't seem so.
> 
> We need `gradle` for debian package publishing.
> 
> I guess we could move eslint and anything else into `gradle` and just do `./gradlew build` in GHA? Alternative?

Ah replied in the wrong PR. Yeah I guess if we're going to need to run gradle on every build anyway it would make sense to put all the yarn commands as tasks in there and condense it down to one thing.

kskewes-sf:
All good, would you have time to help with that? We could PR it to `master` and then I can update #9829 ?

mattgogerly:
> All good, would you have time to help with that? We could PR it to `master` and then I can update #9829 ?

https://github.com/spinnaker/deck/pull/9832

kskewes-sf:
Closing in favour of #9832

PR #9832, opened 2022-03-23
chore(build): remove duplication in build steps

mattgogerly:
Removing duplication between GHA build steps and `./gradlew build`.


kskewes-sf (approved):
LGTM, had to install `xvfb` locally then it was fine. This dependency must already be in the GHA VM.

PR #9841, opened 2022-04-20
Updating email validation to allow for longer domains per issue-6636

ssnorrizaurous:
Updating email validation to allow for longer domains per [issue-6636](https://github.com/spinnaker/spinnaker/issues/6636)


mattgogerly (commented):
Does it make sense to use a proper email Regex such as one from https://emailregex.com? Rather than just fixing this specific issue.

ssnorrizaurous:
> Does it make sense to use a proper email Regex such as one from https://emailregex.com? Rather than just fixing this specific issue.

@mattgogerly makes sense to me. I updated the regex to match the site and initialize it as a const instead of on the fly.

mattgogerly:
@ssnorrizaurous mind running spotless:apply to fix the lint error?

mattgogerly:
@Mergifyio update

PR #9853, opened 2022-05-05
fix(core): apps always display as unconfigured

dmart:
Applications in the config view always show up as unconfigured, fix the
broken reference to the controller so applications properly display
their attributes.

https://github.com/spinnaker/deck/blob/master/packages/core/src/application/config/applicationAttributes.directive.js#L31


mattgogerly (approved):

mattgogerly:

PR #9859, opened 2022-06-03
feat(core): Synchronize the verticalNavExpandedAtom using an atom effect

ovidiupopa07:
Introduce an atom effect which is registering as a listener to CollapsibleSectionStateCache. If there is a listener for the key, the handler function will be called to update the atom value


link108:

ovidiupopa07:
@Mergifyio backport release-1.27.x

PR #9863, opened 2022-06-24
style(core): pipelines adapt to sub pipeline with manual judgment color

OscarMichelH:
## Feature Affected: 
Spinnaker Pipelines UI (Specifically the status/color of a Pipeline while running)

## Details
Feature Details (What should happen): When a child/sub pipeline is running and requires a manual judgment, the parent pipeline should provide some visual representation that the child pipeline has an MJ waiting. 

In the interests of maximum automation and preventing delays - if the parent pipeline knows where an MJ is waiting, it prevents the expenditure of time, effort, and overhead to figure it out. This feature would save a consistent but highly repetitious small amount of time leading to much better user experience when using Spinnaker.

## Changes
The color for status-bar is handled by markerClassName in ExecutionMarker.tsx in deck.
The change always be applied for the parent pipeline only when the child turns in orange for a manual judgment validation, if the child pipeline goes normal without user interaction required the parent pipeline turn in normal color as is working currently. 

Final behavior with changes applied:
 
https://user-images.githubusercontent.com/21283264/175568816-b5eefaa3-ffc0-4a17-819f-a6d53de9ca01.mp4


AlfredoVidalArmory:
Reviewed Oscar's changes and the look good, approve

dogonthehorizon (approved):

OscarMichelH:

dbyron-sf:
This looks like a feature, and so not appropriate to backport.

PR #9865, opened 2022-07-01
fix(aws): Fix Create Server Group button

sergio-quintero:
This line `defaults: { account?: string; region?: string; subnet?: string; mode?: string } = {}` was not creating an empty object when `buildNewServerGroupCommand()` was being called, causing `defaults` to be `null` and break the whole function. 

This adds the manual check and sets `defaults` to empty `{}`if it's null or undefined.


dogonthehorizon (approved):

mrampton (approved):

sergio-quintero:

PR #9871, opened 2022-07-29
perf(pipelinegraph): Improve the performance of the pipeline graph rendering 

ovidiupopa07:
Change how the downstream stages are computed: 

- In the current implementation, the recursive method `PipelineConfigService.getDownstreamStageIds` filtered the complete stages list on every call. The process was slow when dealing with many stages and links between them. 

- With my implementation, I group the list of stages into a `map`. The map's key is the child ref id, and the value is an array containing the child's parents. 
With this approach, retrieval/searching is much faster than the list filtering on each call of the recursive function. 

I also decided to memoize the result of the getDependencyCandidateStages (in stage.module.js). I did it because the method is called four times (with the same input) when clicking on a stage.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -54,7 +54,7 @@
     "lodash": "4.17.21",
     "lodash-decorators": "4.5.0",
     "luxon": "1.23.0",
-    "memoize-one": "4.0.2",
+    "memoize-one": "6.0.0",
     "ngimport": "0.6.1",
     "prop-types": "15.6.1",
     "react": "16.14.0",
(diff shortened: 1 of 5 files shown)


mattgogerly (approved):

perek:
bitmoji amaing

mknous:
@link108 asked in Slack: "are there any tests we can add to this PR? That would make me feel better about backporting it. I see it landed in 1.29, so it will have to be backported if we want to release to 2.27 or 2.28

link108:
I discussed with @ovidiupopa07 and there are already existing tests for this functionality (https://github.com/spinnaker/deck/blob/master/packages/core/src/pipeline/config/services/PipelineConfigService.spec.ts#L145)

link108:
@mergifyio backport release-1.27.x release-1.28.x

PR #9872, opened 2022-08-01
chore(gha): synchronize peer dependencies during bump packages step

abannachZen:
Per conversation in [slack thread 1](https://spinnakerteam.slack.com/archives/CDEGJS7B4/p1659031682070819) and [slack thread 2](https://spinnakerteam.slack.com/archives/CDEGJS7B4/p1659032552772509), the peer dependencies should be synchronized after package bumps have been performed.

I was able to consolidate this into the same stage because of the way that the workflow creates its tags. Unable to run the GHA itself, but was able to run all the commands in order and get the desired outcome locally.

# Summary of old GHA flow
## GHA 1: Bump Packages
1. Lerna bump packages, but do not push the commit to the repo.
2. Create a pull request on those commits.
3. Approve or close (if no changes) PR

## GHA 2: Publish
1. Tag all packages that have changes to the version in `package.json`
2. Publish to NPM

# Summary of new GHA flow
## GHA 1: Bump Packages
1. Lerna bump packages, but do not push the commit to the repo.
2. Navigate down to `packages/pluginsdk-peerdeps`
3. `yarn sync`
4. Commit changes
5. Lerna bump packages again without push, this time should only be `packages/pluginsdk-peerdeps` 
6. Create a pull request on those commits.
7. Approve or close (if no changes) PR

## GHA 2: Publish
1. Tag all packages that have changes to the version in `package.json`
2. Publish to NPM


.github/workflows/package-bump-pr.yml
```diff
@@ -51,12 +51,25 @@ jobs:
...
-            # Update yarn.lock and fix the lerna commit message (to use conventional commits itself)
-            git commit --amend -m "$(git log -1 --pretty=%B | sed -e 's/^Publish/chore(publish): publish packages/')" yarn.lock
+
+            # commit changes
+            git add yarn.lock packages/pluginsdk-peerdeps
+            git commit -m "feat(peerdep-sync): Synchronize peerdependencies"
+
+            # bump version of @spinnaker/pluginsdk-peerdeps
+            npx lerna version --yes --no-push --conventional-commits -m "chore(publish): publish packages"
           fi
```
mattgogerly, line 73: Is it worth moving this logic into a dedicated script? It seems the other steps have followed that convention, e.g. https://github.com/spinnaker/deck/blob/master/scripts/gha_tag_bumped_packages.sh

abannachZen: We could. Is there a standard that is/should be made up to say "If a gha goes beyond x lines change to a script"?

mattgogerly: Not that I'm aware of, just thinking we may as well be consistent. Given this now has (very simple) bash in it, I think it makes sense to extract it.

abannachZen: @mattgogerly Got the script written. Realized that the bumped packages and output changelog needed updating as well. Moved some shared logic to a common script.

scripts/gha_bump_packages.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)
+
+. "$SCRIPT_DIR/gha_common.sh"
+
+# bump versions of packages
+npx lerna version --yes --no-push --conventional-commits -m $PACKAGE_BUMP_COMMIT_MSG
```
abannachZen, line 8: question: is the working directory the directory from where the script was triggered or does this script need to `cd ..` to get up to the project dir?

mattgogerly: Good question, [this](https://github.com/spinnaker/deck/blob/55fa5d00beb576d9d9215923bcc4b7ead64a4fba/scripts/gha_tag_bumped_packages.sh#L4) would suggest it's from the scripts directory.

abannachZen: Doesn't hurt to navigate to the directory 👍 I'll add it in

mattgogerly (approved):

PR #9886, opened 2022-09-09
Fix(core/pipelines): Fix syntax error on ui-select causing stage configuration to not load

bliu2368:
Small change:
* Changed `ui-select` property from "onselect" to "on-select". This was changed about a month ago and seems to cause stage configurations to not render when a new stage is selected. Based on the ui-select documentation the actual property is meant to be "on-select" and this is likely what caused it to break.

Similarly, ui-select tags use "on-select" everywhere else in the repo.
Example: 
Screen Shot 2022-09-09 at 3 08 46 pm

### DIFF
--- a/packages/core/src/pipeline/config/stages/stage.html
+++ b/packages/core/src/pipeline/config/stages/stage.html
@@ -16,7 +16,7 @@
           style="width: 250px"
           class="form-control input-sm"
           autofocus
-          onselect="stageConfigCtrl.selectStageType($item)"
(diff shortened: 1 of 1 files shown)


ovidiupopa07 (approved):

jasonmcintosh (approved):
LGTM :) Wondering long term if there are any better ways to test this kinda stuff...

dogonthehorizon (approved):

ovidiupopa07:
@Mergifyio release-1.28.x

link108:

PR #9888, opened 2022-09-16
fix(search): Error thrown when search version 2 is enabled

ovidiupopa07:
For some results, when building the URL in `UrlBuilderUtils.buildUrl` an error is thrown when the URL is undefined.

If there is no URL value, we should try to build the href URL. 

Current behavior 
current

Behavior after fix

fix

### DIFF
--- a/packages/core/src/navigation/UrlBuilder.ts
+++ b/packages/core/src/navigation/UrlBuilder.ts
@@ -89,7 +89,7 @@ class UrlBuilderUtils {
       });
     });
 
-    if (parts.length > 0) {
+    if (parts.length > 0 && url) {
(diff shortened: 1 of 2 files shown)


kkotula (approved):
Well done 👏🏻

dogonthehorizon (approved):
If it's possible to get in a test here that'd be super :D otherwise LGTM

ovidiupopa07:
> If it's possible to get in a test here that'd be super :D otherwise LGTM
> 
> e: Will add merge label once we've discussed tests a bit more

Will write tests in a different PR as it needs some refactoring to be done before actually writing the tests.

ovidiupopa07:

PR #9890, opened 2022-10-05
fix(core): Do not set static document base URL

jcavanagh:
The previous change to enable optional path-based routing (#9802) added a `<base>` tag, which regressed serving hash-based routing at non-root public paths.

This change returns to AngularJS's default behavior when using hash-based routing.  Note that path-based routing will continue to only function when served from a domain root.

Resolves spinnaker/spinnaker#6723


mattgogerly (approved):

mattgogerly:
@Mergifyio backport release-1.28.x release-1.29.x

PR #9915, opened 2022-11-18
fix(Wercker): Removing wercker reverence

agnihotrigaurav:
Removing wercker references as the same has been deprecated.


mattgogerly:
Hi @agnihotrigaurav, sorry for the slow response - would you mind fixing the commit message?

PR #9930, opened 2022-12-14
feat(core/pipeline): Add missing flag `skipDownstreamOutput` in pipeline stage

vjda:
The https://github.com/spinnaker/orca/pull/3989 introduced a flag to opt-out from passing outputs to child executions. 

However, Deck does not provide any option to select this flag from the UI. 

This PR adds a new checkbox to select if skipping the output from child pipelines in the pipeline stage config settings.

image

The checkbox is disabled by default for backward compatibility and the field `skipDownstreamOutput` is not set unless the user selects the checkbox.

image


mattgogerly (approved):

mattgogerly:
@Mergifyio update

PR #9935, opened 2023-01-27
fix(deck): Apache user should own the settings-local.js file, not spi…

ramyennapusa:
When we enable AWS EC2 Launch templates using custom profiles([Enabling Launch Templates](https://spinnaker.io/docs/setup/other_config/server-group-launch-settings/aws-ec2/launch-templates-setup/)), DeckUI is not displaying the additional features mentioned in `settings-local.js` file because the Apache user `www-data` does not have the necessary permissions on that file.The file `settings.js` has the appropriate permissions.

Please refer to the screenshot below,

image

This change is related to https://github.com/spinnaker/spinnaker/issues/6771

### DIFF
--- a/docker/run-apache2.sh
+++ b/docker/run-apache2.sh
@@ -78,6 +78,8 @@ fi
 if [ -e /opt/spinnaker/config/settings-local.js ];
 then 
 	cp /opt/spinnaker/config/settings-local.js /opt/deck/html/settings-local.js
+	# Apache user must be the owner of settings-local.js file, not spinnaker user, so that it can display custom profiles features on Deck UI
+	chown www-data /opt/deck/html/settings-local.js
 fi
 
 if [ -e /opt/spinnaker/config/plugin-manifest.json ];


ashleykleynhans (approved):

dbyron-sf:
LGTM, though I suspect there's a similar fix elsewhere (in halyard?) for folks who don't use docker.

ramyennapusa:
I guess you are referring to this https://github.com/spinnaker/halyard/pull/1997

ashleykleynhans:
> I guess you are referring to this [spinnaker/halyard#1997](https://github.com/spinnaker/halyard/pull/1997) ? But unfortunately that only fixed the permissions of `settings.js` alone.

Correct!  Except that the "fix" in the PR did not resolve the issue and it is still broken.

mattgogerly (approved):

praseedasathaye:
Thanks @ashleykleynhans for reviewing this and Thanks @Josh-Bonello and @ramyennapusa for helping to fix this!

ashleykleynhans:
> > `settings.js` was never an issue, the permissions were always correct, the fix was supposed to be for `settings-local.js` but made no difference.
> 
> The fix relates to the "lost user data" behavior of [Issue 6771](https://github.com/spinnaker/spinnaker/issues/6771). Other issues are auxiliary.

This does not fix the "lost user data" issue at all.  It fixes 1 of 3 issues:

1. Lost UserData.
2. Adding IPv6 addresses when none of the subnets attached has any, causing instances to fail to launch.
3. Incorrect permissions on `settings-local.js`.

You have fixed the most insignificant issue, which can easily be fixed manually with this PR.

The 2 blocking issues (1) and (2) are still not resolved unfortunately.

ramyennapusa:
> > > `settings.js` was never an issue, the permissions were always correct, the fix was supposed to be for `settings-local.js` but made no difference.
> > 
> > 
> > The fix relates to the "lost user data" behavior of [Issue 6771](https://github.com/spinnaker/spinnaker/issues/6771). Other issues are auxiliary.
> 
> This does not fix the "lost user data" issue at all. It fixes 1 of 3 issues:
> 
>     1. Lost UserData.
> 
>     2. Adding IPv6 addresses when none of the subnets attached has any, causing instances to fail to launch.
> 
>     3. Incorrect permissions on `settings-local.js`.
> 
> 
> You have fixed the most insignificant issue (point 3) with this PR, which can easily be fixed manually, and also only fixed it for Docker and not for people who run Deck on a VM deployed using Debian packages with Halyard.
> 
> The 2 blocking issues (1) and (2) are still not resolved unfortunately.

I am working on fixing the 2 blocking issues as well.

PR #9948, opened 2023-02-20
fix(6755): Resolved issue regarding warning reporting when cloning a server group in AWS

siddhu-opsmx:
As part of resolving issue [6755](https://github.com/spinnaker/spinnaker/issues/6755). Adding changes to reset the instance object properties(i.e, unavailable) whenever server group dialog opens. By doing this will resolved the issues occurs due to incorrect value stored in instance object(i.e, unavailable) while following the same steps given in above issue again and again without refreshing the browser.


siddhu-opsmx:

PR #9960, opened 2023-03-31
fix: UI crashes when running pipeline(s) with many stages.

armory-abedonik:
This change prevents iterating over child nodes as it has already been checked and as a result greatly reduces the number of interactions and increases speed.


dbyron-sf:
BOB-30920 is from some internal ticketing system?  Best to leave it out here.

ovidiupopa07:
@mergify backport release-1.27.x

PR #9971, opened 2023-04-03
feat(provider/google): Added Cloud Run manifest functionality in Deck.

siddhu-opsmx:
This is part of https://github.com/spinnaker/governance/issues/302

Added Cloud Run manifest functionality in Deck.


packages/cloudrun/src/pipeline/stages/deployManifest/manifestStatus/DeployStatusPills.tsx
```diff
@@ -0,0 +1,47 @@
...
+            <span title={statusDescription.message} className={`pill ${statusClass}`}>
+              {statusKey}
+            </span>
+          )}
+          {isUnstableWithMessage && (
+            <span title={statusDescription.message} className="pill warn">
+              {statusKey}
+            </span>
+          )}
+          {(isStable || isUnstableWithMessage) && <span>&nbsp;</span>}
```
mattgogerly, line 42: I'm not sure this does anything useful?

packages/cloudrun/src/pipeline/stages/deployManifest/manifestStatus/ManifestEvents.tsx
mattgogerly, line 1: This seems to duplicate the Kubernetes equivalent. Can we generify that instead of having the same code twice?

packages/cloudrun/src/rolloutStrategy/redblack.strategy.ts
```diff
@@ -0,0 +1,7 @@
+import type { IDeploymentStrategy } from '@spinnaker/core';
+
+export const strategyRedBlack: IDeploymentStrategy = {
+  label: 'Red/Black (Deprecated)',
```
mattgogerly, line 4: Why is red/black deprecated?

packages/core/src/artifact/ArtifactTypes.ts
```diff
@@ -5,6 +5,7 @@ export interface IArtifactTypePatterns {
 export const ArtifactTypePatterns: IArtifactTypePatterns = {
   BITBUCKET_FILE: /bitbucket\/file/,
   CUSTOM_OBJECT: /custom\/object/,
+  CLOUDRUN: /cloudrun\/(.*)/,
```
mattgogerly, line 8: Forgive me if I'm not familiar with Cloud Run. Why does it have its own artifact type if the manifest is just YAML/JSON?

packages/cloudrun/src/pipeline/stages/deployManifest/manifestStatus/DeployStatusPills.tsx
siddhu-opsmx: Removed unwanted condition check

packages/cloudrun/src/rolloutStrategy/redblack.strategy.ts
siddhu-opsmx: Corrected label text.

packages/core/src/artifact/ArtifactTypes.ts
siddhu-opsmx: So based on back-end implementation CLOUDRUN type is not much useful for now like KUBERNETES, So removing it for now will add it if required in future based on enhancement request .

packages/cloudrun/src/pipeline/stages/deployManifest/manifestStatus/ManifestEvents.tsx
siddhu-opsmx: Yes , for now we are using this component in 2 modules (i.e, Kubernetes and Cloud Run) and we can generify that as well But in future if Kubernetes modifies the component based on some new requirement will affect Cloud Run module as well or vice-versa. So we are trying to keep the scope of components to module-centric, So that modification in any one module will not affect other modules.

mattgogerly (approved):

PR #9988, opened 2023-05-02
feat(lambda): Migrate Lambda plugin to OSS

mattgogerly:
Porting the Deck components of https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker to OSS. 

See https://github.com/spinnaker/orca/pull/4449 for corresponding Orca PR.


packages/amazon/src/aws.validators.ts
```diff
@@ -1,18 +1,17 @@
 import { isEmpty } from 'lodash';
 
 export const iamRoleValidator = (value: string, label: string) => {
-  const isIAMRole = value.match(/^arn:aws:iam::\d{12}:role\/?\/[a-zA-Z_0-9+=,.@\-_/]+/);
+  const isIAMRole = value.match(/^arn:aws:iam::\d{12}:role\/?\/[a-zA-Z_0-9+=,.@\-/]+/);
```
dbyron-sf, line 4: Doesn't [the plugin](https://github.com/spinnaker-plugins/aws-lambda-deployment-plugin-spinnaker/blob/ae9aca40cfce86d43b0d372939ae80e1a45a82ba/lambda-deployment-deck/src/utils/aws.validators.ts#L5) have the old value (with the underscore) here?

dbyron-sf:
@mattgogerly would you mind calling out whether the lambda functionality in the UI is disabled by default with this PR, and the config flag that controls it?

packages/amazon/src/aws.validators.ts
mattgogerly: It does, and I'm no Regex expert, but the underscore is already included after A-Z.

dbyron-sf: Look-y there, so it is.

mattgogerly:
> @mattgogerly would you mind calling out whether the lambda functionality in the UI is disabled by default with this PR, and the config flag that controls it?

Thought I had added one but maybe not, hard to tell on mobile. I'll check tomorrow.

mattgogerly:
> > @mattgogerly would you mind calling out whether the lambda functionality in the UI is disabled by default with this PR, and the config flag that controls it?
> 
> Thought I had added one but maybe not, hard to tell on mobile. I'll check tomorrow.

settings-local.js

`feature.lambdaAdditionalStages = true;`

PR #10001, opened 2023-06-02
feat(icons): allow plugins to provide custom icon components

mattgogerly:
Enables plugins to use the Icon component with a custom icon.

Currently the Icon component is limited to only icons defined in iconsByName.

### DIFF
--- a/packages/presentation/src/icons/Icon.tsx
+++ b/packages/presentation/src/icons/Icon.tsx
@@ -1,3 +1,4 @@
+import type { SVGComponent } from '*.svg';
 import React, { memo } from 'react';
 
 import { iconsByName } from './iconsByName';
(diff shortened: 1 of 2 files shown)


packages/presentation/src/icons/Icon.tsx
```diff
@@ -23,16 +25,12 @@ const pxDimensionsBySize: { [size: string]: string } = {
...
 };
 
-export const Icon = memo(({ name, appearance, size, color, className }: IIconProps) => {
-  const Component = iconsByName[name];
-
-  if (!Component) {
-    throwInvalidIconError(name);
-  }
+export const Icon = memo(({ name, reactComponent, appearance, size, color, className }: IIconProps) => {
+  const Component = name ? iconsByName[name] : reactComponent || throwInvalidIconError(name);
```
dbyron-sf, line 33: Maybe a different error if reactComponent is null, and then leave throwInvalidIconError above the same, where it requires a name?

PR #10003, opened 2023-06-13
fix(core/application): Delete Application component loses state while refreshing application.

siddhu-opsmx:
Whenever application has servergroups, then Application cannot be Deleted and Following message shown from deck "You cannot delete this application because it has server groups."
Issue is when we land on to config page directly deck able to show delete option without checking for server groups (happens in deck Ui too when we load config page directly as shown in below recordings).

## Delete Application section before fix:-

https://github.com/spinnaker/deck/assets/61963157/da8c7835-cf8b-4773-9c81-3a042a49eeb6

## Delete Application section after fix:-

https://github.com/spinnaker/deck/assets/61963157/5daa727c-f7b6-4dc1-a7c7-011af9cad618

### DIFF
--- a/packages/core/src/application/config/DeleteApplicationSection.tsx
+++ b/packages/core/src/application/config/DeleteApplicationSection.tsx
@@ -8,10 +8,11 @@ import { ApplicationWriter } from '../service/ApplicationWriter';
 
 export interface IDeleteApplicationSection {
   application: Application;
+  serverGroupsExist: number;
 }
 
 export function DeleteApplicationSection(props: IDeleteApplicationSection) {
-  const { application } = props;
+  const { application, serverGroupsExist } = props;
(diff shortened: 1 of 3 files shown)


siddhu-opsmx:

mattgogerly:
I'm not sure this is the right fix. This change is passing `config.application.serverGroups.data.length` as a separate prop, but `DeleteApplicationSection` is already taking `config.application` as a prop, so this shouldn't be necessary.

`serverGroups.data` is updated as part of the refresh cycle (the orange spinner next to the app name).

Can you check in `AngularJSAdapter` if the props are being updated when the application is refreshed?

siddhu-opsmx:
@mattgogerly ,Sure, I'll check this and let you know once found something.

PR #10009, opened 2023-06-27
fix(core/pipeline): Resolved issue getting during pipeline save with spaces in pipeline name.

siddhu-opsmx:
#### Getting warning message as "shows error as "Could not save pipeline.Reason: Sorry, there was an error retrieving your new pipeline.Please refresh the browser" when we create pipeline by giving spaces in pipeline name, but the pipeline is getting created

https://github.com/spinnaker/deck/assets/61963157/d868f74d-c640-4938-bcc2-71de648f93d5

https://github.com/spinnaker/deck/assets/61963157/dba0943a-c948-493c-bef3-ba23a0b50e1c


siddhu-opsmx:

packages/core/src/pipeline/create/CreatePipelineModal.tsx
```diff
@@ -185,7 +185,7 @@ export class CreatePipelineModal extends React.Component<ICreatePipelineModalPro
       const configs: IPipeline[] = config.strategy
         ? application.strategyConfigs.data
         : application.pipelineConfigs.data;
-      const newPipeline = configs.find((_config) => _config.name === config.name);
+      const newPipeline = configs.find((_config) => _config.name === config.name.trim());
```
mattgogerly, line 188: I think you should do this when the pipeline is initially submitted to be created [L132](https://github.com/spinnaker/deck/blob/c909e83ac33d3975f09e0e231f75d68246cb4bef/packages/core/src/pipeline/create/CreatePipelineModal.tsx#L132).

mattgogerly (commented):
You don't need to tag me for every PR - I check regularly anyway.

packages/core/src/pipeline/create/CreatePipelineModal.tsx
siddhu-opsmx: Updated the code.

mattgogerly (approved):

PR #10011, opened 2023-07-06
feat(artifacts): Add support for artifact store views and calls

xibz:
/artifacts/content-address/{reference} is a new API that allows for requests to be made to get full artifacts.

Also adds support for new artifact type of remote/base64

### DIFF
--- a/packages/core/src/domain/IArtifact.ts
+++ b/packages/core/src/domain/IArtifact.ts
@@ -12,3 +12,6 @@ export interface IArtifact {
   kind?: string; // TODO delete
   customKind?: boolean; // TODO delete
 }
+
+export const ARTIFACT_TYPE_EMBEDDED = 'embedded/base64';
+export const ARTIFACT_TYPE_REMOTE = 'remote/base64';
--- a/packages/core/src/manifest/ManifestYaml.tsx
+++ b/packages/core/src/manifest/ManifestYaml.tsx
@@ -1,26 +1,48 @@
 import React from 'react';
 import { Modal } from 'react-bootstrap';
+import { ArtifactService } from '../pipeline/config/triggers/artifacts/ArtifactService';
+import { decodeUnicodeBase64 } from '../utils';
 
-export interface IManifestYamlProps {
+type IManifestYamlProps = {
   linkName: string;
-  manifestText: string;
   modalTitle: string;
-}
+} & ({ manifestText: string; manifestUri?: never } | { manifestText?: never; manifestUri: string });
 
-export function ManifestYaml(props: IManifestYamlProps) {
+export function ManifestYaml({ linkName, modalTitle, manifestText, manifestUri }: IManifestYamlProps) {
   const [modalVisible, setModalVisible] = React.useState<boolean>(false);
   const toggle = () => setModalVisible(!modalVisible);
+  const [fetchedManifestText, setFetchedManifestText] = React.useState<string>('Loading...');
+  const [error, setError] = React.useState<string | null>(null);
+
+  React.useEffect(() => {
+    if (manifestUri) {
+      ArtifactService.getArtifactByContentReference(manifestUri)
+        .then((manifest) => setFetchedManifestText(decodeUnicodeBase64(manifest.reference)))
+        .catch((e) => setError(`Error: ${typeof e !== 'string' ? e.data?.message ?? JSON.stringify(e) : e}`));
+    }
+  }, []);
+
   return (
     <>
       <a key="modal-link" onClick={toggle} className="clickable">
-        {props.linkName}
+        {linkName}
       </a>
       <Modal key="modal" show={modalVisible} onHide={toggle}>
         <Modal.Header closeButton={true}>
-          <h3>{props.modalTitle}</h3>
+          <h3>{modalTitle}</h3>
         </Modal.Header>
         <Modal.Body>
-          <textarea readOnly={true} rows={15} className="code" value={props.manifestText} />
+          {error ? (
+            <div className="alert alert-warning">
+              <p>{error}</p>
+            </div>
+          ) : null}
+          <textarea
+            readOnly={true}
+            rows={15}
+            className="code"
+            value={manifestUri ? fetchedManifestText : manifestText}
+          />
         </Modal.Body>
         <Modal.Footer>
           <button className="btn btn-primary" onClick={toggle}>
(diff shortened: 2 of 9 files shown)


packages/core/src/manifest/ManifestYaml.tsx
```diff
@@ -1,26 +1,42 @@
...
+export function ManifestYaml({ linkName, modalTitle, manifestText, manifestUri }: IManifestYamlProps) {
   const [modalVisible, setModalVisible] = React.useState<boolean>(false);
   const toggle = () => setModalVisible(!modalVisible);
+  const [fetchedManifestText, setFetchedManifestText] = React.useState<string>('Loading...');
+
+  React.useEffect(() => {
+    if (manifestUri) {
+      ArtifactService.getArtifactByContentReference(manifestUri)
+        .then((manifest) => setFetchedManifestText(decodeUnicodeBase64(manifest.reference)))
+        .catch((e) => setFetchedManifestText(`Error:  ${typeof e !== 'string' ? JSON.stringify(e) : e}`));
```
nemesisOsorio, line 20: if this happens the result is kind of weird 
image
should we use another component to show the error? instead of using the same `ManifestYaml`

xibz: Okay, fixed that. Looks like this now
Screen Shot 2023-07-06 at 5 30 09 PM

packages/core/src/manifest/ManifestYaml.tsx
```diff
@@ -1,26 +1,48 @@
...
-export function ManifestYaml(props: IManifestYamlProps) {
+export function ManifestYaml({ linkName, modalTitle, manifestText, manifestUri }: IManifestYamlProps) {
   const [modalVisible, setModalVisible] = React.useState<boolean>(false);
   const toggle = () => setModalVisible(!modalVisible);
+  const [fetchedManifestText, setFetchedManifestText] = React.useState<string>('Loading...');
+  const [error, setError] = React.useState<string | null>(null);
+
+  React.useEffect(() => {
+    if (manifestUri) {
+      ArtifactService.getArtifactByContentReference(manifestUri)
```
nemesisOsorio, line 19: I'm wondering if we can memoize this call

xibz: We could cache things since it is based on the hash of the artifact. That could be a separate PR for optimization, if necessary. I dont think these calls happpen too too often, so may be fine without it

nemesisOsorio:
Bake (Manifest) was showing a produced artifact, with this change, that artifact is gone.
image

xibz:
> Bake (Manifest) was showing a produced artifact, with this change, that artifact is gone.
> 
> image
> 
> 

Thanks for the review. Ill take a look. I may have a front end engineer internally take a look. The original author of this PR has more context. I only worked on the backend portion. So give me a sec

nemesisOsorio (approved):
Thank you

packages/core/src/domain/IArtifact.ts
```diff
@@ -12,3 +12,6 @@ export interface IArtifact {
   kind?: string; // TODO delete
   customKind?: boolean; // TODO delete
 }
+
+export const ARTIFACT_TYPE_EMBEDDED = 'embedded/base64';
+export const ARTIFACT_TYPE_REMOTE = 'remote/base64';
```
mattgogerly, line 17: Duplicates `ArtifactTypes.ts`?

packages/core/src/pipeline/config/stages/bakeManifest/BakeManifestDetailsTab.tsx
```diff
@@ -3,31 +3,43 @@ import React from 'react';
...
-        {bakedArtifacts.map((artifact, i) => (
-          <ManifestYaml
-            key={i}
-            linkName={bakedArtifacts.length > 1 ? `Baked Manifest ${i} YAML` : 'Baked Manifest YAML'}
-            manifestText={decodeUnicodeBase64(artifact.reference)}
-            modalTitle="Baked Manifest"
-          />
-        ))}
+        {bakedArtifacts.map((artifact, i) => {
+          const linkName = bakedArtifacts.length > 1 ? `Baked Manifest ${i} YAML` : 'Baked Manifest YAML';
```
mattgogerly, line 25: I haven't had a chance to review the whole set of PRs - can a bake stage now bake multiple times?

packages/core/src/pipeline/config/triggers/artifacts/ArtifactService.ts
```diff
@@ -11,4 +11,8 @@ export class ArtifactService {
       .query({ type: type, artifactName: artifactName })
       .get();
   }
+
+  public static getArtifactByContentReference(contentRef: string): PromiseLike<{ reference: string }> {
+    return REST(`/artifacts/content-address/${contentRef}`).get();
```
mattgogerly, line 16: I think the convention is to use `.path()` rather than String templates (see method above).

packages/core/src/domain/IArtifact.ts
xibz: `ArtifactTypes.ts` are used for pattern matching which we don't do here. I mean we could, but it seems unnecessary to pattern match when we know the full type. Honestly, I dont even know why we pattern match at all lol. There could be some reasoning for future proofing, but I've never seen it used, and this type of future proofing doesn't seem too useful, tbh, but I could be wrong. I don't know deck all that well, but the pattern matching is questionable, considering the backend rarely does it. I think it's only in one spot in clouddriver lol, but everything else checks directly against the type.

packages/core/src/pipeline/config/stages/bakeManifest/BakeManifestDetailsTab.tsx
xibz: This is old code from 4 years ago. We just moved it into a variable.

packages/core/src/pipeline/config/triggers/artifacts/ArtifactService.ts
xibz: So `contentRef` is `application/hash`, so the reason it was done this way is to avoid HTTP escaping the `/`.

sergio-quintero (approved):

dbyron-sf (approved):

PR #10012, opened 2023-07-13
feat(kubernetes): Add Deck stage for Rolling Restart

mattgogerly:
The stage already exists but is only exposed as an ad-hoc action on the Infra tabs. Adding an actual pipeline stage for it.

Ref https://github.com/spinnaker/orca/blob/master/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/pipeline/manifest/RollingRestartManifestStage.java


dbyron-sf (approved):
dicey of me to be approving deck code, but LGTM.

PR #10017, opened 2023-07-25
 fix(deck): Removal of API

AbhaHimani:
Removal of loadBalancer API call from the website's loading page


mattgogerly:
Could you explain what you're trying to fix here?

mattgogerly:
Sorry, been slammed recently. Making a note to myself to come back to this - I want to double check the flow as I think the calls made are controlled elsewhere too.

AbhaHimani:
Sure sure! Let me know

PR #10018, opened 2023-07-27
feat(core): set Cancellation Reason to be expanded by default

bliu2368:
image

Changes:
* Set cancellation Reason to be expanded by default

Reason:
Cancellation reason gives context to the user on why a specific pipeline was forced to stop. When it's unexpanded it can be easy to miss the UI element that pops up as a result of the cancellation reason being present.  Showing the reason upfront can prevent foot-guns for Users as it can contain instructions or help the user avoid the issue that caused the cancellation in the first place.


mattgogerly:
@Mergifyio update

jasonmcintosh:
I kinda like this UI change myself :) Though it begs the question of why it's an expansion if it defaults to on but... I can see the appeal for sure

jasonmcintosh (approved):

PR #10020, opened 2023-08-29
feat(core): Add ability to set Default Tag filters for an application…

bliu2368:
… in application config

Summary:
* Added section in Application Config which allows users to define "Default Filters" for their pipelines page. 
<kbd></kbd>
* Each of these Default Filters will automatically be applied when opening the pipelines page, they can still be temporarily disabled
<kbd></kbd>
* Currently only supports filtering by tag (which can be set per pipeline) 
* Added tests

Most of the code changes were based on how bannerConfig was written, so follows the patterns used there. (e.g. the tests are practically identical)

Brief Demo: 

https://github.com/spinnaker/deck/assets/75234625/c43a2a12-570b-4243-aa42-3bf110c22109


mattgogerly (approved):

mattgogerly:
@Mergifyio update

PR #10021, opened 2023-08-30
Revert "fix(core): conditionally hide expression evaluation warning messages (#9771)"

mattgogerly:
Revert "fix(core): conditionally hide expression evaluation warning messages (#9771)"

This reverts commit 7e3dd5053ccdb06ce067303062f90ae82b56bfc8.

This is the cause of https://github.com/spinnaker/spinnaker/issues/6869, which occurs when there are expression evaluation failures as well as actual task failures. 

Deck merges all failure messages found in a stage [into a single string](https://github.com/spinnaker/deck/blob/db06e88bada70fa4065f56fc33af7207943415c5/packages/core/src/orchestratedItem/orchestratedItem.transformer.ts#L143) which is then passed to this component. If there are any expression evaluation failures they are always first in the string, so the filter for `Failed to evaluate` is removing the entire thing rather than just the expression failures as intended.

### DIFF
--- a/packages/core/src/pipeline/details/StageFailureMessage.tsx
+++ b/packages/core/src/pipeline/details/StageFailureMessage.tsx
@@ -73,33 +73,19 @@ export class StageFailureMessage extends React.Component<IStageFailureMessagePro
   }
 
   public render() {
-    const { message, messages, stage } = this.props;
+    const { message, messages } = this.props;
     const { isFailed, failedTask, failedExecutionId, failedStageName, failedStageId } = this.state;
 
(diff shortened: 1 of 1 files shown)


dbyron-sf (approved):

dbyron-sf:
Not sure whether we're gonna get another 1.29.x out the door, but let's backport it there too just in case.

dbyron-sf:
@mergifyio backport release-1.29.x

mattgogerly:
Looks like Mergify and spinnakerbot are chugging today.

PR #10044, opened 2023-10-10
chore(upgrades): OS Upgrades

jasonmcintosh:
I'm NOT at this point rewriting the deck base.  
* The slim is debian based, ubuntu is ubuntu based.
* opnejdk:11-slim is deprecated, migrated to alpine & the openjdk11 installer from that for builds.
If we wanted to match other services and use Alpine which is SLIGHTLY slimmer, save that for different PR/discussion.


mattgogerly:
Fancy making this openjdk17 and saving me a job? :)

(why does the Deck image even have openjdk? Keytool required or something)

jasonmcintosh:
NO clue why it needs the JRE honestly.  I'd ... have strong questions about that.  Even keytool isn't needed since it's apache....  The JDK in theory shouldn't be needed on ANY of the services

mattgogerly (approved):

PR #10026, opened 2023-08-30
fix: Scaling bounds should parse float not int

jasonmcintosh:
An annoyance when trying to use percentages in some of the metrics.

PR #10059, opened 2023-11-09
fix(amazon): Allow scaling bounds to use floats between input steps

sergio-quintero:
Follow up to https://github.com/spinnaker/deck/pull/10026

It is necessary to add the standard HTML `step` prop in `NumberInput` to allow floats as inputs.


nemesisOsorio (approved):

edgarulg (approved):

jasonmcintosh (approved):
Thanks knew I’d missed this!

PR #10073, opened 2023-12-22
Make Wait Stage restartable

Sitkevich88:
I think, It is necessary to make Wait stage restartable, because if you cancel release pipeline, while Wait Stage is running, you won't be able to restart pipeline where it stopped, you will be forced to restart pipeline from the very beginning.

Before:
image

After:
image


jasonmcintosh (approved):
Good with making it restartable.  Kinda surprised there isn't a setting default on this in orca but didn't see much.  I'm not AS familiar with some of the backend logic on this but overall I'd THINK this would make sense to me :)

PR #10076, opened 2024-01-07
feat(ecs): Add support for ECS deployment circuit breaker

jgrumboe:
This is my first contribution, and I tried to implement https://github.com/spinnaker/spinnaker/issues/6225

This change comes together with this PR from clouddriver: https://github.com/spinnaker/clouddriver/pull/6132

It implements a new checkbox to activate ECS Deployment Circuit Breaker for the ECS service; the default is false to not change any existing behavior.

The checkbox looks like this. I thought it should go under "Advanced settings".
image

If enabled, it adds the attribute `enabledDeploymentCircuitBreaker` to the server group:
image

Running the Deploy stage will create an ECS service with Deployment Circuit Breaker enabled:
image

I'm open for your comments and your thoughts, also on the implementation - I mainly "mimicked" how other attributes/feature were implemented and tried my best "copy/pasting". 😏


ovidiupopa07 (approved):
LGTM. I would switch between the checkbox and the text to have the same experience as for the other fields but it looks ok.

jgrumboe:
> LGTM. I would switch between the checkbox and the text to have the same experience as for the other fields but it looks ok.

Good point. I copied it from somewhere.
I will switch it. So the checkbox should be aligned with the text input boxes above, right?

jgrumboe:
@ovidiupopa07 
It looks like this now, much cleaner.

jasonmcintosh (approved):

ovidiupopa07 (approved):

PR #10081, opened 2024-04-15
fix(runJobs): Persist External Log links after the deletion of the pods

christosarvanitis:
After the deletion of the runjob (for any reason) Deck instead of persisting the External system link for the logs it falls back to the default Console Output.

Since the user is linking the logs of a Runjob to an external system it makes sense that the link for the previous executions should persist in Deck. 

Fixes this old issue: https://github.com/spinnaker/spinnaker/issues/5717


ovidiupopa07 (approved):

christosarvanitis:
Backport candidate for all active releases (1.33.x, 1.32.x) if you agree

ovidiupopa07 (approved):

ovidiupopa07:
@Mergifyio backport release-1.33.x release-1.32.x

PR #10085, opened 2024-04-24
fix(lambdaStages): Exporting Lambda stages based on the feature flag setting

christosarvanitis:
When the Lambda Stages where backported from the plugin to Deck the register pipeline stages where not exported.

Also the NumberConcurrencyInput.tsx from the plugin is a duplicate of the NumberInput of Deck so removing it from the codebase


jasonmcintosh (approved):
I'm not as comfortable with UI pieces, but this LOOKS like should fix the lambda issue

christosarvanitis:
@mattgogerly would love your input on this since you did all the heavy lifting on the Lambda import :)

jasonmcintosh (approved):

jasonmcintosh:
https://github.com/Mergifyio backport release-1.34.x release-1.33.x release-1.32.x

PR #10086, opened 2024-04-25
fix(pipelineGraph): Handling exception when requisiteStageRefIds is not defined

christosarvanitis:
Deck prior to 1.27.x was rendering the pipelines when the first stages didnt have defined the requisiteStageRefIds. 

Although this was not ideal it allowed users to template their first stages without the key. After the pipelineGraph updates this is causing an exception and breaking the rendering of the pipelines causing issues to users upgrading.

With this fix it is checked if the key is not defined and defaulting it to a empty array.

Screenshot 2024-04-25 at 09 06 19


ovidiupopa07 (approved):

ovidiupopa07:

PR #10093, opened 2024-05-06
feat(taskView): Implement opt-in paginated request for TaskView

christosarvanitis:
The taskView by default tries to load the full task list (as allowed by the Orca api configuration). However for applications that are heavily used and have a high number of tasks retrieving the full task list in a single request is resulting many times in timeout either because of the backend or because the Deck timeout was reached.

The task controller on the backend supports pagination so this PR attempts to provide the end user control on the max task limit per page to be retrieved from the backend. 
The datasource logic doesnt change in the PR - Deck will still consider the taskView as loaded when all the tasks are retrieved but what changes is that by adding the opt-in setting in settings-local.js
```
SETTINGS.tasksViewLimitPerPage = number; 
```
on the taskView the single request will be splitted in multiple requests with limit=<number>&page=. 

https://github.com/spinnaker/deck/assets/28927773/09c0f28a-79db-41c9-bb8f-a24e36126bd7

### DIFF
--- a/packages/core/src/task/task.dataSource.js
+++ b/packages/core/src/task/task.dataSource.js
@@ -15,17 +15,22 @@ angular.module(CORE_TASK_TASK_DATASOURCE, [CLUSTER_SERVICE]).run([
       return $q.when(angular.isArray(tasks) ? tasks : []);
     };
 
-    const loadTasks = async (application, page = 1) => {
+    const loadPaginatedTasks = async (application, page = 1) => {
+      let limitPerPage = SETTINGS.tasksViewLimitPerPage;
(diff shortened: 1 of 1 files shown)


ovidiupopa07 (approved):
this is to avoid timing out when loading thousands of tasks in one go. For the future I think it would be great to refactor to a fully paginated appraoch

christosarvanitis:
@ovidiupopa07 Since this is behind a feature flag and there is no change in the functionality when not used can we backport it to the active releases?

ovidiupopa07:
@Mergifyio backport release-1.32.x release-1.33.x release-1.34.x

PR #10097, opened 2024-05-07
fix(lambda): Invoke stage excludedArtifactTypes not including the embedded-artifact type

christosarvanitis:
Lambda Invoke Stage after the implementation of the remote/base64 as an artifact type is not including the embedded-artifact account in the expected artifacts since the REMOTE_BASE64 artifact type was not added in the except list.

Before the change:
image

After the change:
image


jasonmcintosh:

PR #10103, opened 2024-05-10
fix(pipeline): Handle render/validation when stageTimeoutMs is a Spel expression

christosarvanitis:
Deck is throwing an exception when stageTimeoutMs is templated with a Spel Expression although Orca can handle the evaluation during runtime. 

This PR fixes the exception in Deck and also modifies the validator to accept a Spel expression. 

Before the change: 
image

After the change: 
image

Recording with a pipeline that uses input with Spel to determine the stageTimeoutMs:

https://github.com/spinnaker/deck/assets/28927773/8f550f45-e2b4-4abf-8ee6-ff6a9fe8c182


ovidiupopa07:

PR #10122, opened 2024-07-09
chore(secrets): adding GHA to sync NPM_AUTH_TOKEN to spinnaker/spinna…

link108:
After this PR is merged, I'll add a personal access token, scoped to spinnaker/deck and spinnaker/spinnaker with the secrets read/write permissions enabled. I'm just going to use a personal token vs an org token, since we don't have that setup at the moment.

### DIFF
--- a/.github/workflows/sync-secrets.yml
+++ b/.github/workflows/sync-secrets.yml
@@ -22,11 +22,11 @@ jobs:
         uses: jpoehnelt/secrets-sync-action@v1.9.0
         with:
           SECRETS: |
-            ^SECRET_TO_SYNC$
+            ^NPM_AUTH_TOKEN$
           REPOSITORIES: |
-            ^link108/sync-secrets-to$
+            ^spinnaker/spinnaker$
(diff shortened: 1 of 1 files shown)


dbyron-sf (approved):

.github/workflows/sync-secrets.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Sync Secrets
+        uses: jpoehnelt/secrets-sync-action@v1.9.0
+        with:
+          SECRETS: |
+            ^SECRET_TO_SYNC$
+          REPOSITORIES: |
+            ^link108/sync-secrets-to$
```
dbyron-sf, line 27: Can we sync to https://github.com/spinnaker/spinnaker directly?

link108: ah yes, that was my intent, updated!

dbyron-sf (approved):

PR #10143, opened 2024-09-11
feat(evaluateVariables stage) Made fetch exec history on evaluateVariables stage configurable

wooseung-sim:
Evaluate Variables stage on pipeline config view loads previous 100 pipeline execution history to show the evaluated variable values on previous execution.

The size of execution context is too huge if the pipeline is lengthy and it has a lot of artifacts and have a lot of previous executions. One of our pipeline is not accessible evaluate variables stage in the pipeline config as it timed out on `executionService.getExecutionsForConfigIds()` executions always due to size of the response (~400MB) and slowness of the VPN network.

phoenix-mcd_·_pipeline_config

Spinning circle UI is keep going and then timed out usually due to slowness of the network and size of the http get response.

This change can make the number of fetching history configurable on `settings-local.js`
```
window.spinnakerSettings.maxFetchHistoryOnEvaluateVariables = 5;
```

Default value of it is `100` if it is not defined on `settings-local.js`.

With this change we can adjust fetch history size for evaluate variable stage config UI, and prevent UI delays and errors.

### DIFF
--- a/packages/core/src/pipeline/config/stages/evaluateVariables/ExecutionAndStagePicker.tsx
+++ b/packages/core/src/pipeline/config/stages/evaluateVariables/ExecutionAndStagePicker.tsx
@@ -26,7 +26,7 @@ export function ExecutionAndStagePicker(props: IExecutionAndStagePickerProps) {
   const fetchExecutions = useData(
     () =>
       executionService.getExecutionsForConfigIds([pipeline.id], {
-        limit: SETTINGS.maxFetchHistoryOnEvaluateVariables
+        limit: SETTINGS.maxFetchHistoryOnEvaluateVariables,
       }),
     [],
     [],
--- a/packages/core/src/pipeline/config/stages/evaluateVariables/ExecutionAndStagePicker.tsx
+++ b/packages/core/src/pipeline/config/stages/evaluateVariables/ExecutionAndStagePicker.tsx
@@ -2,6 +2,7 @@ import { isEqual, keyBy } from 'lodash';
 import React from 'react';
 import type { Option } from 'react-select';
 
+import { SETTINGS } from '../../../../config';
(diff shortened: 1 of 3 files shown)


packages/core/src/pipeline/config/stages/evaluateVariables/ExecutionAndStagePicker.tsx
```diff
@@ -23,7 +24,10 @@ export function ExecutionAndStagePicker(props: IExecutionAndStagePickerProps) {
   const [showStageSelector, setShowStageSelector] = React.useState(false);
 
   const fetchExecutions = useData(
-    () => executionService.getExecutionsForConfigIds([pipeline.id], { limit: 100 }),
+    () =>
+      executionService.getExecutionsForConfigIds([pipeline.id], {
+        limit: SETTINGS.maxFetchHistoryOnEvaluateVariables ?? 100,
```
dbyron-sf, line 29: Is there some way to keep 100 in one place?

richard-timpson: There's some code in packages/core/src/config/settings.ts at line 167 that sets default values for the configs if they are missing. I think that's the right place to put the 100 by default, rather than worrying about it here.

wooseung-sim: Moved default value config to `packages/core/src/config/settings.ts` as per review comment. Thank you.

richard-timpson (approved):

dbyron-sf:
@wooseung-sim can you run `yarn prettier` please?

wooseung-sim:
> @wooseung-sim can you run `yarn prettier` please?

Thank you for letting me know. Ran it and pushed fixed commit.

PR #10149, opened 2024-10-03
feat(gce): add support for hyperdisk in GCE

edgarulg:
Adds support for `hyperdisk-balanced` in Spinnaker for GCE. See: https://cloud.google.com/compute/docs/disks/hyperdisks

In the following changes, I added `hyperdisk-balanced` as option for Persistent Disk in the deploy stage when using GCE in Spinnaker.

Captura de pantalla 2024-10-02 a la(s) 6 27 22 p m

This allows any user to configure their GCE instance to use hyperdisk. The backend already supports this new persistent disk type with the following PR: https://github.com/spinnaker/clouddriver/pull/6288


christosarvanitis (commented):
@edgarulg can you update the branch and the PR title please?

jasonmcintosh (approved):

PR #10158, opened 2024-12-10
fix(stages): accountExtractor in k8s stages should extract data from a context

kbatalin:
For reference, see other stages:
- https://github.com/spinnaker/deck/blob/c96faa581482d251c07132bdcc023e04fc8e869b/packages/cloudfoundry/src/pipeline/stages/runJob/cloudFoundryRunJob.module.ts#L8
- https://github.com/spinnaker/deck/blob/c49d9b545642f511bfd2e14588b78fdf76385c5a/packages/amazon/src/pipeline/stages/cloneServerGroup/awsCloneServerGroupStage.js#L18

During execution, the account is stored in the context

Screenshot after fixing:

<kbd></kbd>


mattgogerly:
@kbatalin mind resolving the conflict so this can hop aboard the merge train?

kbatalin:
@mattgogerly 
👍  Done

mattgogerly (approved):

PR #10160, opened 2025-01-01
feat(amazon): support ALBRequestCountPerTarget scaling policies

chris-h-phillips:
The UI for TargetTracking scaling policies didn't support ALBRequestCountPerTarget. This was the only missing metric documented here: https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-scaling-target-tracking.html

This requires setting the resourceLabel parameter which is a combination of part of both the ALB and target group ARNs. This metric also only works with the Sum statistic. I found that the chart component hard coded average so that had to be fixed up as well.

Here are some screenshots
Summary before and after:
image

image
UpsertModal before and after:
(before the ALBRequestCountPerTarget isn't available)
image

image

### DIFF
--- a/packages/amazon/src/domain/ITargetTrackingPolicy.ts
+++ b/packages/amazon/src/domain/ITargetTrackingPolicy.ts
@@ -21,6 +21,11 @@ export interface ICustomizedMetricSpecification {
 
 export interface IPredefinedMetricSpecification {
   predefinedMetricType: PredefinedMetricType;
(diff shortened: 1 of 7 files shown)


dbyron-sf:
@mattgogerly can you take a look at this please?

mattgogerly (approved):
Caveat that I'm not massively familiar with the ALB scaling policies, but lgtm.

Thanks for the screenshots!

PR #10164, opened 2025-02-19
feat(pipelineRef): conditionally load childExecutions when pipelineRef is enabled

edgarulg:
When pipelineRef is enabled with the following changes: https://github.com/spinnaker/orca/pull/4842
When click on the "View Pipeline Execution" in a execution with PipelineRef trigger Deck is not able to render parentExecution because it does not know how to handle a PipelineRefTrigger.
Captura de pantalla 2025-02-20 a la(s) 3 08 32 p m

As part of this changes I defined a new feature flag in deck that conditionally retrieve parent executions when the user visit the SingleExecutionDetails component in the UI.

**How to enable**
`Setting.js` needs to define the pipelineRefEnabled flag.
`window.spinnakerSettings.feature.pipelineRefEnabled = true;`


danielGz (approved):

jasonmcintosh (approved):
I'm... still not totally comfortable on deck without tests, but.. this LOOKS like it'd work.  I think the "tests" aspect is the only concern, but if rest of the approvers are good with UI stuff like this without tests, I'm good with it...

christosarvanitis (approved):
Thanks @edgarulg! This will keep the a consistent experience with not enabled the feature.

PR #10165, opened 2025-02-28
fix(gcp): when regional eq true, show instances from all existing zones

shlomodaari:
Before the fix: When regional=true, the system did not display all instances across all zones within the selected region.
image

After the fix: When regional=true, all instances from every zone within the selected region are correctly displayed.
image

When only one zone is selected, the behavior remains unchanged, and only the instances from that specific zone are shown.
This ensures accurate instance visibility for both regional and single-zone selections.


christosarvanitis:
This fix is alongside with https://github.com/spinnaker/clouddriver/pull/6351.

Thanks @shlomodaari

christosarvanitis:
@shlomodaari can you please remove the package-lock.json commit from this PR?

shlomodaari:
@christosarvanitis I removed it, thank you for catching this!

edgarulg (approved):
LGTM, thanks for helping with this issue.

PR #10171, opened 2025-04-08
fix(core): Always use clusters parameter for ECS server group URLs

shlomodaari:
# Fix ECS Server Group URLs When onDemandClusterThreshold is Exceeded

## Problem

When viewing ECS server groups in Spinnaker and the `onDemandClusterThreshold` is exceeded, the generated URLs fail to navigate to the correct server group. This happens because:

1. ECS provider requires using the `clusters` parameter for server group navigation
2. The current implementation uses the standard `q`, `acct`, and `reg` parameters which don't work properly for ECS when on-demand caching is disabled

This results in broken links for ECS server groups in the UI when the number of caches exceeds the threshold.

## Solution

This PR modifies the `ServerGroupsUrlBuilder.build` method to use a different URL format specifically for ECS server groups:

// For ECS provider, use clusters parameter format
if (input.provider === 'ecs' || (input.provider && input.provider.includes('ecs'))) {
  // Extract cluster name from server group name (remove version suffix)
  const serverGroupParts = input.serverGroup.split('-');
  let clusterName = input.serverGroup;
  if (serverGroupParts.length > 1 && serverGroupParts[serverGroupParts.length - 1].match(/^v\d+$/)) {
    clusterName = serverGroupParts.slice(0, -1).join('-');
  }

  // Use clusters parameter with account:clusterName format
  return UrlBuilderUtils.buildUrl(href, { clusters: `${input.account}:${clusterName}` });
}

// For other providers, use standard parameters
return UrlBuilderUtils.buildUrl(href, { q: input.serverGroup, acct: input.account, reg: input.region });

## Testing

Testing was performed in an environment where the `onDemandClusterThreshold` was exceeded. 

**Before the fix:**
URL: /#/applications/myapp/serverGroups?q=my-ecs-service-v001&acct=my-account&reg=us-west-2
Result: Empty view, no server group details displayed

**After the fix:**
URL: /#/applications/myapp/serverGroups?clusters=my-account:my-ecs-service
Result: Successfully displays the ECS server group details

The fix has been verified with various ECS server group naming patterns, including those with and without version suffixes.

## Impact

This change:
- Fixes navigation to ECS server groups when `onDemandClusterThreshold` is exceeded
- Maintains backward compatibility for all other cloud providers
- Improves user experience by ensuring links work correctly in all scenarios

### DIFF
--- a/packages/core/src/navigation/UrlBuilder.ts
+++ b/packages/core/src/navigation/UrlBuilder.ts
@@ -244,7 +244,12 @@ class ServerGroupsUrlBuilder implements IUrlBuilder {
       { inherit: false },
     );
 
-    if (input.provider === 'ecs' || (input.provider && input.provider.includes('ecs'))) {
+    if (
+      input.provider === 'ecs' ||
+      (input.provider && input.provider.includes('ecs')) ||
+      input.provider === 'aws' ||
+      (input.provider && input.provider.includes('aws'))
+    ) {
       const serverGroupParts = input.serverGroup.split('-');
       let clusterName = input.serverGroup;
       if (serverGroupParts.length > 1 && serverGroupParts[serverGroupParts.length - 1].match(/^v\d+$/)) {
--- a/packages/core/src/navigation/UrlBuilder.ts
+++ b/packages/core/src/navigation/UrlBuilder.ts
@@ -244,6 +244,16 @@ class ServerGroupsUrlBuilder implements IUrlBuilder {
       { inherit: false },
     );
 
+    if (input.provider === 'ecs' || (input.provider && input.provider.includes('ecs'))) {
+      const serverGroupParts = input.serverGroup.split('-');
+      let clusterName = input.serverGroup;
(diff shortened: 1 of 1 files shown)


christosarvanitis:
thanks @shlomodaari! I think this bug is not only on ECS though. It must be in the AWS as well (since the ECS cluster view is basically the AWS asg with conventions). 
Possibly it is valid  for other providers like GCP or Kubernetes.

shlomodaari:
FYI: The good part is that with this fix it should work when onDemandClusterThreshold is exceeded and also when it is disabled

edgarulg (approved):
Thanks for working on this fix. The implementation looks good to me but I agree with @christosarvanitis this issue might be present in other providers as well.

I am leaving my approve because the initial issue is being fixed but we need to be sure other providers are not affected the same way.

shlomodaari:
Okay, this is fixing now ECS and AWS!
Need to confirm k8s next!

edgarulg:
@Mergify backport release-1.38.x release-1.37.x release-1.36.x

PR #10177, opened 2025-04-25
chore(test/functional): upgrade cypress

cristhian-castaneda:
This PR updates the functional test suite. It removes outdated dependencies, upgrades configurations, and improves developer experience.

#### Removed Deprecated Tools

- **Replaced `cy.route`/`cy.server` with `cy.intercept`**  
  Cypress deprecated `cy.route` in favor of `cy.intercept`. This update ensures forward compatibility.

- **Removed `@uirouter/cypress-runner`**  
  This package was last updated in 2023. It has been removed in favor of a modern dev/test setup using Vite and `start-server-and-test`.

#### Cypress Test Runner Overhaul

- Migrated `cypress.json` to `cypress.config.ts` (recommended format).
- Integrated `@cypress/code-coverage` for tracking test coverage.
- Cleaned up legacy workarounds like `scrollBehavior.js`.

#### New Tooling and Scripts

- Introduced Vite (`vite.config.ts`) for serving the app during test runs.
- Updated `package.json` adding new scripts


test/functional/cypress/integration/google/accelerator_zones.spec.js
```diff
@@ -7,23 +7,28 @@ describe('google: Create Server Group Modal GPU Accelerators', () => {
...
 
   it('should not show NVIDIA Tesla K80 initially', () => {
     cy.visit('#/applications/compute/clusters');
     cy.window().its('angular').should('exist');
-    cy.get('button:contains("Create Server Group")', { timeout: 10000 })
-      .should('be.visible')
-      .click();
+    cy.get('button:contains("Create Server Group")', { timeout: 10000 }).should('be.visible').click();
+
+    cy.contains('ul.steps-indicator li a', 'Advanced Settings').click();
```
jasonmcintosh, line 24: New check looks like?  I like if I'm reading the syntax right :)

test/functional/cypress/integration/google/accelerator_zones.spec.js
```diff
@@ -32,30 +37,25 @@ describe('google: Create Server Group Modal GPU Accelerators', () => {
...
+    cy.get('button:contains("Create Server Group")', { timeout: 10000 }).should('be.visible').click();
 
     // Select region and zone
     cy.get('v2-wizard-page[key=location]').within(() => {
       cy.get('div.form-group:contains("Region") select').select('us-east1');
     });
 
     cy.get('v2-wizard-page[key=zones]').within(() => {
-      cy.get('div.form-group:contains("Zone") select').select('us-east1-c');
+      cy.get('div.form-group:contains("Zone") select').select('us-east1-c', { force: true });
```
jasonmcintosh: is force true needed?

test/functional/cypress/integration/google/accelerator_zones.spec.js
```diff
@@ -32,30 +37,25 @@ describe('google: Create Server Group Modal GPU Accelerators', () => {
...
     });
 
-    cy.wait(1000);
-    cy.get('.btn-primary').first()
-      .should('be.visible')
-      .click({ force: true });  
+    cy.contains('ul.steps-indicator li a', 'Advanced Settings').click();
```
jasonmcintosh, line 51: Intentionally losing the primary button visible check?  And wait?

test/functional/cypress/integration/google/accelerator_zones.spec.js
cristhian-castaneda: Yes, the original test assumed Cypress would automatically scroll the element (the button clicked just below) into the viewport. But in newer Cypress versions, the default scroll behavior doesn’t always bring elements fully into view before interacting with them.

This used to be handled by a `scrollBehavior.js` support script that forced elements to scroll to the center of the viewport:

```js
Cypress.on('scrolled', $el => {
  $el.get(0).scrollIntoView({ block: 'center', inline: 'center' });
});
```

That script was removed in this PR and partially replaced by setting `scrollBehavior: 'center'` in Cypress config, which now achieves the same effect more cleanly.

test/functional/cypress/integration/google/accelerator_zones.spec.js
cristhian-castaneda: Good catch, `{ force: true }` here seemed helpful during testing, but yeah, not actually needed anymore

test/functional/cypress/integration/google/accelerator_zones.spec.js
cristhian-castaneda: Same case as noted here: https://github.com/spinnaker/deck/pull/10177#discussion_r2067836543

cristhian-castaneda:
Superseded by https://github.com/spinnaker/spinnaker/pull/7062

PR #10188, opened 2025-07-22
fix(deck): Fix RunJob external logs when component is not mounted and  interpolation is needed

christosarvanitis:
Deck is failing to render the external logs link when interpolation is needed but the async operation to Clouddriver to mount the manifest. This PR fixes this issue but keeps the ExternalLog link when interpolation is not needed

### DIFF
--- a/packages/core/src/manifest/stage/JobStageExecutionLogs.tsx
+++ b/packages/core/src/manifest/stage/JobStageExecutionLogs.tsx
@@ -1,4 +1,4 @@
-import { template } from 'lodash';
+import { isEmpty, template } from 'lodash';
 import React from 'react';
 import { from as observableFrom, Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


radu-dragon (approved):

packages/core/src/manifest/stage/JobStageExecutionLogs.tsx
```diff
@@ -52,7 +52,7 @@ export class JobStageExecutionLogs extends React.Component<IJobStageExecutionLog
     const { manifest } = this.state;
     const { externalLink, podNamesProviders, location, account } = this.props;
     // prefer links to external logging platforms
-    if (externalLink) {
+    if (externalLink && (!externalLink.includes('{{') || !isEmpty(manifest))) {
```
jasonmcintosh, line 55: I think we need a rewrite or rethink on this... too complicated of logic. I could see passing all of this down a bit and being a bit smarter on the checks internally.

ALSO ideally would have a test on these... spec test that checks the render logic maybe

